PR #88, opened 2023-07-26
[OF-1079] test: add `csp::common::Array` tests

MAG-MichaelK:
**For the reviewer**

* [x] If required, are the changes covered by appropriate tests?
* [x] Are any public-facing API changes well documented?
* [x] Is the code easily readable and extensible and/or follow existing conventions?

### DIFF
--- a/Library/include/CSP/Common/Array.h
+++ b/Library/include/CSP/Common/Array.h
@@ -22,6 +22,7 @@
 #include <cassert>
 #include <cstring>
 #include <initializer_list>
+#include <stdexcept>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


MAG-SamBirley (commented):
Can we add tests that test unhappy paths with `Array` also? This will also lead to a conversation about how to handle index-out-of-bounds errors. Open to whatever we think we should be doing here, but it should be a conscious decision.

MAG-SamBirley (approved):
Love it. Nice work.

MAG-mv (approved):

PR #91, opened 2023-07-27
[NT-0] feat: string splitting by multiple delims

MAG-AW:

Added String Split by multiple delimiters

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Common/Array.h
+++ b/Library/include/CSP/Common/Array.h
@@ -22,7 +22,16 @@
 #include <cassert>
 #include <cstring>
 #include <initializer_list>
+#include <stdexcept>
 #include <utility>
+
+#ifndef CSP_DISABLE_OVERFLOW_CHECKING
+	#ifdef _MSC_VER
+		#include <intrin.h>
+	#else
+		#include <cstdint>
+	#endif
+#endif
 
 
 namespace csp::common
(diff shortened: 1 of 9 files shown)


MAG-SamBirley (changes requested):
Please could we get some tests added as part of this PR covering both the happy and unhappy paths

Library/src/Common/String.cpp
```diff
@@ -163,6 +163,28 @@ List<String> String::Split(char Separator) const
...
+List<String> String::Split(const char* Delimiters) const
+{
+	std::stringstream stringStream(ImplPtr->Text);
+	List<String> StringList;
+	std::string InputLine;
+
+	while (std::getline(stringStream, InputLine))
+	{
+		std::size_t Prev = 0, Pos;
+		while ((Pos = InputLine.find_first_of(Delimiters, Prev)) != std::string::npos)
```
MAG-MichaelK, line 175: Can we get some spacing around this while block and the if blocks?

Library/src/Common/String.cpp
```diff
@@ -163,6 +163,28 @@ List<String> String::Split(char Separator) const
...
+		{
+			if (Pos > Prev)
```
MAG-MichaelK, line 177: Coding convention dictates we should always surround if blocks with curly brackets, even if they are single-line statements.

MAG-MichaelK (changes requested):

PR #94, opened 2023-07-28
[OF-1076] feat: client apps can query whether a space requires a ticket to enter or not

MAG-AW:

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/EventTicketing/EventTicketing.h
+++ b/Library/include/CSP/Systems/EventTicketing/EventTicketing.h
@@ -112,6 +112,29 @@ private:
 };
 
 /// @ingroup Event Ticketing System
+/// @brief Result class holding a collection (array) of TicketedEvents.
+class CSP_API SpaceIsTicketedResult : public csp::services::ResultBase
+{
+	/** @cond DO_NOT_DOCUMENT */
+	CSP_START_IGNORE
+	template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+	CSP_END_IGNORE
+	/** @endcond */
+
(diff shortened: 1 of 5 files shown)


Library/include/CSP/Systems/EventTicketing/EventTicketing.h
```diff
@@ -141,6 +168,9 @@ typedef std::function<void(const TicketedEventResult& Result)> TicketedEventResu
 // @brief Callback providing a ticketed event collection result.
 typedef std::function<void(const TicketedEventCollectionResult& Result)> TicketedEventCollectionResultCallback;
 
+typedef std::function<void(const SpaceIsTicketedResult& Result)> SpaceIsTicketedResultCallback;
+
+
```
MAG-SamBirley, line 173: Two newlines instead of just one?

MAG-SamBirley (changes requested):
Can we also validate that
- when we query if a space is ticketed, it comes back as false, given a space which has never had an event created for it
- when we query if a space is ticketed, it comes back as false, given a space which _has_ had an event created for it, but ticketing is not currently active

MAG-SamBirley (commented):
I can see one of my requested tests has been added (thank you!) but not the other - was there a reason?

- when we query if a space is ticketed, it comes back as false, given a space which has never had an event created for it ❌
- when we query if a space is ticketed, it comes back as false, given a space which has had an event created for it, but ticketing is not currently active ✅

MAG-SamBirley:
I see now that they are both there, just in the same test - would you mind adding some comments to the test just to call out how and what we're testing at each point? If I missed it, feels like others will too.

PR #98, opened 2023-07-28
[OF-1098] test: test `csp::common::String`

MAG-MichaelK:
This PR completes one half of the AC for OF-1098

**For the reviewer**


MAG-SamBirley (approved):

MAG-AdamThorn (approved):

Library/src/Common/String.cpp
```diff
@@ -39,15 +39,22 @@ class String::Impl
...
+		const char* _InText = InText;
+
 		if (InText == nullptr)
 		{
-			return;
+			_InText = "";
 		}
 
-		const size_t Len = strlen(InText);
+		const size_t Len = strlen(_InText);
```
MAG-SamBirley, line 49: Out of interest, what did this fix?

MAG-MichaelK: Text can't be null, so we need to set it to something.

PR #99, opened 2023-07-28
Of 1077 add ticket submit to event ticketing

MAG-pw:

**For the reviewer**

* [x] If required, are the changes covered by appropriate tests?
* [x] Is the code easily readable and extensible and/or follow existing conventions?


Library/include/CSP/Systems/EventTicketing/EventTicketing.h
```diff
@@ -41,6 +48,23 @@ class CSP_API TicketedEvent
...
 };
 
+/// @ingroup Event Ticketing System
+/// @brief Data representation of a ticket for an event
+class CSP_API EventTicket
+{
+public:
+	EventTicket() : Vendor(EventTicketingVendor::Unknown), Status(TicketStatus::Unknown) {};
+
+	csp::common::String Id;
```
MAG-SamBirley, line 58: Can we get inline doxygen formatted comments on these member vars explaining their purpose?

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
```diff
@@ -168,6 +219,9 @@ typedef std::function<void(const TicketedEventResult& Result)> TicketedEventResu
 // @brief Callback providing a ticketed event collection result.
 typedef std::function<void(const TicketedEventCollectionResult& Result)> TicketedEventCollectionResultCallback;
 
+// @brief Callback providing a ticketed event result.
+typedef std::function<void(const EventTicketResult& Result)> EventTicketResultCallback;
```
MAG-SamBirley, line 223: `SubmitEventTicketResultCallback` might be clearer?

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
```diff
@@ -26,6 +26,13 @@ enum class EventTicketingVendor
 	Unknown
 };
 
+enum class TicketStatus
+{
+	Purchased = 0,
```
MAG-SamBirley, line 31: I'm not sure where the best place to leave this documentation is, but writing down somewhere what 'purchased' means versus 'redeemed' would be awesome, as its not clear to me as the reader what the difference is/when I should expect one to be returned over the other.

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
```diff
@@ -35,6 +35,27 @@ csp::systems::EventTicketingVendor VendorNameToEnum(const csp::common::String& V
...
+		Status = csp::systems::TicketStatus::Purchased;
+	}
+	if (DtoStatus.GetValue() == chs::TicketStatus::eTicketStatus::REDEEMED)
+	{
+		Status = csp::systems::TicketStatus::Redeemed;
+	}
+	else
+	{
+		FOUNDATION_LOG_MSG(csp::systems::LogLevel::Warning,
+						   "Encountered an unknown ticket status when parsing a response from services. Defaulting to 'Unknown'");
```
MAG-SamBirley, line 53: Possibly this should be error-level? If this happens it's bad for the user as they'll likely hit a scenario where they've tried to use a ticket they got for a space, but due to a code error, they won't able to enter the space.

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
```diff
@@ -81,6 +102,73 @@ void SpaceEventDtoToTicketedEvent(const chs::SpaceEventDto& Dto, csp::systems::T
 	}
 }
 
+void SpaceTicketDtoToEventTicket(const chs::SpaceTicketDto& Dto, csp::systems::EventTicket& Ticket)
+{
+	Ticket.Id	   = Dto.GetId();
```
MAG-SamBirley, line 107: If you assigned `Ticket` a default-constructed `EventTIcket` object at the top here, you could avoid having to assign default values for all cases where a dto property was missing?

MAG-SamBirley (changes requested):

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
MAG-pw: `EventTicket` is a mapping of the DTO for the ticket resource and this is a result class that holds an `EventTicket`. It's ppossible this could be used again in another operation, e.g. `GetTickets` so I don't think we should name it specifically for the operation that uses it. Similarly we call the result for `Create/Update/GetTicketedEvents` `TicketedEventResult`.

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
MAG-pw: My understanding is that for the Submit, we should only expect to see redeemed. I think it would be for another API endpoint that we could see a purchased value but that's not something we have yet. However, it is part of the API that purchased can be a value so we have to cover it. 

My guess is that purchased means we do have knowledge of the ticket but a user has not actually submitted it for redemption yet. 

I think this sort of nuanced documentation probably belongs at source in CHS docs rather than in CSP where we are only wrapping their interpretations which could change over time.

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
MAG-pw: Agreed!

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
MAG-pw: This is following a pattern that we use in many similar places, will take a look and see if it makes sense to do this.

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
MAG-SamBirley: Ok that makes sense, thanks!

Library/include/CSP/Systems/EventTicketing/EventTicketing.h
MAG-SamBirley: I think there is still value in documenting this in the public API to some degree (at your discretion) - because it is valuable information for those seeking to build their own services that interact with CSP. For those people, knowing how CSP expects to interact with services and how it articulates the data it gets back via the API is important context.

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
MAG-SamBirley: Understood, don't be swayed by the established pattern if this approach appears to be more future-proof!

MAG-pw: Looking around at a few other `*DtoTo*` functions that we have, we actually often do not even log missing values. However, I think there is value in logging and setting defaults for values that we expect. I think it makes sense for those defaults to be set here as I don't think this code should assume defaults that are set in the constructor are correct in this situation.

Library/src/CSP/Systems/EventTicketing/EventTicketingSystem.cpp
```diff
@@ -121,6 +121,34 @@ void EventTicketingSystem::GetTicketedEvents(const csp::common::Array<csp::commo
...
+	{
+		RequestOnBehalfOfUserId = *OnBehalfOfUserId;
+	}
+
+	static_cast<chs::TicketedSpaceApi*>(EventTicketingAPI)
+		->apiV1SpacesSpaceIdVendorsVendorNameEventsVendorEventIdTicketsVendorTicketIdPut(SpaceId,
+																						 GetVendorNameString(Vendor),
+																						 VendorEventId,
+																						 VendorTicketId,
+																						 RequestOnBehalfOfUserId,
```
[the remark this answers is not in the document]
MAG-pw: Good point. I'll make sure to get a version of the test for that and run through it.

Library/src/CSP/Systems/EventTicketing/EventTicketing.cpp
MAG-SamBirley: Ok

MAG-SamBirley (approved):

Library/src/CSP/Systems/EventTicketing/EventTicketingSystem.cpp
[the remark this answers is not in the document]
MAG-pw: I've added that test now and it has been run and passed. Thanks for catching it.

PR #107, opened 2023-08-02
[OB-2795] fix: properly allocate and free strings

MAG-MichaelK:
String copies were incorrectly pairing `CSP_NEW` with `DllFree` in the C interface, which was causing OOB errors when freeing strings returned from to any of our wrappers.

**For the reviewer**

### DIFF
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -404,7 +404,7 @@ void CSPFoundation::SetClientUserAgentInfo(const csp::ClientUserAgent& ClientUse
 
 void Free(void* Pointer)
 {
-	csp::memory::DllFree(Pointer);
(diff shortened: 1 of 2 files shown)


MAG-SamBirley:
Tested with a web client that was exhibiting the deadlocking behaviour and verified that it resolves it?

MAG-SamBirley:
Also, please share a build with a web client developer before we go ahead and merge to verify it solves the issue for them.

MAG-mv (approved):

MAG-MichaelK:
I verified this using the web example project we have in our repository, which Tom verified shows the same issues. I would sometimes get a deadlock, and sometimes get an OOB error.

Mag-JB (approved):

MAG-SamBirley:
Ok, happy with that - thanks!

PR #115, opened 2023-08-07
[OB-2804] fix: Handle non-JSON error responses

MAG-pw:
This adds a function to HttpPayload to check if the payload has a JSON mime-type. This is then used when printing error messages to check if an error response is JSON formatted before trying to parse it as JSON. If it is not JSON then it is simply printed as is.

**For the reviewer**

### DIFF
--- a/Library/src/Web/HttpPayload.cpp
+++ b/Library/src/Web/HttpPayload.cpp
@@ -207,4 +207,18 @@ void HttpPayload::SetBoundary(const csp::common::String& InBoundary)
 	Boundary = InBoundary;
 }
 
+const bool HttpPayload::IsJsonPayload() const
+{
+	const auto ContentTypeHeader = Headers.find("content-type");
+	if (ContentTypeHeader == Headers.end())
+	{
+		return false;
+	}
+	else
+	{
+		csp::common::String ContentType = ContentTypeHeader->second.c_str();
+		return ContentType.Split(';')[0] == "application/json";
+	}
+}
+
 } // namespace csp::web
--- a/Library/src/Web/HttpPayload.cpp
+++ b/Library/src/Web/HttpPayload.cpp
@@ -24,6 +24,7 @@
 #include <cstdio>
 #include <rapidjson/stringbuffer.h>
 #include <rapidjson/writer.h>
+#include <regex>
 #include <sstream>
 
 namespace csp::web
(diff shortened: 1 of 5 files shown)


Library/src/Web/HttpPayload.cpp
```diff
@@ -207,4 +207,18 @@ void HttpPayload::SetBoundary(const csp::common::String& InBoundary)
 	Boundary = InBoundary;
 }
 
+const bool HttpPayload::IsJsonPayload() const
+{
+	const auto ContentTypeHeader = Headers.find("content-type");
```
MAG-mv, line 212: Should we add a reference here to prevent a copy?

Library/src/Web/HttpPayload.cpp
```diff
@@ -207,4 +207,18 @@ void HttpPayload::SetBoundary(const csp::common::String& InBoundary)
...
+	if (ContentTypeHeader == Headers.end())
+	{
+		return false;
+	}
+	else
+	{
+		csp::common::String ContentType = ContentTypeHeader->second.c_str();
+		return ContentType.Split(';')[0] == "application/json";
```
MAG-mv, line 220: Do we need an aditional check here to ensure the value is valid before splitting / accessing the value?

Library/src/Web/HttpPayload.h
```diff
@@ -71,6 +71,8 @@ class HttpPayload
 
 	void SetBoundary(const csp::common::String& InBoundary);
 
+	const bool IsJsonPayload() const;
```
MAG-mv, line 74: Returning a const here inst necessary as the return value isn't a pointer / ref

MAG-mv (changes requested):

MAG-SamBirley (approved):

MAG-pw:
Converted back to draft as I've discovered an issue with this.

Library/src/Web/HttpPayload.cpp
MAG-pw: Changed how this is done so resolving.

Library/src/Web/WebClient.cpp
```diff
@@ -380,38 +381,81 @@ void WebClient::DestroyRequest(HttpRequest* Request)
...
+		// the full response will be printed later.
 		if (ResponseJson.HasMember("errors"))
 		{
-			const auto ResponseArray = ResponseJson["errors"].GetObject().FindMember("")->value.GetArray();
+			if (ResponseJson["errors"].IsArray())
+			{
+				const auto& ResponseArray = ResponseJson["errors"].GetArray();
 
-			for (uint32_t i = 0; i < ResponseArray.Size(); i++)
+				for (uint32_t i = 0; i < ResponseArray.Size(); i++)
```
MAG-mv, line 416: This doesn't look like a safe call, we have declared i twice

MAG-mv (changes requested):

MAG-mv (approved):

PR #117, opened 2023-08-07
[OF-1111] feat: GetCheckoutInformation

Mag-JB:

**For the reviewer**

### DIFF
--- a/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
+++ b/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
@@ -73,7 +73,7 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetProductInformationTest)
 			"isEcommerceActive": true,
 			"privateAccessToken": "string"
 		}
-	3. check sopify has synced with your namespace
+	3. check shopify has synced with your namespace
 		Endpoint: /api/v1/vendors/shopify/validate
 		{
 			"storeName": "string",
--- a/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
+++ b/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
@@ -39,8 +39,8 @@ csp::common::Map<csp::common::String, csp::common::String> GetShopifyDetails()
 {
 	if (!std::filesystem::exists("ShopifyCreds.txt"))
 	{
-		LogFatal("ShopifyCreds.txt not found! This file must exist and must contain the following information:\nSpaceId "
-				 "<SpaceId>\nProductId <ProductId>");
+		LogFatal("ShopifyCreds.txt not found! This file must exist and must contain a minimum of the following information:\nSpaceId "
+				 "<SpaceId>\nProductId <ProductId> but may also need \nProductId <ProductId>\nCartId <CartId>\n");
 	}
 
 	csp::common::Map<csp::common::String, csp::common::String> OutMap;
--- a/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
+++ b/Tests/src/PublicAPITests/ECommerceSystemTests.cpp
@@ -153,7 +153,7 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetCheckoutInformationTest)
 	const csp::common::String FalseSpaceId = "abcdefghijk1234567891011";
 	const csp::common::String FalseCartId  = "B1-1234567891011121314151617e8e21er";
 
-	// The additional info inside of this mask need to be added to the ShopifyCred.txt file on a new line as: "Key Value"
+	// The additional info such as "CartId" inside of this test need to be added to the ShopifyCred.txt file on a new line as: <Key> <Value>
 	auto [Result] = AWAIT_PRE(ECommerceSystem, GetCheckoutInformation, RequestPredicate, Details["SpaceId"], Details["CartId"]);
 	EXPECT_EQ(Result.GetResultCode(), csp::services::EResultCode::Success);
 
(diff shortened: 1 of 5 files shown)


MAG-mv (dismissed):

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
```diff
@@ -136,4 +136,31 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetProductInformationTest)
 
 	LogOut(UserSystem);
 }
-#endif
\ No newline at end of file
+#endif
+
+#if RUN_ECOMMERCE_TESTS || RUN_ECOMMERCE_GET_CHECKOUT_INFORMATION_TEST
+CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetCheckoutInformationTest)
```
MAG-SamBirley, line 142: Can we also test unhappy paths here? Like what happens if an invalid space ID is passed in, other invalid parameters etc

MAG-SamBirley (changes requested):

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
Mag-JB: sorted

MAG-SamBirley (approved):

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
```diff
@@ -136,4 +136,45 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetProductInformationTest)
...
+	auto* UserSystem	  = SystemsManager.GetUserSystem();
+	auto* ECommerceSystem = SystemsManager.GetECommerceSystem();
+
+	csp::common::String UserId;
+	LogIn(UserSystem, UserId);
+	auto Details						   = GetShopifyDetails();
+	const csp::common::String FalseSpaceId = "abcdefghijk1234567891011";
+	const csp::common::String FalseCartId  = "B1-1234567891011121314151617e8e21er";
+
+	// The additional info inside of this mask need to be added to the ShopifyCred.txt file on a new line as: "Key Value"
```
MAG-pw, line 156: Can you add a little more details about what this comment is actually saying. I don't really get it.

MAG-mv (approved):

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
Mag-JB: sorry typo, should be sorted

MAG-SamBirley (approved):

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
```diff
@@ -136,4 +136,45 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetProductInformationTest)
 
 	LogOut(UserSystem);
 }
-#endif
\ No newline at end of file
+#endif
+
+#if RUN_ECOMMERCE_TESTS || RUN_ECOMMERCE_GET_CHECKOUT_INFORMATION_TEST
+CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetCheckoutInformationTest)
```
MAG-pw, line 142: Since this is a manual test, can we add some documentation to this to explain what needs to be done before this test can be run. This could also be done for the test above also. There is a lot of setup that needs to be done that I'm currently struggling with using the internal google doc but that is not available externally. We should also note that doing the setup for the tests requires swagger usage and we currently have swagger behind the VPN so not everyone will be able to run these tests. 

Unrelated but I think that the non-happy path stuff could exist in it's own test since that could run without the manual intervention.

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
```diff
@@ -136,4 +136,45 @@ CSP_PUBLIC_TEST(CSPEngine, ECommerceSystemTests, GetProductInformationTest)
...
+	auto* UserSystem	  = SystemsManager.GetUserSystem();
+	auto* ECommerceSystem = SystemsManager.GetECommerceSystem();
+
+	csp::common::String UserId;
+	LogIn(UserSystem, UserId);
+	auto Details						   = GetShopifyDetails();
+	const csp::common::String FalseSpaceId = "abcdefghijk1234567891011";
+	const csp::common::String FalseCartId  = "B1-1234567891011121314151617e8e21er";
+
+	// The additional info such as "CartId" inside of this test need to be added to the ShopifyCred.txt file on a new line as: <Key> <Value>
```
MAG-pw, line 156: I think if the format of the `ShopifyCreds.txt` file has changed we should update the error message we print above.

Mag-JB: sorted

Tests/src/PublicAPITests/ECommerceSystemTests.cpp
```diff
@@ -56,18 +56,37 @@ csp::common::Map<csp::common::String, csp::common::String> GetShopifyDetails()
...
+
+	1. Create a space (Add to shopify Creds)
+	2. Connected your shopify.dev account to your space using the "Private Access Token" and store name
+		Endpoint : /api/v1/spaces/{spaceId}/vendors/shopify
+		{
+			"storeName": "string",
+			"isEcommerceActive": true,
+			"privateAccessToken": "string"
+		}
+	3. check sopify has synced with your namespace
```
MAG-pw, line 76: typo: Sopify -> Shopify

MAG-pw (dismissed):
One tiny typo but otherwise looks like all is addressed.

MAG-pw (approved):

MAG-SamBirley (approved):

PR #122, opened 2023-08-10
[OF-1108] feat: Adds ECommerce Create/Get Cart

MAG-pw:

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/ECommerce/ECommerce.h
+++ b/Library/include/CSP/Systems/ECommerce/ECommerce.h
@@ -228,7 +228,7 @@ private:
 
 
 /// @ingroup ECommerce System
-/// @brief Data class used to contain information when attempting to get Product Info.
+/// @brief Data class used to contain information when attempting to get a Cart.
 class CSP_API CartInfoResult : public csp::services::ResultBase
 {
 	/** @cond DO_NOT_DOCUMENT */
--- a/Library/include/CSP/Systems/ECommerce/ECommerce.h
+++ b/Library/include/CSP/Systems/ECommerce/ECommerce.h
@@ -136,6 +136,43 @@ public:
 };
 
 /// @ingroup ECommerce System
+/// @brief Represents a line in a cart.
+class CSP_API CartLine
+{
+public:
+	CartLine() : Quantity(0) {};
+
+	/// @brief ID of the line in the cart.
+	csp::common::String CartLineId;
+
+	/// @brief ID of the variant of the product.
+	csp::common::String ProductVariantId;
+
+	/// @brief Quantity of the product in the cart.
+	int Quantity;
+};
+
+/// @ingroup ECommerce System
+/// @brief Represents a cart.
+class CSP_API CartInfo
+{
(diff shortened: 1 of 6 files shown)


Mag-JB (approved):

Library/include/CSP/Systems/ECommerce/ECommerce.h
```diff
@@ -189,8 +226,38 @@ class CSP_API CheckoutInfoResult : public csp::services::ResultBase
 	CheckoutInfo CheckoutInformation;
 };
 
+
+/// @ingroup ECommerce System
+/// @brief Data class used to contain information when attempting to get Product Info.
```
MAG-SamBirley, line 231: Is this comment right? Feels like it was copy/pasted from elsewhere?

Library/src/Systems/ECommerce/ECommerce.cpp
```diff
@@ -113,6 +115,84 @@ void ProductInfoDtoToProductInfo(const chs_aggregation::ShopifyProductDto& Dto,
...
+		Cart.SpaceId = CartDto.GetSpaceId();
+	}
+	else
+	{
+		FOUNDATION_LOG_ERROR_MSG("ShopifyCartDto missing SpaceId");
+	}
+
+	if (CartDto.HasShopifyCartId())
+	{
+		// There is a prefix added to the ID within the CHS system but it is not needed for our purposes.
```
MAG-SamBirley, line 131: Please can we frame this in a way that is services-agnostic. Hopefully others will build services that can talk to CSP.
Something like, "We expect services may add a prefix to the ID, but it is not needed for our purposes". That will be more helpful

Library/src/Web/WebClient.cpp
```diff
@@ -389,9 +389,32 @@ void WebClient::PrintClientErrorResponseMessages(const HttpResponse& Response)
 	const uint16_t ResponseCode				   = static_cast<uint16_t>(Response.GetResponseCode());
 	const csp::common::String& ResponsePayload = Response.GetPayload().GetContent();
 
+	csp::common::String Verb = "";
+	switch (Response.GetRequest()->GetVerb())
```
MAG-SamBirley, line 393: Love this

MAG-SamBirley (changes requested):

Library/include/CSP/Systems/ECommerce/ECommerce.h
MAG-pw: It was! 🤦 Will fix.

Library/src/Systems/ECommerce/ECommerce.cpp
MAG-pw: I don't see this as services-agnostic. The bit we are stripping out is specific to Magnopus Services and leaves only the raw shopify cart. I think it would be good to update this comment to reflect that and make it clearer.

Mag-JB (approved):

MAG-SamBirley (approved):

PR #147, opened 2023-09-01
[OF-1027] feat!: Update ResetUserPassword to token-change-password endpoint

Mag-JB:

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -162,7 +162,7 @@ public:
 	CSP_ASYNC_RESULT void ConfirmUserEmail(NullResultCallback Callback);
 
 	/// @brief Reset the users password.
-	/// @param Token csp::common::String : Bearer Token of the user changing their password
+	/// @param Token csp::common::String : Token received through email by user
 	/// @param NewPassword csp::common::String : The new password for the associated account
 	/// @param Callback NullResultCallback : callback to call when a response is received
 	CSP_ASYNC_RESULT void ResetUserPassword(const csp::common::String& Token, const csp::common::String& NewPassword, NullResultCallback Callback);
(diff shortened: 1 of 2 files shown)


Mag-JB:
This has been manually tested as integration tests are not possible

MAG-SamBirley:
@MAG-DanielMurdolo with this change, we're removing the parameter that lets client code specify the redirect URL associated with a reset password invocation. Wanted to check that is expected from your side?

MAG-SamBirley (dismissed):
Great work, thanks!

MAG-AdrianMeredith:
I can confirm that the Web Client doesn't currently use this function and as I understand it, not redirect url should be necessary.  Should be good our side

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -162,8 +162,10 @@ class CSP_API CSP_NO_DISPOSE UserSystem : public SystemBase
 	CSP_ASYNC_RESULT void ConfirmUserEmail(NullResultCallback Callback);
 
 	/// @brief Reset the users password.
+	/// @param Token csp::common::String : Bearer Token of the user changing their password
```
MAG-pw, line 165: I think of bearer token as the OAuth access token in the headers. Is that what this token is or is it something else?

MAG-pw (commented):
Looks generally ok, just one quick question on the comments.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-SamBirley: Good catch. According to swagger this is actually a token that is received via an automated email that is sent

MAG-pw: I thought that might be what this was. I think we should update the comment to make that more clear before shipping this.

Mag-JB: sorted

MAG-SamBirley (approved):

MAG-pw (approved):

PR #153, opened 2023-09-06
[OF-1096] test: added tests for common Map

MAG-AW:

**For the reviewer**


Tests/Map.cpp
```diff
@@ -0,0 +1,84 @@
...
+public:
+	MapTest()
+	{
+		MyMap[1] = "One";
+		MyMap[2] = "Two";
+		MyMap[3] = "Three";
+	};
+
+	Map<int, csp::common::String> MyMap;
+};
```
MAG-MichaelK, line 40: We should try to adhere to the same testing style in all of our tests. Other csp::common tests create the test object locally. Relying on a class having the test object as a field changes the lifetime and behaviour of said test object.

Tests/Map.cpp
```diff
@@ -0,0 +1,84 @@
...
+	auto MapTestLocal = MapTest();
+
+	// Update an element in the map
+	MapTestLocal.MyMap[2] = "NewTwo";
+
+	// Check if the updated element has the correct value
+	EXPECT_EQ(MapTestLocal.MyMap[2], "NewTwo");
+}
+
+#endif
```
MAG-MichaelK, line 84: As these are meant to be unit tests, we should be testing every public function of Map with various value types. This means constructors, operator overloads, copy assignment, etc. You can refer to the existing tests for Array, Optional, String, etc.

MAG-MichaelK (changes requested):

Tests/Map.cpp
```diff
@@ -0,0 +1,123 @@
...
+// Test case to check if elements are initialised with a list correctly
+CSP_INTERNAL_TEST(CSPEngine, CommonMapTests, MapInitialiseListTest)
+{
+	Map<int, csp::common::String> MyMap;
+	MyMap = {{1, "One"}, {2, "Two"}, {3, "Three"}};
+
+	// Check if the map contains the initial values
+	EXPECT_TRUE(MyMap.HasKey(1));
+	EXPECT_TRUE(MyMap.HasKey(2));
+	EXPECT_TRUE(MyMap.HasKey(3));
```
MAG-MichaelK, line 53: It's worth checking the values too.

Tests/Map.cpp
MAG-MichaelK: I see that the copy constructor is being tested, but not the move constructor. Copy and move assignment are also missing, as are tests for Keys(), Values(), and Clear().

Tests/Map.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+	Map<int, csp::common::String> OldMap;
+	OldMap = {{1, "One"}, {2, "Two"}, {3, "Three"}};
+
+	Map<int, String> MyMap(OldMap);
+
+	// Check if the map contains the initial values
+	EXPECT_TRUE(MyMap.HasKey(1));
+	EXPECT_TRUE(MyMap.HasKey(2));
+	EXPECT_TRUE(MyMap.HasKey(3));
```
MAG-MichaelK, line 70: It's worth checking if Size() returns what we expect here.

MAG-MichaelK (changes requested):

Tests/src/InternalTests/CommonTypeTests/Map.cpp
```diff
@@ -0,0 +1,193 @@
...
+	EXPECT_FALSE(MyMap.Values()->Data() == nullptr);
+
+	MyMap.Clear();
+
+	EXPECT_EQ(MyMap.Size(), 0);
+	EXPECT_EQ(MyMap.Keys()->Size(), 0);
+	EXPECT_TRUE(MyMap.Keys()->Data() == nullptr);
+	EXPECT_EQ(MyMap.Values()->Size(), 0);
+	EXPECT_TRUE(MyMap.Values()->Data() == nullptr);
+}
```
MAG-MichaelK, line 191: This should be split into 3 separate tests. We want to keep our unit tests as small as possible and only test a single unit at a time.

MAG-MichaelK (changes requested):

Tests/src/InternalTests/CommonTypeTests/Map.cpp
```diff
@@ -0,0 +1,228 @@
...
+	EXPECT_FALSE(MyMap.Keys()->Data() == nullptr);
+}
+
+CSP_INTERNAL_TEST(CSPEngine, CommonMapTests, MapValuesTest)
+{
+	Map<int, String> MyMap;
+	MyMap = {{1, "One"}, {2, "Two"}, {3, "Three"}};
+
+	EXPECT_EQ(MyMap.Values()->Size(), 3);
+	EXPECT_FALSE(MyMap.Values()->Data() == nullptr);
```
MAG-MichaelK, line 213: `Keys()` and `Values()` returns a newly-allocated array, so we should store and delete it.

MAG-MichaelK (changes requested):

MAG-MichaelK (approved):

MAG-SamBirley (approved):

PR #161, opened 2023-09-11
[OF-1145] feat: Expose some Logging APIs

MAG-pw:
This exposes the get and set for logging level for the wrapper generator.

Additionally it adds the ability to log a test message so clients have something they can use to test logging.

**For the reviewer**

### DIFF
--- a/Library/src/Systems/ECommerce/ECommerce.cpp
+++ b/Library/src/Systems/ECommerce/ECommerce.cpp
@@ -47,15 +47,32 @@ void ProductInfoDtoToProductInfo(const chs_aggregation::ShopifyProductDto& Dto, 
 			ProductInfo.Variants[i].AvailableForSale = VariantProductInformation[i]->GetAvailableForSale();
 			if (VariantProductInformation[i]->HasImage())
 			{
-				if (VariantProductInformation[i]->GetImage()->HasMediaContentType())
+				auto VariantProductImage = VariantProductInformation[i]->GetImage();
+
+				if (VariantProductImage->HasMediaContentType())
 				{
-					ProductInfo.Variants[i].Media.MediaContentType = VariantProductInformation[i]->GetImage()->GetMediaContentType();
+					ProductInfo.Variants[i].Media.MediaContentType = VariantProductImage->GetMediaContentType();
 				}
 
-				ProductInfo.Variants[i].Media.Alt	 = VariantProductInformation[i]->GetImage()->GetAlt();
-				ProductInfo.Variants[i].Media.Url	 = VariantProductInformation[i]->GetImage()->GetUrl();
-				ProductInfo.Variants[i].Media.Width	 = VariantProductInformation[i]->GetImage()->GetWidth();
-				ProductInfo.Variants[i].Media.Height = VariantProductInformation[i]->GetImage()->GetHeight();
+				if (VariantProductImage->HasAlt())
+				{
+					ProductInfo.Variants[i].Media.Alt = VariantProductImage->GetAlt();
+				}
+
+				if (VariantProductImage->HasUrl())
+				{
+					ProductInfo.Variants[i].Media.Url = VariantProductImage->GetUrl();
+				}
+
+				if (VariantProductImage->HasWidth())
+				{
+					ProductInfo.Variants[i].Media.Width = VariantProductImage->GetWidth();
+				}
+
+				if (VariantProductImage->HasHeight())
+				{
+					ProductInfo.Variants[i].Media.Height = VariantProductImage->GetHeight();
+				}
 			}
 
 			if (Dto.GetVariants()[i]->HasSelectedOptions())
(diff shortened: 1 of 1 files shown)


Library/include/CSP/Systems/Log/LogSystem.h
```diff
@@ -71,19 +71,27 @@ class CSP_API CSP_NO_DISPOSE LogSystem
...
 
 	/// @brief Check if we currently log a specified log verbosity level.
-	CSP_NO_EXPORT bool LoggingEnabled(const csp::systems::LogLevel Level);
+	/// @param Level The level to check.
+	bool LoggingEnabled(const csp::systems::LogLevel Level);
 
 	/// @brief Log a message at a specific verbosity level.
+	/// @param Level The level to log this message at.
+	/// @param InMessage The message to be logged.
 	CSP_NO_EXPORT void LogMsg(const csp::systems::LogLevel Level, const csp::common::String& InMessage);
```
MAG-SamBirley, line 87: Any reason to keep this as NOEXPORT?

Library/include/CSP/Systems/Log/LogSystem.h
```diff
@@ -71,19 +71,27 @@ class CSP_API CSP_NO_DISPOSE LogSystem
...
 	/// @brief Log an event.
+	/// @param InEvent The event to be logged.
 	CSP_NO_EXPORT void LogEvent(const csp::common::String& InEvent);
```
MAG-SamBirley, line 90: Any reason to keep this as NOEXPORT?

Library/include/CSP/Systems/Log/LogSystem.h
```diff
@@ -71,19 +71,27 @@ class CSP_API CSP_NO_DISPOSE LogSystem
...
 
+	/// @brief Logs a test message at the set system verbosity level.
+	void LogTestMessage();
+
 	/// @brief Specify a 'Marker' event which can be used to communicate a certain process occurring, usually for debugging.
 	CSP_NO_EXPORT void BeginMarker(const csp::common::String& InMarker);
```
MAG-SamBirley, line 96: Any reason to keep this as NOEXPORT?

Library/include/CSP/Systems/Log/LogSystem.h
MAG-pw: For this and `LogEvent` it seems like it would be a significant change to do as part of this ticket as it would open up the log system for clients to use for their own logging too. I don't really have an opinion on whether that's good/bad/otherwise, just that it's a fair bit beyond the scope of this ticket.

Library/include/CSP/Systems/Log/LogSystem.h
MAG-pw: See comment above on `LogMsg`

Library/include/CSP/Systems/Log/LogSystem.h
MAG-pw: See comment above on `LogMsg`

Library/include/CSP/Systems/Log/LogSystem.h
MAG-SamBirley: Ok

MAG-SamBirley (approved):

Mag-JB (approved):

Library/include/CSP/Systems/Log/LogSystem.h
MAG-MichaelK: These functions are already exposed to anything that uses the C++ DLL directly, like Unreal

Library/src/Systems/Log/LogSystem.cpp
```diff
@@ -150,4 +150,9 @@ void LogSystem::ClearAllCallbacks()
 	Callbacks->Clear();
 }
 
+void LogSystem::LogTestMessage()
```
MAG-MichaelK, line 153: We should keep testing code inside the test project and not include it in CSP itself.

Tests/Web/html/tests/logsystem_tests.js
```diff
@@ -0,0 +1,58 @@
+import { test, assert } from '../test_framework.js';
+
+import { Systems } from '../connected_spaces_platform.js';
+
+function logCallback(message) {
```
MAG-MichaelK, line 5: This function is never used

MAG-MichaelK (changes requested):

Tests/Web/html/tests/logsystem_tests.js
MAG-pw: Ah yeah, I changed to the below way of doing it to include the `messageLogged` boolean. Will remove

Library/src/Systems/Log/LogSystem.cpp
MAG-pw: I was more thinking of this as a way even clients could specifically trigger a test message rather than just our own testing since they would otherwise be waiting for some other log message to be able to verify. If we remove the no export on the `LogMsg` function then this would become unnecessary anyways.

Library/include/CSP/Systems/Log/LogSystem.h
MAG-pw: Should we remove all `CSP_NO_EXPORT` from this so? It's quite a change to the scope of the ticket but if we're happy to do so then I'm happy make the change. @MAG-SamBirley, happy to update the ticket?

MAG-SamBirley: I'm ok treating it as out of scope for this change, but @MAG-pw could you create another one for us to get rid of the other NOEXPORTs?

Library/src/Systems/Log/LogSystem.cpp
MAG-MichaelK: I think that using some other function that we know sends a log message would be a better idea here, as `LogTestMessage` is unlike any other code we have in CSP and is only ever probably going to be used by us. Having a public function for testing functionality feels wrong.

MAG-pw: It feels like it would be an ugly way of testing the log system to have a dependency on calling something else that we know, for now, logs a message. We couldn't be sure that in the future it would still do so.

MAG-pw: Would it help if I renamed this to something like `LogDefaultMessage`?

MAG-pw: Since the actual logging functionality is outside the scope of this ticket I'm going to just remove this and the test that I introduced.

Library/include/CSP/Systems/Log/LogSystem.h
MAG-pw: Will do.

MAG-pw: @MAG-SamBirley ticket created an assigned to you for prioritization.

MAG-SamBirley (approved):

PR #194, opened 2023-09-27
[OF-1119] feat: add age verification flag

MAG-MichaelK:
This change adds `HasVerifiedAge` to `CreateUser` and `Login` functions. It also adds a way of getting back errors returned by CHS via the `X-ErrorCode` response header.

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/Users/Authentication.h
+++ b/Library/include/CSP/Systems/Users/Authentication.h
@@ -97,9 +97,10 @@ public:
 };
 
 
-enum ELoginStateResultFailureReason
+enum class ELoginStateResultFailureReason
 {
-	None,
+	Unknown = -1,
+	None	= 0,
 	AgeNotVerified
 };
 
--- a/Library/include/CSP/Systems/Users/Authentication.h
+++ b/Library/include/CSP/Systems/Users/Authentication.h
@@ -97,6 +97,13 @@ public:
 };
 
 
+enum ELoginStateResultFailureReason
+{
+	None,
+	AgeNotVerified
+};
+
+
 /// @brief Result structure for a login state request.
 class CSP_API LoginStateResult : public csp::services::ResultBase
 {
@@ -111,13 +118,15 @@ public:
 	LoginState& GetLoginState();
 	const LoginState& GetLoginState() const;
 
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
+
 private:
 	LoginStateResult();
 	LoginStateResult(LoginState* InStatePtr);
 
 	void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
 
-private:
 	LoginState* State;
 };
 
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -62,10 +62,12 @@ public:
 	/// @param UserName csp::common::String
 	/// @param Email csp::common::String
 	/// @param Password csp::common::String
+	/// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
 	/// @param Callback LoginStateResultCallback : callback to call when a response is received
 	CSP_ASYNC_RESULT void Login(const csp::common::String& UserName,
 								const csp::common::String& Email,
 								const csp::common::String& Password,
+								const csp::common::Optional<bool>& UserHasVerifiedAge,
 								LoginStateResultCallback Callback);
 
 	/// @brief Log in to Magnopus Connected Services using a login token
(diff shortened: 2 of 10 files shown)


MAG-MGN:
I see we changed Login and CreateUser, but how do we plan to update an existing user? 
If he didn't say he's over 18 yet, he should have the opportunity to check a checkbox in the client, which would set his verified age.

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -97,6 +97,13 @@ class CSP_API LoginTokenInfo
 };
 
 
+enum ELoginStateResultFailureReason
+{
+	None,
```
lyubomirlichev, line 102: If `None` is an option for a failure reason then perhaps it should be instead 'Undefined' or "Unknown" ? If we get this on a client level, it would be interpreted as "it failed but there is no reason for the fail" which doesn't have enough info to explain itself...

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -62,10 +62,12 @@ class CSP_API CSP_NO_DISPOSE UserSystem : public SystemBase
 	/// @param UserName csp::common::String
 	/// @param Email csp::common::String
 	/// @param Password csp::common::String
+	/// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
 	/// @param Callback LoginStateResultCallback : callback to call when a response is received
 	CSP_ASYNC_RESULT void Login(const csp::common::String& UserName,
 								const csp::common::String& Email,
 								const csp::common::String& Password,
+								const csp::common::Optional<bool>& UserHasVerifiedAge,
```
lyubomirlichev, line 70: Considering the age should be explicitly 18 and over, it would be nice to have it captured it as part of the variable name.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -141,6 +140,7 @@ class CSP_API CSP_NO_DISPOSE UserSystem : public SystemBase
 									 const csp::common::String& Email,
 									 const csp::common::String& Password,
 									 bool ReceiveNewsletter,
+									 bool HasVerifiedAge,
```
lyubomirlichev, line 143: Should there be an xml comment to describe this parameter as well ?

Library/include/CSP/Systems/Users/Authentication.h
MAG-MichaelK: Would you be OK with an `Unknown` member with a value of -1 being added instead? `None` in this context means no `X-ErrorCode` header was present.

Tests/CSharp/src/ExtensionMethods.cs
```diff
@@ -24,13 +24,10 @@ public static void TestLogOut(this Systems.UserSystem userSystem)
 
         public static string TestLogIn(this Systems.UserSystem userSystem, string email = null, string password = null, Services.EResultCode expectedResult = Services.EResultCode.Success, bool pushCleanupFunction = true)
         {
-            if (email == null)
-                email = UserSystemTests.DefaultLoginEmail;
+            email ??= UserSystemTests.DefaultLoginEmail;
+            password ??= UserSystemTests.DefaultLoginPassword;
```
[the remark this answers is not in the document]
MAG-MichaelK: These are pretty common in modern C# code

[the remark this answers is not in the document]
MAG-MichaelK: Also, these tests will be going away.

Library/include/CSP/Systems/Users/UserSystem.h
lyubomirlichev: If that's the case, perhaps we could just match the comment with the variable name ? As the comment suggests 18

MAG-MichaelK: I deliberately chose not mention a specific age here, as the plan is for us to eventually allow 13+.

Library/include/CSP/Systems/Users/Authentication.h
lyubomirlichev: Yes i think that's a bit better , thanks 🙂

Library/include/CSP/Systems/Users/UserSystem.h
MAG-MichaelK: I must have missed that one

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -104,17 +107,13 @@ class CSP_API CSP_NO_DISPOSE UserSystem : public SystemBase
...
 	CSP_ASYNC_RESULT void LoginToThirdPartyAuthenticationProvider(const csp::common::String& ThirdPartyToken,
 																  const csp::common::String& ThirdPartyStateId,
+																  const csp::common::Optional<bool>& UserHasVerifiedAge,
 																  LoginStateResultCallback Callback);
 
-	/// @brief Log in to Magnopus Connected Services services as a guest.
-	/// @param DeviceId csp::common::String : The device Id to use when logging in
-	/// @param Callback LoginStateResultCallback : callback to call when a response is received
-	[[deprecated("This is no longer needed as we now have proper device ID generation! Please use LoginAsGuest() instead.")]] CSP_ASYNC_RESULT void
-		LoginAsGuestWithId(const csp::common::String& DeviceId, LoginStateResultCallback Callback);
```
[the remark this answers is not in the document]
MAG-MichaelK: I chose to remove it here instead of updating this function to pass the new boolean flag. It has been deprecated for a long time now, and no clients seem to still use it, although the wrapper that the Unity team does use uses it. I can modify the merge commit to include the breaking change marker.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-SamBirley: Unless we feel strongly otherwise, let's keep it general for now as it does give us latitude in the future.

lyubomirlichev: Happy with either to be honest, up to you. One small note for specificity would be that the CHS endpoint variable name is verifiedAgeEighteen based on their integration doc. This value aims to wrap that in a sense, so it would make sense to match its name instead of generalising it. Either way, it feels like the xml comment should match the value on whether '18>' is mentioned or not.

MAG-SamBirley: Agreed on that last part.

MAG-DanielMurdolo (dismissed):

MAG-DanielMurdolo (approved):

PR #205, opened 2023-10-03
[NT-0] refac: ensure platform define always set

MAG-MichaelK:
**For the reviewer**

### DIFF
--- a/Library/include/CSP/CSPCommon.h
+++ b/Library/include/CSP/CSPCommon.h
@@ -18,9 +18,9 @@
 // Enable this define if you are using the CSP library in DLL form
 // #define USING_CSP_DLL
 
-#if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


MAG-AdamThorn (dismissed):

MAG-SamBirley (approved):
This is neat! This will however break C++ clients (like our internal Unreal client) where they have these already defined. 

If you agree, please can we telegraph this as a breaking change?

cc @MAG-AdamThorn

MAG-MichaelK:
Ah I should have caught that. I'll add protection against that by using something like `ifndef CSP_WINDOWS def CSP_WINDOWS`

MAG-SamBirley (changes requested):
Setting review state to `request changes` to avoid this getting accidentally merged.

MAG-SamBirley (approved):

PR #220, opened 2023-10-13
[NT-0] refac: Root directory hygiene

MAG-SamBirley:
In an effort to improve our root directory hygiene, this change moves several markdowns to the .github directory (which GH will still pick up as it did before), moves one pipeline file to the `teamcity` directory, and removes one legacy sonar config file that is no longer in use.


Mag-JB (approved):

MAG-MichaelK:
Why was requirements.txt moved to the teamcity folder? It's unrelated to TC.

PR #221, opened 2023-10-17
[OW-1437] feat: Add support for automatic garbage collection and rejecting failed http requests

MAG-AdrianMeredith:
[OW-1437] feat: Add support for automatic garbage collection and rejecting failed http requests

- Added support for the client side javascript automatically deleting the csp object when garbage collected
- Ensured that failed http requests reject the returned promise

**For the reviewer**

[OW-1437]: https://magnopus.atlassian.net/browse/OW-1437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tools/WrapperGenerator/Templates/TypeScript/Partials/Class.mustache
+++ b/Tools/WrapperGenerator/Templates/TypeScript/Partials/Class.mustache
@@ -33,6 +33,30 @@ export class {{ name }}
     /** @internal */
     constructor(pointer: NativePointer) {
         super(pointer);
+        const ptr = pointer.pointer;
+        const ownsPointer = this.ownsPointer;
+        {{# should_dispose }}
+        // Free WASM pointer when this object is garbage collected
+        registry.register(this, {
+            fn: (dFn, p, ownsPointer) => {
+                if (ownsPointer) {
+                    const systemsManager = Systems.SystemsManager.get();
+                    const logSystem = systemsManager.getLogSystem();
+                    if (logSystem.loggingEnabled(Systems.LogLevel.Verbose)) {
+                        console.log("Deleting unreferences object",dFn, p);
+                    }
+                    try {
+                        Module.ccall(dFn, "void", ["number"], [p]);
+                    } catch(e) {
+                        // Nothing to do
+                    }
+                }
+            },
+            destructorName: "csp_{{# namespace_underscore }}{{namespace_underscore}}_{{/ namespace_underscore }}{{name}}_Dtor",
+            ptr,
+            ownsPointer
+        })
+        {{/ should_dispose }}
     }
 
 {{# has_base_type }}
--- a/Tools/WrapperGenerator/Templates/TypeScript/Partials/Method/TaskMethod.mustache
+++ b/Tools/WrapperGenerator/Templates/TypeScript/Partials/Method/TaskMethod.mustache
@@ -11,12 +11,19 @@ async {{ name }}(
 {{/ is_async_result_with_progress }}
 ): Promise<{{# results }}{{# type }}{{> TypeName }}{{/ type }}{{/ results }}> {
     var _resolve;
-
-    var _promise = new Promise<{{# results }}{{# type }}{{> TypeName }}{{/ type }}{{/ results }}>(
-        (_r) => {
-            _resolve = _r
+    {{# results }}
+    {{# type.is_result_base }}
+    var _reject;
+    {{/ type.is_result_base }}
+    var _promise = new Promise<{{# type }}{{> TypeName }}{{/ type }}>(
+        (_r, {{# type.is_result_base }}_rj{{/ type.is_result_base }}) => {
+            _resolve = _r;
+            {{# type.is_result_base }}
+            _reject = _rj;
+            {{/ type.is_result_base }}
         }
     );
+    {{/ results }}
 
     var _callbackPtr: number;
     var _callback = (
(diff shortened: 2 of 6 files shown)


Tools/WrapperGenerator/Templates/TypeScript/Partials/Class.mustache
```diff
@@ -33,6 +33,30 @@ export class {{ name }}
...
+                        console.log("Deleting unreferences object",dFn, p);
+                    }
+                    try {
+                        Module.ccall(dFn, "void", ["number"], [p]);
+                    } catch(e) {
+                        // Nothing to do
+                    }
+                }
+            },
+            destructorName: "csp_{{# namespace_underscore }}{{namespace_underscore}}_{{/ namespace_underscore }}{{name}}_Dtor",
```
MAG-MichaelK, line 55: This new `namespace_underscore` property is unnecessary, as each type has a `full_safe_type_name` property that is used when generating C interface function names.

Tools/WrapperGenerator/Templates/TypeScript/Partials/Class.mustache
```diff
@@ -33,6 +33,30 @@ export class {{ name }}
...
+        const ptr = pointer.pointer;
+        const ownsPointer = this.ownsPointer;
+        {{# should_dispose }}
+        // Free WASM pointer when this object is garbage collected
+        registry.register(this, {
+            fn: (dFn, p, ownsPointer) => {
+                if (ownsPointer) {
+                    const systemsManager = Systems.SystemsManager.get();
+                    const logSystem = systemsManager.getLogSystem();
+                    if (logSystem.loggingEnabled(Systems.LogLevel.Verbose)) {
```
MAG-MichaelK, line 45: We shouldn't be directly calling console.log like this in the generated code, nor manually checking the current log level. Instead, call the `LogSystem` functions.

Tools/WrapperGenerator/Templates/TypeScript/Partials/Method/TaskMethod.mustache
```diff
@@ -61,24 +68,20 @@ async {{ name }}(
...
-        )
+{{# type.is_result_base }}
+    if ({{> ResultName }}.getResultCode() == Services.EResultCode.Success) {
+        
+{{/ type.is_result_base }}
+        _resolve({{> ResultName }})
+        
+{{# type.is_result_base }}
+    } else {
+        _reject(new Error({{> ResultName }}.getResponseBody()));
```
MAG-MichaelK, line 80: This does not return enough information to the clients. We expose not only the response body, but the HTTP error code and the parsed X-ErrorCode value. This also prevents result objects from being manually disposed as they are not returned to the caller.

Tools/WrapperGenerator/Templates/TypeScript/Partials/Method/TaskMethod.mustache
```diff
@@ -11,12 +11,19 @@ async {{ name }}(
 {{/ is_async_result_with_progress }}
 ): Promise<{{# results }}{{# type }}{{> TypeName }}{{/ type }}{{/ results }}> {
     var _resolve;
-
-    var _promise = new Promise<{{# results }}{{# type }}{{> TypeName }}{{/ type }}{{/ results }}>(
-        (_r) => {
-            _resolve = _r
+    {{# results }}
+    {{# type.is_result_base }}
+    var _reject;
```
MAG-MichaelK, line 16: Why would we only reject the promise for ResultBase-derived types?

Tools/WrapperGenerator/Templates/TypeScript/Partials/ResultName.mustache
```diff
@@ -0,0 +1,12 @@
+
```
MAG-MichaelK, line 1: This partial looks to only be used in a single place, so it's wasteful to have it in a separate file.

Tools/WrapperGenerator/TypeScriptWrapperGenerator.py
```diff
@@ -49,6 +49,21 @@ def __translate_namespace(self, obj: Any) -> None:
 
         setattr(obj, "translated_namespace", ".".join(filter(None, namespaces)))
 
+    def __namespace_underscore(self, obj: Any) -> None:
```
MAG-MichaelK, line 52: Not needed.

MAG-MichaelK:
It's worth noting that we are planning to migrate fully to Jinja soon(ish), so heavy changes to Mustache cause pain and grief, especially on my soul.

MAG-AdrianMeredith:
@MAG-MichaelK This PR is basically a Proof of concept rather than a something I expect to be merged.  It shows something very close to what we're looking for

Tools/WrapperGenerator/Templates/TypeScript/Partials/Method/TaskMethod.mustache
MAG-AdrianMeredith: Ideally what we need here is an simple javascript interface for Errors.  Ideally with an agreed error code type and message.  http code should be redundant and not exposed to the client (as it is internal implementation). The error code should cover the type of error already.
{
errorCode: 3,
message: "Resource not found"
}

MAG-SamBirley:
Having discussed with @MAG-AdrianMeredith, the plan to solve the problem of automated memory management for JS in CSP involves adding support for JS smart pointers, which will be done after we have migrated the CSP templating engine from Mustache to Jinja2.

Closing this one out 👍

PR #227, opened 2023-10-20
[OB-2823] fix: add checks for codes to retries

MAG-AW:
Currently allowed codes for retry:
429
408
5xx

**For the reviewer**

### DIFF
--- a/Library/src/Web/HttpRequest.cpp
+++ b/Library/src/Web/HttpRequest.cpp
@@ -162,27 +162,10 @@ HttpProgress& HttpRequest::GetProgress()
 
 bool ResultCodeValidForRetry(csp::web::EResponseCodes Status)
 {
-	if (Status == csp::web::EResponseCodes::ResponseTooManyRequests					 // 429
-		|| Status == csp::web::EResponseCodes::ResponseRequestTimeout				 // 408
-		|| Status == csp::web::EResponseCodes::ResponseInternalServerError			 // 500
-		|| Status == csp::web::EResponseCodes::ResponseNotImplemented				 // 501
-		|| Status == csp::web::EResponseCodes::ResponseBadGateway					 // 502
-		|| Status == csp::web::EResponseCodes::ResponseServiceUnavailable			 // 503
-		|| Status == csp::web::EResponseCodes::ResponseGatewayTimeout				 // 504
-		|| Status == csp::web::EResponseCodes::ResponseVersionNotSupported			 // 505
-		|| Status == csp::web::EResponseCodes::ResponseVariantAlsoNegotiates		 // 506
-		|| Status == csp::web::EResponseCodes::ResponseInsufficientStorage			 // 507
-		|| Status == csp::web::EResponseCodes::ResponseLoopDetected					 // 508
-		|| Status == csp::web::EResponseCodes::ResponseNotExtended					 // 510
-		|| Status == csp::web::EResponseCodes::ResponseNetworkAuthenticationRequired // 511
-	)
-	{
-		return true;
-	}
-	else
-	{
-		return false;
-	}
+	return (Status == csp::web::EResponseCodes::ResponseTooManyRequests	  // 429
+			|| Status == csp::web::EResponseCodes::ResponseRequestTimeout // 408
+			|| static_cast<int>(Status) >= 500							  // 500
+	);
 }
 
 /// @brief Retry this request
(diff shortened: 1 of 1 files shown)


MAG-AdamThorn (approved):

Library/src/Web/HttpRequest.cpp
```diff
@@ -160,6 +160,30 @@ HttpProgress& HttpRequest::GetProgress()
 	return Progress;
 }
 
+bool ResultCodeValidForRetry(csp::web::EResponseCodes Status)
+{
+	if (Status == csp::web::EResponseCodes::ResponseTooManyRequests					 // 429
```
MAG-SamBirley, line 165: We can simply return the conditional here instead of an actual if/else.

Also, is there a reason to check each 500 response? Why not something like:

```
return 
    Status == csp::web::EResponseCodes::ResponseTooManyRequests ||
    Status == csp::web::EResponseCodes::ResponseRequestTimeout ||
    Status  >= 500;
```

Library/src/Web/HttpRequest.cpp
```diff
@@ -171,7 +195,7 @@ HttpProgress& HttpRequest::GetProgress()
 /// @return true if retry succeeded, false if retry limit was reached
 bool HttpRequest::Retry(const uint32_t MaxRetries)
 {
-	if (RetryCount < MaxRetries)
+	if (ResultCodeValidForRetry(this->Response.GetResponseCode()) && RetryCount < MaxRetries)
```
MAG-SamBirley, line 198: Why do we need the `this->` here?

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -1346,7 +1346,7 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, SetUserRoleOnInviteTest)
 	CreateSpace(SpaceSystem, UniqueSpaceName, TestSpaceDescription, SpaceAttributes::Private, nullptr, nullptr, nullptr, Space);
 
 	// Invite second test account as a Moderator Role user
-	auto [Result] = AWAIT_PRE(SpaceSystem, InviteToSpace, RequestPredicate, Space.Id, AlternativeLoginEmail, true, "https://www.magnopus.com/");
+	auto [Result] = AWAIT_PRE(SpaceSystem, InviteToSpace, RequestPredicate, Space.Id, AlternativeLoginEmail, true, "");
```
MAG-SamBirley, line 1349: Why did this line need changing?

Library/src/Web/HttpRequest.cpp
MAG-SamBirley: Yeah I get that, but the cast would allow for us to avoid human error (missing codes etc) both now and in the future. Think its worth it in this instance.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-SamBirley: Hm, wouldn't that mean other PR's checks would be failing similarly?

MAG-SamBirley: Yeah I'd love to know why - any ideas @jeromagnopus?

MAG-SamBirley: Marking this as resolved to keep things moving, but I would really like to understand this!

MAG-SamBirley (approved):

MAG-MichaelK (approved):

Tests/src/PublicAPITests/SpaceSystemTests.cpp
jeromagnopus: I'm not sure what this parameter is meant to be - but I'm assuming it's the `emailLinkUrl` that we use to construct the environment and release-build-specific URL?  If so, we have an allow-list for these and `https://www.magnopus.com` is not in them.  This is somewhat recent, but was merged and deployed weeks ago at minimum.  

I'm not sure why you're suddenly seeing failures, however, a solution would be to change to `https:\\dev.magnoverse.space`

jeromagnopus: @MAG-SamBirley Please see what I wrote above - not sure if it answers all questions, however.

PR #229, opened 2023-10-23
[NT-0] fix: Revert "[OF-815] test: (wrapper gen) unit test async functions"

MAG-SamBirley:
Checking out behaviour with this change reverted, as most recent C# builds appear to fail in a related area.


MAG-SamBirley:
Note this change is not intended to be merged, but a means of sanity checking whether the PR being reverted introduced issues when building wrappers for C# CSP

MAG-SamBirley:
Shutting this one down now that we've found & fixed the issue in https://github.com/magnopus-opensource/connected-spaces-platform/pull/230

PR #232, opened 2023-10-25
[OB-2975] fix: foundation Shutdown process

MAG-AW:
* added initialisation check in tick to prevent tick operating when uninitialised (such as when shutting down)
* added function to deregister all listeners to prevent events being fired after shutdown is called

**For the reviewer**


MAG-MichaelK (approved):

MAG-AdamThorn (approved):

MAG-SamBirley:
@MAG-AW worth adding a test that CSP gracefully handles cases now where tick is called when CSP is uninitialized?

Mag-JB (approved):

PR #251, opened 2023-11-13
[OB-3040] fix: Resolving a QuotaSystem C# crash

MAG-SamBirley:
A bug in the CSP wrapper generator means that reference-to-enum parameters will be considered to be int pointers at the C# wrapper level.

These int pointers were then passed to the generated C wrapper, which attempted to dereference them before passing them to CSP's C++ API.

Those dereferences were the cause of the crash.

The fix here is to alter the signature of the QuotaSystem API such that it passes enum params by value rather than reference.

A separate issue for the wrapper generator bug will be raised and tracked.

### DIFF
--- a/Library/include/CSP/Systems/Quota/QuotaSystem.h
+++ b/Library/include/CSP/Systems/Quota/QuotaSystem.h
@@ -36,7 +36,7 @@ class WebClient;
 
 namespace csp::systems
 {
-/// @ingroup QuotaSystem
+/// @ingroup Quota System
 /// @brief Public facing system that allows interfacing with Magnopus Connect Services' Quota Server.
 /// Offers methods for receiving Quota Queries.
 class CSP_API CSP_NO_DISPOSE QuotaSystem : public SystemBase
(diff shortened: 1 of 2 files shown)


lyubomirlichev (dismissed):

Library/include/CSP/Systems/Quota/QuotaSystem.h
```diff
@@ -83,12 +83,12 @@ class CSP_API CSP_NO_DISPOSE QuotaSystem : public SystemBase
 	/// @param TierName TierNames : Name of the tier
 	/// @param FeatureName TierFeatures : Name of the feature
 	/// @param Callback FeatureQuotaCallback : callback when asynchronous task finishes
-	CSP_ASYNC_RESULT void GetTierFeatureQuota(const TierNames& TierName, const TierFeatures& FeatureName, FeatureQuotaCallback Callback);
+	CSP_ASYNC_RESULT void GetTierFeatureQuota(const TierNames TierName, const TierFeatures FeatureName, FeatureQuotaCallback Callback);
```
MAG-MichaelK, line 86: `const` is not needed here due to this being passed by value.

MAG-SamBirley: Yup agreed, thanks!

MAG-SamBirley: Fixed in latest.

MAG-MichaelK (approved):

MAG-MichaelK (approved):

MAG-MichaelK (approved):

PR #254, opened 2023-11-14
[OF-1094] test: Cancellation token tests

MAG-SamBirley:
Introducing test coverage for `csp::common::CancellationToken`.

### DIFF
--- a/Tests/src/PublicAPITests/CancellationTokenTests.cpp
+++ b/Tests/src/PublicAPITests/CancellationTokenTests.cpp
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2023 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Awaitable.h"
+#include "CSP/CSPFoundation.h"
+#include "CSP/Common/CancellationToken.h"
+#include "TestHelpers.h"
+
+#include "gtest/gtest.h"
+
+namespace
+{
+
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_CANCEL_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, CancelStateTest)
+{
+	csp::common::CancellationToken CancellationToken;
+	CancellationToken.Cancel();
+	EXPECT_TRUE(CancellationToken.Cancelled());
+}
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_ASYNCREF_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, CancelAsyncRefTest)
+{
+	csp::common::CancellationToken CancellationToken;
+
+	std::thread Thread(
+		[&CancellationToken]()
+		{
+			while (CancellationToken.Cancelled() == false)
+			{
+			}
+		});
+
+	CancellationToken.Cancel();
+	Thread.join();
+}
+#endif
+
+} // namespace
--- a/Tests/src/PublicAPITests/CancellationTokenTests.cpp
+++ b/Tests/src/PublicAPITests/CancellationTokenTests.cpp
@@ -21,8 +21,6 @@
 
 #include "gtest/gtest.h"
 
-#define RUN_CANCELLATION_TOKEN_TESTS 1
-
 namespace
 {
 
--- a/Tests/src/PublicAPITests/CancellationTokenTests.cpp
+++ b/Tests/src/PublicAPITests/CancellationTokenTests.cpp
@@ -37,10 +37,10 @@ CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, CancelStateTest)
 	csp::common::CancellationToken CancellationToken;
 	EXPECT_FALSE(CancellationToken.Cancelled());
 
-	EXPECT_NO_THROW(CancellationToken.Cancel());
+	CancellationToken.Cancel();
 	EXPECT_TRUE(CancellationToken.Cancelled());
 
-	CancellationToken.Cancel(); // Test that multiple cancellations don't affect the state
+	EXPECT_NO_THROW(CancellationToken.Cancel()); // Test that multiple cancellations don't affect the state
 	EXPECT_TRUE(CancellationToken.Cancelled());
 }
 #endif
--- a/Tests/src/PublicAPITests/CancellationTokenTests.cpp
+++ b/Tests/src/PublicAPITests/CancellationTokenTests.cpp
@@ -27,8 +27,7 @@ namespace
 #if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_CONSTRUCTION_TEST
 CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, ConstructionAndDestructionTest)
 {
-	csp::common::CancellationToken CancellationToken;
-	// No need to assert anything, just ensuring that construction and destruction don't crash
+	EXPECT_NO_THROW(csp::common::CancellationToken());
 }
 #endif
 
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+#include "Awaitable.h"
+#include "CSP/CSPFoundation.h"
+#include "CSP/Common/CancellationToken.h"
+#include "TestHelpers.h"
+
+#include "gtest/gtest.h"
+
+#define RUN_CANCELLATION_TOKEN_TESTS 1
```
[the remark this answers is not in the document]
MAG-SamBirley: Whoops! Thanks (again)! Fixed in latest

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, AsyncRefTest)
+{
+	csp::common::CancellationToken CancellationToken;
+
+	std::thread Thread(
+		[&CancellationToken]()
+		{
+			while (CancellationToken.Cancelled() == false)
+			{
+			}
```
MAG-MichaelK, line 70: Adding a sleep in here would greatly reduce CPU usage during this loop. We also prefer to not explicitly compare against boolean constants. `!CancellationToken.Cancelled()` would look better here.

Also, is there any point in testing this? I can't think of a situation where holding a reference to an object in another thread would fail while said object is still alive.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "gtest/gtest.h"
+
+namespace
+{
+
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_CONSTRUCTION_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, ConstructionAndDestructionTest)
+{
+	csp::common::CancellationToken CancellationToken;
```
MAG-MichaelK, line 30: Scoping this and catching any exceptions that occur during destruction would result in a better testing experience. Otherwise, an exception thrown here would result in an early process exit, meaning no further tests would run.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_CANCEL_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, CancelStateTest)
+{
+	csp::common::CancellationToken CancellationToken;
+	EXPECT_FALSE(CancellationToken.Cancelled());
+
+	CancellationToken.Cancel();
+	EXPECT_TRUE(CancellationToken.Cancelled());
+
+	CancellationToken.Cancel(); // Test that multiple cancellations don't affect the state
```
MAG-MichaelK, line 44: This should also not throw, so worth testing that.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+	EXPECT_TRUE(CancellationToken.Cancelled());
+}
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_COPYMOVE_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, CopyMoveTest)
+{
+	// Ensure copy and move operations are deleted
+	ASSERT_FALSE(std::is_copy_constructible<csp::common::CancellationToken>::value);
```
MAG-MichaelK, line 53: Prefer the `std::is_*_v<T>` forms over `std::is_*<T>::value`

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_ASYNCREF_TEST
+CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, AsyncRefTest)
+{
+	csp::common::CancellationToken CancellationToken;
+
+	std::thread Thread(
+		[&CancellationToken]()
+		{
+			while (CancellationToken.Cancelled() == false)
```
MAG-MichaelK, line 68: This needs a timeout in case it fails.

MAG-MichaelK (changes requested):

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: I generally agree, but also quite like it being there as a demonstration of the intended usage, as a functional test.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Fair point!

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: No issue this, but interested in the rationale, since the former is a typedef of the latter. Is there a portability reason?

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Addressed in latest.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Addressed in latest.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Addressed in latest.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Also addressed in latest.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Test nuked in latest.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-MichaelK: The C++ committee suggests using them since they were introduced, as it's less to type and also more immediately obvious what it is you're doing. Plus, you only ever use these structs when wanting to get the value from them, so it makes sense to use the less verbose form.

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-MichaelK: The test for throwing should be on the 2nd cancel, not the 1st.

MAG-SamBirley: How come?

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Got it, thanks! 👍

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-MichaelK: Because the first Cancel shouldn't fail, but the second might as it's already cancelled. That's what we're checking here.

MAG-SamBirley: Gotcha

MAG-SamBirley: Fixed in latest.

Mag-JB (approved):

Tests/src/PublicAPITests/CancellationTokenTests.cpp
```diff
@@ -27,8 +27,7 @@ namespace
 #if RUN_ALL_UNIT_TESTS || RUN_CANCELLATION_TOKEN_TESTS || RUN_CANCELLATIONTOKEN_CONSTRUCTION_TEST
 CSP_PUBLIC_TEST(CSPEngine, CancellationTokenTests, ConstructionAndDestructionTest)
 {
-	csp::common::CancellationToken CancellationToken;
-	// No need to assert anything, just ensuring that construction and destruction don't crash
+	EXPECT_NO_THROW(csp::common::CancellationToken());
```
MAG-MichaelK, line 30: This will only catch exceptions in the constructor and not the destructor. You should use a try-catch with an explicit scope here to guarantee the destructor is called within the try block.

MAG-MichaelK: It's also worth noting that destructors by default are marked as `nothrow` in c++11 and up, so we should avoid throwing at all costs.

MAG-MichaelK (changes requested):

Tests/src/PublicAPITests/CancellationTokenTests.cpp
MAG-SamBirley: Looking at the macro underneath `EXCEPT_NO_THROW` I think we are good as-is, since the try block it expands to is where the statement gets invoked, so both the ctor and dtor will be in there.
image

MAG-MichaelK (approved):

PR #257, opened 2023-11-16
[OF-1193] feat: Add XErrorCodeType

Mag-JB:
* Add XErrorCodeCatergory

**For the reviewer**

### DIFF
--- a/Library/src/Systems/Quota/Quota.cpp
+++ b/Library/src/Systems/Quota/Quota.cpp
@@ -38,6 +38,17 @@ void FeaturesLimitResult::OnResponse(const csp::services::ApiResponseBase* ApiRe
 
 	auto* FeatureProgressResponse		   = static_cast<csp::services::DtoArray<chs::QuotaFeatureLimitProgressDto>*>(ApiResponse->GetDto());
 	const csp::web::HttpResponse* Response = ApiResponse->GetResponse();
+	const auto& Headers					   = Response->GetPayload().GetHeaders();
+
+	if (Headers.count("x-errorcode") > 0 && !Headers.at("x-errorcode").empty())
+	{
+		XErrorCodeType = StringToXErrorCodeTypeEnum(Headers.at("x-errorcode").c_str());
+
+		if (Headers.count("x-errorcategory") > 0 && !Headers.at("x-errorcategory").empty())
+		{
+			XErrorCategory = StringToXErrorCatergoriesEnum(Headers.at("x-errorcategory").c_str());
+		}
+	}
 
 	if (ApiResponse->GetResponseCode() == csp::services::EResponseCode::ResponseSuccess)
 	{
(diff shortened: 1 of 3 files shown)


Library/src/Systems/Quota/Quota.cpp
```diff
@@ -367,4 +422,67 @@ const TierFeatures StringToTierFeatureEnum(const csp::common::String& Value)
 	return TierFeatures::SpaceOwner;
 }
 
+XErrorCodeTypes StringToXErrorCodeTypeEnum(const csp::common::String& Value)
+{
+	if (Value == "group_spaceownerquota")
```
MAG-AdamThorn, line 427: The string case here does not match what is listed in the CHS TDD, is what you have here correct? Eg: **Group_SpaceOwnerQuota**

MAG-AdamThorn: Same for the ones below

Mag-JB: The TDD seems to be wrong

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -367,4 +422,67 @@ const TierFeatures StringToTierFeatureEnum(const csp::common::String& Value)
...
+		return XErrorCodeTypes::TicketedSpacesUserQuota;
+	}
+
+	CSP_LOG_ERROR_FORMAT("QuotaSystem XErrorCodeType not recognized: %s. Defaulting to Unknown", Value.c_str());
+	return XErrorCodeTypes::Unknown;
+}
+
+XErrorCategories StringToXErrorCatergoriesEnum(const csp::common::String& Value)
+{
+	if (Value == "quota_limitreached")
```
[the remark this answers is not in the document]
Mag-JB: The TDD seems to be wrong

Mag-JB:
Not auto merging until there is an answer about capitalisation

MAG-MichaelK (changes requested):
We should be using the already available `FailureReason` value for exposing `X-ErrorCode` values, and completely ignoring the category as it doesn't matter to us. `ResultBase` already takes care of checking for the `X-ErrorCode` header for us. Please re-implement this by overriding `ParseErrorCode()` in the QuotaSystem result classes so that `ResultBase` can get an integer value representing the error from the string CHS gives us. For reference, please check https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/src/Systems/Users/Authentication.cpp#L102

Library/src/Systems/Quota/Quota.cpp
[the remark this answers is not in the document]
MAG-MichaelK: We modify any incoming header value to be lowercase. CHS does indeed send it in the way Adam described, but that does not matter to us.

PR #258, opened 2023-11-16
[OB-2323] fix: CSP crashing on M1 macs

MAG-SamBirley:
Similar to recent crashes that have been fixed for Quest & Android, mimalloc crashes when CSP is initialized within an application running on an M1 Mac.

The short-term fix for this is the same as was applied for Quest. We fall back to STL allocator behaviour and avoid the usage of mimalloc entirely when we're running on macosx.

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -185,7 +185,8 @@ if not Project then
             }
         filter "platforms:macosx"
             defines { 
-                "CSP_MACOSX", 				
+                "CSP_MACOSX",
+                "USE_STD_MALLOC=1",
                 "JS_STRICT_NAN_BOXING"
             }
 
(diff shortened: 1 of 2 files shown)


Mag-JB (dismissed):

MAG-AdamThorn (dismissed):

MAG-MichaelK (changes requested):
We also need to exclude mimalloc from the linker arguments otherwise this will not build correctly when building a shared library.

MAG-SamBirley:
> We also need to exclude mimalloc from the linker arguments otherwise this will not build when building a shared library.

Great catch! Thanks

MAG-SamBirley:
> > We also need to exclude mimalloc from the linker arguments otherwise this will not build when building a shared library.
> 
> Great catch! Thanks

Fixed in latest.

Mag-JB (approved):

MAG-MichaelK (approved):

PR #260, opened 2023-11-16
[OF-1193] feat: Add FailureReason for Quota

Mag-JB:

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/Quota/Quota.h
+++ b/Library/include/CSP/Systems/Quota/Quota.h
@@ -69,6 +69,21 @@ enum class TierNames
 	Enterprise
 };
 
+enum class EQuotaResultFailureReason
+{
+	Unknown = -1,
+	None	= 0,
+	SpaceOwnerQuota,
+	ConcurrentUsersQuota,
+	ObjectCaptureQuota,
+	AudioVideoQuota,
+	TotalUploadSizeInKilobytes,
+	AgoraOwnerQuota,
+	OpenAIUserQuota,
+	ShopifyUserQuota,
+	TicketedSpacesUserQuota
+};
+
 /// @ingroup Quota System
 /// @brief Data representation of a progress of a specific feature.
 /// Limit Value of -1 means unlimited usage
@@ -124,6 +139,9 @@ public:
 	/// @return csp::common::Array<FeatureProgress> : const array of feature progress class
 	const csp::common::Array<FeatureLimitInfo>& GetFeaturesLimitInfo() const;
 
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
+
 private:
 	FeaturesLimitResult(void*) {};
 
@@ -146,6 +164,9 @@ public:
 	/// @brief Retrieves the feature progress result.
 	/// @returnFeatureProgress : const ref to feature progress class
 	const FeatureLimitInfo& GetFeatureLimitInfo() const;
+
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
 
 private:
 	FeatureLimitResult(void*) {};
@@ -170,6 +191,9 @@ public:
 	/// @returnFeatureProgress : const ref to user tier information class
 	const UserTierInfo& GetUserTierInfo() const;
 
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
+
 private:
 	UserTierResult(void*) {};
 
@@ -193,6 +217,9 @@ public:
 	/// @returnFeatureProgress : const ref to feature quota class
 	const FeatureQuotaInfo& GetFeatureQuotaInfo() const;
 
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
+
 private:
 	FeatureQuotaResult(void*) {};
 
@@ -215,6 +242,9 @@ public:
 	/// @brief Retrieves an array of feature quota results.
 	/// @returnFeatureProgress : const ref to feature quota class
 	const csp::common::Array<FeatureQuotaInfo>& GetFeaturesQuotaInfo() const;
+
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
 
 private:
 	FeaturesQuotaResult(void*) {};
--- a/Library/src/Systems/Quota/Quota.cpp
+++ b/Library/src/Systems/Quota/Quota.cpp
@@ -68,6 +68,38 @@ void FeaturesLimitResult::OnResponse(const csp::services::ApiResponseBase* ApiRe
 	}
 }
 
+int FeaturesLimitResult::ParseErrorCode(const csp::common::String& Value)
+{
+	if (Value == "group_spaceownerquota")
+		return (int) EQuotaResultFailureReason::SpaceOwnerQuota;
+
+	if (Value == "scopes_concurrentusersquota")
+		return (int) EQuotaResultFailureReason::ConcurrentUsersQuota;
+
+	if (Value == "assetdetail_objectcapturequota")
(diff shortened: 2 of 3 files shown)


Library/include/CSP/Systems/Quota/Quota.h
```diff
@@ -124,6 +139,9 @@ class CSP_API FeaturesLimitResult : public csp::services::ResultBase
 	/// @return csp::common::Array<FeatureProgress> : const array of feature progress class
 	const csp::common::Array<FeatureLimitInfo>& GetFeaturesLimitInfo() const;
 
+protected:
+	int ParseErrorCode(const csp::common::String& Value) override;
```
MAG-SamBirley, line 143: Would it be ok to add the `virtual` qualifier to these overrrides? For me personally, it helps a little with readability.

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -68,6 +68,38 @@ void FeaturesLimitResult::OnResponse(const csp::services::ApiResponseBase* ApiRe
 	}
 }
 
+int FeaturesLimitResult::ParseErrorCode(const csp::common::String& Value)
+{
+	if (Value == "group_spaceownerquota")
+		return (int) EQuotaResultFailureReason::SpaceOwnerQuota;
```
MAG-SamBirley, line 74: Prefer static_cast over C-style cast?

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -104,6 +136,38 @@ void FeatureLimitResult::OnResponse(const csp::services::ApiResponseBase* ApiRes
...
+	if (Value == "assetdetail_totaluploadsizeinkilobytes")
+		return (int) EQuotaResultFailureReason::TotalUploadSizeInKilobytes;
+
+	if (Value == "agoraoperation_groupownerquota")
+		return (int) EQuotaResultFailureReason::AgoraOwnerQuota;
+
+	if (Value == "openaioperation_userquota")
+		return (int) EQuotaResultFailureReason::OpenAIUserQuota;
+
+	if (Value == "shopify_userquota")
```
MAG-SamBirley, line 162: With errors like these that can be returned for a variety of requests, is there a way we could have only one place that parses them, to avoid the duplicate code?

If there's a good reason for the duplicate code, I could be perusaded to keep it as-is.

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -138,6 +202,38 @@ void UserTierResult::OnResponse(const csp::services::ApiResponseBase* ApiRespons
 	}
 }
 
+int UserTierResult::ParseErrorCode(const csp::common::String& Value)
+{
+	if (Value == "group_spaceownerquota")
```
MAG-SamBirley, line 207: Our [programming guidelines](https://github.com/magnopus-opensource/connected-spaces-platform/wiki/Programming-guidelines) closely follow the Epic coding guidelines, which ask for us to always include braces (see the **Braces** section [here](https://docs.unrealengine.com/5.2/en-US/epic-cplusplus-coding-standard-for-unreal-engine/)

MAG-SamBirley (changes requested):

Library/src/Systems/Quota/Quota.cpp
Mag-JB: sorted

Library/src/Systems/Quota/Quota.cpp
Mag-JB: sorted

Library/src/Systems/Quota/Quota.cpp
Mag-JB: sorted

Library/include/CSP/Systems/Quota/Quota.h
Mag-JB: sorted

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -68,6 +68,18 @@ void FeaturesLimitResult::OnResponse(const csp::services::ApiResponseBase* ApiRe
 	}
 }
 
+int FeaturesLimitResult::ParseErrorCode(const csp::common::String& Value)
+{
+	const int32_t ReturnValue = StringToEQuotaResultFailureReasonIntValue(Value);
```
MAG-SamBirley, line 73: Now that these four overrides look exactly the same, and the base implementation doesn't really do anything atm, we could make this the base implementation and avoid these four overrides entirely.

MAG-SamBirley (changes requested):

Library/include/CSP/Systems/Quota/Quota.h
MAG-MichaelK: We should avoid using virtual where we don't want functions to be overridden by further derived classes. As there should never be a need for us to derive from `FeaturesLimitResult`, `ParseErrorCode` should only be marked as `override` and not `virtual`.

Library/src/Systems/Quota/Quota.cpp
MAG-MichaelK: The duplicate code is useful because each enum is unique to the result. I can't think of a way we can guarantee developers will not use conflicting enum values.

Library/src/Systems/Quota/Quota.cpp
MAG-MichaelK: The `ResultBase` implementation of ParseErrorCode is empty on purpose. It is not meant to do anything. There is not a single failure reason that is shared amongst all responses. Also, the enum differs for each result type, and for a good reason. Please leave this as it is.

Library/include/CSP/Systems/Quota/Quota.h
MAG-SamBirley: Yep all good, happy with that

Library/src/Systems/Quota/Quota.cpp
```diff
@@ -229,6 +277,19 @@ void FeaturesQuotaResult::OnResponse(const csp::services::ApiResponseBase* ApiRe
...
 	}
 }
+
+int FeaturesQuotaResult::ParseErrorCode(const csp::common::String& Value)
+{
+	const int32_t ReturnValue = StringToEQuotaResultFailureReasonIntValue(Value);
+
+	if (ReturnValue == 0)
+	{
+		return ResultBase::ParseErrorCode(Value);
```
MAG-MichaelK, line 287: Aaaaaaaaand here

Tests/src/PublicAPITests/QuotaSystemTests.cpp
```diff
@@ -436,4 +446,25 @@ CSP_PUBLIC_TEST(CSPEngine, QuotaSystemTests, GetTotalSpaceSizeinKilobytes)
 	// Delete space
 	DeleteSpace(SpaceSystem, Space.Id);
 }
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_QUOTASYSTEM_TESTS || RUN_QUOTASYSTEM_STRINGTOEQUOTERESULTFAILUREREASONINTVALUE_TEST
+CSP_PUBLIC_TEST(CSPEngine, QuotaSystemTests, StringToEQuotaResultFailureReasonIntValueTest)
+{
+	EXPECT_EQ(StringToEQuotaResultFailureReasonIntValue(""), static_cast<int>(EQuotaResultFailureReason::None));
```
MAG-MichaelK, line 454: This should be equal to `Unknown` and not `None`. `None` is _only_ used when there is no error.

MAG-MichaelK (changes requested):

PR #270, opened 2023-11-22
[OF-1195] feat!: Add XErrorCodes to WebService

Mag-JB:

**For the reviewer**


Library/src/Services/WebService.cpp
```diff
@@ -104,9 +105,16 @@ void ResultBase::SetResult(csp::services::EResultCode ResCode, uint16_t HttpResC
...
+EResultBaseFailureReason ResultBase::ParseErrorCode(const csp::common::String& Value)
 {
-	return (int) EResultBaseFailureReason::Unknown;
+	if (XErrorCodeToFailureReason.HasKey(Value))
+	{
+		return XErrorCodeToFailureReason[Value];
+	}
+	else
+	{
+		return EResultBaseFailureReason::Unknown;
```
MAG-SamBirley, line 116: Worth including a log message here to help us catch error codes that we don't yet track in CSP?

MAG-SamBirley (changes requested):

Library/include/CSP/Services/WebService.h
```diff
@@ -105,14 +139,14 @@ class CSP_API ResultBase
...
-	int GetFailureReason() const;
+	EResultBaseFailureReason GetFailureReason() const;
 
 protected:
 	ResultBase(csp::services::EResultCode ResCode, uint16_t HttpResCode);
 
 	void SetResult(csp::services::EResultCode ResCode, uint16_t HttpResCode);
 
-	virtual int ParseErrorCode(const csp::common::String& Value);
+	virtual EResultBaseFailureReason ParseErrorCode(const csp::common::String& Value);
```
MAG-AdamThorn, line 149: Does this still need to be virtual?

Library/src/Services/WebService.cpp
MAG-SamBirley: Saw the new error message. Super useful! Thanks :)

Library/include/CSP/Services/WebService.h
MAG-MichaelK: No, and it should also be made private.

Library/include/CSP/Services/WebService.h
```diff
@@ -121,7 +155,44 @@ class CSP_API ResultBase
 	float ResponseProgress = 0.0f;
 
 	csp::common::String ResponseBody;
-	int FailureReason;
+	EResultBaseFailureReason FailureReason;
+
+private:
+	csp::common::Map<csp::common::String, EResultBaseFailureReason> XErrorCodeToFailureReason = {
```
MAG-MichaelK, line 161: This map should be moved to WebService.cpp in an anonymous namespace, changes to a `std::map`, and be made static. We don't want to re-create this every time a new ResultBase instance is made. We should also only be mapping the quota errors and any error codes we already handle as that is all we need to cover in this ticket.

Library/include/CSP/Services/WebService.h
```diff
@@ -62,7 +63,40 @@ enum class EResultCode : uint8_t
 enum class EResultBaseFailureReason
```
MAG-MichaelK, line 63: Can we get a better name for this now that it contains all error codes?

MAG-MichaelK (changes requested):

Library/include/CSP/Services/WebService.h
MAG-MichaelK: Something like `EFailureReason`

MAG-SamBirley (approved):
Happy with the changes made to address my requests, so approved from that point of view. I would like to also see the feedback from @MAG-MichaelK addressed too though!

MAG-MichaelK (approved):

Mag-JB:
I will test this before merging!

Library/include/CSP/Services/WebService.h
```diff
@@ -105,14 +138,14 @@ class CSP_API ResultBase
...
-	int GetFailureReason() const;
+	ERequestFailureReason GetFailureReason() const;
 
 protected:
 	ResultBase(csp::services::EResultCode ResCode, uint16_t HttpResCode);
 
 	void SetResult(csp::services::EResultCode ResCode, uint16_t HttpResCode);
 
-	virtual int ParseErrorCode(const csp::common::String& Value);
+	ERequestFailureReason ParseErrorCode(const csp::common::String& Value);
```
MAG-MichaelK, line 148: Can we also make this private as it doesn't need to be accessed by any derived types?

MAG-MichaelK (changes requested):

MAG-SamBirley (dismissed):

Library/include/CSP/Services/WebService.h
MAG-MichaelK: Just waiting on this to be done then I can approve this PR

Mag-JB: oops didn't see this one sneak in after the approval

MAG-MichaelK (approved):

MAG-AdamThorn (approved):

PR #299, opened 2024-01-02
[NT-0] feat: Copying asset collections from one space to another

MAG-SamBirley:
This PR introduces the ability for a client application to invoke the duplication of a set of asset collections from one source space to another destination space. It requires that the set of source asset collections all come from the same space.

Some things to be aware of about this change:
* From a CSP API point of view, It changes the notion that an asset collection can belong to _many_ spaces to just _one_. This is to align with the copying API, and also to strengthen this concept in general.
* It also changes the behaviour of one of the constructors for `csp::common::Array` - when an array is constructed from a raw ptr, it now performs a per-element copy (via the = operator) rather than memcpying the block of memory - this resolves issues around inadvertent shared-memory ownership


Library/src/Systems/Assets/AssetCollection.cpp
```diff
@@ -93,12 +94,15 @@ void PrototypeDtoToAssetCollection(const chs::PrototypeDto& Dto, csp::systems::A
...
+		auto& GroupIds = Dto.GetGroupIds();
+		if (GroupIds.size() > 0)
+		{
+			AssetCollection.SpaceId = GroupIds[0];
+		}
+		else
 		{
-			AssetCollection.SpaceIds[i] = GroupIds[i];
+			CSP_LOG_MSG(csp::systems::LogLevel::Error,
+						"Encountered an asset collection that did not belong to any space whilst processing a response from services.");
```
MAG-SamBirley, line 105: It is valid for an asset collection to not be associated with a space, so this should not log an error

MAG-SamBirley: Addressed in latest. Please can I get a re-review @MAG-AW @MAG-AdamThorn?

Library/include/CSP/Common/Array.h
```diff
@@ -74,7 +74,15 @@ template <typename T> class CSP_API Array
...
-			memcpy(ObjectArray, Buffer, Size * sizeof(T));
+
+			// Perform the copy per-element via the assignment operator, rather than memcpy.
+			// This is to avoid introducing inadvertent shared memory ownership between elements in different arrays.
+			// We do not, for example, want strings in an array to hold the same pointers to
+			// memory as another array containing strings that it was constructed from.
+			for (size_t i = 0; i < Size; i++)
+			{
+				ObjectArray[i] = Buffer[i];
+			}
```
MAG-MichaelK, line 85: The purpose of this constructor was to be able to "move" a raw C-style array into a `csp::common::Array`. The raw array would then be freed by the caller.

Library/include/CSP/Systems/Assets/AssetCollection.h
```diff
@@ -71,16 +71,37 @@ class CSP_API AssetCollection
...
+
+	// The unique identifier of the POI this asset collection relates to. Empty if it does not relate to a POI.
 	csp::common::String PointOfInterestId;
+
+	// The unique ID of the asset collection that is a parent to this asset collection. Empty if there is no parent relationship.
 	csp::common::String ParentId;
-	csp::common::Array<csp::common::String> SpaceIds;
+
+	// Where the asset collection belongs to a space, the unique ID of the space that this asset collection belongs to. Empty otherwise.
+	csp::common::String SpaceId;
```
MAG-MichaelK, line 93: What's the rationale behind this change? This essentially limits us in terms of sharing assets between spaces, and requires a unique copy on CHS for each space. I would rather we instead give the user the _option_ to copy the assets instead of referencing the original.

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -76,6 +76,18 @@ class CSP_API CSP_NO_DISPOSE AssetSystem : public SystemBase
 	/// @param Callback NullResultCallback : callback when asynchronous task finishes
 	CSP_ASYNC_RESULT void DeleteAssetCollection(const AssetCollection& AssetCollection, NullResultCallback Callback);
 
+	/// @brief Copies an array of asset collections to another space. Note that all source asset collections must belong to the same space.
```
MAG-MichaelK, line 79: Is this a CHS limitation?

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -76,6 +76,18 @@ class CSP_API CSP_NO_DISPOSE AssetSystem : public SystemBase
...
+	/// @param SourceAssetCollections csp::common::Array<AssetCollection> : The array of asset collections to copy. They must all belong to the same
+	/// space.
+	/// @param DestSpaceId  const csp::common::String : The unique identifier of the space to copy these asset collections to.
+	/// @param CopyAsync const csp::common::Optional<bool> : Whether to instruct the services to perform the copy of the asset collections
+	/// asynchronously.
+	/// @param Callback NullResultCallback : callback when asynchronous task finishes
+	CSP_ASYNC_RESULT void CopyAssetCollectionsToSpace(csp::common::Array<AssetCollection>& SourceAssetCollections,
```
MAG-MichaelK, line 86: What is this in aid of? I see this PR has no ticket, so is this for a personal project?

Library/src/Systems/Assets/AssetCollection.cpp
```diff
@@ -93,12 +94,10 @@ void PrototypeDtoToAssetCollection(const chs::PrototypeDto& Dto, csp::systems::A
 
 	if (Dto.HasGroupIds())
 	{
-		auto& GroupIds			 = Dto.GetGroupIds();
-		AssetCollection.SpaceIds = csp::common::Array<csp::common::String>(GroupIds.size());
-
-		for (int i = 0; i < GroupIds.size(); ++i)
+		auto& GroupIds = Dto.GetGroupIds();
+		if (GroupIds.size() > 0)
```
MAG-MichaelK, line 98: Can we get empty lines before and after scope blocks, please?

Library/src/Systems/Assets/AssetCollection.cpp
```diff
@@ -241,7 +240,11 @@ void AssetCollectionsResult::FillResultTotalCount(const csp::common::String& Jso
 	{
 		JsonDoc.Parse(JsonContent.c_str());
 
-		if (JsonDoc.HasMember("itemTotalCount"))
+		if (JsonDoc.IsArray())
```
MAG-MichaelK, line 243: Why was this change needed? It has worked up until now without this. Did something change in CHS?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -210,6 +210,80 @@ void AssetSystem::DeleteAssetCollection(const AssetCollection& AssetCollection,
...
+
+		INVOKE_IF_NOT_NULL(Callback, MakeInvalid<AssetCollectionsResult>());
+		return;
+	}
+
+	csp::common::String SourceSpaceId					= SourceAssetCollections[0].SpaceId;
+	std::vector<csp::common::String> AssetCollectionIds = {SourceAssetCollections[0].Id};
+
+	bool AssetCollectionsBelongToSameSpace = true;
+	for (size_t i = 1; i < SourceAssetCollections.Size(); ++i)
```
MAG-MichaelK, line 230: Spacing

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -210,6 +210,80 @@ void AssetSystem::DeleteAssetCollection(const AssetCollection& AssetCollection,
...
+		CSP_LOG_MSG(
+			LogLevel::Error,
+			"An asset with no space ID was provided whilst attempting to perform a copy to another space. All assets must have a valid space ID.");
+
+		INVOKE_IF_NOT_NULL(Callback, MakeInvalid<AssetCollectionsResult>());
+		return;
+	}
+
+	// Verify that all source asset collections belong to the same space. If not, this qualifies as an unsupported operation.
+	if (AssetCollectionsBelongToSameSpace == false)
```
MAG-MichaelK, line 248: Prefer `!` over `== false`

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -210,6 +210,80 @@ void AssetSystem::DeleteAssetCollection(const AssetCollection& AssetCollection,
...
+			std::nullopt,							// ParentId
+			std::nullopt,							// GroupIds
+			std::nullopt,							// Types
+			true,									// HasGroup
+			std::nullopt,							// CreatedBy
+			std::nullopt,							// PrototypeOwnerIds
+			std::nullopt,							// ReadAccessFilters
+			std::nullopt,							// WriteAccessFilters
+			SourceSpaceId,							// originalGroupId
+			DestSpaceId,							// newGroupId
```
MAG-MichaelK, line 280: Ah, I see we're limited by CHS in regards to group ID.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -385,14 +459,8 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
 												const csp::common::Map<csp::common::String, csp::common::String>& NewMetadata,
 												AssetCollectionResultCallback Callback)
 {
-	csp::common::Optional<csp::common::String> SpaceId;
```
MAG-MichaelK, line 388: We need this to be optional because not all AssetCollections have a space ID. Specifically, I think space thumbnails don't.

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -234,7 +234,7 @@ public static void CreateAssetCollectionNoSpaceTest()
             GetAssetCollectionByName(assetSystem, testAssetCollectionName, out var assetCollection);
 
             Assert.AreEqual(assetCollection.Name, testAssetCollectionName);
-            Assert.IsTrue(assetCollection.SpaceIds.IsEmpty());
```
MAG-MichaelK, line 237: Again, this assertion was here specifically to ensure we could create AssetCollections with no space IDs.

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
             Assert.AreEqual(assets[0].GetThirdPartyPlatformType(), Systems.EThirdPartyPlatform.UNITY);
 
         }
+#endif
+#if RUN_ALL_UNIT_TESTS || RUN_ASSETSYSTEM_TESTS || RUN_ASSETSYSTEM_COPY_ASSET_COLLECTION_TEST
+        [Test]
+        public static void copyAssetCollectionTest()
```
MAG-MichaelK, line 1386: Casing

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
...
+                using var result = assetSystem.UploadAssetData(sourceAssetCollection, sourceAsset, fileSource).Result;
+                var resCode = result.GetResultCode();
+
+                Assert.AreEqual(resCode, Systems.EResultCode.Success);
+            }
+
+            // Create 'dest' space and invoke the copy
+            var destSpace = SpaceSystemTests.CreateSpace(spaceSystem, destSpaceName, testSpaceDescription, Systems.SpaceAttributes.Private, null, null, null);
+            var destAssetCollections = new Common.Array<Systems.AssetCollection>();
+            {
```
MAG-MichaelK, line 1423: Spacing

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
...
+                assetSystem.UploadAssetDataOnProgress += (s, e) => LogDebug($"Uploading asset... ({e.Progress}%)");
+                using var result = assetSystem.UploadAssetData(sourceAssetCollection, sourceAsset, fileSource).Result;
+                var resCode = result.GetResultCode();
+
+                Assert.AreEqual(resCode, Systems.EResultCode.Success);
+            }
+
+            // Create 'dest' space and invoke the copy
+            var destSpace = SpaceSystemTests.CreateSpace(spaceSystem, destSpaceName, testSpaceDescription, Systems.SpaceAttributes.Private, null, null, null);
+            var destAssetCollections = new Common.Array<Systems.AssetCollection>();
```
MAG-MichaelK, line 1422: This needs to be disposed once you are done with it. It can initially be set to `null` as it looks like it's just being re-assigned later without the initial value being used anyway.

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
...
+
+            // Validate the copied asset collection and its data
+	        {
+                Assert.AreEqual(destAssetCollections.Size(), 1UL);
+		        Assert.AreNotEqual(destAssetCollections[0].Id, sourceAssetCollection.Id);
+		        Assert.AreEqual(destAssetCollections[0].SpaceId, destSpace.Id);
+		        Assert.AreEqual(destAssetCollections[0].Type, sourceAssetCollection.Type);
+		        Assert.AreEqual(destAssetCollections[0].Tags.Size(), 1UL);
+
+                GetAssetsInCollection(assetSystem, destAssetCollections[0], out var destAssets);
```
MAG-MichaelK, line 1441: Dispose `destAssets` when finished with it.

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
...
+                byte[] downloadedData = new byte[result.GetDataLength()];
+                Marshal.Copy(result.GetData(), downloadedData, 0, (int)result.GetDataLength());
+
+                var fileData = File.ReadAllBytes(Path.GetFullPath("assets/test.json"));
+                Assert.IsTrue(downloadedData.SequenceEqual(fileData));
+            }
+
+            // Validating that we must have at least one asset collection to copy
+	        {
+		        Common.Array<Systems.AssetCollection> sourceAssetCollections = new Common.Array<Systems.AssetCollection>(1);
```
MAG-MichaelK, line 1457: `Common.Array<Systems.AssetCollection> sourceAssetCollections` should be `using var sourceAssetCollections` to both ensure it gets disposed and to reduce repetition.

Tests/CSharp/src/Tests/AssetSystemTests.cs
```diff
@@ -1380,6 +1380,110 @@ public static void thirdPartyPackagedAssetIdentifierTest()
...
+                using var result = assetSystem.CopyAssetCollectionsToSpace(sourceAssetCollections, destSpace.Id, false).Result;
+                Assert.AreEqual(result.GetResultCode(), Systems.EResultCode.Failed);
+	        }
+
+            // Validating we cannot perform a copy if the asset has no space ID
+            {
+                Common.Array<Systems.AssetCollection> assetCollections = new Common.Array<Systems.AssetCollection>(1);
```
MAG-MichaelK, line 1464: Same here as above in regards to `using var`.

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2168,4 +2162,156 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, DownloadAssetDataInvalidURLTest)
...
+		printf("Uploading source asset data...\n");
+
+		csp::common::String Uri;
+		UploadAssetData(AssetSystem, SourceAssetCollection, Asset, Source, Uri);
+	}
+
+	// Create 'dest' space and invoke the copy
+	csp::systems::Space DestSpace;
+	csp::common::Array<csp::systems::AssetCollection> DestAssetCollections;
+	{
```
MAG-MichaelK, line 2229: Spacing

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2168,4 +2162,156 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, DownloadAssetDataInvalidURLTest)
...
+		printf("Creating dest space and invoking the copy...\n");
+
+		CreateSpace(SpaceSystem, DestSpaceName, SpaceDescription, csp::systems::SpaceAttributes::Private, nullptr, nullptr, nullptr, DestSpace);
+
+		csp::common::Array<csp::systems::AssetCollection> SourceAssetCollections(&SourceAssetCollection, 1);
+		auto [Result] = AWAIT_PRE(AssetSystem, CopyAssetCollectionsToSpace, RequestPredicate, SourceAssetCollections, DestSpace.Id, false);
+
+		EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+		DestAssetCollections = Result.GetAssetCollections();
```
MAG-MichaelK, line 2238: Surround test macros with blank lines to separate them from other parts of the test.

Library/include/CSP/Systems/Assets/AssetCollection.h
MAG-SamBirley: This was discussed in yesterday's engineering meeting with all client teams, CSP and CHS present. We agreed on this approach unanimously.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-SamBirley: Yep

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-SamBirley: It's related to the upcoming asset library feature, which I have fleshing out in Unreal to help guide the PRD.

Library/src/Systems/Assets/AssetCollection.cpp
MAG-SamBirley: It allows us to reuse the same object type `AssetCollectionResult` for both queries that return a single asset collection, and those that return multiple asset collections.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-SamBirley: We unilaterally agreed in the all-teams engineering meeting yesterday to represent asset collections that do not belong to a space with an empty string.

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Please see the above notes regarding the decisions made in the engineering sync

Library/include/CSP/Common/Array.h
MAG-SamBirley: I don't remember the specific instance that highlighted the issue here, but the issue is that if the memory being copied describes a set of non-PoD objects which hold pointers to memory that they control the lifetime over, then this constructor was leading to inadvertently shared memory lifetime management (because the pointer addresses would be directly copied to the new array's objects).

So then if one array was destroyed and the objects it described had their destructors called, and had their memory freed, then the other array's objects would be pointing at memory that was no longer valid.

To address this, this change replaces the memcpy with a per-element assignment, which makes use of the object's  assignment operator instead (making it the object's concern to safely handle memory management & object lifetimes).

Library/src/Systems/Assets/AssetCollection.cpp
MAG-SamBirley: Addressed in latest

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Addressed in latest

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Addressed in latest

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Addressed in latest

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Addressed in latest

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Addressed in latest

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-MichaelK: I would prefer if we had a PRD ready to review before making changes required for said PRD.

Library/src/Systems/Assets/AssetCollection.cpp
MAG-MichaelK: I'm not sure I follow. This change looks like 2 different ways of handling JSON that returns multiple results.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-MichaelK: Is this change supported in CHS yet, though? I believe we were having permission issues before, which is what led to us needing to remove space ID for certain asset types.

Library/include/CSP/Common/Array.h
MAG-SamBirley: Since this change isn't directly related to the overall goals of the PR, I've restored it to using memcpy. This change is no longer part of the PR, so resolving the conversation.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-SamBirley: Hey! The PRD is already in review, and is pretty clear on the requirement for asset duplication, so this is definitely required.

Library/include/CSP/Systems/Assets/AssetCollection.h
MAG-SamBirley: Understood, however since the group consensus was to adopt this approach, and our services team are also considering doing similarly at their API level, I don't see any harm in including this part in the PR. It helps keeps the new `Copy` function cleaner too!

MAG-AdamThorn (approved):

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Resolved!

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Resolved!

Tests/CSharp/src/Tests/AssetSystemTests.cs
MAG-SamBirley: Resolved!

Library/src/Systems/Assets/AssetSystem.cpp
MAG-SamBirley: Resolved!

Library/src/Systems/Assets/AssetSystem.cpp
MAG-SamBirley: Resolved!

Library/src/Systems/Assets/AssetCollection.cpp
MAG-SamBirley: That's right. This function can now handle two different payloads, one which describes a single asset collection, and one that can return many.

PR #304, opened 2024-01-04
[OF-793] feat: add Avatar info settings

MAG-MichaelK:
This change adds `SettingsSystem::SetAvatarInfo()` and `SettingsSystem::GetAvatarInfo()` to allow clients to set persistent Avatar values that can be retrieved outside of the Multiplayer system.

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/Settings/SettingsSystem.h
+++ b/Library/include/CSP/Systems/Settings/SettingsSystem.h
@@ -135,9 +135,17 @@ public:
 														 const csp::systems::BufferAssetDataSource& NewAvatarPortrait,
 														 NullResultCallback Callback);
 
+	/// @brief Sets the avatar type and identifier for a user.
+	/// @param InUserId csp::common::String : The ID of the user to set avatar info for.
+	/// @param InType csp::systems::AvatarType : The type of avatar (predefined, Ready Player Me, or custom).
+	/// @param InIdentifier csp::common::Variant : A value used to identify or locate the avatar. Differs depending on the value of InType.
+	/// @param Callback NullResultCallback : Callback to call when task finishes.
 	CSP_ASYNC_RESULT void
 		SetAvatarInfo(const csp::common::String& InUserId, AvatarType InType, const csp::common::Variant& InIdentifier, NullResultCallback Callback);
 
+	/// @brief Retrieves the avatar type and identifier for a user.
+	/// @param InUserId csp::common::String : The ID of the user to get avatar info for.
+	/// @param Callback NullResultCallback : Callback to call when task finishes.
 	CSP_ASYNC_RESULT void GetAvatarInfo(const csp::common::String& InUserId, AvatarInfoResultCallback Callback);
 
 private:
--- a/Library/include/CSP/Systems/Settings/SettingsSystem.h
+++ b/Library/include/CSP/Systems/Settings/SettingsSystem.h
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #pragma once
 
 #include "CSP/Systems/Assets/Asset.h"
@@ -20,6 +21,7 @@
 #include "CSP/Systems/Settings/SettingsCollection.h"
 #include "CSP/Systems/SystemBase.h"
 #include "CSP/Systems/SystemsResult.h"
+
 
 namespace csp::services
 {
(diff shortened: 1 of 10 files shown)


Library/include/CSP/Systems/Settings/SettingsSystem.h
```diff
@@ -133,6 +135,11 @@ class CSP_API CSP_NO_DISPOSE SettingsSystem : public SystemBase
 														 const csp::systems::BufferAssetDataSource& NewAvatarPortrait,
 														 NullResultCallback Callback);
 
+	CSP_ASYNC_RESULT void
+		SetAvatarInfo(const csp::common::String& InUserId, AvatarType InType, const csp::common::Variant& InIdentifier, NullResultCallback Callback);
```
MAG-SamBirley, line 139: Don't forget inline documentation here!

Library/src/Systems/Settings/SettingsSystem.cpp
```diff
@@ -868,4 +872,66 @@ void SettingsSystem::RemoveAvatarPortrait(const csp::common::String& UserId, Nul
...
+	switch (InIdentifier.GetValueType())
+	{
+		case csp::common::VariantType::Integer:
+			Json.AddMember("identifier", InIdentifier.GetInt(), Json.GetAllocator());
+			break;
+		case csp::common::VariantType::String:
+			Json.AddMember("identifier", rapidjson::Value(InIdentifier.GetString().c_str(), InIdentifier.GetString().Length()), Json.GetAllocator());
+			break;
+		default:
+			throw std::exception("Unsupported Avatar Identifier type");
```
MAG-SamBirley, line 895: CSP doesn't currently follow a pattern of throwing exceptions when passed unsupported types. 

Since the current pattern is to print to the Error log and early out (random example [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/src/Systems/Users/UserSystem.cpp#L270)), could we change it follow the same pattern instead?

MAG-SamBirley (changes requested):
Lgtm, just a couple small change requests

MAG-DanielMurdolo:
> cc @MAG-DanielMurdolo - opens up the possibility of CSP clients giving users a means to have persistent avatar selections across devices & other CSP clients.

I did wonder if that's what it was for. Sign me up...

Library/include/CSP/Systems/Settings/SettingsSystem.h
MAG-MichaelK: Address in latest commit

Library/src/Systems/Settings/SettingsSystem.cpp
MAG-MichaelK: Address in latest commit

MAG-SamBirley (approved):

Library/src/Systems/Spaces/SpaceSystemHelpers.h
```diff
@@ -43,26 +43,27 @@ constexpr const char* SPACE_THUMBNAIL_ASSET_NAME_PREFIX			   = "SPACE_THUMBNAIL_
 
 constexpr const char* AVATAR_THUMBNAIL_ASSET_NAME_PREFIX = "AVATAR_THUMBNAIL";
 
-String GetSpaceMetadataAssetCollectionName(const String& SpaceId);
-Map<String, String> ConvertSpaceMetadataToAssetCollectionMetadata(const String& Metadata);
-String GetSpaceIdFromMetadataAssetCollectionName(const String& MetadataAssetCollectionName);
+csp::common::String GetSpaceMetadataAssetCollectionName(const csp::common::String& SpaceId);
```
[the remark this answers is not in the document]
MAG-MichaelK: This file was included in multiple source files, so the `using namespace` statement was also included with it. We typically try to avoid this as it can make namespaces unclear if the developer doesn't know he's importing a `using namespace`. This file went against that pattern and so was refactored.

Tests/src/InternalTests/SpaceHelperTests.cpp
```diff
@@ -25,6 +25,7 @@
 	#include "TestHelpers.h"
 
 
+using namespace csp::common;
```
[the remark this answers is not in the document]
MAG-MichaelK: This was needed due to it being removed from SpaceSystemHelpers.h

PR #312, opened 2024-01-24
[OB-3073] fix: ensure const `Map` is immutable

MAG-MichaelK:
Indexing into a const `Map` with a key that is not already present will no longer cause an allocation and will instead throw.

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Common/Map.h
+++ b/Library/include/CSP/Common/Map.h
@@ -60,7 +60,7 @@ public:
 		new (Container) MapType(*Other.Container);
 	}
 
-	/// @brief Constructs an map from an initializer_list.
+	/// @brief Constructs a map from a `std::initializer_list`.
 	/// @param Values const std::initializer_list<std::pair<const TKey, const TValue>> : Elements to construct the map from
 	CSP_NO_EXPORT Map(const std::initializer_list<std::pair<const TKey, const TValue>> Values)
 	{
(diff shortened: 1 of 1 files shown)


Library/include/CSP/Common/Map.h
```diff
@@ -81,19 +81,26 @@ template <typename TKey, typename TValue> class CSP_API Map
...
-	/// @param Key const TKey& : Elements key to access
+	/// @brief Returns a const reference to the element with the given key in this map.
+	///        Throws if the given key is not present.
+	/// @param Key const TKey& : Key of element in this map
 	/// @return const TValue& : Map element
 	const TValue& operator[](const TKey& Key) const
 	{
+		if (Container->count(Key) == 0)
+		{
+			throw std::runtime_error("Key not present in Map. Please ");
```
MAG-SamBirley, line 101: Looks like there's some missing text after 'Please '?

MAG-SamBirley: Perhaps the second sentence could read. "Please validate the key first exists using `HasKey` ".

MAG-MichaelK: I have no idea how this happened. I distinctly remember completing that sentence. Will fix now.

MAG-SamBirley (approved):
lgtm!

Library/include/CSP/Common/Map.h
```diff
@@ -129,22 +136,23 @@ template <typename TKey, typename TValue> class CSP_API Map
...
 	/// @param Key const TKey& : key to check if the map contains
 	/// @return bool
 	bool HasKey(const TKey& Key) const
 	{
 		return Container->count(Key) > 0;
 	}
 
-	/// @brief Returns all keys in the map.
+	/// @brief Returns a copy of all keys in this map.
+	///        This copy should be disposed by the caller once it is no longer needed.
```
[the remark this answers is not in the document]
MAG-MichaelK: It depends entirely on which client is using it, which is why I didn't include that information in the comments. Web and Unity should use the established disposal process, while Unreal should call `oly::memory::DllFree()`.

PR #320, opened 2024-01-26
[NT-0] fix: Remove distutils dependency

MAG-AdamThorn:
The build agents have been updated to Python 3.12
and support distutils has been removed with that release. Updating the script to use shutil.copy_tree instead.

**For the reviewer**

### DIFF
--- a/teamcity/PublishCSPForUnity.py
+++ b/teamcity/PublishCSPForUnity.py
@@ -8,7 +8,6 @@ import uuid
 
 import git
 
-from distutils import dir_util
 from pathlib import Path
 
 from jinja2 import Environment, FileSystemLoader
@@ -163,7 +162,7 @@ def main():
         f.write(minimal_meta_template.render(guid=generate_guid(), importer_type="Default"))
 
     for path in args.csharp_artifact_paths:
-        dir_util.copy_tree(path, source_output_directory)
+        shutil.copytree(path, source_output_directory)
 
     for path, dirs, files in os.walk(source_output_directory):
         for dir in dirs:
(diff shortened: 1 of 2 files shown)


teamcity/PublishCSPForUnity.py
```diff
@@ -163,7 +162,7 @@ def main():
         f.write(minimal_meta_template.render(guid=generate_guid(), importer_type="Default"))
 
     for path in args.csharp_artifact_paths:
-        dir_util.copy_tree(path, source_output_directory)
+        shutil.copytree(path, source_output_directory)
```
MAG-SamBirley, line 165: `shutil.copytree` is notorious for failing if the target directories already exist. Since Python 3.8 though, this can be accounted for by passing a `dirs_exit_ok` parameter. More info [here](https://stackoverflow.com/a/64609884).

I expect we want to change this line to be:
`shutil.copytree(path, source_output_directory, dirs_exist_ok=True)`

MAG-SamBirley: (I would expect our scripts to fail without making this change).

MAG-SamBirley (changes requested):

teamcity/PublishCSPForUnity.py
MAG-AdamThorn: Agreed, change made.

MAG-SamBirley (approved):

teamcity/PublishCSPForUnity.py
MAG-SamBirley: Looks good, thanks!

MAG-SamBirley (changes requested):
Ah just checked, you should make the same change to use `shutil` in `BuildNPMWebPackage.py` also, as this will otherwise lead to a similar failure on publish

PR #324, opened 2024-01-26
[OF-627] fix: IsModifiable flag impl

MAG-AW:
This PR includes functionality to prevent entity properties (and component properties) being modified when they should not be allowed to, when flagged as non transferable and the entity is not owned by the user actioning the change.

**For the reviewer**

### DIFF
--- a/Library/src/Multiplayer/ComponentBase.cpp
+++ b/Library/src/Multiplayer/ComponentBase.cpp
@@ -80,7 +80,7 @@ void ComponentBase::SetProperty(uint32_t Key, const ReplicatedValue& Value)
 	}
 
 	// If the entity is not owned by us, and not a transferable entity, it is not allowed to modify the entity.
-	if (GetParent()->IsModifiable() == false)
+	if (!GetParent()->IsModifiable())
 	{
 		CSP_LOG_ERROR_FORMAT("Error: Update attempted on a non-owned entity that is marked as non-transferable. Skipping update. Entity name: %s",
 							 GetParent()->GetName());
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -1066,7 +1066,14 @@ bool SpaceEntity::Deselect()
 
 bool SpaceEntity::IsModifiable()
 {
-	return (OwnerId == EntitySystem->GetMultiplayerConnection()->GetClientId() || IsTransferable);
+    if (EntitySystem != nullptr && EntitySystem->GetMultiplayerConnection() != nullptr)
+    {
+	    return (OwnerId == EntitySystem->GetMultiplayerConnection()->GetClientId() || IsTransferable);
+    }
+    else
+    {
+        return true;
+    }
 }
 
 bool SpaceEntity::InternalSetSelectionStateOfEntity(const bool SelectedState, uint64_t ClientID)


Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -145,6 +145,11 @@ const csp::common::String& SpaceEntity::GetName() const
 
 void SpaceEntity::SetName(const csp::common::String& Value)
 {
+	if (!IsModifiable())
+	{
+		return;
```
MAG-SamBirley, line 150: Should we print errors to the log as otherwise this may come cross to the client application a silent (and unexpected) failures? Appreciate that's a lot of code repetition though...

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -717,6 +717,14 @@ void SpaceEntitySystem::QueueEntityUpdate(SpaceEntity* EntityToUpdate)
 		return;
 	}
 
+	// If the entity is not owned by us, and not a transferable entity, it is not allowed to modify the entity.
+	if (EntityToUpdate->IsModifiable() == false)
```
MAG-MichaelK, line 721: Prefer `!value` over `value == false`

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1138,6 +1146,17 @@ void SpaceEntitySystem::ProcessPendingEntityOperations()
 
 			if (CurrentTime - PendingEntity->TimeOfLastPatch >= EntityPatchRate || !EntityPatchRateLimitEnabled)
 			{
+				// If the entity is not owned by us, and not a transferable entity, it is not allowed to modify the entity.
+				if (PendingEntity->IsModifiable() == false)
```
MAG-MichaelK, line 1150: Prefer `!value` over `value == false`

MAG-SamBirley (dismissed):

Library/src/Multiplayer/ComponentBase.cpp
```diff
@@ -79,6 +79,14 @@ void ComponentBase::SetProperty(uint32_t Key, const ReplicatedValue& Value)
 							 static_cast<uint32_t>(Value.GetReplicatedValueType()));
 	}
 
+	// If the entity is not owned by us, and not a transferable entity, it is not allowed to modify the entity.
+	if (GetParent()->IsModifiable() == false)
```
MAG-MichaelK, line 83: Another `!x` vs `x == false` here. Sorry, didn't see this one earlier.

Tests/src/InternalTests/SerialisationTests.cpp
```diff
@@ -55,35 +115,15 @@ CSP_INTERNAL_TEST(CSPEngine, SerialisationTests, SpaceEntityUserSignalRSerialisa
 	User->Serialise(Serialiser);
 	auto SerialisedUser = Serialiser.Finalise();
 
-	/*
```
MAG-MichaelK, line 58: Why was this comment removed? It's useful for if future developers need to modify these tests.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1066,7 +1066,14 @@ bool SpaceEntity::Deselect()
...
 bool SpaceEntity::IsModifiable()
 {
-	return (OwnerId == EntitySystem->GetMultiplayerConnection()->GetClientId() || IsTransferable);
+    if (EntitySystem != nullptr && EntitySystem->GetMultiplayerConnection() != nullptr)
+    {
+	    return (OwnerId == EntitySystem->GetMultiplayerConnection()->GetClientId() || IsTransferable);
+    }
+    else
+    {
+        return true;
```
MAG-SamBirley, line 1075: Suggestion - Perhaps a comment here to explain the rationale? That having no connection means that the local application has permission to mutate any state on any entity, whereas when connected the application can only mutate those entities which can be owned by the local user?

MAG-SamBirley (approved):

MAG-AdamThorn (approved):

PR #351, opened 2024-03-04
[NT-0] VisionOS Support for CSP

MAG-SamBirley:
This change introduces support for VisionOS to CSP.

Notable changes include:
* A new shell script for generating VisionOS xcodeprojects: `generate_solution_visionos.sh`. Run this, open the project and build to create visionos-compatible CSP libraries.
* Since premake does not yet support the device, the approach taken is to instruct premake to build for iOS, with an additional argument flagging that it should actually generate a project targeting visionos. Responding to this argument is entirely handled within the `premake5_helpers.lua` script.
* Some refactorings have been made around the Unity package generation pipeline to allow for Unity-compatible packages to be generated via MacOS. This greatly helps with local debugging and iteration.
* Mimalloc has been disabled for iOS and VisionOS. Whilst this has not been reported to be an issue on iOS, it causes problems on VIsionOS similar to what we have seen in the past for Android and WASM. I would recommend we strip it out entirely in the near future.
* A fix for a MacOS bug in our project generation scripts which always led to a loose file named '142' being created at the root directory level.

image


Library/premake5.lua
```diff
@@ -206,17 +206,14 @@ if not Project then
             defines { 
                 "CSP_IOS",
                 "USE_STD_MALLOC=1",
-                "JS_STRICT_NAN_BOXING"
+                "JS_STRICT_NAN_BOXING",
+                "USE_STD_MALLOC=1"
```
MAG-AdamThorn, line 210: You have "USE_STD_MALLOC" repeated.

MAG-AdamThorn (changes requested):

generate_solution_visionos.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ ! -f /modules/premake/README.md ]
+then
+	git submodule update --recursive
+fi
+
+
+if [ ! -f /modules/premake/bin/release/premake5 ]
+then
+	MIN_XCODE_VERSION="14.2"
```
lyubomirlichev, line 19: The first public release of visionOS was [introduced](https://xcodereleases.com/) in Xcode 15.2, it's very likely this won't work as expected in previous versions.

Library/premake5.lua
MAG-SamBirley: Thank you for saving me from myself 😅 Fixed in latest!

generate_solution_visionos.sh
MAG-SamBirley: I can dig that

MAG-SamBirley: Fixed in latest.

MAG-AdamThorn (approved):

PR #369, opened 2024-03-20
[OB-3308] refac: RefreshSession split into new function LogInWithToken

MAG-AW:

* RefreshSession is now internal and returns a NullResultCallback, it is not for `Login` usage.
* A new API function LogInWithRefreshToken is for clients to login using a refresh token and now includes MultiplayerConnection initialisation.

**For the reviewer**

### DIFF
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -565,6 +565,32 @@ void MultiplayerConnection::SetScopes(csp::common::String InSpaceId, ErrorCodeCa
 	Connection->Invoke("SetScopes", Params, LocalCallback);
 }
 
+void MultiplayerConnection::ResetScopes(ErrorCodeCallbackHandler Callback)
+{
+	if (Connection == nullptr || !Connected)
+	{
+		INVOKE_IF_NOT_NULL(Callback, ErrorCode::NotConnected);
+
+		return;
+	}
+
+	std::function<void(signalr::value, std::exception_ptr)> LocalCallback = [this, Callback](signalr::value Result, std::exception_ptr Except)
+	{
+		if (Except != nullptr)
+		{
+			auto Error = ParseError(Except);
+			INVOKE_IF_NOT_NULL(Callback, Error);
+
+			return;
+		}
+
+		INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
+	};
+
+    std::vector<signalr::value> ParamsVec;
+	signalr::value Params = signalr::value(std::move(ParamsVec));
+	Connection->Invoke("ResetScopes", Params, LocalCallback);
+}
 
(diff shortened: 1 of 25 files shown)


MAG-SamBirley (approved):

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -632,7 +632,14 @@ ConversationSystem* MultiplayerConnection::GetConversationSystem() const
 
 ConnectionState MultiplayerConnection::GetConnectionState() const
 {
-	return static_cast<ConnectionState>(Connection->GetConnectionState());
+	if (Connection != nullptr)
+	{
+		return static_cast<ConnectionState>(Connection->GetConnectionState());
```
lyubomirlichev, line 637: If my understanding is correct, this will not create a multiplayer connection for returning users. How will this affect a user opening and logging in to a client application with a non-expired, valid authentication or refresh token from a previous login?

Using this API:
Screenshot 2024-03-21 at 08 47 59

lyubomirlichev: Similar question for the Exchange Key login as well:
Screenshot 2024-03-21 at 09 00 13

MAG-ThomasGreenhalgh: I think there could be an issue as things stand. For example on web in case where:
1. User logs in to OKO
2. User closes the tab
3. User opens OKO in another tab

At step 3, currently the web client will restore the session using `refreshSession` with token information the web client previously saved to local storage. There won't be any call to login.

Is it the case that no multiplayer connection would be created in this scenario, unless login is called?

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -177,17 +178,86 @@ void UserSystem::Login(const csp::common::String& UserName,
...
 
-void UserSystem::RefreshSession(const csp::common::String& UserId, const csp::common::String& RefreshToken, LoginStateResultCallback Callback)
+void UserSystem::LoginWithRefreshToken(const csp::common::String& UserId, const csp::common::String& RefreshToken, LoginStateResultCallback Callback)
 {
-	auto Request = std::make_shared<chs_user::RefreshRequest>();
-	Request->SetDeviceId(csp::CSPFoundation::GetDeviceId());
-	Request->SetUserId(UserId);
-	Request->SetRefreshToken(RefreshToken);
+	if (CurrentLoginState.State == ELoginState::LoggedOut || CurrentLoginState.State == ELoginState::Error)
+	{
```
lyubomirlichev, line 184: Does this also need a `CurrentLoginState.State = ELoginState::LoginRequested;` at the start ?

lyubomirlichev (approved):
Looks good to me 🙂

MAG-SamBirley (approved):
Please update the PR title and description, but otherwise approved!

PR #372, opened 2024-03-26
[NT-0] fix: Local entities are destroyed upon exiting a space

MAG-SamBirley:
As part of a recent broad refactor to the multiplayer connection's application lifetime, the connection now persists for the duration of the application's lifetime. Entering a space now narrows the connections focus to listen to entities within the space, and exiting stops the connection from listening to entities within the space.

A bug has arisen however where whilst leaving a space stops the application receiving updates regarding entities, it does not clear CSP's local view of currently known entities.

This change introduces a new function to SpaceEntitySystem, `LocalDestroyAllEntities` which is invoked by the SpaceSystem after it has finished listening for entities within the space. It clear's CSP's local view of all currently known entities.

Tests have been updated to validate this behaviour.

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
@@ -264,6 +264,11 @@ public:
 	/// It is highly advised not to call this function unless you know what you are doing.
 	void RetrieveAllEntities();
 
+	/// @brief Destroys the client's local view of all currently known entities.
+	///
+	/// They still reside on the server, however they will not be accessible in the client application.
+	void LocalDestroyAllEntities();
+
 	/// @brief Sets the selected state of an entity, if the operation is acceptable.
 	///
 	/// Criteria:
--- a/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
@@ -325,6 +325,10 @@ protected:
 	using SpaceEntityList = csp::common::List<SpaceEntity*>;
 
 	SpaceEntityList Entities;
+	SpaceEntityList Avatars;
+	SpaceEntityList Objects;
+	SpaceEntityList SelectedEntities;
+
 	std::recursive_mutex* EntitiesLock;
 
 private:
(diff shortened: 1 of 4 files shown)


Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -264,6 +264,11 @@ class CSP_API SpaceEntitySystem
 	/// It is highly advised not to call this function unless you know what you are doing.
 	void RetrieveAllEntities();
 
+	/// @brief Destroys the client's local view of all currently known entities.
+	///
+	/// They still reside on the server, however they will not be accessible in the client application.
+	void LocalDestroyAllEntities();
```
lyubomirlichev, line 270: Does a client app need to call this manually ?

MAG-AdamThorn: @lyubomirlichev it is called below in the SpaceSystem.

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -320,6 +325,10 @@ class CSP_API SpaceEntitySystem
 	using SpaceEntityList = csp::common::List<SpaceEntity*>;
 
 	SpaceEntityList Entities;
+	SpaceEntityList Avatars;
```
MAG-AdamThorn, line 328: Are these being made public for the multiplayertests? Can that test class not be made a friend rather?

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
MAG-SamBirley: That's right, a client does _not_ need to call this, but SpaceSystem needs access to it in order to invoke it on exiting a space.

I elected to make it public as it's the symmetric opposite of `RetreiveAllEntities` just above, which is also public; it felt neater than adding a new `friend` qualifier.

lyubomirlichev: Thank you, sounds good 👍

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
MAG-AdamThorn: Ah right, missed that they were protected.

PR #385, opened 2024-04-05
[NT-0] feat: gaussian splat asset type

MAG-SamBirley:
In service of affording users with the ability to upload gaussian splat files, CSP introduces a new asset type `GAUSSIAN_SPLAT`.

The change also includes test coverage to validate that the new asset type is returned as expected when an asset of this type is created. It does not validate we can actually upload a splat file, since they're typically in the order of 100s of MBs.


MAG-AdrianMeredith:
Looks good to me

Tests/src/PublicAPITests/ComponentTests/GaussianSplatComponentTests.cpp
```diff
@@ -226,4 +215,69 @@ CSP_PUBLIC_TEST(CSPEngine, GaussianSplatTests, GaussianSplatScriptInterfaceTest)
...
+
+	char UniqueAssetName[256];
+	SPRINTF(UniqueAssetName, "%s-%s", TestAssetName, GetUniqueString().c_str());
+
+	// Create asset collection
+	csp::systems::AssetCollection AssetCollection;
+	CreateAssetCollection(AssetSystem, Space.Id, nullptr, UniqueAssetCollectionName, nullptr, nullptr, AssetCollection);
+
+	// Create asset
+	auto [Result] = Awaitable(&csp::systems::AssetSystem::CreateAsset,
```
MAG-MichaelK, line 258: I don't think we really need to test this every time. As long as you've tested it locally, I'm fine with excluding this test.

MAG-SamBirley:
> Before merging this, do we have the correct mime data type in CSP to handle Gaussian Splats? What file type/s do they use and what mime type is required?
> 
> CSP mime type helpers are here:
> 
> https://github.com/magnopus-opensource/connected-spaces-platform/blob/9f15b459d8f1c1c54b46a49086d35d634d960883/Library/include/CSP/Common/MimeTypeHelper.h#L42

This is an excellent question and I don't know the answer. @MAG-AdrianMeredith @MAG-AlessioRegalbuto any insights here?

Tests/src/PublicAPITests/ComponentTests/GaussianSplatComponentTests.cpp
MAG-SamBirley: I'm on the fence - there's a good argument about the redundant aspect of most of the test. Equally from an integration test viewpoint, there's some value to it.

Whilst I agree with @MAG-AW that the component will likely evolve fast in the near future, I don't expect this test itself to change.

Going to make the call and elect to remove it.

MAG-SamBirley: New test is gone on latest revision. 🤘

MAG-AdrianMeredith:
> > Unfortunately it doesn't have one https://en.wikipedia.org/wiki/PLY_(file_format)
> 
> That link shows it as "text/plain", which we already support (csp::common::MimeTypes::TEXT_PLAIN), so this is fine.

My point is Unity loves to use the mimetype as the definitive source of what the type of object it is.  This wont work here, even the ply file itself  doesn't say anything about being gaussian splat.  We'd have to parse the fields and check to see whether they line up with what we're expecting

MAG-MichaelK (approved):

PR #390, opened 2024-04-10
[OB-3269] fix: Iterate over components correctly

MAG-ME:
[OB-3269] fix: Iterate over components correctly

- Fix issue in `SpaceEntity::FindFirstComponentOfType` which would only work correctly if a component had not been deleted from the entity
- The code was updated to iterate over the maps in question in a way consistent with other similar use cases in this class
- STL style iterators were added to the Map class as part of the TravelZoo work in this [PR](https://github.com/magnopus/connected-spaces-platform-internal/pull/4)
- These would probably be the best way to iterate over maps going forward

[OB-3269]: https://magnopus.atlassian.net/browse/OB-3269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -27,6 +27,7 @@
 #include "CSP/Multiplayer/Components/ExternalLinkSpaceComponent.h"
 #include "CSP/Multiplayer/Components/FiducialMarkerSpaceComponent.h"
 #include "CSP/Multiplayer/Components/FogSpaceComponent.h"
+#include "CSP/Multiplayer/Components/GaussianSplatSpaceComponent.h"
 #include "CSP/Multiplayer/Components/ImageSpaceComponent.h"
 #include "CSP/Multiplayer/Components/LightSpaceComponent.h"
 #include "CSP/Multiplayer/Components/PortalSpaceComponent.h"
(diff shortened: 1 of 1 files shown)


Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1131,32 +1131,45 @@ void SpaceEntity::DestroyComponent(uint16_t Key)
...
-		for (int i = 0; i < DirtyComponents.Size(); ++i)
+		const csp::common::Array<uint16_t>* DirtyComponentKeys = DirtyComponents.Keys();
+
+		for (i = 0; i < DirtyComponentKeys->Size(); ++i)
 		{
-			DirtyComponent Component = DirtyComponents[i];
+			const DirtyComponent& Component = DirtyComponents[DirtyComponentKeys->operator[](i)];
 
-			if (Component.Component->GetComponentType() == ComponentType::ScriptData)
+			if (Component.Component->GetComponentType() == Type)
```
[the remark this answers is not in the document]
MAG-ME: Thanks, I've added the check

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1131,32 +1131,45 @@ void SpaceEntity::DestroyComponent(uint16_t Key)
 
 ComponentBase* SpaceEntity::FindFirstComponentOfType(ComponentType Type, bool SearchDirtyComponents) const
 {
-	auto& CheckComponents = *GetComponents();
+	auto& CheckComponents							  = *GetComponents();
+	const csp::common::Array<uint16_t>* ComponentKeys = Components.Keys();
 
-	for (int i = 0; i < Components.Size(); ++i)
+	int i							= 0;
```
MAG-AdamThorn, line 1137: Why are you declaring 'i' here and not in the for loop itself?

MAG-ME: Old habit, refactored as suggested

MAG-AdamThorn (approved):

PR #397, opened 2024-04-17
[OF-1272] feat: Altered variant collection api

MAG-AW:
* Changed GetProductInfoCollectionByVariantIds to return an array which contains a product with a single variant, per variantID given.

**For the reviewer**

### DIFF
--- a/Tests/src/PublicAPITests/UserSystemTests.cpp
+++ b/Tests/src/PublicAPITests/UserSystemTests.cpp
@@ -98,7 +98,7 @@ void LogIn(csp::systems::UserSystem* UserSystem,
 
 void LogInAsGuest(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, csp::systems::EResultCode ExpectedResult)
 {
-	auto [Result] = Awaitable(&csp::systems::UserSystem::LoginAsGuest, UserSystem, nullptr).Await(RequestPredicate);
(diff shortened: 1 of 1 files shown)


Library/src/Systems/ECommerce/ECommerce.cpp
```diff
@@ -189,6 +189,175 @@ void ProductInfoDtoToProductInfo(const chs_aggregation::ShopifyProductDto& Dto,
...
+					CSP_LOG_MSG(LogLevel::Log, "ShopifyProductDto missing Image");
+				}
+
+				if (VariantProductInformation->HasSelectedOptions())
+				{
+					auto VariantOptionInformation = VariantProductInformation->GetSelectedOptions();
+
+					ProductInfoCollection[TotalVariantIndex].Variants[0].Options = common::Array<VariantOptionInfo>(VariantOptionInformation.size());
+
+					for (int n = 0; n < VariantOptionInformation.size(); ++n)
```
MAG-AdamThorn, line 266: You have inconsistent placement of the increment expression. Here and below it is prefixed, above for VariantCount it is postfixed. Also is there a reason you are calling the incremented variable 'n' here and 'j' below? There is no 'i' that I can see.

MAG-SamBirley (approved):
@MAG-DanielMurdolo has confirmed that the change works as expected, and the change looks good to me. Approved!

MAG-AdamThorn (approved):

PR #403, opened 2024-04-29
[OB-3444] fix: Log messages emitted under multiplayer connection failures do not crash the application

MAG-SamBirley:
Fixed a number of cases where the format string passed to CSP log messages was inconsistent with the type of the arguments being passed in.

Log messages emitted when multiplayer errors are encountered now include a string representation of the error code.


Library/src/Multiplayer/ErrorCodeStrings.h
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+#include <string>
+#include "CSP/Multiplayer/MultiPlayerConnection.h"
+
+namespace csp::multiplayer
+{
+    static std::string ToString(csp::multiplayer::ErrorCode ErrorCode)
```
MAG-MichaelK, line 23: This could be done more efficiently by using the enum value to index into a static string array.

MAG-MichaelK: I'm also not sure this should be in this namespace if it's just called "ToString". Best to call it "ErrorCodeToString" or make this a generic solution.

MAG-MichaelK: Also, make it inline :)

MAG-MichaelK (changes requested):

Library/src/Multiplayer/ErrorCodeStrings.h
MAG-SamBirley: All great feedback. Thank you both!

MAG-SamBirley:
> What's the deal with the commit history here? Seems to have lots of unrelated commits, though those changes must already be in develop. I would appreciate if this was tidied in case we end up with unnecessary history.

Roger that. Will squash & merge when we get there if that works for you.

Library/src/Multiplayer/ErrorCodeStrings.h
MAG-SamBirley: @MAG-MichaelK if you don't mind, I'll not convert to an array of strings indexed into via the enum, because it would mean we cannot as neatly respond to scenarios where the function is unaware of new error code enum values we introduce in the future.

image

Since this isn't a hot code path, and since I wouldn't expect the perf characteristics of the switch statement approach to be much of a problem, I'd prefer to keep the function as-is.

MAG-SamBirley: Most recent commit addresses function name & inline feedback. 👍

MAG-AdamThorn (approved):

PR #406, opened 2024-05-09
[NT-0] refac: Refactor Org test accounts

MAG-AdamThorn:
I have made the following quality of life improvements to remove the requirement to create/add new org test accounts to the test_account_creds.txt file: 
- I have reverted the test_account_creds.txt so that it only requires the default and alt email address and password information.
- I have reduced the Organization test account requirements down to a single test account with the member and admin roles.
- I have introduced a new organization_test_account_creds.txt file which contains the credentials for this new test account.
- I have updated the Functional Test TeamCity deployment steps to account for this.

**For the reviewer**

### DIFF
--- a/Tests/src/PublicAPITests/OrganizationSystemTests.cpp
+++ b/Tests/src/PublicAPITests/OrganizationSystemTests.cpp
@@ -18,18 +18,19 @@
 #include "CSP/Common/Array.h"
 #include "CSP/Multiplayer/MultiPlayerConnection.h"
 #include "CSP/Systems/Assets/AssetSystem.h"
-#include "CSP/Systems/Spaces/SpaceSystem.h"
 #include "CSP/Systems/Organizations/OrganizationSystem.h"
+#include "CSP/Systems/Spaces/SpaceSystem.h"
 #include "CSP/Systems/SystemsManager.h"
 #include "CSP/Web/HTTPResponseCodes.h"
+#include "Debug/Logging.h"
 #include "SpaceSystemTestHelpers.h"
 #include "TestHelpers.h"
 #include "UserSystemTestHelpers.h"
-#include "Debug/Logging.h"
 
 #include "gtest/gtest.h"
 #include <algorithm>
 #include <filesystem>
+#include <fstream>
 
 
 using namespace csp::common;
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/OrganizationSystemTests.cpp
```diff
@@ -59,16 +85,16 @@ void GetUsersRoles(::OrganizationSystem* OrganizationSystem, const csp::common::
...
 	InviteOrganizationRoleInfo InviteUser2;
-	InviteUser2.UserEmail = AltUser2NonMemberEmail;
-	InviteUser2.OrganizationRoles  = {EOrganizationRole::Member};
+	InviteUser2.UserEmail		  = AlternativeLoginEmail;
+	InviteUser2.OrganizationRoles = {EOrganizationRole::Member};
 
 	InviteOrganizationRoleCollection OrganizationInvites;
-	OrganizationInvites.EmailLinkUrl = "https://dev.magnoverse.space";
-	OrganizationInvites.SignupUrl = "https://dev.magnoverse.space";
+	OrganizationInvites.EmailLinkUrl	 = "https://dev.magnoverse.space";
```
MAG-SamBirley, line 96: Question - do we think this would be better data-driven, in the same way as account creds? The test assumes the developer is working with magnopus-internal backend environments, which is why I ask.

MAG-SamBirley (commented):
Looks great other than that one question!

MAG-MichaelK:
Why do we need 3 test accounts for this? I would much rather we not introduce another test accounts txt file requirement and just use the existing 2 accounts.

Tests/src/PublicAPITests/OrganizationSystemTests.cpp
MAG-AdamThorn: Yes I agree that could/should be data driven. Will require changes to a few tests (at least User System Tests), so I would rather create a separate ticket for that work. You happy with that?

MAG-SamBirley: Yup!

MAG-SamBirley (approved):

MAG-AdamThorn:
We can't do that unfortunately. We need an account that belongs to the organization and has at least admin privileges so that they can perform organization operations such as invite other members. We can also not grant one of existing accounts that role as part of the test as they need to be invited by another user who is an admin, owner or superuser.

MAG-MichaelK (approved):

PR #409, opened 2024-05-14
[OB-3438] fix: ensure context clear on ExitSpace

MAG-MichaelK:
This change ensures that entities are deleted and leader election is reset when exiting a space.

**For the reviewer**

### DIFF
--- a/Library/src/Systems/Spaces/SpaceSystem.cpp
+++ b/Library/src/Systems/Spaces/SpaceSystem.cpp
@@ -19,13 +19,13 @@
 #include "CSP/CSPFoundation.h"
 #include "CSP/Common/StringFormat.h"
 #include "CSP/Multiplayer/MultiPlayerConnection.h"
-#include "Multiplayer/ErrorCodeStrings.h"
 #include "CSP/Systems/Assets/AssetSystem.h"
 #include "CSP/Systems/SystemsManager.h"
 #include "CSP/Systems/Users/UserSystem.h"
 #include "CallHelpers.h"
 #include "Debug/Logging.h"
 #include "Events/EventSystem.h"
+#include "Multiplayer/ErrorCodeStrings.h"
 #include "Services/AggregationService/Api.h"
 #include "Services/AggregationService/Dto.h"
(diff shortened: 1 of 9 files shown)


Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -260,44 +269,50 @@ void SpaceSystem::ExitSpace(NullResultCallback Callback)
...
-						if (Error != csp::multiplayer::ErrorCode::None)
+						if (Error != multiplayer::ErrorCode::None)
 						{
-						    CSP_LOG_ERROR_FORMAT("Error on exiting spaces whilst clearing scopes, ErrorCode: %s", csp::multiplayer::ErrorCodeToString(Error).c_str());
-                            INVOKE_IF_NOT_NULL(Callback, MakeInvalid<NullResult>());
+							CSP_LOG_ERROR_FORMAT("Error on exiting spaces whilst clearing scopes, ErrorCode: %s",
+												 multiplayer::ErrorCodeToString(Error).c_str());
+							INVOKE_IF_NOT_NULL(Callback, MakeInvalid<NullResult>());
+
+							return;
```
MAG-SamBirley, line 301: Nitpick - please could we follow an if/else pattern here rather than earlying out?

MAG-SamBirley (approved):
Great change! Loads of good app lifecycle hygiene here. 🎉

MAG-AdamThorn (dismissed):

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-MichaelK: We specifically don't do that in most of our code. We always early out for errors.

MAG-SamBirley: That's fair. From a quick scan of `SpaceSystemc.pp` I agree it is used more frequently. Resolving!

MAG-SamBirley (approved):

MAG-AdamThorn (approved):

PR #420, opened 2024-05-29
[OF-1311] feat: Text components

MAG-IY:
Client applications can now create, update and delete text components that belong to an entity.

Text components allow the client application to reason about a body of text that should appear at a particular location in a space, along with its billboarding behaviour, size, background and text color.


Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
```diff
@@ -0,0 +1,155 @@
...
+	Height,
+	BillboardMode,
+	IsVisible,
+	IsARVisible,
+	Num
+};
+
+
+/// @brief The billboard mode supported by this text space component.
+enum class TextSpaceComponentBillboardMode
```
MAG-SamBirley, line 52: Should we refactor [this enum](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/include/CSP/Multiplayer/Components/ImageSpaceComponent.h#L51) so that it becomes shared with image components and text components, rather than introducing a new one with the same values?

MAG-SamBirley (changes requested):
One change request, but it's looking good! @MAG-DanielMurdolo @lyubomirlichev mind checking this meets your expectations?

MAG-SamBirley (changes requested):
I think we're also missing a TextComponent section in [EntityScriptBinding.cpp](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/src/Multiplayer/Script/EntityScriptBinding.cpp#L460), which is needed to associate the defined component properties with the strings creators will use to get/set the properties via script.

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-IY: I wasn't sure what to do about this, as there is indeed a BillboardMode enum class elsewhere. Clean code practices dictate promotion to it's own file, but I was trying to exercise a light touch on this ticket. If the consensus is to refactor, I am happy to do it.

MAG-SamBirley: Promoting to its own file makes sense to me!

MAG-AdamThorn:
While you were away Alex refactored the [ComponentBase type getters](https://github.com/magnopus-opensource/connected-spaces-platform/commit/e2d030cbec15130fe96447f11049932efe901d2c). You will need to get latest and update your TextSpaceComponent getters to align with that. Thanks

MAG-SamBirley (commented):
All previous feedback addressed 🎉. One more question on the latest commit.

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-SamBirley: Changes in latest look great, thanks!

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
```diff
@@ -249,34 +166,27 @@ void TextSpaceComponent::SetIsVisible(bool Value)
...
+	return GetBooleanProperty(static_cast<uint32_t>(TextPropertyKeys::IsARVisible));
 }
 
 void TextSpaceComponent::SetIsARVisible(bool Value)
 {
 	SetProperty(static_cast<uint32_t>(TextPropertyKeys::IsARVisible), Value);
 }
 
-TextSpaceComponentBillboardMode TextSpaceComponent::GetBillboardMode() const
+BillboardMode TextSpaceComponent::GetBillboardMode() const
```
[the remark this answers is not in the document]
MAG-IY: Could you explain?

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
```diff
@@ -0,0 +1,147 @@
...
+	/// @brief Sets the Text Width.
+	/// @param Value The Text Width.
+	void SetWidth(float InValue);
+
+	/// @brief Sets the Text Height.
+	/// @param Value The Text Height.
+	float GetHeight() const;
+	/// @brief Sets the Text Height.
+	/// @param Value The Text Height.
+	void SetHeight(bool InValue);
```
MAG-DanielMurdolo, line 104: InValue should be float?

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
```diff
@@ -0,0 +1,193 @@
...
+{
+	SetProperty(static_cast<uint32_t>(TextPropertyKeys::Width), InValue);
+}
+
+float TextSpaceComponent::GetHeight() const
+{
+	return GetFloatProperty(static_cast<uint32_t>(TextPropertyKeys::Height));
+}
+
+void TextSpaceComponent::SetHeight(bool InValue)
```
MAG-DanielMurdolo, line 102: InValue should be float?

MAG-DanielMurdolo (changes requested):

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
```diff
@@ -0,0 +1,147 @@
...
+	/// @param Value The text color.
+	void SetTextColor(const csp::common::Vector3& Value);
+
+	/// @brief Gets the background color.
+	/// @return The background color.
+	const csp::common::Vector3& GetBackgroundColor() const;
+
+	/// @brief Sets the background color.
+	/// @param Value The background color.
+	void SetBackgroundColor(const csp::common::Vector3& Value);
```
lyubomirlichev, line 83: I wonder whether it's worth having colour(for background and text) as a Vector4 so we open up a channel for transparency ? Not saying we'd use it right away but for background for example it could be a meaningful factor as part of a creator workflow.

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
```diff
@@ -0,0 +1,147 @@
...
+
+	/// @brief Sets the background visibility.
+	/// @param Value The background visibility.
+	bool GetIsBackgroundVisible() const;
+	/// @brief Sets the background visibility.
+	/// @param Value The background visibility.
+	void SetIsBackgroundVisible(float InValue);
```
lyubomirlichev, line 90: The name suggests a bool parameter, a binary yes or no action. If it's meant to be a float, my suggestion would be to reword the name to SetBackgroundVisibility, include comments to describe whether its a 0-1 range. If it's meant to be a bool, then I digress.

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
```diff
@@ -0,0 +1,186 @@
...
+namespace csp::multiplayer
+{
+
+TextSpaceComponent::TextSpaceComponent(SpaceEntity* Parent) : ComponentBase(ComponentType::Image, Parent)
+{
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Position)]			 = csp::common::Vector3::Zero();
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Rotation)]			 = csp::common::Vector4 {0, 0, 0, 1};
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Scale)]				 = csp::common::Vector3::One();
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Text)]				 = "";
+	Properties[static_cast<uint32_t>(TextPropertyKeys::TextColor)]			 = csp::common::Vector3::Zero();
```
lyubomirlichev, line 32: I believe the default text colour should be white as per the [PRD](https://magnopus.atlassian.net/wiki/spaces/OL/pages/2737963046/Text+Component#:~:text=Default%20will%20be%20black), however a vector3 zero is black I believe.

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
```diff
@@ -0,0 +1,147 @@
...
+	/// @return The text this text component refers to.
+	const csp::common::String& GetText() const;
+
+	/// @brief Sets the text this text component refers to.
+	/// @param Value The text this text component refers to.
+	void SetText(const csp::common::String& Value);
+
+	/// @brief Gets the text color.
+	/// @return The text color.
+	const csp::common::Vector3& GetTextColor() const;
```
lyubomirlichev, line 71: Can we please have more details on the format of the colour similarly to how GaussianSplatComponent details the tint colour ? It would definitely help with integration.

Example from GaussianSplatComponent
`/// @brief Sets the tint that should be globally applied to the Gaussian Splat.
	/// @param Value The tint value, expected to be in RGB color space, with each value normalised between 0...1.
	/// Defaults to 1,1,1.`

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-MGN: Also the corresponding getter is returning a bool, so I'd expect `InValue` to be a bool.

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
```diff
@@ -0,0 +1,186 @@
...
+	Properties[static_cast<uint32_t>(TextPropertyKeys::BackgroundColor)]	 = csp::common::Vector3::Zero();
+	Properties[static_cast<uint32_t>(TextPropertyKeys::IsBackgroundVisible)] = true;
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Width)]				 = 1.0f;
+	Properties[static_cast<uint32_t>(TextPropertyKeys::Height)]				 = 1.0f;
+	Properties[static_cast<uint32_t>(TextPropertyKeys::BillboardMode)]		 = static_cast<int64_t>(BillboardMode::Off);
+	Properties[static_cast<uint32_t>(TextPropertyKeys::IsVisible)]			 = false;
```
lyubomirlichev, line 38: I think this should be visible by default ? Unless I am misunderstanding and these values are not defaults.

lyubomirlichev (changes requested):
Requested a few changes, and have a few suggestions.

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-SamBirley: No you're right, visible should be true by default, as should visible in AR.

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-SamBirley: Good catch!

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-IY: Seems reasonable, but I would personally wait until there is a call for it in a ticket.

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-IY: How are color channels represented in data in CSP? 0.0f-1.0f, or 0.0f-255.0f?

MAG-SamBirley: 0-1, where 0 is black and 1 is white.

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-SamBirley: Addressed in latest 👍

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-SamBirley: Addressed in latest 👍

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-SamBirley: Addressed in latest 👍

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
lyubomirlichev: Fair enough 👍

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
```diff
@@ -29,13 +29,13 @@ TextSpaceComponent::TextSpaceComponent(SpaceEntity* Parent) : ComponentBase(Comp
...
-	Properties[static_cast<uint32_t>(TextPropertyKeys::TextColor)]			 = csp::common::Vector3::Zero();
+	Properties[static_cast<uint32_t>(TextPropertyKeys::TextColor)]			 = csp::common::Vector3(1.0f, 1.0f, 1.0f);
 	Properties[static_cast<uint32_t>(TextPropertyKeys::BackgroundColor)]	 = csp::common::Vector3::Zero();
 	Properties[static_cast<uint32_t>(TextPropertyKeys::IsBackgroundVisible)] = true;
 	Properties[static_cast<uint32_t>(TextPropertyKeys::Width)]				 = 1.0f;
 	Properties[static_cast<uint32_t>(TextPropertyKeys::Height)]				 = 1.0f;
 	Properties[static_cast<uint32_t>(TextPropertyKeys::BillboardMode)]		 = static_cast<int64_t>(BillboardMode::Off);
-	Properties[static_cast<uint32_t>(TextPropertyKeys::IsVisible)]			 = false;
+	Properties[static_cast<uint32_t>(TextPropertyKeys::IsVisible)]			 = true;
 	Properties[static_cast<uint32_t>(TextPropertyKeys::IsARVisible)]		 = false;
```
MAG-SamBirley, line 39: Please could we make this one default to true too?

Library/include/CSP/Multiplayer/Components/TextSpaceComponent.h
MAG-SamBirley: Addressed in latest 👍

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-SamBirley: Addressed in latest 👍

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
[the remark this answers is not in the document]
MAG-SamBirley: Addressed in latest 👍

Library/src/Multiplayer/Components/TextSpaceComponent.cpp
MAG-SamBirley: Addressed in latest (opened a separate conversation for the AR one) 👍

lyubomirlichev (approved):
LGTM 🎉

PR #421, opened 2024-05-31
[OF-1322] refac: Update generate shell scripts

MAG-AdamThorn:
As of Python 3.12+ it is a requirement that we use a venv to install modules. The generate_solution_X.sh scripts for mac, ios and visionos have all been updated to check for the version of Python installed, and if >= 3.12, to create and activate a venv before proceeding with module installation.

**For the reviewer**


MAG-SamBirley (approved):
Confirmed works on python 3.11.6 too! 🎉

MAG-IY (approved):
Accepted. No comments

PR #422, opened 2024-05-31
[OF-1172] feat: Querying for ecommerce product variants now also returns the available stock count

MAG-AW:
CSP ECommerce API endpoints that return product variants now also include a new member variable that tells the client application how many items the store currently has in stock.

### DIFF
--- a/Library/include/CSP/Systems/Organizations/Organization.h
+++ b/Library/include/CSP/Systems/Organizations/Organization.h
@@ -91,7 +91,6 @@ public:
 	csp::common::String CreatedAt;
 	csp::common::String CreatedBy;
 	csp::common::String Name;
-	csp::common::String Description;
 	csp::common::Array<OrganizationRoleInfo> Members;
 	int32_t SpaceCount;
 };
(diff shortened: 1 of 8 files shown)


Library/include/CSP/Systems/ECommerce/ECommerce.h
```diff
@@ -103,6 +102,8 @@ class CSP_API ProductVariantInfo
 	csp::common::Array<VariantOptionInfo> Options;
 	/// @brief Unit price for the variant
 	CurrencyInfo UnitPrice;
+    /// @brief Quantity of variant available
+	int32_t QuantityAvailable;
```
MAG-SamBirley, line 106: One small change request, the [AC](https://magnopus.atlassian.net/browse/OF-1172) asks for the variable exposed via the public API to be named `AvailableStock` - please could we update it to that?

MAG-SamBirley (changes requested):

MAG-MichaelK (approved):

PR #436, opened 2024-07-04
[OF-1303] feat: support for maintainance window url

MAG-IY:
* Add running switch to Maintainance system (active by default)
* Add maintainance URL to ClientUserAgent info
* Change maintainance API to consume client supplied url
* turn off(deactivate) maintainance system if query fails.

### DIFF
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -55,6 +55,9 @@ public:
 
 	// @brief Magnopus services environment. e.g odev.
 	csp::common::String CHSEnvironment;
+
+	// @brief Maintainance Window Url.
+	csp::common::String MaintainanceWindowURI;
 };
 
 /// @brief Main entry point for interacting with Foundation.
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -67,11 +67,9 @@ public:
 	/// systems required for foundation to operate.
 	/// @param EndpointRootURI csp::common::String : Root URI for service endpoints
 	/// @param Tenant csp::common::String : Tenant for Magnopus Services. Data is not shared between tenants so clients using separate tenants cannot
-	/// @param Maintainance Window Url.
 	/// interact with each other.
 	/// @return bool : True for successful initialisation.
-	static bool
-		Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& MaintainanceWindowURI, const csp::common::String& Tenant);
+	static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant);
 
 	/// @brief This should be used at the end of the application lifecycle.
 	/// Clears event queues and destroys foundation systems.
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -32,6 +32,7 @@ public:
 	csp::common::String MultiplayerServiceURI;
 	csp::common::String AggregationServiceURI;
 	csp::common::String TrackingServiceURI;
+	csp::common::String MaintainanceWindowURI;
 };
 
 /// @brief Holds client data used in requests for all Magnopus Serives.
@@ -55,9 +56,6 @@ public:
 
 	// @brief Magnopus services environment. e.g odev.
 	csp::common::String CHSEnvironment;
-
-	// @brief Maintainance Window Url.
-	csp::common::String MaintainanceWindowURI;
 };
 
 /// @brief Main entry point for interacting with Foundation.
@@ -69,9 +67,11 @@ public:
 	/// systems required for foundation to operate.
 	/// @param EndpointRootURI csp::common::String : Root URI for service endpoints
 	/// @param Tenant csp::common::String : Tenant for Magnopus Services. Data is not shared between tenants so clients using separate tenants cannot
+	/// @param Maintainance Window Url.
 	/// interact with each other.
 	/// @return bool : True for successful initialisation.
-	static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant);
+	static bool
+		Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& MaintainanceWindowURI, const csp::common::String& Tenant);
 
 	/// @brief This should be used at the end of the application lifecycle.
 	/// Clears event queues and destroys foundation systems.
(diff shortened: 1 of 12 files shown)


Library/include/CSP/CSPFoundation.h
```diff
@@ -55,6 +55,9 @@ class CSP_API ClientUserAgent
 
 	// @brief Magnopus services environment. e.g odev.
 	csp::common::String CHSEnvironment;
+
+	// @brief Maintainance Window Url.
+	csp::common::String MaintainanceWindowURI;
```
MAG-SamBirley, line 60: The intent with this struct is that it is used to derive the useragent header we sent with HTTP requests. I can see why we went with this as a home for the maintenance window URI, but perhaps `EndpointURIs` would be a better home?

Library/src/Web/MaintenanceApi/MaintenanceApi.cpp
```diff
@@ -31,28 +31,14 @@ MaintenanceApi::~MaintenanceApi()
...
-	// S3 bucket URLS are case Sensitive so the CHS Environment name is set to lower case.
-	std::string CHSEnvironmentLower = std::string(CHSEnvironment.c_str());
-	std::transform(CHSEnvironmentLower.begin(), CHSEnvironmentLower.end(), CHSEnvironmentLower.begin(), ::tolower);
-	csp::web::Uri Uri;
-
-	// For Magnopus cloud hosted services, maintenance windows come from a different URL in production
-	if (CHSEnvironmentLower != "oprod")
-	{
-		Uri = csp::web::Uri(
-			csp::common::StringFormat("https://maintenance-windows.magnoboard.com/%s/maintenance-windows.json", CHSEnvironmentLower.c_str()).c_str());
```
MAG-SamBirley, line 48: ♥

MAG-SamBirley (changes requested):

Library/src/Systems/Maintenance/MaintenanceSystem.cpp
```diff
@@ -45,27 +47,32 @@ MaintenanceSystem::~MaintenanceSystem()
 
 void MaintenanceSystem::GetMaintenanceInfo(MaintenanceInfoCallback Callback)
 {
-	const MaintenanceInfoCallback GetMaintenanceInfoCallback = [Callback](const MaintenanceInfoResult& Result)
+	if (Running == true)
```
MAG-SamBirley, line 50: Could we rename `Running` to be something that helps the reader understand it's purpose a little clearer? I'm imaging something like `AllowMaintenanceInfoRequests`

MAG-SamBirley (changes requested):

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: That was my suggestion (clientagentinfo) as an alternative to adding it as an argument to the initialise call, but I agree  EndpointURIs would be better still.

Library/src/Systems/Maintenance/MaintenanceSystem.cpp
MAG-IY: That's reasonable.

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: Yes I agree

MAG-IY: Ok, changes incoming soon then

MAG-SamBirley: Appreciate that's a breaking change. I'd be fine with it though given the nature of the request.

MAG-IY: Either one is a breaking change, as both APIs are public facing

MAG-SamBirley: That's fair, and since my original suggestion of the endpoint route seems non-viable (due to it being immutable for CSP client devs), I guess we go with the maintenance system approach.

MAG-IY: I've made the suggested rework changes and are now ready for review

Library/include/CSP/Systems/Maintenance/MaintenanceSystem.h
```diff
@@ -21,6 +21,7 @@
 #include "CSP/Systems/SystemBase.h"
 #include "CSP/Systems/SystemsResult.h"
 
+#include <string>
```
MAG-SamBirley, line 24: Is this needed? Generally we abstract C++ STL headers in CSP public API headers (in this case, favouring `csp::common::Stirng`)

Library/src/Systems/Maintenance/MaintenanceSystem.cpp
```diff
@@ -31,23 +31,23 @@ namespace csp::systems
...
 MaintenanceSystem::~MaintenanceSystem()
 {
 	CSP_DELETE(MaintenanceAPI);
 }
 
-void MaintenanceSystem::GetMaintenanceInfo(MaintenanceInfoCallback Callback)
+void MaintenanceSystem::GetMaintenanceInfo(const csp::common::String& MaintainanceURL, MaintenanceInfoCallback Callback)
 {
-	if (Running == true)
+	if (AllowMaintenanceInfoRequests == true)
```
MAG-SamBirley, line 50: Could we emit a message to the log if `AllowMaintenanceInfoRequests` is false in this function? I'm imagining a client application developer calling this function, getting it wrong the first time and then wondering why they don't get callbacks on subsequent invocations. In that instance, giving them feedback via the log feels right?

MAG-SamBirley (changes requested):

Library/include/CSP/Systems/Maintenance/MaintenanceSystem.h
MAG-IY: Yes that is something that needed cleaned up

Library/include/CSP/CSPFoundation.h
MAG-SamBirley: Love it

Library/src/Systems/Maintenance/MaintenanceSystem.cpp
MAG-IY: Done

MAG-SamBirley (approved):
Nice work! 🎉

Library/include/CSP/Systems/Maintenance/MaintenanceSystem.h
MAG-SamBirley: Looks good!

Library/src/Systems/Maintenance/MaintenanceSystem.cpp
MAG-SamBirley: Thanks!

PR #440, opened 2024-07-15
[OF-1324] fix: path validation for CartId

MAG-AW:
* GetCheckoutInformation checks CartId for forward and backward slashes, early outs if found

Unsure if this is the best approach, but wrapper gen changes for API level coverage were too complex to achieve in a reasonable time-frame. It's questionable if this is better than no change at all, so I expect discussion either here in PR or externally.

### DIFF
--- a/Library/src/Systems/ECommerce/ECommerceSystem.cpp
+++ b/Library/src/Systems/ECommerce/ECommerceSystem.cpp
@@ -99,7 +99,7 @@ void ECommerceSystem::GetProductInfoCollectionByVariantIds(const common::String&
 void ECommerceSystem::GetCheckoutInformation(const common::String& SpaceId, const common::String& CartId, CheckoutInfoResultCallback Callback)
 {
 	std::string CharacterCheck = CartId.c_str();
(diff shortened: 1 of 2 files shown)


Library/src/Systems/ECommerce/ECommerceSystem.cpp
```diff
@@ -98,6 +98,16 @@ void ECommerceSystem::GetProductInfoCollectionByVariantIds(const common::String&
 
 void ECommerceSystem::GetCheckoutInformation(const common::String& SpaceId, const common::String& CartId, CheckoutInfoResultCallback Callback)
 {
+	std::string CharacterCheck = CartId.c_str();
+    if (CharacterCheck.find("/") != std::string::npos && CharacterCheck.find(R"(\)") != std::string::npos)
```
MAG-SamBirley, line 102: I think the `&&` should be `||`?

MAG-SamBirley: Fix looks good 👍

MAG-AdamThorn (approved):

MAG-SamBirley (approved):

PR #441, opened 2024-07-16
[OF-1334] feat: Sequence System - Generic system for storing shared sorted lists.

MAG-mv:
Commit contains:
- SequenceSystem
- Tests
- Multiplayer event
- Additional functions from: https://magnopus.atlassian.net/browse/OF-1340

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Multiplayer/EventParameters.h
+++ b/Library/include/CSP/Multiplayer/EventParameters.h
@@ -91,4 +91,24 @@ public:
 	csp::common::String UserId;
 };
 
+enum class ESequenceUpdateType
+{
+	Create,
+	Update,
+	Delete
+};
+
+class CSP_API SequenceChangedParams
+{
+public:
+	/// @brief The type of update to the sequence.
+	ESequenceUpdateType UpdateType;
+
+	/// @brief The key of the sequence which was updated.
+	csp::common::String Key;
+
+	/// @brief If the update type is "Update", this will be the new key.
+	csp::common::String NewKey;
+};
+
 } // namespace csp::multiplayer
--- a/Library/include/CSP/Multiplayer/EventParameters.h
+++ b/Library/include/CSP/Multiplayer/EventParameters.h
@@ -107,7 +107,7 @@ public:
 	/// @brief The key of the sequence which was updated.
 	csp::common::String Key;
 
-	/// @brief If the update type is "Update", this will be the new key.
+	/// @brief If a sequence is renamed using the RenameSequence function, this will be the new key.
 	csp::common::String NewKey;
 };
 
--- a/Library/include/CSP/Systems/Sequence/SequenceSystem.h
+++ b/Library/include/CSP/Systems/Sequence/SequenceSystem.h
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2023 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/String.h"
+#include "CSP/Systems/Sequence/Sequence.h"
+#include "CSP/Systems/SystemBase.h"
+
+namespace csp::web
+{
+
+class WebClient;
+
+} // namespace csp::web
+
+
+namespace csp::memory
+{
+
+CSP_START_IGNORE
+template <typename T> void Delete(T* Ptr);
+CSP_END_IGNORE
+
+} // namespace csp::memory
+
+namespace csp::systems
+{
+/// @ingroup Sequence System
+/// @brief Public facing system that allows the management of groupings of hotspots in a space.
+class CSP_API SequenceSystem : public SystemBase
+{
+	CSP_START_IGNORE
+	/** @cond DO_NOT_DOCUMENT */
+	friend class SystemsManager;
+	friend void csp::memory::Delete<SequenceSystem>(SequenceSystem* Ptr);
+	/** @endcond */
+	CSP_END_IGNORE
+public:
+	/// @brief
+	/// @param SequenceKey csp::common::String : The unique grouping name
+	/// @param ReferenceType csp::common::String : The type of reference (GroupId, SpaceId etc.)
+	/// @param ReferenceId csp::common::String : The id of the reference
+	/// @param Items csp::common::Array<csp::common::String> : An ordered array of members
+	/// @param Callback SequenceResultCallback : callback to call when a response is received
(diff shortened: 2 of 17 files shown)


Library/src/Multiplayer/EventSerialisation.cpp
```diff
@@ -71,6 +71,11 @@ csp::multiplayer::ReplicatedValue EventDeserialiser::ParseSignalRComponent(uint6
 {
 	csp::multiplayer::ReplicatedValue ReplicatedValue;
 
+	if (Component.type() == signalr::value_type::null)
```
MAG-mv, line 74: I had to add this line to the base class serializer to prevent as crash, due to the third event param being null unless the UpdateType == Update

Tests/src/PublicAPITests/SequenceSystemTests.cpp
```diff
@@ -0,0 +1,630 @@
...
+
+	const char* TestSequenceKey = "OLY-UNITTEST-SEQUENCE-REWIND";
+	char UniqueSequenceName[256];
+	SPRINTF(UniqueSequenceName, "%s-%s", TestSequenceKey, GetUniqueString().c_str());
+
+	csp::systems::Sequence Sequence;
+	CreateSequence(SequenceSystem, UniqueSequenceName, "GroupId", Space.Id, SequenceItems, Sequence);
+
+	// Rename sequence
+	const char* TestUpdatedSequenceKey = "OLY-UNITTEST-SEQUENCE-REWIND-UPDATED";
```
MAG-AdamThorn, line 401: While we're here do you mind removing any mention of 'REWIND'?

MAG-AdamThorn (changes requested):

Library/include/CSP/Multiplayer/EventParameters.h
```diff
@@ -91,4 +91,24 @@ class CSP_API UserPermissionsParams
...
+{
+public:
+	/// @brief The type of update to the sequence.
+	ESequenceUpdateType UpdateType;
+
+	/// @brief The key of the sequence which was updated.
+	csp::common::String Key;
+
+	/// @brief If the update type is "Update", this will be the new key.
+	csp::common::String NewKey;
```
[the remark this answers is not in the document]
MAG-mv: Ah, yes. I can change this comment to be more specific to key change updates

Library/include/CSP/Systems/Sequence/SequenceSystem.h
```diff
@@ -0,0 +1,113 @@
...
+{
+	CSP_START_IGNORE
+	/** @cond DO_NOT_DOCUMENT */
+	friend class SystemsManager;
+	friend void csp::memory::Delete<SequenceSystem>(SequenceSystem* Ptr);
+	/** @endcond */
+	CSP_END_IGNORE
+public:
+	/// @brief
+	/// @param SequenceKey csp::common::String : The unique grouping name
```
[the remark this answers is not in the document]
MAG-mv: I think its worth mentioning, I think ti will prevent a bunch of questions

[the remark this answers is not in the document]
MAG-mv: Does this sound ok? @param SequenceKey csp::common::String : The unique grouping name. The suggested convention is: Type:[Id]. I left the naming general as the Type could be hotspot, hierarchy and Id could be a specId of parent entity id. Also, we may add more usages in the future

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-mv: I'll change OLY to CSP as well.

MAG-mv: probably something we should enfore in all future PRs

Library/include/CSP/Systems/Sequence/SequenceSystem.h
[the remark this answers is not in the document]
MAG-AdamThorn: Yes that looks good to me.

MAG-AdamThorn (approved):

Library/src/Multiplayer/EventSerialisation.cpp
MAG-SamBirley: Worth including a comment in the code to this effect? Thinking of future us encountering this line and not understanding the context around why it is necessary,

MAG-SamBirley (changes requested):
One small request for an additional comment, otherwise looks great!

MAG-SamBirley (dismissed):
Approved from me, but maybe also wait on feedback from @MAG-ME

Library/include/CSP/Systems/Sequence/SequenceSystem.h
```diff
@@ -0,0 +1,113 @@
...
+	/** @endcond */
+	CSP_END_IGNORE
+public:
+	/// @brief
+	/// @param SequenceKey csp::common::String : The unique grouping name. The suggested convention is: Type:[Id]
+	/// @param ReferenceType csp::common::String : The type of reference (GroupId, SpaceId etc.)
+	/// @param ReferenceId csp::common::String : The id of the reference
+	/// @param Items csp::common::Array<csp::common::String> : An ordered array of members
+	/// @param Callback SequenceResultCallback : callback to call when a response is received
+	void CreateSequence(const csp::common::String& SequenceKey,
```
MAG-ME, line 60: Is it worth adding a note for the API consumer that reusing a sequence key would not be considered an error but will overwrite that sequence?

Tests/src/PublicAPITests/SequenceSystemTests.cpp
```diff
@@ -0,0 +1,663 @@
...
+		});
+
+	// Delete space
+	DeleteSpace(SpaceSystem, Space.Id);
+
+	// Log out
+	LogOut(UserSystem);
+}
+
+#endif
```
MAG-ME, line 663: Is it possible to test the case where DeleteSequences will fail because  `This call will fail if the user isn't a creator of the space`

Library/include/CSP/Systems/Sequence/SequenceSystem.h
```diff
@@ -0,0 +1,113 @@
...
+								const csp::common::Optional<csp::common::String>& ReferenceType,
+								const csp::common::Optional<csp::common::Array<csp::common::String>>& ReferenceIds,
+								SequencesResultCallback Callback);
+
+	/// @brief Gets a sequence by it's key
+	/// @param SequenceKey csp::common::String : The unique grouping name
+	/// @param Callback SequenceResultCallback : callback to call when a response is received
+	void GetSequence(const csp::common::String& SequenceKey, SequenceResultCallback Callback);
+
+	/// @brief Deletes the given sequences. This call will fail if the user isn't a creator of the space
```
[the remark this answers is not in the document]
MAG-mv: CHS have mentioned that any calls which modify data will have this restriction. I will add this comment to create and update too

Library/include/CSP/Systems/Sequence/SequenceSystem.h
MAG-mv: Good idea!

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-mv: Added a test which tests sequence key change, edit, and delete from a user not in the space

MAG-SamBirley (approved):

MAG-SamBirley (approved):

PR #446, opened 2024-07-18
[OB-3650] fix: reverse shutdown order

magnopus-swifty:
Systems must be shut down in reverse order to `CreateSystems()` to ensure that any dependencies continue to exist until each system is successfully shut down.

This resolves a potential crash on shutdown, as it prevents the script system from being deleted before script entities.

Found via a Unity client where CSP is shutdown when terminating a play session in-editor.

### DIFF
--- a/Library/src/Systems/SystemsManager.cpp
+++ b/Library/src/Systems/SystemsManager.cpp
@@ -15,6 +15,7 @@
  */
 #include "CSP/Systems/SystemsManager.h"
 
+#include "CSP/Multiplayer/MultiplayerConnection.h"
 #include "CSP/Systems/Analytics/AnalyticsSystem.h"
 #include "CSP/Systems/Assets/AssetSystem.h"
 #include "CSP/Systems/ECommerce/ECommerceSystem.h"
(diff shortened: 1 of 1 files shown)


MAG-SamBirley (approved):
Nice find & fix. Thank you!

MAG-SamBirley:
> Has this been tested as a build in a client? This order was specifically set when we last did a refactor to the connection flow, alternative orders broke some systems, so this needs testing to ensure it has no issues on clients.

That's interesting. Could you share a link to the PR/change for viz?

MAG-SamBirley:
> Has this been tested as a build in a client? This order was specifically set when we last did a refactor to the connection flow, alternative orders broke some systems, so this needs testing to ensure it has no issues on clients.

I would expect shutdown issues to also manifest via the CSP tests, but it's still a good call, just to be sure. I can do that in the background today.

MAG-DanielMurdolo:
I would probably like the change to be tested in _all_ clients, to @MAG-AW's point, as we did have some fun and games across the clients during this change.

MAG-DanielMurdolo:
Tested on Web using the build `DEV_240718_080150_368597973edd304c98af0e0d4bba6a023b35d3a3` and saw no issues.

MAG-SamBirley:
Tested on Unreal (using local Windows build of CSP generated from branch) and saw no issues.

MAG-SamBirley:
Since this change is all about resolving a Unity crash; I'm happy to skip validating that engine.

PR #451, opened 2024-07-22
[OF-1313] feat: Add name property to component base

MAG-mv:
- Added GetComponentName and SetComponentName to ComponentBase
- Deprecated Name properties and accessors on derived classes
- Setup script accessor for aname
- Added tests for name in ComponentBase

### DIFF
--- a/Library/src/Multiplayer/ComponentBase.cpp
+++ b/Library/src/Multiplayer/ComponentBase.cpp
@@ -271,17 +271,17 @@ void ComponentBase::InvokeAction(const csp::common::String& InAction, const csp:
 
 const csp::common::String& ComponentBase::GetComponentName() const
 {
-	return GetStringProperty(COMPONENT_KEY_VIEW_NAME);
+	return GetStringProperty(COMPONENT_KEY_NAME);
 }
 
 void ComponentBase::SetComponentName(const csp::common::String& Value)
 {
-	SetProperty(COMPONENT_KEY_VIEW_NAME, Value);
+	SetProperty(COMPONENT_KEY_NAME, Value);
 }
 
 void ComponentBase::InitialiseProperties()
 {
-	Properties[COMPONENT_KEY_VIEW_NAME] = "";
+	Properties[COMPONENT_KEY_NAME] = "";
 }
 
 } // namespace csp::multiplayer
--- a/Library/src/Multiplayer/ComponentBase.cpp
+++ b/Library/src/Multiplayer/ComponentBase.cpp
@@ -18,6 +18,7 @@
 #include "CSP/Common/List.h"
 #include "CSP/Multiplayer/Script/EntityScript.h"
 #include "CSP/Multiplayer/SpaceEntity.h"
+#include "ComponentBaseKeys.h"
 #include "Debug/Logging.h"
 #include "Memory/Memory.h"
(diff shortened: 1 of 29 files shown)


Library/src/Multiplayer/Script/ComponentScriptInterface.cpp
```diff
@@ -64,6 +64,24 @@ int64_t ComponentScriptInterface::GetComponentType() const
 	return (int64_t) ComponentType::Invalid;
 }
 
+void ComponentScriptInterface::SetComponentName(std::string name)
+{
+	if (Component)
```
MAG-SamBirley, line 69: Should we emit an error log if component == nullptr for the setter and getter?

Library/src/Multiplayer/ComponentBase.cpp
```diff
@@ -18,6 +18,7 @@
 #include "CSP/Common/List.h"
 #include "CSP/Multiplayer/Script/EntityScript.h"
 #include "CSP/Multiplayer/SpaceEntity.h"
+#include "ComponentBaseKeys.h"
```
MAG-SamBirley, line 21: It looks like this header is not part of this PR. I can see it's causing failures [on the PR checks](https://uk-teamcity.magnopus.global/buildConfiguration/Olympus_Foundation_X64_ReleaseDLLCSharp/229426?buildTab=log&linesState=312.13556&logView=flowAware&focusLine=14061).

MAG-SamBirley (changes requested):

Library/src/Multiplayer/Script/ComponentScriptInterface.cpp
MAG-mv: We currently dont do this for any other script component props

MAG-SamBirley: Understood. Do you think we should?

MAG-mv: We should further investigate this by understanding the circumstances where the COmponent would be null to find the best method to handle this

MAG-SamBirley: Makes sense. Thanks!

Tests/src/PublicAPITests/ComponentTests/ComponentTests.cpp
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+#include "../SpaceSystemTestHelpers.h"
```
MAG-IY, line 17: Should these paths not be relative to the project root like the rest? Or is there a reason for the file relative path?

MAG-mv: image
Component tests is in a nested folder and we dont have an include dir setup at the root

MAG-IY: OK fair enough. Thought that might be your answer, but just wanted to check. It's not something I feel strongly about anyway.

MAG-SamBirley (approved):
Looks good to me.

One minor suggestion for the future would be to note in the new base keys header that adding extra default values requires updating the component tests (where the number of default values is hard coded). Not significant enough to block the PR though.

PR #453, opened 2024-07-24
[OF-1371] Added ParentId property to SpaceEntity

MAG-mv:
- Added ParentId property to SpaceEntity
- Added serialisation/deserialisation
- Updated serialisation tests to support this
- Correctly constructed EntityLock in all constructors to prevent SerializationTest crashes


MAG-IY (dismissed):
No comments seems fine. Approved

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -341,6 +343,7 @@ class CSP_API SpaceEntity
 	bool IsTransferable;
 	bool IsPersistant;
 	uint64_t OwnerId;
+	uint64_t ParentId;
```
MAG-SamBirley, line 346: Observation - we are serialising null for no-parent, but representing in CSP as 0.
Suggestion - Should we align those two things? Perhaps by using csp::Optional<uint64_t> for the type of `ParentId` and using null when there's no parent?

MAG-mv: I'm happy to do that

MAG-mv: Updated

MAG-AdamThorn (approved):

MAG-SamBirley (commented):
Love the nullable implementation. Reads much nicer. DId we test that it works as expected in C# and web? Not sure we've done an optional member variable before, so might be worth checking before it goes in.

MAG-mv:
> Love the nullable implementation. Reads much nicer. DId we test that it works as expected in C# and web? Not sure we've done an optional member variable before, so might be worth checking before it goes in.

We have now decided to keep the ids private, and no accessors are needed for this ticket

MAG-SamBirley (approved):

PR #457, opened 2024-07-26
[OB-3664] Fixed CustomComponentTest from hanging

MAG-mv:
- Added a test timeout to the loop to prevent hanging in the future

### DIFF
--- a/Tests/src/PublicAPITests/ComponentTests/CustomComponentTests.cpp
+++ b/Tests/src/PublicAPITests/ComponentTests/CustomComponentTests.cpp
@@ -190,7 +190,10 @@ CSP_PUBLIC_TEST(CSPEngine, CustomTests, CustomComponentTest)
 		EntitySystem->QueueEntityUpdate(CreatedObject);
 		EntitySystem->ProcessPendingEntityOperations();
 
-		SpaceSystem->ExitSpace([](const csp::systems::NullResult& Result){});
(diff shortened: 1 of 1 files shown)


MAG-SamBirley (approved):
We're introducing a workaround for a test failure with this change, but that said it's a test for a feature with low-utilisation, and it's actively blocking high-prio work. 

I'd like to find and fix the true issue, but I'm ok doing that in a future PR.

MAG-mv:
@MAG-SamBirley The true issue was because of an error on my development branch, which I am fixing in my PR

MAG-SamBirley:
> @MAG-SamBirley The true issue was because of an error on my development branch, which I am fixing in my PR

Given that if we had this PR merged, we would not have caught the true issue, as we sure we want to merge this change?

MAG-mv:
> Given that if we had this PR merged, we would not have caught the true issue, are we sure we want to merge this change?

This PR contains a check to ensure the callback is called: EXPECT_TRUE(GotAllEntities);

This will make the issue more clear instead of hangning

MAG-SamBirley:
> > Given that if we had this PR merged, we would not have caught the true issue, are we sure we want to merge this change?
> 
> This PR contains a check to ensure the callback is called: EXPECT_TRUE(GotAllEntities);
> 
> This will make the issue more clear instead of hanging

Ah, yes, sorry for missing that. That's perfect!

PR #458, opened 2024-07-29
[OF-1370] Added ParentId support to SpaceEntitySystem and SpaceEntity

MAG-mv:
- Fixed bug with setting Optional to nullptr
- Added ability to set, get and remove ParentId
- Added serialisation/deserialisation for ParentId
- Added patch serialisation/deserialisation for ParentId
- Added tests for all functionality
- Added new SpaceEntityUpdateFlag for ParentId
- Also completed ApplyIncomingPatch work for OF-1372 to prevent deserialisation crashes

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -90,6 +90,7 @@ enum SpaceEntityUpdateFlags
 	UPDATE_FLAGS_SELECTION_ID		  = 32,
 	UPDATE_FLAGS_THIRD_PARTY_REF	  = 64,
 	UPDATE_FLAGS_THIRD_PARTY_PLATFORM = 128,
+	UPDATE_FLAGS_PARENT				  = 256,
 };
 
 /// @brief Primary multiplayer object that can have associated scripts and many multiplayer components created within it.
@@ -205,6 +206,13 @@ public:
 	/// @brief Get SpaceEntitySystem Object
 	/// @return SpaceEntitySystem
 	SpaceEntitySystem* GetSpaceEntitySystem();
+
+	void SetParentEntity(SpaceEntity* Parent);
+	void RemoveParentEntity();
+
+	SpaceEntity* GetParentEntity() const;
+
+	const csp::common::List<SpaceEntity*>* GetChildEntities() const;
 
 	/// @brief Queues an update which will be executed on next Tick() or ProcessPendingEntityOperations(). Not a blocking or async function.
 	void QueueUpdate();
@@ -336,6 +344,10 @@ private:
 
 	ComponentBase* FindFirstComponentOfType(ComponentType Type, bool SearchDirtyComponents = false) const;
 
+	void AddChildEntitiy(SpaceEntity* ChildEntity);
+
+	bool ResolveEntityHierarchy();
+
 	SpaceEntitySystem* EntitySystem;
 
 	SpaceEntityType Type;
@@ -344,11 +356,16 @@ private:
 	bool IsPersistant;
 	uint64_t OwnerId;
 	csp::common::Optional<uint64_t> ParentId;
+	bool ShouldUpdateParent;
+
 	csp::common::String Name;
 	SpaceTransform Transform;
 	csp::systems::EThirdPartyPlatform ThirdPartyPlatform;
 	csp::common::String ThirdPartyRef;
 	uint64_t SelectedId;
+
+	SpaceEntity* Parent;
+	csp::common::List<SpaceEntity*> ChildEntities;
 
 	UpdateCallback EntityUpdateCallback;
 	DestroyCallback EntityDestroyCallback;
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -86,9 +86,11 @@ SpaceEntity::SpaceEntity()
 	, RefCount(CSP_NEW std::atomic_int(0))
 	, SelectedId(0)
 	, ParentId(nullptr)
+	, ShouldUpdateParent(false)
 	, ThirdPartyRef("")
 	, ThirdPartyPlatform(csp::systems::EThirdPartyPlatform::NONE)
 	, TimeOfLastPatch(0)
+	, Parent(nullptr)
 {
 }
 
@@ -109,9 +111,11 @@ SpaceEntity::SpaceEntity(SpaceEntitySystem* InEntitySystem)
 	, RefCount(CSP_NEW std::atomic_int(0))
 	, SelectedId(0)
 	, ParentId(nullptr)
+	, ShouldUpdateParent(false)
 	, ThirdPartyRef("")
 	, ThirdPartyPlatform(csp::systems::EThirdPartyPlatform::NONE)
 	, TimeOfLastPatch(0)
+	, Parent(nullptr)
 {
 }
 
@@ -315,6 +319,28 @@ SpaceEntitySystem* SpaceEntity::GetSpaceEntitySystem()
 	return EntitySystem;
 }
 
+void SpaceEntity::SetParentEntity(SpaceEntity* InParent)
+{
+	ParentId		   = InParent->GetId();
+	ShouldUpdateParent = true;
+}
+
+void SpaceEntity::RemoveParentEntity()
+{
+	ParentId		   = nullptr;
+	ShouldUpdateParent = true;
+}
+
+SpaceEntity* SpaceEntity::GetParentEntity() const
+{
+	return Parent;
+}
+
+const csp::common::List<SpaceEntity*>* SpaceEntity::GetChildEntities() const
+{
+	return &ChildEntities;
+}
+
 void SpaceEntity::QueueUpdate()
 {
 	if (EntitySystem)
@@ -439,8 +465,8 @@ void SpaceEntity::SerialisePatch(IEntitySerialiser& Serialiser) const
 		Serialiser.WriteBool(false); // Destroy
 		Serialiser.BeginArray();	 // ParentId
 		{
-			Serialiser.WriteBool(false);
-			Serialiser.WriteNull();
+			Serialiser.WriteBool(ShouldUpdateParent);
+			ParentId.HasValue() ? Serialiser.WriteUInt64(*ParentId) : Serialiser.WriteNull();
 		}
 		Serialiser.EndArray();
 
@@ -645,16 +671,41 @@ void SpaceEntity::Deserialise(IEntityDeserialiser& Deserialiser)
 
 void SpaceEntity::DeserialiseFromPatch(IEntityDeserialiser& Deserialiser)
 {
+	SpaceEntityUpdateFlags UpdateFlags = SpaceEntityUpdateFlags(0);
+
+	if (ShouldUpdateParent)
+	{
+		uint32_t size = 0;
+		Deserialiser.EnterArray(size);
+		{
+			ShouldUpdateParent = Deserialiser.ReadBool();
+
+			if (Deserialiser.NextValueIsNull())
+			{
+				Deserialiser.Skip();
+			}
+			else
+			{
+				ParentId = Deserialiser.ReadUInt64();
+			}
+
+			UpdateFlags = SpaceEntityUpdateFlags(UpdateFlags | UPDATE_FLAGS_PARENT);
+		}
+		Deserialiser.LeaveArray();
+
+		ResolveEntityHierarchy();
+	}
+
 	std::scoped_lock<std::mutex> ComponentsLocker(*ComponentsLock);
+
+	csp::common::Array<ComponentUpdateInfo> ComponentUpdates(0);
 
 	if (!Deserialiser.NextValueIsNull()) // It is valid for entities to not have components
 	{
 		Deserialiser.EnterComponents();
 		{
 			auto RealComponentCount = Deserialiser.GetNumRealComponents();
-
-			SpaceEntityUpdateFlags UpdateFlags = SpaceEntityUpdateFlags(0);
-			csp::common::Array<ComponentUpdateInfo> ComponentUpdates(RealComponentCount);
+			ComponentUpdates		= csp::common::Array<ComponentUpdateInfo>(RealComponentCount);
 
 			uint16_t ComponentKey;
 			uint64_t _ComponentType;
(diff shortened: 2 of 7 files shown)


Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -206,6 +207,13 @@ class CSP_API SpaceEntity
 	/// @return SpaceEntitySystem
 	SpaceEntitySystem* GetSpaceEntitySystem();
 
+	void SetParentEntity(SpaceEntity* Parent);
+	void RemoveParentEntity();
```
MAG-SamBirley, line 211: If you feel that this would result in a cleaner API, could we remove this if we were to allow `MyEntity->SetParentEntity(nullptr)`?

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -206,6 +207,13 @@ class CSP_API SpaceEntity
...
+
+	SpaceEntity* GetParentEntity() const;
```
MAG-SamBirley, line 213: Remember that all public API functions should be accompanied with doxygen-styled comments!

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -439,8 +465,8 @@ void SpaceEntity::SerialisePatch(IEntitySerialiser& Serialiser) const
 		Serialiser.WriteBool(false); // Destroy
 		Serialiser.BeginArray();	 // ParentId
 		{
-			Serialiser.WriteBool(false);
-			Serialiser.WriteNull();
+			Serialiser.WriteBool(ShouldUpdateParent);
```
MAG-SamBirley, line 468: What effect does serializing this state have?

MAG-SamBirley: Oh I see, it tells other connected CSP client applications that they too need to update the entity's parent 👍

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -645,16 +671,41 @@ void SpaceEntity::Deserialise(IEntityDeserialiser& Deserialiser)
 
 void SpaceEntity::DeserialiseFromPatch(IEntityDeserialiser& Deserialiser)
 {
+	SpaceEntityUpdateFlags UpdateFlags = SpaceEntityUpdateFlags(0);
+
+	if (ShouldUpdateParent)
```
MAG-SamBirley, line 676: It looks like we're testing for this value before we've actually deserialized the state of the value (in line 681)?

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1191,6 +1247,52 @@ ComponentBase* SpaceEntity::FindFirstComponentOfType(ComponentType Type, bool Se
...
+		// Set our new parent
+		Parent = GetSpaceEntitySystem()->FindSpaceEntityById(*ParentId);
+
+		if (Parent != nullptr)
+		{
+			Parent->ChildEntities.Append(this);
+		}
+		else
+		{
+			CSP_LOG_ERROR_MSG("SpaceEntity unable to find parent");
```
MAG-SamBirley, line 1280: Worth including the parent ID in the message to aid debugging?

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1191,6 +1247,52 @@ ComponentBase* SpaceEntity::FindFirstComponentOfType(ComponentType Type, bool Se
 	return LocatedComponent;
 }
 
+void SpaceEntity::AddChildEntitiy(SpaceEntity* ChildEntity)
+{
+	ChildEntities.Append(ChildEntity);
+}
+
+bool SpaceEntity::ResolveEntityHierarchy()
```
MAG-SamBirley, line 1255: Possibly better named `ResolveParent`? It's all about parentage, and it'd make the call-site on line #972 make more sense too.

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -849,6 +849,21 @@ void SpaceEntitySystem::OnAllEntitiesCreated()
 	// Ensure entity list is up to date
 	ProcessPendingEntityOperations();
 
+	// Resolve entity hierarchy
+	for (size_t i = 0; i < Entities.Size(); ++i)
+	{
+		SpaceEntity* Entity = Entities[i];
+
+		if (Entity->ParentId.HasValue())
```
MAG-SamBirley, line 857: Any reason to not invoke the new `ResolveEntityHierarchy` function? Could be an opportunity to follow the DRY pattern.

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1368,7 +1383,6 @@ void SpaceEntitySystem::ApplyIncomingPatch(const signalr::value* EntityMessage)
 		const uint64_t EntityID = Deserialiser.ReadUInt64();
 		const uint64_t OwnerID	= Deserialiser.ReadUInt64();
 		const bool Destroy		= Deserialiser.ReadBool();
-		Deserialiser.Skip(); // ParentId //TODO: Support this if required
```
MAG-SamBirley, line 1371: How come we no longer need to at least skip the parent ID value?

MAG-SamBirley (changes requested):

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -2066,4 +2066,287 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, FindComponentByIdTest)
...
+		EntitySystem->ProcessPendingEntityOperations();
+		std::this_thread::sleep_for(50ms);
+		WaitForTestTimeoutCountMs += 50;
+	}
+
+	EXPECT_TRUE(ChildEntityUpdated);
+
+	// Check entity is unparented correctly
+	EXPECT_EQ(CreatedChildEntity->GetParentEntity(), nullptr);
+	EXPECT_EQ(CreatedParentEntity->GetChildEntities()->Size(), 0);
```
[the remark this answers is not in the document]
MAG-SamBirley: That's a good point. I didn't catch what we do in the PR when a parent entity is destroyed. How do the child entities react?

Library/src/Multiplayer/SpaceEntitySystem.cpp
MAG-mv: We handle this later on in DeserialiseFromPatch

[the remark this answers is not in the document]
MAG-mv: This is now handled later in DeserialiseFromPatch

Library/src/Multiplayer/SpaceEntity.cpp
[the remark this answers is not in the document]
MAG-mv: Do I also need to handle this in the local deserialisation?

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: I like it

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: Good spot. This is only working because it is using my local value

Library/src/Multiplayer/SpaceEntity.cpp
MAG-SamBirley: You're right, it updates lists of children on the parent too. To me that still feels appropriate to describe as parent resolution.

Library/src/Multiplayer/SpaceEntity.cpp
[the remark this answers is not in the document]
MAG-mv: I'm referring to SpaceEntity::ApplyLocalPatch, which is called for the local tests. This function now sets the UPDATE_FLAGS_PARENT. So I'm confused about what you mean by "This prevents an entity receiving a parent change update if it is locally queued for a parent change."

Library/src/Multiplayer/SpaceEntitySystem.cpp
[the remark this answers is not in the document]
MAG-mv: That makes sense. I will update this now

Library/src/Multiplayer/SpaceEntity.cpp
MAG-SamBirley: Fair enough. Would you prefer to leave as-is?

MAG-mv: ResolveParentChildRelationship?

MAG-SamBirley: I'm sold.

Library/src/Multiplayer/SpaceEntitySystem.cpp
MAG-mv: ResolveEntityHierarchy relies on ShouldUpdateParent being set, which isn't the case in this circumstance. I could manually set this on deserialisation if we want to go down this route, but I think we should consider the following:

. While I understand it removes some repeated code, ResolveEntityHierarchy also contains logic that isn't necessary in this case (removing the parent and removing previous parents), which may make it harder to understand/debug.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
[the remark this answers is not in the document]
MAG-mv: Good call. I will:

- Update the current tests to account for multiple children
- Add another test which tests deleting a parent

Let me know if there's any more that are needed

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: @MAG-SamBirley @MAG-AW. The GetObject call will make it more robust. The only downside is that this is another async call which will add additional time/complexity.

Is this something we feel should be done in this PR?

Library/src/Multiplayer/SpaceEntitySystem.cpp
MAG-SamBirley: Makes sense. Thanks for the explanation!

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-SamBirley: Latest changes look good, thanks!

Library/src/Multiplayer/SpaceEntity.cpp
MAG-SamBirley: Latest changes look good, thanks!

Library/src/Multiplayer/SpaceEntity.cpp
MAG-SamBirley: So should we just go with the parent ID for this PR?

Library/src/Multiplayer/SpaceEntity.cpp
MAG-SamBirley: Latest change looks good to me 🙌

MAG-AdamThorn (dismissed):

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1277,7 +1261,8 @@ bool SpaceEntity::ResolveEntityHierarchy()
 		}
 		else
 		{
-			CSP_LOG_ERROR_MSG("SpaceEntity unable to find parent");
+			CSP_LOG_ERROR_FORMAT("SpaceEntity unable to find parent for entity: %s. Please report if this issue is encountered.",
+								 std::to_string(GetId()));
 		}
 	}
 	else
```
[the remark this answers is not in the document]
MAG-mv: I'm happy to update to return false here. The error signifies that there is a problem for callers

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -808,6 +808,19 @@ void SpaceEntitySystem::RemoveEntity(SpaceEntity* EntityToRemove)
 	PendingOutgoingUpdateUniqueSet->erase(EntityToRemove);
 }
 
+bool SpaceEntitySystem::EntityExists(SpaceEntity* Entity)
```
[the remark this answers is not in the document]
MAG-mv: I'm happy to remove EntityExists althogether in favour of FindSpaceEntityById

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -2348,5 +2465,208 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ParentEntityEnterSpaceReplicationTe
...
+		while (!DestroyCalled && WaitForTestTimeoutCountMs < WaitForTestTimeoutLimit)
+		{
+			EntitySystem->ProcessPendingEntityOperations();
+			std::this_thread::sleep_for(50ms);
+			WaitForTestTimeoutCountMs += 50;
+		}
+
+		EXPECT_TRUE(DestroyCalled);
+
+		// Ensure parent and child is deleted
```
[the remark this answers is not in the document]
MAG-mv: Updated

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -321,13 +321,15 @@ SpaceEntitySystem* SpaceEntity::GetSpaceEntitySystem()
 
 void SpaceEntity::SetParentEntity(SpaceEntity* InParent)
```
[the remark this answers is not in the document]
MAG-mv: Also, we woulod have to resolve the child for that parent at this point too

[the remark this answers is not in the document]
MAG-mv: Another drawback is that we will have no way of testing the replication actually worked correctly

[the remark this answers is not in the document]
MAG-mv: Sounds good!

MAG-AdamThorn (dismissed):

MAG-AdamThorn (dismissed):

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -123,6 +123,9 @@ class CSP_API SpaceEntity
 	// General callback providing success/fail boolean.
 	typedef std::function<void(bool)> CallbackHandler;
 
+	// Callback that will provide a pointer to a SpaceEntity object.
+	typedef std::function<void(SpaceEntity*)> EntityCreatedCallback;
```
MAG-SamBirley, line 127: When should the developer expect to bind to this callback?

MAG-mv: This had to be duplicated from SpaceEntitySystem so the wrapper generator didnt break. This is for CreateChildEntity

MAG-SamBirley (approved):
lgtm!

MAG-AdamThorn (approved):

PR #462, opened 2024-08-09
Hotspot sequence system addition

MAG-IY:
Added Hotspot sequence system, with the following API:

void CreateHotspotGroup(String GroupName, Array<String> HotspotIds, HotSpotGroupResultCallback Callback)
void UpdateHotspotGroup(String GroupName, Array<String> HotspotIds, HotSpotGroupResultCallback Callback)
void CreateHotspotGroup(String GroupName, Array<String> HotspotIds, HotSpotGroupResultCallback Callback)
void UpdateHotspotGroup(String GroupName, Array<String> HotspotIds, HotSpotGroupResultCallback Callback)
void DeleteHotspotGroup(String GroupName, HotSpotGroupResultCallback Callback)

Added test suite with the following cases:

CreateHotspotGroupTest (Tests a new group can be created with hotspot ids)
CreateHotspotGroupNoHotspotsTest (Tests a new group can be created with no ids)
UpdateHotspotGroupTest (Test a groups hospot ids can be modified)
GetHotspotGroupTest (Test we can get a hotspot by its group name)
GetHotspotNoGroupTest (Test we get an error when we search for a hitspot group)
GetHotspotsTest (Create 2 hotspot groups in a space and ensure we retrieve both)
GetHotspotsNoGroupsTest (Create 0 hotspot groups in a space and ensure we retrieve nothing)
DeleteHotspotNoGroupTest (Ensure we error trying to delete an invalid group)


Library/include/CSP/Systems/HotSpotSequence/HotSpotGroup.h
```diff
@@ -0,0 +1,87 @@
...
+private:
+	HotSpotGroupResult(void*) {};
+
+	void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
+
+	HotSpotGroup Group;
+};
+
+/// @ingroup HotSpotSequence System
+/// @brief Data class used to contain information when attempting to get a Hot Spot Group.
```
MAG-mv, line 58: This is a duplicate comment from HotSpotGroupResult. This should probably say "get an array of Hot Spot Groups"

Library/include/CSP/Systems/HotSpotSequence/HotSpotGroup.h
```diff
@@ -0,0 +1,87 @@
...
+{
+	/** @cond DO_NOT_DOCUMENT */
+	CSP_START_IGNORE
+	template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+	CSP_END_IGNORE
+	/** @endcond */
+
+public:
+	const csp::common::Array<HotSpotGroup>& GetHotSpotGroups() const;
+	void setHotSpotGroups(csp::common::Array<HotSpotGroup>& groups);
```
MAG-mv, line 69: Lowercase s doesn't match the rest of our API

Library/include/CSP/Systems/HotSpotSequence/HotSpotSequenceSystem.h
```diff
@@ -0,0 +1,97 @@
...
+	/** @endcond */
+	CSP_END_IGNORE
+
+	HotSpotSequenceSystem(csp::systems::SequenceSystem* SequenceSystem, csp::systems::SpaceSystem* SpaceSystem);
+	/// @brief Create a Hotspot group
+	/// @param GroupName csp::common::String : The unique grouping name
+	/// @param HotspotIds csp::common::Array<csp::common::String> : set of Hotspot ids to add to the group
+	/// @param Callback HotSpotGroupResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void
+		CreateHotspotGroup(csp::common::String GroupName, csp::common::Array<csp::common::String> HotspotIds, HotSpotGroupResultCallback Callback);
```
MAG-mv, line 66: Should be const Array& to prevent copy

Library/include/CSP/Systems/HotSpotSequence/HotSpotSequenceSystem.h
```diff
@@ -0,0 +1,97 @@
...
+
+	/// @brief Update a Hotspot group
+	/// @param GroupName csp::common::String : The unique grouping name
+	/// @param HotspotIds csp::common::Array<csp::common::String> : set of Hotspot ids to add to the group
+	/// @param Callback HotSpotGroupResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void
+		UpdateHotspotGroup(csp::common::String GroupName, csp::common::Array<csp::common::String> HotspotIds, HotSpotGroupResultCallback Callback);
```
MAG-mv, line 78: Should be const Array& to prevent copy

MAG-mv, line 78: const& for string too

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "CSP/Systems/Sequence/SequenceSystem.h"
+#include "CSP/Systems/Spaces/SpaceSystem.h"
+
+#include <string>
+
+namespace csp::systems
+{
+HotSpotSequenceSystem::HotSpotSequenceSystem(csp::systems::SequenceSystem* SequenceSystem, csp::systems::SpaceSystem* SpaceSystem)
+{
+	this->sequenceSystem = SequenceSystem;
```
MAG-mv, line 29: We dont use Lowercamel case inertnally anywhere else, either

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+	csp::common::String key = spaceID + ":" + GroupName;
+
+	auto CB = [Callback, key](SequenceResult result)
+	{
+		auto data = result.GetSequence();
+		HotSpotGroup group;
+		group.Items = data.Items;
+		group.Name	= data.Key;
+		HotSpotGroupResult returnValue(result.GetResultCode(), result.GetHttpResultCode());
+		returnValue.setHotSpotGroup(group);
```
MAG-ME, line 47: Could we pass the group into the result via the constructor and remove the need for a separate method?

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+	auto spaceID			= spaceSystem->GetCurrentSpace().Id;
+	csp::common::String key = GroupName + spaceID;
+
+	auto CB = [Callback, key](SequenceResult result)
+	{
+		HotSpotGroupResult returnValue(result.GetResultCode(), result.GetHttpResultCode());
+		HotSpotGroup group;
+		group.Items = result.GetSequence().Items;
+		group.Name	= result.GetSequence().Key;
+		returnValue.setHotSpotGroup(group);
```
MAG-ME, line 68: Same comment as above about using the constructor to set the group in the result

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+	this->sequenceSystem = SequenceSystem;
+	this->spaceSystem	 = SpaceSystem;
+}
+
+void HotSpotSequenceSystem::CreateHotspotGroup(csp::common::String GroupName,
+											   csp::common::Array<csp::common::String> HotspotIds,
+											   HotSpotGroupResultCallback Callback)
+{
+	auto spaceID			= spaceSystem->GetCurrentSpace().Id;
+	csp::common::String key = spaceID + ":" + GroupName;
```
MAG-mv, line 38: Key should include "Hotspots:" as a prefix as shown in the spec:
Hotspots:[SomeSpaceId]:[SomeName]

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+	auto CB = [Callback, key](SequenceResult result)
+	{
+		auto data = result.GetSequence();
+		HotSpotGroup group;
+		group.Items = data.Items;
+		group.Name	= data.Key;
```
MAG-mv, line 45: The name isn't the entire key, it's only the end section:

csp::common::String key = spaceID + ":" + GroupName;

You will need to extract the GroupName from the key. I did this in the previous work using:


const std::regex Expression("^(?:[^:]*\:){2}([^:]*)");
std::smatch Match;
const bool Found = std::regex_search(std::begin(SequenceKeyString), std::end(SequenceKeyString), Match, Expression);

if (Found == false)
{
           return nullptr;
}

std::string Name = Match[1];

MAG-mv (changes requested):

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+	this->sequenceSystem = SequenceSystem;
+	this->spaceSystem	 = SpaceSystem;
+}
+
+void HotSpotSequenceSystem::CreateHotspotGroup(csp::common::String GroupName,
+											   csp::common::Array<csp::common::String> HotspotIds,
+											   HotSpotGroupResultCallback Callback)
+{
+	auto spaceID			= spaceSystem->GetCurrentSpace().Id;
+	csp::common::String key = spaceID + ":" + GroupName;
```
MAG-ME, line 38: It might be useful to have a single method to format the key name as this code is replicated a few times, If the format were to change it would be easy to miss one instance.

Library/src/Systems/HotSpotSequenceSystem/HotSpotSequenceSystem.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+	this->SequenceSystem = SequenceSystem;
+	this->SpaceSystem	 = SpaceSystem;
+}
+
+void HotSpotSequenceSystem::CreateHotspotGroup(const csp::common::String& GroupName,
+											   const csp::common::Array<csp::common::String>& HotspotIds,
+											   HotSpotGroupResultCallback Callback)
+{
+	auto CB = [Callback](SequenceResult Result)
```
MAG-mv, line 66: Callback parameters should be const&

MAG-mv (changes requested):

Library/include/CSP/Systems/HotSpotSequence/HotSpotGroup.h
```diff
@@ -0,0 +1,89 @@
...
+{
+	/** @cond DO_NOT_DOCUMENT */
+	CSP_START_IGNORE
+	template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+	CSP_END_IGNORE
+	/** @endcond */
+
+public:
+	const HotSpotGroup& GetHotSpotGroup() const;
+	CSP_NO_EXPORT HotSpotGroupResult(HotSpotGroup& group, csp::systems::EResultCode ResCode, uint16_t HttpResCode)
```
MAG-ME, line 46: Group could be a const ref

Library/include/CSP/Systems/HotSpotSequence/HotSpotGroup.h
```diff
@@ -0,0 +1,89 @@
...
+{
+	/** @cond DO_NOT_DOCUMENT */
+	CSP_START_IGNORE
+	template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+	CSP_END_IGNORE
+	/** @endcond */
+
+public:
+	const csp::common::Array<HotSpotGroup>& GetHotSpotGroups() const;
+	CSP_NO_EXPORT HotSpotGroupsResult(csp::common::Array<HotSpotGroup>& groups, csp::systems::EResultCode ResCode, uint16_t HttpResCode)
```
MAG-ME, line 70: groups could be a const ref

Library/include/CSP/Systems/HotSpotSequence/HotSpotSequenceSystem.h
```diff
@@ -0,0 +1,105 @@
...
+											 const csp::common::Array<csp::common::String>& HotspotIds,
+											 HotSpotGroupResultCallback Callback);
+
+	/// @brief Rename a Hotspot group
+	/// @param CurrentGroupName csp::common::String : The unique grouping name
+	/// @param NewGroupName csp::common::String : The unique grouping name
+	/// @param Callback HotSpotGroupResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void
+		RenameHotspotGroup(const csp::common::String& GroupName, const csp::common::String& NewGroupName, HotSpotGroupResultCallback Callback);
+	/// @brief Update a Hotspot group
```
MAG-ME, line 75: Blank line for readability

Library/include/CSP/Systems/HotSpotSequence/HotSpotSequenceSystem.h
```diff
@@ -0,0 +1,105 @@
...
+	/// @param Callback HotSpotGroupResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void GetHotspotGroup(const csp::common::String& GroupName, HotSpotGroupResultCallback Callback);
+
+
+	/// @brief Get all Hotspot groups for the current space and logged in user.
+	/// @param Callback HotSpotGroupsResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void GetHotspotGroups(HotSpotGroupsResultCallback Callback);
+
+
+	/// @brief Get a Hotspot group by name
```
MAG-ME, line 94: Looks like a copy/paste comment...replace Get with Delete

Library/include/CSP/Systems/HotSpotSequence/HotSpotSequenceSystem.h
```diff
@@ -0,0 +1,105 @@
...
+	/// @param GroupName csp::common::String : The unique grouping name
+	/// @param Callback HotSpotGroupResultCallback : callback to call when a response is received
+	CSP_ASYNC_RESULT void DeleteHotspotGroup(const csp::common::String& GroupName, HotSpotGroupResultCallback Callback);
```
MAG-ME, line 97: Is this the right callback? The response would be success or failure here I would think? No need to return a hotspot group.

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
```diff
@@ -0,0 +1,648 @@
...
+	auto [Result]
+		= Awaitable(&csp::systems::HotSpotSequenceSystem::CreateHotspotGroup, HotspotSequenceSystem, GroupName, Items).Await(RequestPredicate);
+
+	EXPECT_EQ(Result.GetResultCode(), ExpectedResultCode);
+	EXPECT_EQ(Result.GetFailureReason(), ExpectedResultFailureCode);
+
+	if (ExpectedResultCode == csp::systems::EResultCode::Success)
+	{
+		csp::systems::HotSpotGroup group = Result.GetHotSpotGroup();
+
```
MAG-ME, line 70: Check the group name?

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
```diff
@@ -0,0 +1,648 @@
...
+						csp::systems::EResultCode ExpectedResultCode				  = csp::systems::EResultCode::Success,
+						csp::systems::ERequestFailureReason ExpectedResultFailureCode = csp::systems::ERequestFailureReason::None)
+{
+	auto [Result]
+		= Awaitable(&csp::systems::HotSpotSequenceSystem::RenameHotspotGroup, HotSpotSequenceSystem, GroupName, NewGroupName).Await(RequestPredicate);
+
+	EXPECT_EQ(Result.GetResultCode(), ExpectedResultCode);
+	EXPECT_EQ(Result.GetFailureReason(), ExpectedResultFailureCode);
+
+	if (ExpectedResultCode == csp::systems::EResultCode::Success)
```
MAG-ME, line 151: Verify the new name matches the requested name

MAG-ME:
The tests verify things work as expected, but don't seem to test that invalid values are rejected correctly. Things like renaming or deleting groups that don't exist for example.

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
MAG-IY: We check metadata tags now for this

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
MAG-IY: We check metadata tags now for this

MAG-mv (approved):

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
MAG-ME: What I meant was is it worth confirming the group name we asked for is the one CHS returned? They could, in theory, return a different name.

MAG-ME:
The class names have been consistently (re)named (HotSpot vs Hotspot) but the folder and file names for those classes did not get renamed in the same way, and we probably should for consistency.

Tests/src/PublicAPITests/HotSpotSequenceSystemTests.cpp
MAG-IY: Yes, we check this, as when the group is retrieved, we extract the group name from metadata (as opposed to a constructed unique key)

MAG-IY:
> The class names have been consistently (re)named (HotSpot vs Hotspot) but the folder and file names for those classes did not get renamed in the same way, and we probably should for consistency.

I have tried to rename the files/directory, but git does not recognise the change.

MAG-ME (approved):

MAG-mv (approved):

PR #467, opened 2024-08-16
[OF-1394] fix: prevent using / in sequence keys

MAG-IY:
* added sequence key validation to all functions
* added invalid key test for all functions
* added new non public constructor for SequenceResult


MAG-SamBirley (approved):

Tests/src/PublicAPITests/SequenceSystemTests.cpp
```diff
@@ -391,6 +432,46 @@ CSP_PUBLIC_TEST(CSPEngine, SequenceSystemTests, GetSequenceTest)
...
+
+	// Create sequence
+	csp::common::Array<csp::common::String> SequenceItems {"Hotspot1", "Hotspot2", "Hotspot3"};
+	const char* TestSequenceKey = "CSP-UNITTEST/SEQUENCE-MAG";
+	char UniqueSequenceName[256];
+	SPRINTF(UniqueSequenceName, "%s-%s", TestSequenceKey, GetUniqueString().c_str());
+
+
+	csp::systems::Sequence RetrievedSequence;
+	GetSequence(SequenceSystem, UniqueSequenceName, RetrievedSequence, csp::systems::EResultCode::Failed);
```
MAG-AdamThorn, line 465: Won't this fail anyway as you haven't created a sequence to actually get?

Tests/src/PublicAPITests/SequenceSystemTests.cpp
```diff
@@ -443,6 +524,58 @@ CSP_PUBLIC_TEST(CSPEngine, SequenceSystemTests, UpdateSequenceTest)
...
+	SPRINTF(UniqueSequenceName, "%s-%s", TestSequenceKey, GetUniqueString().c_str());
+
+	csp::common::Map<csp::common::String, csp::common::String> MetaData;
+
+	// Update sequence
+	csp::common::Array<csp::common::String> UpdatedSequenceItems {"Hotspot4", "Hotspot5"};
+
+	csp::systems::Sequence UpdatedSequence;
+	MetaData["Foo"] = "Bar";
+	UpdateSequence(SequenceSystem,
```
MAG-AdamThorn, line 562: Won't this test fail anyway as you haven't first created a sequence to update?

Tests/src/PublicAPITests/SequenceSystemTests.cpp
```diff
@@ -588,6 +774,48 @@ CSP_PUBLIC_TEST(CSPEngine, SequenceSystemTests, GetSequencesByCriteriaTest)
...
+
+	// Test searches
+	csp::common::Array<csp::systems::Sequence> RetrievedSequences;
+
+	// Test Sequence key search
+	const char* TestSequenceKey = "CSP-UNITTEST/SEQUENCE-MAG";
+	char UniqueSequenceName[256];
+	SPRINTF(UniqueSequenceName, "%s-%s", TestSequenceKey, GetUniqueString().c_str());
+	// Get the first sequence
+	GetSequencesByCriteria(SequenceSystem, {UniqueSequenceName}, nullptr, nullptr, {}, RetrievedSequences, csp::systems::EResultCode::Failed);
```
MAG-AdamThorn, line 809: You don't seem to have created a sequence to actually retrieve so it will fail anyway regardless of the invalid name.

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-AdamThorn: Here the comment says create sequence but it doesn't, it just creates an invalid name.

MAG-IY: Yes, it will indeed fail, but with a different error code: 
Attempting to get a non-existent sequence will result in a 404 error from CHS.
Attempting to get a sequence with an invalid character name (which is what we are testing for) will respond with an error code of 0. 

Thus, no sequence actually needs to be created to test that an invalid name is rejected with the correct error code. If you look at the call signature for GetSequence you will see that the default value for the expected return code is 0, so it's sort of implied.

That being said, it's only implied, and so not clear, so I will tidy up the comments and make it explicit.

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-IY: Yes, but it's ok. See my other response

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-IY: See my other comment

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-AdamThorn: And this morning we agreed that you would add some additional comments to make it clear what was going on in each case.

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -79,6 +106,17 @@ void SequenceSystem::UpdateSequence(const String& SequenceKey,
...
 void SequenceSystem::RenameSequence(const String& OldSequenceKey, const String& NewSequenceKey, SequenceResultCallback Callback)
 {
+	if (!ValidateKey(OldSequenceKey))
+	{
+		CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Cannot rename Sequence. Old Key: %s contains invalid characters", OldSequenceKey.c_str());
+		return;
+	}
+	if (!ValidateKey(NewSequenceKey))
+	{
+		CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Cannot create Sequence. New Key: %s contains invalid characters", NewSequenceKey.c_str());
```
MAG-AdamThorn, line 116: Copy/paste error, this should be "Cannot rename Sequence..."

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -97,6 +135,16 @@ void SequenceSystem::GetSequencesByCriteria(const Array<String>& InSequenceKeys,
 											const csp::common::Map<csp::common::String, csp::common::String>& MetaData,
 											SequencesResultCallback Callback)
 {
+	for (size_t i = 0; i < InSequenceKeys.Size(); i++)
+	{
+		if (!ValidateKey(InSequenceKeys[i]))
+		{
+			CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Cannot get Sequence. New Key: %s contains invalid characters", InSequenceKeys[i].c_str());
```
MAG-AdamThorn, line 142: I don't think the log message should say "New Key:", just "Key:" as it is not a new key.

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -127,6 +175,12 @@ void SequenceSystem::GetSequencesByCriteria(const Array<String>& InSequenceKeys,
 
 void SequenceSystem::GetSequence(const String& SequenceKey, SequenceResultCallback Callback)
 {
+	if (!ValidateKey(SequenceKey))
+	{
+		CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Cannot get Sequence. New Key: %s contains invalid characters", SequenceKey.c_str());
```
MAG-AdamThorn, line 180: I don't think the log message should say "New Key:", just "Key:" as it is not a new key.

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -139,6 +193,17 @@ void SequenceSystem::GetSequence(const String& SequenceKey, SequenceResultCallba
 
 void SequenceSystem::DeleteSequences(const Array<String>& InSequenceKeys, NullResultCallback Callback)
 {
+	for (size_t i = 0; i < InSequenceKeys.Size(); i++)
+	{
+		if (!ValidateKey(InSequenceKeys[i]))
+		{
+			CSP_LOG_FORMAT(csp::systems::LogLevel::Error,
+						   "Cannot delete Sequence. New Key: %s contains invalid characters",
```
MAG-AdamThorn, line 201: I don't think the log message should say "New Key:", just "Key:" as it is not a new key.

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -21,6 +21,8 @@
 #include "Services/AggregationService/Api.h"
 #include "Systems/ResultHelpers.h"
 
+#include <sstream>
```
MAG-AdamThorn, line 24: This include is not required.

MAG-AdamThorn (changes requested):

Tests/src/PublicAPITests/SequenceSystemTests.cpp
MAG-IY: Done

MAG-AdamThorn (dismissed):

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -69,10 +76,8 @@ void SequenceSystem::CreateSequence(const String& SequenceKey,
 {
 	if (!ValidateKey(SequenceKey))
 	{
-		std::stringstream Ss;
-		Ss << "Cannot Create Sequence:  Key " << SequenceKey.c_str() << " contains invalid '/' characters";
-		CSP_LOG_ERROR_MSG({Ss.str().c_str()});
-		INVOKE_IF_NOT_NULL(Callback, MakeInvalid<SequenceResult>());
+		CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Cannot create Sequence. Key: %s contains invalid characters", SequenceKey.c_str());
```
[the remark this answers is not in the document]
MAG-IY: agreed

Library/include/CSP/Systems/Sequence/SequenceSystem.h
```diff
@@ -53,6 +53,7 @@ class CSP_API SequenceSystem : public SystemBase
 public:
 	// Hotspot:[SpaceId]:[GroupName]
 	/// @brief Creates a new sequence. If a sequence already exists with this key, it will overwrite the current one.
+	/// @brief This call will fail (Reason InvalidSequenceKey) if the SequenceKey parameter contains invalid keys, such as spaces, '/' or '%'
```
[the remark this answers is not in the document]
MAG-IY: Let's do the best practice then

MAG-AdamThorn (approved):

MAG-AdamThorn (approved):

PR #468, opened 2024-08-20
[NT-0] test: Ensure tests don't self-message

MAG-AW:
* Unless specified in the test, tests won't inherit self-messaging flags

### DIFF
--- a/Tests/src/PublicTestBase.cpp
+++ b/Tests/src/PublicTestBase.cpp
@@ -15,6 +15,7 @@
  */
 #include "PublicTestBase.h"
 
+#include "Awaitable.h"
 #include "CSP/CSPFoundation.h"
 #include "CSP/Systems/Log/LogSystem.h"
 #include "CSP/Systems/SystemsManager.h"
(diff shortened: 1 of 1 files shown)


Tests/src/PublicTestBase.cpp
```diff
@@ -36,12 +37,20 @@ void PublicTestBase::SetUp()
...
+	AWAIT(Connection, SetAllowSelfMessagingFlag, false);
 }
 
 void PublicTestBase::TearDown()
 {
 	::testing::Test::TearDown();
 
+	auto Connection = csp::systems::SystemsManager::Get().GetMultiplayerConnection();
+
+	AWAIT(Connection, SetAllowSelfMessagingFlag, false);
```
MAG-AdamThorn, line 52: Why do we set it false again here in teardown, won't it already be false?

MAG-SamBirley: Isn't the multiplayer connection object destroyed (along with CSP itself) when CSP is [deinitialised ](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Tests/src/PublicTestBase.cpp#L55) at the end of each test?

MAG-SamBirley: Right, I think I get it now.

Next stupid question - why do we need to set it to false in teardown if we also set it to true before the start of each test? Won't that also cover the case where a test sets it to true and forgets to set it back to false?

MAG-SamBirley: @MAG-AW pointed out to me on a call that:
* This flag persists across user sessions
* And so setting it to false at the end of the test also ensures it gets set back to the expected value after the last test runs.

MAG-SamBirley (approved):

MAG-AdamThorn (approved):

PR #469, opened 2024-08-20
[OF-1384] refac: Moved Conversation Functions

MAG-AW:
* Moved functions to Component from ConversationSystem
* Deleted ConversationSystem and references
* Updated tests
* Some tests involving network events are commented out for now, pending changes to the events in a later ticket

**For the reviewer**

### DIFF
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -90,7 +90,7 @@ public:
 	CSP_ASYNC_RESULT void DeleteAssetCollection(const AssetCollection& AssetCollection, NullResultCallback Callback);
 
 	/// @brief Deletes a given array of asset collections.
-	/// @param  Array<AssetCollection> AssetCollections : The array of asset collections to delete
+	/// @param AssetCollections csp::common::Array<AssetCollection> : The array of asset collections to delete
 	/// @param Callback NullResultCallback : callback when asynchronous task finishes
 	CSP_ASYNC_RESULT void DeleteMultipleAssetCollections(csp::common::Array<AssetCollection>& AssetCollections, NullResultCallback Callback);
 


Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -274,4 +751,121 @@ const csp::common::String& ConversationSpaceComponent::GetConversationId() const
...
+				++*DeletionCounter;
+
+				if (*DeletionCounter == MessagesCount)
+				{
+					INVOKE_IF_NOT_NULL(Callback, DeleteCommentResult);
+				}
+			}
+		};
+
+		AssetSystem->DeleteAssetCollection(Messages[idx], DeleteCommentCallback);
```
MAG-AdamThorn, line 867: Is there a reason we are doing it this way and not just calling the DELETE /api/v1/prototypes CHS endpoint? That would allow us to just pass in an array of the prototype Ids we want to delete and delete them all in one go.

Looking at the AssetSystem we haven't yet exposed that endpoint but it's easy enough to do.
https://ogs-odev-internal.magnoboard.com/mag-prototype/swagger/index.html

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -205,17 +495,204 @@ void ConversationSpaceComponent::SetConversationInfo(const MessageInfo& Conversa
...
+				= ConversationSystemHelpers::GetMessageInfoFromMessageAssetCollection(GetConversationResult.GetAssetCollection());
+			NewConversationData.EditedTimestamp = CurrentConversationData.EditedTimestamp;
+		}
+
+		AssetSystem->UpdateAssetCollectionMetadata(GetConversationResult.GetAssetCollection(),
+												   ConversationSystemHelpers::GenerateConversationAssetCollectionMetadata(NewConversationData),
+												   GetUpdatedConversationCallback);
+	};
+
+	AssetSystem->GetAssetCollectionById(ConversationId, GetConversationCallback);
```
MAG-AdamThorn, line 579: Do we need to make this call? We currently need it as the UpdateAssetCollectionMetadata method takes the AssetCollection object, but it only requires the Id. Can we not rather create an alternate method for updating metadata by Id or replace and deprecate the existing one? This feels inefficient as it stands.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-AdamThorn: Yes I definitely think we should take this as an opportunity to improve things if we can. Thanks.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -274,4 +751,121 @@ const csp::common::String& ConversationSpaceComponent::GetConversationId() const
...
+	}
+
+	const csp::systems::AssetCollectionResultCallback AddCommentCallback = [=](const csp::systems::AssetCollectionResult& AddCommentResult)
+	{
+		if (AddCommentResult.GetResultCode() == csp::systems::EResultCode::InProgress)
+		{
+			return;
+		}
+
+		if (AddCommentResult.GetResultCode() == csp::systems::EResultCode::Success)
```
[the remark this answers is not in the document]
MAG-AdamThorn: It would just mean having an if block for the failure case and then moving success out of the nested if block. It doesn't change anything about the execution, just matches the flow convention used everywhere else.

MAG-AdamThorn (approved):

PR #471, opened 2024-08-21
[OF-1375] feat: Implement Sequence Hierarchies

MAG-mv:
- Added the following functions to SpaceEntitySystem:
    - CreateSequenceHierarchy
    - UpdateSequenceHierarchy
    - GetSequenceHierarchy
    - GetAllSequenceHierarchies
    - DeleteSequenceHierarchy
    - SetSequenceHierarchyChangedCallback
- Added tests
- Added SequenceHierarchy data classes
- Added event deserializer


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -442,14 +442,29 @@ void MultiplayerConnection::StartEventMessageListening()
 		}
 		else if (EventType == "SequenceChanged")
 		{
-			if (!SequenceChangedCallback)
+			SequenceChangedEventDeserialiser SequenceDeserialiser;
+			SequenceDeserialiser.Parse(EventValues);
+
+			if (SequenceChangedCallback)
```
MAG-mv, line 448: Created another event here, as we dont want to overwrite the SequenceChanged event, as this could be being used by clients

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1067,6 +1074,116 @@ const csp::common::List<SpaceEntity*>* SpaceEntitySystem::GetRootHierarchyEntiti
...
+void SpaceEntitySystem::GetAllSequenceHierarchies(SequenceHierarchiesResultCallback Callback)
+{
+	const auto* SpaceSystem		 = csp::systems::SystemsManager::Get().GetSpaceSystem();
+	const common::String SpaceId = SpaceSystem->GetCurrentSpace().Id;
+
+	auto GetSequencesCallback = [Callback](const systems::SequencesResult& GetSequenceResult)
+	{
+		SequenceHierarchiesResult Result(GetSequenceResult.GetResultCode(), GetSequenceResult.GetHttpResultCode());
+
+		if (GetSequenceResult.GetResultCode() != systems::EResultCode::InProgress)
```
MAG-AdamThorn, line 1151: What about if it fails?

Library/include/CSP/Multiplayer/SequenceHierarchy.h
```diff
@@ -0,0 +1,92 @@
...
+	SequenceHierarchyResult(void*) {};
+
+	void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
+
+	SequenceHierarchy SequenceHierarchy;
+};
+
+/// @ingroup Space Entity System
+/// @brief Data class used to contain information when attempting to get an array of sequence hierachies
+class CSP_API SequenceHierarchiesResult : public csp::systems::ResultBase
```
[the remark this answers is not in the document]
MAG-AdamThorn: I like that idea.

Tests/src/PublicAPITests/New Text Document.txt
MAG-IY, line 1: Does this file need to exist?

Library/src/Multiplayer/SequenceHierarchy.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+}
+
+common::String CreateSequenceKey(common::Optional<uint64_t> ParentId, const common::String& SpaceId)
+{
+	common::String Key = SequenceHierarchyName + ":" + SpaceId;
+
+	if (ParentId.HasValue())
+	{
+		Key += ":" + systems::SequenceIdPrefix + std::to_string(*ParentId).c_str();
```
MAG-IY, line 25: Shouldn't we be checking these for invalid (% / ) characters?

MAG-IY (changes requested):
Most looks fine. Just a couple of comments

Library/include/CSP/Multiplayer/SequenceHierarchy.h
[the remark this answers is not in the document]
MAG-mv: I like it

Library/src/Multiplayer/SequenceHierarchy.cpp
MAG-mv: ParentId is a uint64_t, and SpaceId is a hex string, so this would never happen unless a user passes in a random string for the SpaceId

MAG-mv: This function isn't used internally by the SequenceHierachy, so there is no way they could pass in an invalid space

MAG-mv: This function is only exposed for testing purposes

MAG-AdamThorn (dismissed):

MAG-AdamThorn (approved):

PR #473, opened 2024-09-04
[NT-0] fix: crash on DLL load for CPP apps

MAG-SamBirley:
### Bug & Fix

This change concerns moving a static string declaration in the global scope out of a public header. Specifically `SequenceHierarchyName`.

Global variable strings within CSP can cause apps using CSP dlls to *crash on load* of the DLL, as it leads to the application attempting to instantiate the global-scope string variable on DLL-load, _before_ CSP has initialised its memory allocators, which the string class depends upon. The end result is a null dereference crash.

The fix though is to _not_ declare this string variable at the global scope, because this will always be risky with respect to CSP's memory allocators. Instead it is retrieved through a global-scope static function, which more or less guarantees that it will only ever be called after the allocators have been initialised.

### Additionally...

Since in this case, since the variable is only used internally within CSP, it has also been moved to an internal source file.

To avoid creating a brand new header for a single constant, the `MultiplayerKeyConstants.h` file has been used for this and renamed to the more generic `MultiplayerConstants.h`.

### DIFF
--- a/Library/include/CSP/Multiplayer/SequenceHierarchy.h
+++ b/Library/include/CSP/Multiplayer/SequenceHierarchy.h
@@ -92,10 +92,6 @@ private:
 	csp::common::Array<SequenceHierarchy> SequenceHierarchyCollection;
 };
 
-CSP_START_IGNORE
-static inline const csp::common::String SequenceHierarchyName = "EntityHierarchy";
-CSP_END_IGNORE
-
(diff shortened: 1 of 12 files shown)


MAG-SamBirley:
Updated the PR to also address a similar issue with `SequenceIdPrefix` in `Sequence.h` following the same pattern.

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #476, opened 2024-09-09
[OF-1373] feat: Add global transform support

MAG-IY:
* Added operators for vector classes
* Added addition operator for SpaceTranform
* Added global transform data getters
* Added transform equality operator
* Added simple test to verify above

### DIFF
--- a/Library/src/Common/Vector.cpp
+++ b/Library/src/Common/Vector.cpp
@@ -44,6 +44,31 @@ const Vector3& Vector3::One()
 	return _One;
 }
 
+Vector3 Vector3::operator+(const Vector3& Other) const
+{
+	return Vector3(X + Other.X, Y + Other.Y, Z + Other.Y);
+}
+
+Vector3 Vector3::operator-(const Vector3& Other) const
+{
+	return Vector3(X - Other.X, Y - Other.Y, Z - Other.Y);
+}
+
+Vector3 Vector3::operator/(float Divisor) const
+{
+	return Vector3(X / Divisor, Y / Divisor, Z / Divisor);
+}
+
+Vector3 Vector3::operator*(const Vector3& Other) const
+{
+	return Vector3(X * Other.X, Y * Other.Y, Z * Other.Z);
+}
+
+Vector3 Vector3::operator*(float Scalar) const
+{
+	return Vector3(X * Scalar, Y * Scalar, Z * Scalar);
+}
+
 bool Vector3::operator==(const Vector3 Other) const
 {
 	return IsNearlyEqual(X, Other.X) && IsNearlyEqual(Y, Other.Y) && IsNearlyEqual(Z, Other.Z);
@@ -74,6 +99,31 @@ const Vector4& Vector4::Identity()
 	static Vector4 _Identity = {0, 0, 0, 1};
 
 	return _Identity;
+}
+
+Vector4 Vector4::operator+(const Vector4& Other) const
+{
+	return Vector4(X + Other.X, Y + Other.Y, Z + Other.Z, W + Other.W);
+}
+
+Vector4 Vector4::operator-(const Vector4& Other) const
+{
+	return Vector4(X - Other.X, Y - Other.Y, Z - Other.Z, W - Other.W);
+}
+
+Vector4 Vector4::operator/(float Divisor) const
+{
+	return Vector4(X / Divisor, Y / Divisor, Z / Divisor, W / Divisor);
+}
+
+Vector4 Vector4::operator*(const Vector4& Other) const
+{
+	return Vector4(X * Other.X, Y * Other.Y, Z * Other.Z, W * Other.W);
+}
+
+Vector4 Vector4::operator*(float Scalar) const
+{
+	return Vector4(X * Scalar, Y * Scalar, Z * Scalar, W * Scalar);
 }
 
 bool Vector4::operator==(const Vector4 Other) const
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -179,9 +179,24 @@ const SpaceTransform& SpaceEntity::GetTransform() const
 	return Transform;
 }
 
+const SpaceTransform& SpaceEntity::GetGlobalTransform() const
+{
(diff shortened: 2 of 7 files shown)


Library/src/Common/Vector.cpp
```diff
@@ -44,6 +44,31 @@ const Vector3& Vector3::One()
 	return _One;
 }
 
+Vector3 Vector3::operator+(const Vector3& Other) const
+{
+	return Vector3(X + Other.X, Y + Other.Y, Z + Other.Y);
```
MAG-AdamThorn, line 49: Typo, the last one should be 'Z + Other.Z'

Library/src/Common/Vector.cpp
```diff
@@ -44,6 +44,31 @@ const Vector3& Vector3::One()
...
+}
+
+Vector3 Vector3::operator-(const Vector3& Other) const
+{
+	return Vector3(X - Other.X, Y - Other.Y, Z - Other.Y);
```
MAG-AdamThorn, line 54: Typo, the last one should be 'Z + Other.Z'

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -2346,6 +2346,103 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ParentEntityReplicationTest)
...
+	// Enter space
+	auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id);
+	EXPECT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+	// Create Entities
+	csp::common::String ParentEntityName = "ParentEntity";
+	csp::common::String ChildEntityName	 = "ChildEntity";
+	SpaceTransform ObjectTransformParent = {csp::common::Vector3 {1, 2, 3}, csp::common::Vector4 {4, 5, 6, 7}, csp::common::Vector3 {9, 8, 7}};
+	SpaceTransform ObjectTransformChild	 = {csp::common::Vector3 {1, 1, 1}, csp::common::Vector4 {1, 1, 1, 1}, csp::common::Vector3 {1, 1, 1}};
+	// SpaceTransform ObjectTransformExpected = {csp::common::Vector3 {2, 3, 4}, csp::common::Vector4 {5, 6, 7, 8}, csp::common::Vector3 {10, 9, 8}};
```
MAG-AdamThorn, line 2386: Can we remove this commented line?

Library/src/Multiplayer/SpaceTransform.cpp
```diff
@@ -25,3 +25,13 @@ csp::multiplayer::SpaceTransform::SpaceTransform(const csp::common::Vector3& Pos
 	: Position(Position), Rotation(Rotation), Scale(Scale)
 {
 }
+
+csp::multiplayer::SpaceTransform csp::multiplayer::SpaceTransform::operator+(const SpaceTransform& Transform)
+{
+	return SpaceTransform(Position + Transform.Position, Rotation + Transform.Rotation, Scale + Transform.Scale);
```
MAG-SamBirley, line 31: This isn't a mathematically correct transformation from one space to another. 

Position/rotation/scale cannot be simply be added together, as each element is not just dependent upon itself, but also the other two elements.

We need to do the equivalent of a matrix multiplication here to derive a transform relative to the entity's space.

MAG-SamBirley: A good reference for what converting from a (vector,quat,vector) to a matrix representation of a transform may look like is Epic's implementation [here].(https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Math/TransformNonVectorized.h#L241). They follow a similar pattern to CSP where transform data is stored as (vector,quat,vector), and converted to a matrix when there is a need to combine two transforms together.

Note you will need to register with Epic if you haven't already in order to view engine source.

Library/src/Multiplayer/SpaceTransform.cpp
```diff
@@ -25,3 +25,13 @@ csp::multiplayer::SpaceTransform::SpaceTransform(const csp::common::Vector3& Pos
...
+}
+
+bool csp::multiplayer::SpaceTransform::operator==(const SpaceTransform& Transform) const
+{
+	return Position == Transform.Position && Rotation == Transform.Rotation && Scale == Transform.Scale;
```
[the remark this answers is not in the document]
MAG-IY: The equality operators already have that built in: Look at Vector.cpp, lines 74, and 129

Library/src/Multiplayer/SpaceTransform.cpp
MAG-IY: The orientation and positions are fine to just add, but I agree the scale needs looking at.

MAG-SamBirley: Sounds good

MAG-mv (approved):

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -179,11 +180,26 @@ const SpaceTransform& SpaceEntity::GetTransform() const
...
+
 const csp::common::Vector3& SpaceEntity::GetPosition() const
 {
 	return Transform.Position;
 }
 
+csp::common::Vector3 SpaceEntity::GetGlobalPosition() const
+{
+	if (Parent != nullptr)
+		return Parent->GetTransform().Position + Transform.Position;
```
MAG-AdamThorn, line 198: Does the global position not need to take into account the parents rotation?

MAG-mv: Doesn't scale need to eb taken into account in some circumstances, too?

MAG-AdamThorn: Another issue with the above is we should be calling the parent's GetGlobalTransform here, as otherwise you're just returning the transform relative to the parent (which isn't necessarily the root of the hierarchy).

MAG-AdamThorn (approved):

MAG-mv (approved):

PR #497, opened 2024-09-26
[OF-1419] feat: hotspot names support the space character

MAG-SamBirley:
CSP client applications can now reason about sequence keys that include characters usually reserved in URIs, such as spaces.

As part of this work, we have introduced a new set of encoding source files for future encode/decode behaviour.

During development, we discovered that there are some service endpoints currently expecting the provided keys to be _doubly_ encoded. The current approach takes this into account and double-encodes where necessary from a services point of view.

### DIFF
--- a/Tests/src/InternalTests/EncodeTests.cpp
+++ b/Tests/src/InternalTests/EncodeTests.cpp
@@ -45,6 +45,14 @@ CSP_INTERNAL_TEST(CSPEngine, EncodeTests, EncodeURI)
 		csp::common::String DecodedString = csp::common::Decode::URI(EncodedString);
 		EXPECT_EQ(DecodedString, OriginalURL);
 	}
+
+	{
+		// Double encoding and double decoding should produce the same end-result.
+		const csp::common::String OriginalURL("abc defghij*zlmnopqrst#uvwxyz");
+		csp::common::String EncodedString = csp::common::Encode::URI(OriginalURL, true);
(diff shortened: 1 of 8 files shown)


MAG-SamBirley:
WASM check is failing as this change depends on Poco URL formatting (we strip Poco out of WASM builds).

MAG-SamBirley:
Poco dependency for this change has now been stripped out. Instead of using Poco URI encoding, we use simple hand-rolled version instead.

MAG-mv (approved):

MAG-mv (dismissed):

Tests/src/InternalTests/EncodeTests.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#ifndef SKIP_INTERNAL_TESTS
+
+#include "TestHelpers.h"
+#include "Common/Encode.h"
+
+#include "gtest/gtest.h"
+
+CSP_INTERNAL_TEST(CSPEngine, EncodeTests, EncodeURI)
```
MAG-AdamThorn, line 24: As we are using double encoding for the sequence keys I think our tests should cover that.

MAG-AdamThorn (changes requested):

MAG-AdamThorn (approved):

MAG-mv (approved):

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #498, opened 2024-09-27
[OF-1421] feat: Add llvm version in prerequisites

MAG-ChristopherAtkinson:
install_prerequisites.ps1 now specifies the version of llvm to use to reduce the inconsistencies in clang-format for new environments. 

As part of the changes included, we have introduced space delimiter string and array expansion to format the required command, this change will allow for any available argument applicable to chocolatey install to be provided in future, but is currently only used to specify version formation for a single package. 

More information on `choco install` can be found [here](https://docs.chocolatey.org/en-us/choco/commands/install/).


MAG-AdamThorn (approved):

MAG-SamBirley (approved):
Pulled and ran the script locally with no issue. Installed the expected version. Nice work.

PR #501, opened 2024-09-30
[OB-3719] fix: static initialization

MAG-mv:
- Internal static component id now checks if it has been initialized to prevent it concantinating
- Test has been updates to test for this case

### DIFF
--- a/Library/src/Multiplayer/Components/HotspotSpaceComponent.cpp
+++ b/Library/src/Multiplayer/Components/HotspotSpaceComponent.cpp
@@ -69,9 +69,14 @@ void HotspotSpaceComponent::SetIsSpawnPoint(bool Value)
 
 const csp::common::String& HotspotSpaceComponent::GetUniqueComponentId() const
 {
-	static csp::common::String UniqueComponentId = std::to_string(Parent->GetId()).c_str();
-	UniqueComponentId += ":";
-	UniqueComponentId += std::to_string(Id).c_str();
+	static csp::common::String UniqueComponentId;
+
+	if (UniqueComponentId.Length() == 0)
+	{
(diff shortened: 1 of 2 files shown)


MAG-SamBirley:
Do we know why it's static at all? What I always interpreted that function to be was a unique identifier per-component, in which case having it be just a local variable initialised every time would make more sense?

Might be missing something though.

MAG-SamBirley:
> Do we know why it's static at all? What I always interpreted that function to be was a unique identifier per-component, in which case having it be just a local variable initialised every time would make more sense?
> 
> Might be missing something though.

Al

MAG-SamBirley:
Sorry about that, misclick! 😱

MAG-AdamThorn (dismissed):

MAG-SamBirley (commented):
Do we need to consider scenarios where the entity is a child of another entity which is itself a child? Do we need to recurse up the parent tree calling `GetUniqueComponentId` in order to be certain it's truly unique?

MAG-mv:
> Do we need to consider scenarios where the entity is a child of another entity which is itself a child? Do we need to recurse up the parent tree calling `GetUniqueComponentId` in order to be certain it's truly unique?

Each entitys id is unique, so this should ensure uniqueness

MAG-SamBirley:
> > Do we need to consider scenarios where the entity is a child of another entity which is itself a child? Do we need to recurse up the parent tree calling `GetUniqueComponentId` in order to be certain it's truly unique?
> 
> Each entitys id is unique, so this should ensure uniqueness

Right, yes of course! Thanks

MAG-SamBirley (dismissed):

MAG-SamBirley (approved):

MAG-AdamThorn (approved):

PR #502, opened 2024-09-30
[OF-1407] feat Hotspot group CRUD events

MAG-SamBirley:
The CSP API now includes surface area allowing the client to bind to a new event callback, via `SetHotspotSequenceChangedCallback` on the multiplayer connection.

The callback will fire when a hotspot group has a CRUD operation applied to it. It describes the change type (Create, Update, Rename, Delete), the hotspot group name and the ID of the space that the hotspot group relates to.


Library/src/Multiplayer/EventSerialisation.h
```diff
@@ -146,4 +146,20 @@ class SequenceHierarchyChangedEventDeserialiser : public EventDeserialiser
 	SequenceHierarchyChangedParams EventParams;
 };
 
+/// A deserialiser for getting SequenceHotspot data from an event:
+/// TODO!!
```
MAG-mv, line 150: Whats this todo?

MAG-mv (changes requested):

Library/src/Multiplayer/EventSerialisation.cpp
```diff
@@ -28,26 +28,30 @@ namespace
 {
 ESequenceUpdateType ESequenceUpdateIntToUpdateType(uint64_t UpdateType)
 {
+    ESequenceUpdateType SequenceUpdateType = ESequenceUpdateType::Invalid;
+
 	if (UpdateType == 0)
```
MAG-ME, line 33: Curious why we don't use a switch statement here? It's more efficient apparently.

Library/src/Multiplayer/EventSerialisation.cpp
```diff
@@ -60,12 +64,19 @@ std::string RemoveIdPrefix(const std::string& Id)
...
+	return csp::common::Decode::URI(RawValue.GetString());
+}
+
 } // namespace
 
 csp::common::String csp::multiplayer::GetSequenceKeyIndex(const csp::common::String& SequenceKey, int i)
 {
 	const std::string SequenceKeyString(SequenceKey.c_str());
-	// Match item after second ':' to get our parent id
+	// Match item after second ':' to get our parent Id.
```
MAG-ME, line 78: "i" should probably be unsigned

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -450,21 +455,29 @@ void MultiplayerConnection::StartEventMessageListening()
...
-				}
-			}
+            // There are a variety of sequence types.
+            // Other CSP callbacks may also need to fire if the sequence change relates to a particular sequence type.
+			const csp::common::String Key = SequenceDeserialiser.GetEventParams().Key;
+            const csp::common::String SequenceType = GetSequenceKeyIndex(Key, 0);
+            if (SequenceType == "EntityHierarchy")
+            {
+				auto EntitySystem = csp::systems::SystemsManager::Get().GetSpaceEntitySystem();
+                if(EntitySystem->SequenceHierarchyChangedCallback)
```
MAG-ME, line 465: picky but inconsistent use of space after if statement (Michael picked me up on something like this so blame him)

Tests/src/PublicAPITests/HotspotSequenceTests.cpp
```diff
@@ -220,11 +204,40 @@ CSP_PUBLIC_TEST(CSPEngine, HotspotSequenceTests, CreateHotspotGroupTest)
...
+		EXPECT_EQ(Params.SpaceId, Space.Id);
+		EXPECT_EQ(Params.HotspotGroupName, TestGroupName);
+		CallbackCalled = true;
+	});
+
 	// Delete sequence
 	DeleteHotspotGroup(HotspotSystem, TestGroupName);
+
+	// Clear out the callback as we have validated what we came here for.
+	WaitForCallback(CallbackCalled);
```
MAG-ME, line 237: What if we get more callbacks than expected?

Library/src/Multiplayer/EventSerialisation.h
MAG-SamBirley: An oversight! Will fix.

Library/src/Multiplayer/EventSerialisation.cpp
MAG-SamBirley: No good reason I know of. Will update to use a switch statement.

Library/src/Multiplayer/EventSerialisation.cpp
MAG-SamBirley: I agree with @MAG-ME in this case. Making it unsigned will provide additional protection in the form of warnings at compile time should a developer attempt to pass a signed integer in. Will make the change.

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-SamBirley: You got it!

Tests/src/PublicAPITests/HotspotSequenceTests.cpp
MAG-SamBirley: It's a valid question, but given that this approach was taken to conform with the general approach followed currently in CSP tests, I'd prefer to avoid addressing it in this change and save it for one more targeted towards upgrading this pattern.

Library/src/Multiplayer/EventSerialisation.h
MAG-SamBirley: Fixed in latest.

Library/src/Multiplayer/EventSerialisation.cpp
MAG-SamBirley: Fixed in latest.

Library/src/Multiplayer/EventSerialisation.cpp
MAG-SamBirley: Fixed in latest.

MAG-mv (approved):

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-SamBirley: Fixed in latest.

MAG-ME: Still seeing a few 'if's without spaces. Not a big deal fi we aren't that concerned.

MAG-SamBirley: Oh, that. Yes I'm not personally concerned and it's not actually called out in the coding standards. If you're happy, I'm happy.

MAG-ME (approved):

PR #503, opened 2024-10-01
[NT-0] refac: CSP VS solutions now target VS2022 

MAG-SamBirley:
The premake scripts responsible for generating Visual Studio solutions now generate VS2022 solutions by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ ARM64/*
 **/node_modules/**
 *.log
 *.npmrc
+app.config
 
 # Allows Project files for Examples
 !/Examples/**/**.sln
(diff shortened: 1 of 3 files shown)


MAG-SamBirley:
As part of this change, we discovered that there was a bug in quickjs that led to Release builds under VS2022 failing to compile.
The issues had already been addressed [over here](https://github.com/quickjs-ng/quickjs/issues/309).

I briefly sought to use this as an opportunity to refactor CSP's usage of quickjs to be git module based, however that led to other build issues, so eventually sought the faster path of directly integrating the fix into CSP's version of quickjs.

MAG-AdamThorn (approved):

magnopus-swifty:
LGTM. Anecdotally I've been running CSP on 2022 since I started contributing and haven't encountered any problems.

MAG-mv (approved):

PR #509, opened 2024-10-07
[nt-0] fix: remove missing submodules

magnopus-swifty:
`modules/msgpack` and `modules/openssl` don't have corresponding `gitlink`s so their entries in `.gitmodules` are incorrect

Further discussion in [this Slack thread](https://magnopus.slack.com/archives/C037VC209QT/p1728074303456579)


MAG-mv (approved):

MAG-SamBirley (approved):
LGTM

PR #512, opened 2024-10-14
[OB-3663] feat: added uri validation

MAG-mv:
- Update CSP Initialise function to pass in Agent info to prevent ordering issues
- Adding Ping to CSP Initialise to ensure uri is valid
- Added test for failure


Library/src/CSPFoundation.cpp
```diff
@@ -382,12 +390,19 @@ const csp::common::String& CSPFoundation::GetTenant()
 
 void CSPFoundation::SetClientUserAgentInfo(const csp::ClientUserAgent& ClientUserAgentHeader)
```
MAG-AdamThorn, line 391: Do we still need this method if the ClientUserAgent info is being passed in via the Initialize method?

MAG-mv: I kept it just in case clients want to update again. But I can make private if we dont think its necessary

MAG-mv: It might be better to hide it tbf. I dont know why they would want to do this. What do you think?

MAG-AdamThorn: Yes I agree, I think we should make it private

PR #517, opened 2024-10-15
CSP 5.3.1

MAG-AdamThorn:
- Remove manual GetMusubiLODTest test.
- Introduce new python script to direct users to the GitHub release notes via the npmjs package change log.


MAG-mv (approved):

MAG-OR (commented):
I'm not going to review this because I am a newbie, but I did have a look at the change :-)
I see this PR is from develop to Main, if I understand correctly this means it will go live?

MAG-AdamThorn:
@MAG-OR Once it has gone to Main it can then be tagged and published as a release.

PR #519, opened 2024-10-18
Reseat example links from main -> develop branch

Demorro:
Something I noticed whilst browsing for getting started examples.

Links to the getting started examples were outdated, and hitting 404's, presumably from a prior branch name switch from main -> develop.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -83,16 +83,16 @@ For more information about tenants and tenant IDs, visit our [FAQ.](https://www.
 To help you get started using the Connected Spaces Platform, we have a set of examples for each supported language. We keep these examples up to date with each new major version of the library.
 
 - WebAssembly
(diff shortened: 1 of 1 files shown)


MAG-AdamThorn:
Thank you for your PR, we really appreciate you taking the time to raise one for this issue.

We recently encountered a problem with our main branch and were forced to create a new main branch, which we called `Main` (with a capital 'M') to avoid confusion. We plan to rename this back to main (with a small 'm') soon, but are in the middle of some infrastructure changes which include the introduction of a new public Team City server, so have elected to wait until that work is complete before renaming the branch.

We clearly missed a place that needed updating though, so thank you for drawing that to our attention. The fix in this case is not to reseat the urls from `main` to `develop`, but from `main` to `Main`. If you are happy to make that change to your PR that would be great and we can approve it to be merged.

Thank you

MAG-AdamThorn (approved):

MAG-AdamThorn:
That's great thank you. I have approved, just need another member of the team to do the same and it will get merged!

MAG-mv (approved):

Issue #524, opened 2024-10-23
Generated `SpaceEntity.GetParentEntity()` returns wrapped null pointer, causing hard crash


magnopus-swifty:
**Describe the bug**

In the generated C# code, `SpaceEntity.GetParentEntity()` looks like this:

```csharp
public Csp.Multiplayer.SpaceEntity GetParentEntity()
{
	var _result = new Csp.Multiplayer.SpaceEntity(
		csp_multiplayer_SpaceEntity_GetParentEntityC_SpaceEntityP(_ptr)
	);

	return _result;
}
```

In the C++ code ([SpaceEntity.h](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/include/CSP/Multiplayer/SpaceEntity.h)) `GetParentEntity()` returns `nullptr` when the entity has no parent.

The generated code should do a null check when wrapping a `NativePointer` and return a `null` managed object.

**To Reproduce**

In C#, call `SpaceEntity.GetParentEntity()` in a space on any entity that doesn't have a parent.

**Expected behaviour**

`SpaceEntity.GetParentEntity()` should return `null`.

**CSP version (please complete the following information):**

`v5.3.1+544`

**Language:**

C#

magnopus-swifty:
[this line](https://github.com/magnopus-opensource/connected-spaces-platform/blob/933e731936ee411f8314f530e9efb52af836fcd2/Tools/WrapperGenerator/Templates/CSharp/Partials/Template/Method.mustache#L42) of `Method.mustache` needs modifying to check the return value of the called function before wrapping it.

PR #525, opened 2024-10-23
#524 Check for null when calling functions that return a pointer

magnopus-swifty:
Modify the templates so functions that return values store the result in a temporary local var. This is then checked for `null` so a C# `null` can be returned instead of a wrapped `null` (which would cause a hard crash if dereferenced).

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -63,30 +63,26 @@ validate_commit_msg() {
     local second_word=$(echo "${subject}" | awk '{print $2;}')
     local is_allowed=false
 
-    
     # Validate the ticket ID
-    re='^[0-9]+$'
-    first_character=${first_word:0:1}
-    last_character=${first_word: -1}
-    if [[ "$first_character" == *"["* ]] && [[ "$last_character" == *"]"* ]] && [[ "$first_word" == *"-"* ]]  ; then
-        truncated_first_word=${first_word#"["}
-        truncated_first_word=${truncated_first_word%"]"}
-        IFS='-' read -ra ADDR <<< "$truncated_first_word"
-        if ! [[ ${ADDR[0]} =~ $re ]] ; then
-            if [[ ${ADDR[1]} =~ $re ]] ; then
-                is_allowed=true
-            fi
-        fi
+    if [[ "$first_word" =~ ^\[([[:alpha:]]+-|#)([[:digit:]]+)\]$ ]]; then
+        is_allowed=true
     fi
 
-    if [[ "${is_allowed}" == true ]]; then
-        is_allowed=false
-        for verb in "${allowed_subject_verbs[@]}"; do
-            if [[ "${verb}" == "${second_word}" ]]; then
-                is_allowed=true
-            fi
-        done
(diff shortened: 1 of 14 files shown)


magnopus-swifty:
This PR changes the output for functions that returns pointers (like `GetParentEntity()`) from:

```csharp
        public Csp.Multiplayer.SpaceEntity GetParentEntity()
        {
            var _result = new Csp.Multiplayer.SpaceEntity(
                csp_multiplayer_SpaceEntity_GetParentEntityC_SpaceEntityP(_ptr)
            );

            return _result;
        }
```

to

```csharp
        public Csp.Multiplayer.SpaceEntity GetParentEntity()
        {
            var _value = csp_multiplayer_SpaceEntity_GetParentEntityC_SpaceEntityP(_ptr);

            if (_value.Equals(NativePointer.Zero))
                return null;

            var _result = new Csp.Multiplayer.SpaceEntity(_value);

            return _result;
        }
```

MAG-SamBirley:
Firstly, this change is awesome. Secondly, it looks like this change may have brought some of your other recent changes along for the ride (it has changes to the commit message template and chevron module removal). Can we get rid of those for this PR?

magnopus-swifty:
> Firstly, this change is awesome. Secondly, it looks like this change may have brought some of your other recent changes along for the ride (it has changes to the commit message template and chevron module removal). Can we get rid of those for this PR?

Those will get diffed out once those changes are merged in. Unfortunately there isn't a nice way to build on top of other in-progress work with Github.

mag-richardzirbes:
Do we need to put any methods that could return a null value inside a try-catch block now, with the usage of the helper CheckNativePointer(pointer)? (I only sort-of understand the mustache syntax)

magnopus-swifty:
> Do we need to put any methods that could return a null value inside a try-catch block now, with the usage of the helper CheckNativePointer(pointer)? (I only sort-of understand the mustache syntax)

No - if `null` is a valid return (e.g. in the case of `GetParentEntity()`) then the C# wrapper will also just return null (rather than an invalid object wrapping a null).

If a method has a return type that's a reference, e.g. `const SpaceTransform& GetTransform() const` then the `CheckNativePointer` will _ensure_ that a non-null pointer is returned (it shouldn't ever do so) and will _always_ return a valid object.

magnopus-swifty:
Methods returning a reference will check the reference for non-nullness (only in Debug builds):

e.g. `SpaceEntity.GetTransform()` calling `const SpaceTransform& GetTransform() const`:

```csharp
        public Csp.Multiplayer.SpaceTransform GetTransform()
        {
            var _value = csp_multiplayer_SpaceEntity_GetTransformC_SpaceTransformRC(_ptr);

            WrapperHelper.CheckNativePointer(_value);

            var _result = new Csp.Multiplayer.SpaceTransform(_value);

            return _result;
        }
```

Methods where null is a valid return will also return `null` on the C# side, e.g. `GetParentEntity()`.

MAG-SamBirley (approved):
Change looks good!

MAG-mv (approved):

MAG-mv (approved):

PR #526, opened 2024-10-24
[OF-1432] refac: remove sorted hierarchy feature

MAG-ChristopherAtkinson:
The sorted hierarchy feature is being pushed back, so the current work related to this feature should be removed.  The following changes have been made:
 * Removed the use of Sequence Hierarchy
 * Removed the use of Sequence Hierarchy Tests
 * Removed event parameter and deserialization


Library/src/Multiplayer/EventSerialisation.cpp
```diff
@@ -28,38 +28,38 @@ namespace
...
-    		break;
-    	}
-    	default:
-    	{
-            CSP_LOG_ERROR_MSG("SequenceChangedEvent - Detected an unsupported update type.");
-    		break;
-    	}
-    }
-
-    return SequenceUpdateType;
```
MAG-ChristopherAtkinson, line 62: this should be spaces not tab, need to check my vs2022 in case the setting have been missed from setup

MAG-AdamThorn (approved):

MAG-OR (approved):

PR #528, opened 2024-10-29
[OF-1439] feat: Added map generic map support to ReplicatedValue

MAG-mv:
- Map<ReplicatedValue, ReplicatedValue> can now be set as a ReplicatedValue
- Added ReplicatedValue operators to support maps
- Tests exist to validate this
- Maps can now be serialized to signalr value through a custom structure
- Maps can also be deserialized
- Tests exist to validate this
-  Bug fixes for map copy assignments to work correctly with unions

### DIFF
--- a/Library/include/CSP/Common/Map.h
+++ b/Library/include/CSP/Common/Map.h
@@ -114,7 +114,15 @@ public:
 			return *this;
 		}
 
-		Container->~MapType();
+		if (Container)
+		{
+			Container->~MapType();
+		}
+		else
+		{
+			Container = (MapType*) csp::memory::DllAlloc(sizeof(MapType));
+		}
+
 		new (Container) MapType(*Other.Container);
 
 		return *this;
@@ -130,10 +138,29 @@ public:
 			return *this;
 		}
 
-		Container->~MapType();
+		if (Container)
+		{
+			Container->~MapType();
+		}
+		else
+		{
+			Container = (MapType*) csp::memory::DllAlloc(sizeof(MapType));
+		}
+
 		new (Container) MapType(*Other.Container);
 
 		return *this;
+	}
+
+
+	CSP_NO_EXPORT bool operator==(const Map<TKey, TValue>& OtherValue) const
+	{
+		return *Container == *OtherValue.Container;
+	}
+
+	CSP_NO_EXPORT bool operator!=(const Map<TKey, TValue>& OtherValue) const
+	{
+		return *Container != *OtherValue.Container;
 	}
 
 	/// @brief Returns the number of elements in this map.
--- a/Library/src/Multiplayer/ReplicatedValue.cpp
+++ b/Library/src/Multiplayer/ReplicatedValue.cpp
@@ -35,6 +35,10 @@ ReplicatedValue::~ReplicatedValue()
 	{
 		Value.String.~String();
 	}
+	else if (ReplicatedType == ReplicatedValueType::Map)
+	{
+		Value.Map.~Map();
+	}
 }
 
 ReplicatedValue::ReplicatedValue(bool InValue) : ReplicatedType(ReplicatedValueType::Boolean)
@@ -75,6 +79,11 @@ ReplicatedValue::ReplicatedValue(const csp::common::Vector3& InValue) : Replicat
 ReplicatedValue::ReplicatedValue(const csp::common::Vector4& InValue) : ReplicatedType(ReplicatedValueType::Vector4)
 {
 	Value.Vector4 = InValue;
+}
+
+ReplicatedValue::ReplicatedValue(const csp::common::Map<ReplicatedValue, ReplicatedValue>& InMapValue) : ReplicatedType(ReplicatedValueType::Map)
+{
+	Value.Map = InMapValue;
 }
 
 ReplicatedValue::ReplicatedValue(const ReplicatedValue& OtherValue)
@@ -119,6 +128,11 @@ ReplicatedValue& ReplicatedValue::operator=(const ReplicatedValue& InValue)
 		case ReplicatedValueType::Vector4:
 		{
 			SetVector4(InValue.GetVector4());
+			break;
+		}
+		case ReplicatedValueType::Map:
+		{
+			SetMap(InValue.GetMap());
 			break;
 		}
 		case ReplicatedValueType::InvalidType:
@@ -178,6 +192,10 @@ bool ReplicatedValue::operator==(const ReplicatedValue& OtherValue) const
 				IsEqual = GetVector4() == OtherValue.GetVector4();
 				break;
 			}
+			case ReplicatedValueType::Map:
+			{
+				IsEqual = GetMap() == OtherValue.GetMap();
+			}
 			default:
 			{
 				assert(0 && "Unhandled replicated value type!");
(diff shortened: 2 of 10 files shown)


MAG-SamBirley (commented):
Worth testing that if you serialize other values/value types to a component within the same test where we're serializing a map that those also come back as expected? i.e. - validating that maps don't stomp over any other values.

MAG-mv:
@MAG-SamBirley Thanks, I will test that

Library/src/Multiplayer/ReplicatedValue.cpp
```diff
@@ -192,6 +212,66 @@ bool ReplicatedValue::operator!=(const ReplicatedValue& OtherValue) const
...
+	return false;
+}
+
+bool ReplicatedValue::operator>(const ReplicatedValue& OtherValue) const
+{
+	switch (OtherValue.GetReplicatedValueType())
+	{
+		case ReplicatedValueType::Boolean:
+		{
+			return GetBool() > OtherValue.GetBool();
```
MAG-OR, line 251: I'm genuinely curious if there are cases where checking if bools or greater or lower than each other is useful? can't we do everything we need with equals and  not equals?

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
```diff
@@ -27,6 +27,176 @@ namespace csp::multiplayer
...
+				   std::vector {signalr::value(VValue.X), signalr::value(VValue.Y), signalr::value(VValue.Z), signalr::value(VValue.W)});
+			break;
+		}
+		case ReplicatedValueType::Map:
+		{
+			ValueType = ItemComponentData::GENERIC_MAP;
+
+			// Creates a custom signalr data structure to hold an array of key-value pairs
+			const auto& Map = Value.GetMap();
+			auto Keys		= Map.Keys();
```
MAG-OR, line 83: (very minor) your spaces are misaligned

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
```diff
@@ -27,6 +27,176 @@ namespace csp::multiplayer
...
+
+			// Root of map which contains the type information
+			// and an array of key-value pairs
+			std::vector<signalr::value> MapRoot;
+			MapRoot.reserve(3);
```
MAG-OR, line 88: why 3? (it might be obvious, please forgive me for asking)

MAG-mv: 0 = Type information of the key
1 = Type information of value
2 = vector of key-value pairs

I will update to make this more clear

Library/src/Multiplayer/ReplicatedValue.cpp
MAG-mv: This is needed to support keys in C++ maps, as they are internally ordered using these operators. However, there won't be many cases where bools will be used as keys.

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-mv: This is Github formatting. It looks correct on my end
image

MAG-OR (approved):

Library/include/CSP/Common/Map.h
```diff
@@ -114,7 +114,15 @@ template <typename TKey, typename TValue> class CSP_API Map
...
 		}
 
-		Container->~MapType();
+		if (Container)
+		{
+			Container->~MapType();
+		}
+		else
+		{
+			Container = (MapType*) csp::memory::DllAlloc(sizeof(MapType));
```
MAG-SamBirley, line 123: When could `Container == nullptr` at the time where the `=` operator is invoked? A cursory glace at the constructors suggest it should always have been allocated at this point?

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
```diff
@@ -27,6 +27,178 @@ namespace csp::multiplayer
...
+				if (First)
+				{
+					First	  = false;
+					KeyType	  = KeyValue.first;
+					ValueType = Value.first;
+				}
+				else
+				{
+					// Ensure key and values are all the same type
+					assert(KeyType == KeyValue.first && "Key types dont match!");
```
MAG-SamBirley, line 116: No strong preference, more curious - why assert instead of logging an error?

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
```diff
@@ -27,6 +27,178 @@ namespace csp::multiplayer
...
+}
+
+ReplicatedValue SignalRValueToReplicatedValue(ItemComponentData Type, const signalr::value& Value)
+{
+	switch (Type)
+	{
+		case ItemComponentData::BOOL:
+			return Value.as_bool();
+		case ItemComponentData::INT64:
+			return (Value.is_integer()) ? ReplicatedValue(Value.as_integer()) : ReplicatedValue((int64_t) Value.as_uinteger());
```
MAG-SamBirley, line 147: c-style casting from a uint to an int may lead to value overflow/undefined behavior - worth doing something slightly more verbose/elaborate to avoid that?

MAG-SamBirley (changes requested):
Amazing work. Adding support for replicated maps is huge! No major concerns, just a few nitpicky things I noticed.

Library/include/CSP/Common/Map.h
MAG-mv: It was crashing for union assignment

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-mv: I was following the convention of this file, which always asserts (except 1 instance)

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-mv: This is code moved from another function, I didn't write it in this PR.

MAG-mv: Happy to update if you think it'll be an issue. Are you imagining some kind of bounds checking?

Library/include/CSP/Common/Map.h
MAG-mv: I assume because the constructor hasnt been called and it was assigning to empty memory

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-SamBirley: Do you think it's the right one to follow here? (Not a leading question)

Library/include/CSP/Common/Map.h
MAG-SamBirley: Ah, I see.

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-SamBirley: No, I wouldn't want to slow you down if this isn't a new issue. Mind logging an issue for it though?

Library/src/Multiplayer/SignalRMsgPackEntitySerialiser.cpp
MAG-mv: Thinking about it more I think it makes sense to error, if clients could setup the map incorrectly,  I think they should be notified through an error.

MAG-mv: Updated!

MAG-OR (approved):
I have looked at the changes made since my last review and they look fine to me

MAG-AdamThorn (approved):

MAG-SamBirley (approved):

PR #529, opened 2024-10-30
[OF-1455] feat: Set tags on space metadata

MAG-DanielMurdolo:
- Adding support for setting and updating tags on the space metadata prototype
- This is so we can use it to identify types of spaces using the first class property from the cloud services endpoint


Tests/src/PublicAPITests/SequenceHierarchyTests.cpp
```diff
@@ -197,7 +197,15 @@ CSP_PUBLIC_TEST(CSPEngine, SequenceHierarchyTests, CreateSequenceHierarchyTest)
 	SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
 
 	csp::systems::Space Space;
-	CreateSpace(SpaceSystem, UniqueSpaceName, TestSpaceDescription, csp::systems::SpaceAttributes::Private, nullptr, nullptr, nullptr, Space);
+	CreateSpace(SpaceSystem,
```
MAG-AdamThorn, line 200: Formatting with these methods also odd.

MAG-AdamThorn (dismissed):

MAG-mv (approved):

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #530, opened 2024-11-01
[OF-1441] feat: Video Material Classes

MAG-mv:
- Added GLTFMaterial class
- Added TextureInfo class
- Json serialization
- Tests


Library/src/Systems/Assets/Material.cpp
```diff
@@ -0,0 +1,280 @@
...
+	Deserializer.DeserializeMember("metallicRoughnessTexture", Obj.MetallicRoughnessTexture);
+	Deserializer.DeserializeMember("normalTexture", Obj.NormalTexture);
+	Deserializer.DeserializeMember("occlusionTexture", Obj.OcclusionTexture);
+	Deserializer.DeserializeMember("emissiveTexture", Obj.EmissiveTexture);
+}
+
+namespace csp::systems
+{
+
+GLTFMaterial::GLTFMaterial(const csp::common::String& InAssetCollectionId, const csp::common::String& InAssetId)
```
MAG-AdrianMeredith, line 119: I think this should either inherit from a base class or have a common interfact for the base "Material". I'd argue gltf material should be in its own file.
This is because future shader types may share almost 0 properties with gltf shader (other than name, type etc)

MAG-AdrianMeredith (changes requested):
Looks like a good start! just a few small comments

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -0,0 +1,195 @@
...
+	void SetAlphaCutoff(float Mode);
+	/// TODO: comment
+	float GetAlphaCutoff() const;
+
+	/// TODO: comment
+	void SetDoubleSided(bool DoubleSided);
+	/// TODO: comment
+	bool GetDoubleSided() const;
+
+	/// @brief Sets the factor of the base color texture
```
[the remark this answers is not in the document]
MAG-mv: Updated to use GLTF spec comments

Library/src/Systems/Assets/Material.cpp
MAG-mv: I assume Material base should have:

- Name
- Type 
- AssetCollectionId
- AssetId

MAG-AdrianMeredith (dismissed):

MAG-AdamThorn (changes requested):

Library/include/CSP/Systems/Assets/GLTFMaterial.h
```diff
@@ -0,0 +1,205 @@
...
+enum class EAlphaMode
+{
+	Opaque,
+	Mask,
+	Blend
+};
+
+/// @ingroup Asset System
+/// @brief Data class which represents a GLTF material.
+class CSP_API GLTFMaterial : public Material
```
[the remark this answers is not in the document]
MAG-AdrianMeredith: Could be anything, these would be separate shaders. for example a fog card material that only appears when looking straight at it and smooth fades when close, or a spinny portal effect. 

Essentially because we cannot have runtime shaders, we will have to create a library of built in ones.  I'm keen to ensure adding these later is painless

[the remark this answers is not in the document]
MAG-AdamThorn: In that case I wonder if 'GLTFMaterial' the right name for the default standard material?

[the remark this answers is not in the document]
MAG-AdrianMeredith: This name is intentional so thats its clear to implementors that they should be using their existing gltf shaders to implement it.  Any questions on how to render this shader can then defer to the gltf material spec rather than some proprietary one

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -142,6 +150,22 @@ std::shared_ptr<chs::PrototypeDto> CreatePrototypeDto(const Optional<String>& Sp
 	return PrototypeInfo;
 }
 
+
+csp::common::String CreateUniqueMaterialAssetCollectionName(const csp::common::String& Name, const csp::common::String& SpaceId)
+{
+	return MATERIAL_ASSET_COLLECTION_NAME_PREFIX + SpaceId + "_" + Name;
```
MAG-AdamThorn, line 156: Are we confident that these names will actually be unique? For example if I passed in an asset name of car_texture. It feels like in a car themed space that might be quite a common choice. Do we do anything to guard against name collisions our end? I will probably find out the answer to that last question further down.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -847,4 +881,305 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
...
+		// 2. Delete asset collection
+		auto DeleteAssetCollectionCB = [this, Callback](const NullResult& DeleteAssetCollectionResult)
+		{
+			Callback(DeleteAssetCollectionResult);
+		};
+
+		DeleteAssetCollectionById(Material.GetAssetCollectionId(), DeleteAssetCollectionCB);
+	};
+
+	DeleteAssetById(Material.GetAssetCollectionId(), Material.GetAssetId(), DeleteAssetCB);
```
MAG-AdamThorn: If the associated AssetCollection only had this one material asset does that get cleaned up by CHS? If not do we need to clear it up?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -847,4 +881,305 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
...
+			std::future<void> CallbackFuture = CallbackPromise.get_future();
+
+			// 3. Download asset data for each material asset
+			auto DownloadMaterialCallback
+				= [this, Callback, &AssetsToDownload, &AssetsDownloaded, &SpaceMaterials, &CallbackPromise](const AssetDataResult& DownloadResult)
+			{
+				if (DownloadResult.GetResultCode() != EResultCode::Success)
+				{
+					Callback(GLTFMaterialsResult(DownloadResult.GetResultCode(), DownloadResult.GetHttpResultCode()));
+					return;
```
MAG-AdamThorn, line 1073: Do we not need to increment `AssetsDownloaded` even when download unsuccessful?

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+	// Log in
+	csp::common::String UserId;
+	LogIn(UserSystem, UserId);
+
+	// Create space to search for materials
+	::Space Space;
+	CreateDefaultTestSpace(SpaceSystem, Space);
+
+	// Create 3 materials
```
MAG-AdamThorn, line 345: Typo: YOu are creating 2 not 3.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Each material consists of one asset collection and one material

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: It would only be a space specific I believe. So you couldn't have 2 textures of the same name in the same space

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Updated to a better system to handle failures better

Tests/src/PublicAPITests/ComponentTests/AnimatedModelComponentTests.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+	const char* TestSpaceName		 = "CSP-UNITTEST-SPACE-MAG";
+	const char* TestSpaceDescription = "CSP-UNITTEST-SPACEDESC-MAG";
+
+	char UniqueSpaceName[256];
+	SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+	// Log in
+	csp::common::String UserId;
+	LogIn(UserSystem, UserId);
```
MAG-OR, line 66: The tests will need to be updated following the test user account generation story. Use `LogInAsNewTestUser(UserSystem, UserId)` whenever you do not need to know or reuse the user account details. If you do, create a user (CreateTestUser) and then log in with the old LogIn function (which no longer accepts defaults). Happy to help if needed

MAG-OR (changes requested):

Tests/src/PublicAPITests/ComponentTests/AnimatedModelComponentTests.cpp
MAG-mv: Thanks. I will also have to change my multiplayer event to use the event bus

MAG-AdamThorn (approved):

Library/docs/source/learn/learn.md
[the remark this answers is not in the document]
MAG-ElliotMorris: These are changes from Sams work, I think you want to rebase your branch atop latest to remove them from the changeset, because they shouldn't be muddying this PR.

MAG-ElliotMorris (approved):
Reviewed the last few commits. (Up to where I think I stopped reviewing on the feature branch.)

This is a tricky review experience, I hope we can make this process cleaner. Feature branches add annoyance for sure, but even in those constraints I'm hoping there may be ways to be clearer with the storytelling in PR's.

MAG-OR (approved):
PR has been updated

PR #532, opened 2024-11-04
[nt-0] fix: move dispose to NativeClassWrapper

magnopus-swifty:
Ensure that the destructor is called on wrapped classes during finalization if
* they own the pointer
* CSP is still initialized

### DIFF
--- a/Library/CSharpWrapper/src/NativeClassWrapper.cs
+++ b/Library/CSharpWrapper/src/NativeClassWrapper.cs
@@ -21,7 +21,7 @@ namespace Csp
         public static NativePointer Zero => new NativePointer(IntPtr.Zero, 0);
     }
 
-    public class NativeClassWrapper
+    public class NativeClassWrapper : IDisposable
     {
         internal IntPtr _ptr;
         internal bool _ownsPtr;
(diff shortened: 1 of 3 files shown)


Library/CSharpWrapper/src/NativeClassWrapper.cs
```diff
@@ -38,5 +38,52 @@ internal NativeClassWrapper(NativePointer ptr)
...
+                        {
+                            //TODO: Log that the object wasn't disposed and is being destroyed by finalizer
+                        }
+
+                        if (_ptr == IntPtr.Zero)
+                        {
+                            throw new InvalidOperationException("Internal pointer was null");
+                        }
+
+                        if (CSPFoundation.GetIsInitialised())
```
magnopus-swifty, line 62: This line makes Unity tests fail since `CSPFoundation` is a generated class that doesn't exist in those tests. There might be a more elegant way to handle this.

Library/CSharpWrapper/src/NativeClassWrapper.cs
```diff
@@ -21,7 +21,7 @@ internal NativePointer(IntPtr pointer, byte ownsOwnData)
         public static NativePointer Zero => new NativePointer(IntPtr.Zero, 0);
     }
 
-    public class NativeClassWrapper
+    public class NativeClassWrapper : IDisposable
```
magnopus-swifty, line 24: The drawback here is that it makes all NativeClassWrapper-derived classes also IDisposable, which is incorrect for e.g. the *System singletons.

MAG-ElliotMorris:
@magnopus-swifty Is this PR still active? Would rather not have long-term draft PR's sitting around indefinitely against the main repo.

MAG-ElliotMorris:
@magnopus-swifty If there is no more activity or pleas to keep it open on this PR by tommorow, it will be closed.

PR #533, opened 2024-11-05
[OF-1437] refac: Introduce a new Event Bus for SignalR Multiplayer events

MAG-OR:
The current event handling in MultiplayerConnection is being replaced by a more intuitive and extensible EventBus implementation. 
A new EventBus class is introduced, which manages the event handling in place of MultiplayerConnection.
The SystemBase class is extended to handle default event registration, deregistration and default deserialisation. Each system can override these functions as needed.
Clients are still able to register using callbacks, as before, the only difference being that the EventBus is now handling it.


Library/src/Systems/SystemBase.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void SystemBase::RegisterSystemCallback()
+{
+	if (!SystemCallback)
+	{
+		return;
+	}
+
+	if (EventBusPtr)
+		EventBusPtr->ListenEvent("DefaultEvent", this);
```
MAG-OR, line 52: It  may make more sense to make RegisterSystemCallback and DeregisterSystemCallback purely virtual because this DefaultEvent does nothing. However, this would mean that we have to implement events for all the systems that inherit from this class and are currently not using the event bus (about 10).
It may be more elegant to just have this function do nothing at all, as opposed to registering a default event.

Library/src/Systems/ECommerce/ECommerceSystem.cpp
```diff
@@ -99,14 +99,14 @@ void ECommerceSystem::GetProductInfoCollectionByVariantIds(const common::String&
...
+	if (CharacterCheck.find("/") != std::string::npos || CharacterCheck.find(R"(\)") != std::string::npos)
+	{
 		CSP_LOG_ERROR_MSG("Call to GetCheckoutInformation failed. CartId must NOT include path characters forward or back slash.");
 
 		INVOKE_IF_NOT_NULL(Callback, MakeInvalid<CheckoutInfoResult>());
 
-        return;
-    }
+		return;
+	}
```
MAG-OR, line 109: not good, will fix.

Library/src/Systems/ECommerce/ECommerceSystem.cpp
```diff
@@ -56,11 +56,11 @@ void RemoveUrl(csp::common::String& Url)
 namespace csp::systems
 {
 
-ECommerceSystem::ECommerceSystem() : SystemBase(), ShopifyAPI(nullptr)
+ECommerceSystem::ECommerceSystem() : SystemBase(nullptr, nullptr), ShopifyAPI(nullptr)
 {
 }
 
-ECommerceSystem::ECommerceSystem(csp::web::WebClient* InWebClient) : SystemBase(InWebClient)
+ECommerceSystem::ECommerceSystem(csp::web::WebClient* InWebClient) : SystemBase(InWebClient, nullptr)
```
MAG-OR, line 63: I am aware that @MAG-ChristopherAtkinson isn't keen on setting the EventBus pointer to nullptr, however we would have to modify the 10 classes that currently derive from SystemBase and are not currently using the EventBus. 
I am not quite sure what the ramifications of such a change would be -- I am happy to do it if people think it is worth doing.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+namespace csp::multiplayer
+{
+
+extern ErrorCode ParseError(std::exception_ptr Exception);
+
+constexpr const uint64_t ALL_CLIENTS_ID = -1;
+
+EventBus::~EventBus()
+{
+	CSP_DELETE(MultiplayerConnectionInst);
```
MAG-mv, line 34: I dont think we should be deleting this here

Library/include/CSP/Multiplayer/EventBus.h
```diff
@@ -0,0 +1,104 @@
...
+
+	// The callback used to register to listen to network events.
+	typedef std::function<void(bool, const csp::common::Array<ReplicatedValue>&)> ParameterisedCallbackHandler;
+
+	/// @brief Registers a callback to listen for the named event.
+	/// @param EventName csp::common::String : The identifying name for the event to listen for.
+	/// @param Callback ParameterisedCallbackHandler : A callback to register for the event which contains the parameter payload data.
+	void ListenEvent(const csp::common::String& EventName, ParameterisedCallbackHandler Callback);
+
+	/// @brief Registers a system to listen for the named event.
```
MAG-AdamThorn, line 70: I think it would be worth expanding the comment brief to explain why Systems are being handled uniquely.

Library/include/CSP/Multiplayer/EventBus.h
```diff
@@ -0,0 +1,104 @@
...
+	void StopListenEvent(const csp::common::String& EventName);
+
+	/// @brief Instructs the event bus to start listening to messages
+	void StartEventMessageListening();
+
+private:
+	EventBus();
+	~EventBus();
+
+	EventBus(const EventBus& InBoundConnection);
```
MAG-AdamThorn, line 92: This may become clearer later, but why is there a ctor that takes in a const ref to another EventBus instance?

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+	MultiplayerConnectionInst = InMultiplayerConnection;
+	EventMap				  = {};
+}
+
+void EventBus::ListenEvent(const csp::common::String& EventName, ParameterisedCallbackHandler Callback, csp::systems::SystemBase* System)
+{
+	if (MultiplayerConnectionInst->Connection == nullptr || !MultiplayerConnectionInst->Connected)
+	{
+		return;
```
MAG-AdamThorn, line 53: Add logging here and parenthesis and logging for the check below.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -251,9 +252,24 @@ class CSP_API UserSystem : public SystemBase
...
 
+	// Callback to receive access permission changes Data when a message is sent.
+	typedef std::function<void(const csp::multiplayer::UserPermissionsParams&)> UserPermissionsChangedCallbackHandler;
+
+	/// @brief Sets a callback for an access control changed event.
+	/// @param Callback UserPermissionsChangedCallbackHandler: Callback to receive data for the user permissions that has been changed.
+	CSP_EVENT void SetUserPermissionsChangedCallback(UserPermissionsChangedCallbackHandler Callback);
+
+	/// @brief Registers the system to listen for the named event.
+	void RegisterSystemCallback();
```
MAG-mv, line 263: should be marked override

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+		{
+			return;
+		}
+
+		std::vector<signalr::value> EventValues = Result.as_array()[0].as_array();
+		const csp::common::String EventType(EventValues[0].as_string().c_str());
+
+		if (EventMap.find(EventType) == EventMap.end())
+		{
+			return;
```
MAG-AdamThorn, line 106: Consider adding logging here to aid with debugging.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		if (EventMap[EventType].second)
+		{
+			EventMap[EventType].second->Deserialise(EventValues);
+		}
+		else
```
MAG-AdamThorn, line 113: This means that if a system has been registered any registered callbacks will be ignored, is that correct? If that is by design then I would not have a ListenEvent method that takes both a callback and an event.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void EventBus::ListenEvent(const csp::common::String& EventName, ParameterisedCallbackHandler Callback, csp::systems::SystemBase* System)
+{
+	if (MultiplayerConnectionInst->Connection == nullptr || !MultiplayerConnectionInst->Connected)
+	{
+		return;
+	}
+
+	if (!Callback && !System)
```
MAG-mv, line 56: Tiny thing but can we add brackets here for consistency?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -148,22 +146,22 @@ MultiplayerConnection::~MultiplayerConnection()
 		CSP_DELETE(WebSocketClient);
 		CSP_DELETE(NetworkEventManager);
 		CSP_DELETE(ConversationSystemPtr);
+		EventBusPtr = nullptr;
```
MAG-AdamThorn, line 149: We need to call CSP_DELETE on the EventBus

Library/include/CSP/Systems/SystemBase.h
```diff
@@ -25,17 +26,52 @@ class WebClient;
...
+
+	/// @brief Registers the system to listen for the default event.
+	virtual void RegisterSystemCallback();
+
+	/// @brief Deregisters the system from listening for the default event.
+	virtual void DeregisterSystemCallback();
+
+	/// @brief Deserialises the event values of the system.
+	/// @param EventValues std::vector<signalr::value> : event values to deserialise
+	CSP_NO_EXPORT virtual void Deserialise(const std::vector<signalr::value>& EventValues);
```
MAG-mv, line 65: Would this make more sense to be called "OnEvent", as we dont know what the system is going to do wth the event

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: This should be done with all of them

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void EventBus::ListenEvent(const csp::common::String& EventName, ParameterisedCallbackHandler Callback)
+{
+	ListenEvent(EventName, Callback, nullptr);
+}
+
+void EventBus::StopListenEvent(const csp::common::String& EventName)
+{
+	if (!EventMap.empty() && EventMap.find(EventName) != EventMap.end())
```
MAG-AdamThorn, line 81: The issue here is that it doesn't care if multiple callbacks or a system have been registered. So if multiple callbacks have been registered and a registered system calls this, it will mean the registered callbacks are also removed.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -847,4 +851,36 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
...
+	{
+		return;
+	}
+
+	EventBusPtr->ListenEvent("AssetDetailBlobChanged", this);
+}
+
+void AssetSystem::DeregisterSystemCallback()
+{
+	if (EventBusPtr)
```
MAG-AdamThorn, line 872: Add parenthesis here, and in the method below - it's one of our conventions.

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
```diff
@@ -318,9 +327,52 @@ void HotspotSequenceSystem::RemoveItemFromGroups(const csp::common::String& Item
...
+		return;
+	}
+
+	EventBusPtr->ListenEvent("SequenceChanged", this);
+}
+
+void HotspotSequenceSystem::DeregisterSystemCallback()
+{
+	if (EventBusPtr)
+		EventBusPtr->StopListenEvent("SequenceChanged");
```
MAG-AdamThorn, line 355: Add parenthesis.

Library/src/Systems/SystemBase.cpp
MAG-AdamThorn, line 52: Add parenthesis here and below.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -792,4 +800,38 @@ void UserSystem::ResetAuthenticationState()
...
+		return;
+	}
+
+	EventBusPtr->ListenEvent("AccessControlChanged", this);
+}
+
+void UserSystem::DeregisterSystemCallback()
+{
+	if (EventBusPtr)
+		EventBusPtr->StopListenEvent("AccessControlChanged");
```
MAG-AdamThorn, line 822: Add parenthesis

Library/include/CSP/Systems/SystemBase.h
MAG-OR: Could do, I don't have a preference, @MAG-AdamThorn do you?

Library/src/Multiplayer/EventBus.cpp
MAG-OR: that is a very good point, I did not realise this would be a use case. what is the best solution, passing the callback or system?

Library/src/Multiplayer/EventBus.cpp
MAG-OR: that is a good point. I was not sure if this would ever be a use-case, so I kind of left it half-baked. apologies! so is there a use case where the same event would be listened to both by a System and by one or more ParameterisedCallbackHandlers?
If not, would it make sense to decouple the two completely and have 2 separate event maps? 
If yes, I need to remove the else here and see if other changes are required

Library/src/Multiplayer/EventBus.cpp
MAG-OR: done

Library/src/Multiplayer/EventBus.cpp
MAG-OR: done

Library/src/Systems/Assets/AssetSystem.cpp
MAG-OR: done

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
MAG-OR: done

Library/src/Systems/SystemBase.cpp
MAG-OR: done

Library/src/Systems/Users/UserSystem.cpp
MAG-OR: done

Library/src/Multiplayer/EventBus.cpp
MAG-OR: done

Library/include/CSP/Systems/Users/UserSystem.h
MAG-OR: done

Library/include/CSP/Multiplayer/EventBus.h
MAG-OR: done, please let me know if the new description is satisfactory

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-OR: done

Library/src/Multiplayer/EventBus.cpp
MAG-AdamThorn: It would also be an issue if just multiple callbacks were registered and no system. If this method were called for one callback then all callbacks would be removed.

I think this could be a valid method - to stope everything listening for an event, but would likely need to rename/make it very clear what it did.

If we are supporting registering multiple callbacks (or a callback/system combo) with a single event, then we will need a method for unregistering them individually.

Library/src/Multiplayer/EventBus.cpp
MAG-AdamThorn: Hmmm. Thihking about it the whole point of being able to register systems is because they have unique deserialisation requirements. In which case there would be no point sending those events to a system to be deserialised and to then apply the default deserialiser for any registered callbacks.

So I think it should be that you either register one or many callbacks for an event (which then use the default deserialiser), or you register a single system (which then uses its unique deserialiser). You don't have both.

If we are doing that then we should ensure that people do not register a system and calback/s for the same event.

What do you think @MAG-mv

Library/src/Systems/ECommerce/ECommerceSystem.cpp
MAG-OR: I tried to fix this several times, it just stays like this

Library/src/Multiplayer/EventBus.cpp
MAG-mv: Yep. That sounds good

Library/include/CSP/Multiplayer/EventBus.h
```diff
@@ -0,0 +1,110 @@
...
+
+	/// @brief Stops the event bus from listening for a particular event, for any system or callbacks
+	/// @brief that were registered.
+	/// @param EventName csp::common::String : The identifying name for the event to stop listening for.
+	void StopAllListenEvent(const csp::common::String& EventName);
+
+	/// @brief Stops the event bus from listening for a particular event, for the specified system.
+	/// @param EventName csp::common::String : The identifying name for the event to stop listening for.
+	/// @param System csp::systems::SystemBase : The system that will stop being registered.
+	void StopSystemListenEvent(const csp::common::String& EventName, csp::systems::SystemBase* System);
```
MAG-mv, line 84: The "ListenEvent" both have the same naming convention. Is there a reason the stop listening have different names?

Library/include/CSP/Systems/SystemBase.h
MAG-mv: @MAG-AdamThorn In the context of the event bus sending the event to the system, imo "OnEvent" feels more clear

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -0,0 +1,185 @@
...
+		EventMap.erase(EventName);
+	}
+}
+
+void EventBus::StopSystemListenEvent(const csp::common::String& EventName, csp::systems::SystemBase* System)
+{
+	if (System && !EventMap.empty() && EventMap.find(EventName) != EventMap.end() && EventMap[EventName].second == System)
+	{
+		// Remove system
+		EventMap[EventName].second = nullptr;
```
MAG-mv, line 111: Should we actually remove the element here?

MAG-mv (changes requested):

Library/src/Multiplayer/EventBus.cpp
MAG-OR: yeah I don't think this is necessary. the two lines below are enough. I will remove it

Library/include/CSP/Multiplayer/EventBus.h
MAG-OR: the reasoning was the following: there is a function that deregisters everything, called StopAllListenEvent. This function has a different name following Adam's comment that it should be crystal clear that it deletes everything (including all callbacks if several are defined). From this, it seemed a little odd to to me have 2 overloaded functions, one for a system, one for a callback. you are right that it is not following ListenEvent, and it is best to do the same for both. I will only keep StopAllListenEvent, and rename the other 2 to StopListenEvent

MAG-mv: Sounds good!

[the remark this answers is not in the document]
MAG-OR: yes absolutely, well spotted

Library/include/CSP/Multiplayer/EventBus.h
```diff
@@ -0,0 +1,110 @@
...
+	~EventBus();
+
+	EventBus(MultiplayerConnection* InMultiplayerConnection);
+
+	class MultiplayerConnection* MultiplayerConnectionInst;
+
+	// TODO: Replace these with pointers! We can't use STL containers as class fields due to the fact that the class size will
+	//   change depending on which runtime is used.
+	typedef std::vector<ParameterisedCallbackHandler> Callbacks;
+	typedef std::map<csp::common::String, std::pair<Callbacks, csp::systems::SystemBase*>> EventMaps;
```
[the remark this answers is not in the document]
MAG-mv: I think separating would be more clear and easier to handle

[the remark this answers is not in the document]
MAG-OR: the rationale for the single container was to support registering both system(s) and callback(s) for the same event, but since then we've agreed that this was off the table. I agree that in the case where we only support either/or, 2 containers is probably best

Library/include/CSP/Systems/SystemBase.h
MAG-AdamThorn: Sorry missed this. Yes I think OnEvent makes more sense here.

Library/include/CSP/Multiplayer/EventBus.h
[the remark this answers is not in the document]
MAG-AdamThorn: Yes ok makes sense. Thanks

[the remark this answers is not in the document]
MAG-ChristopherAtkinson: Yeah it that sounds good, and think it would help improve readability and clarity

Library/src/Systems/ECommerce/ECommerceSystem.cpp
MAG-OR: I created a story for this (OF-1464)

Library/include/CSP/Multiplayer/EventBus.h
```diff
@@ -0,0 +1,111 @@
...
+	/** @endcond */
+
+	// The callback used to register to listen to events.
+	typedef std::function<void(bool, const csp::common::Array<ReplicatedValue>&)> ParameterisedCallbackHandler;
+
+	/// @brief Registers a system to listen for the named event, where the system can define its
+	/// @brief own callback and deserialiser.
+	/// @param EventName csp::common::String : The identifying name for the event to listen for.
+	/// @param System csp::systems::SystemBase* : A pointer to the system which wants to register for the event.
+	CSP_NO_EXPORT void ListenEvent(const csp::common::String& EventName, csp::systems::SystemBase* System);
```
MAG-OR, line 70: could you please confirm whether the use of CSP_NO_EXPORT for the 2 system-related ListenEvent and StopListenEvent methods is correct? @MAG-AdamThorn @MAG-mv

MAG-ElliotMorris (dismissed):
Only reviewed the lastest commit removing multiple callbacks. LGTM

Library/include/CSP/Multiplayer/EventBus.h
MAG-OR: confirmed during a call

MAG-AdamThorn (dismissed):

MAG-mv (dismissed):

MAG-ElliotMorris (approved):

MAG-mv (approved):

PR #538, opened 2024-11-13
[OF-1456] feat: Children now reparent after parent deletion

MAG-mv:
- Updated deletion patch to include children parent changes
- Local entity update callback is called when the children reparent
- Moved resolving of hierarchies into ProcessPendingEntityUpdate to fix issues with self messaging tests
- Added tests


MAG-OR (commented):
LGTM however I don't have detailed knowledge of the space entity code

MAG-ChristopherAtkinson (dismissed):
Changes look good to me, glad to see the edge case is handled for pending entities.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -3040,4 +3040,224 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, CreateObjectParentTest)
...
+	auto* Connection	 = SystemsManager.GetMultiplayerConnection();
+	auto* EntitySystem	 = SystemsManager.GetSpaceEntitySystem();
+
+	// Log in
+	csp::common::String UserId;
+	LogIn(UserSystem, UserId);
+
+	// Create space
+	const char* TestSpaceName		 = "CSP-UNITTEST-SPACE-MAG";
+	const char* TestSpaceDescription = "CSP-UNITTEST-SPACEDESC-MAG";
```
MAG-ElliotMorris, line 3062: [Nit] Odd alignment. Could also be `constexpr char*`

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -3040,4 +3040,224 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, CreateObjectParentTest)
...
+
+	char UniqueSpaceName[256];
+	SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
```
MAG-ElliotMorris, line 3065: [Consider] 
Is C++20 on the table? You might prefer this rather than allocating buffers.
`std::string UniqueSpaceName = std::format("{}-{}", TestSpaceName, GetUniqueString());`

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -3040,4 +3040,224 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, CreateObjectParentTest)
...
+
+	// Test setting the parent for the first child
+	{
+		bool ChildEntityUpdated = false;
+
+		CreatedChildEntity1->SetUpdateCallback(
+			[&ChildEntityUpdated,
+			 ChildEntityName1](SpaceEntity* Entity, SpaceEntityUpdateFlags Flags, csp::common::Array<ComponentUpdateInfo>& UpdateInfo)
+			{
+				if (Entity->GetName() == ChildEntityName1 && Flags & SpaceEntityUpdateFlags::UPDATE_FLAGS_PARENT)
```
MAG-ElliotMorris, line 3101: [Nit] Stupid tired people like me might thank you for separating logical and bitwise logic more clearly.
```suggestion
				if (Entity->GetName() == ChildEntityName1 && (Flags & SpaceEntityUpdateFlags::UPDATE_FLAGS_PARENT))
```

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -3040,4 +3040,224 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, CreateObjectParentTest)
...
+
+	// Delete the parent
+	{
+		bool DestroyCalled = false;
+
+		auto DestroyCb = [&DestroyCalled](bool Success)
+		{
+			DestroyCalled = true;
+			EXPECT_TRUE(Success);
+		};
```
MAG-ElliotMorris, line 3152: [Nit] This CB is defined out of line, whilst the update callbacks are all inline? Could be consistent.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -3040,4 +3040,224 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, CreateObjectParentTest)
...
+		EntitySystem->SetInitialEntitiesRetrievedCallback(EntitiesReadyCallback);
+
+		auto [EnterResult2] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id);
+		EXPECT_EQ(EnterResult2.GetResultCode(), csp::systems::EResultCode::Success);
+
+		WaitForCallbackWithUpdate(EntitiesCreated, EntitySystem);
+		EXPECT_TRUE(EntitiesCreated);
+	}
+
+	// Ensure children have been parented to the root
```
MAG-ElliotMorris, line 3226: [Question] Is "parented to the root" the correct terminology, considering that `getParentEntity` is null now? Is this actually a deparent? Maybe :
```suggestion
	// Ensure children have been unparented and are now root entities
```

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1339,6 +1339,20 @@ void SpaceEntity::ResolveParentChildRelationship()
...
+			// so check pending entities
+			for (const auto& PendingParent : *EntitySystem->PendingAdds)
+			{
+				if (PendingParent->Id == *ParentId)
+				{
+					Parent = PendingParent;
+					break;
+				}
+			}
+		}
```
MAG-ElliotMorris, line 1354: [Consider]
Could `FindSpaceEntityById` be a better place to put this pending check behaviour? It strikes me that this might be the sensible default behaviour?

Perhaps an interface like
`SpaceEntity* SpaceEntitySystem::FindSpaceEntityById(uint64_t EntityId, bool includePending = true);`

MAG-ElliotMorris (approved):
Just minor nits, nothing blocking. Nice thorough tests.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-mv: I think we're locked to 17 atm

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-mv: This is better!

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: I believe this will cause a breaking change for clients as I dont think c++ optionals are supported by the wrapper gen

MAG-mv: I want to address the issue with handling entities when we start planning the multiplayer refactor

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-mv: This is clang format doing this. It's worth reviewing our usage at some point

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-mv: Alignment looks correct on my end. I'm not sure why its being shown like this

MAG-ChristopherAtkinson: I have had issue previously where github is not correctly rendering the file but is fine in checkout, like same thing here

MAG-OR (commented):
looks fine to me but I'll let "grown-ups" do the approval

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-ElliotMorris: It'll be because github is set to render tabs as 8 spaces by default, your editors probably have it at 4, which aligns correctly.

Imo we should have it so tabs are automatically converted to spaces in our linters, but it's not an issue for right now 🤷

MAG-AdamThorn (approved):

MAG-ChristopherAtkinson (approved):
Changed looks good to me since last comment, happy for this to go to approval.

PR #541, opened 2024-11-20
[OB-3674] fix: Do not encode already encoded URLs

MAG-ElliotMorris:
CSP made a change in March to encode URLs in asset details. We added this encode earlier than that to workaround the issue for a client team, but now it's causing double encode errors, so we should remove it.

Added a couple of tests so we'd know if this changes again.

(Magnopus slack confirming this change : https://magnopus.slack.com/archives/C02T07E8P2B/p1732097572110679)

### DIFF
--- a/Tests/assets/TestWith Space.json
+++ b/Tests/assets/TestWith Space.json
@@ -1,0 +1,4 @@
+{
+  "is_test_data": "true",
+  "meaning_of_life": 42
+}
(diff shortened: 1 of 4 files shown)


Tests/assets/TestWith Space.json
```diff
@@ -0,0 +1,4 @@
+{
+  "is_test_data": "true",
+  "meaning_of_life": 42
```
MAG-OR, line 3: [Comment] 🦭 (seal of approval)

MAG-ChristopherAtkinson (approved):
Ah that make sense, thank you for confirm this has been moved to the asset details,. And glad to see the introduction for testing in this area should help to identify any changes in future.

PR #545, opened 2024-11-22
[OF-1479] R&D: Experimentally remove custom allocators

MAG-ElliotMorris:
This is an experimental branch to research the impacts of moving away from custom allocators and back to standard memory management.

Not for formal review and definitely not for merge.


MAG-ElliotMorris:
Looks like I missed some `CSP_NEWs` in the WASM templates. Not going to fix, this is just an experiment 🤷

MAG-ElliotMorris:
Closing this as it was just an experiment. Annoyingly I had hoped the tests would stay around in this view for reference? They were proof that this had worked :P

PR #547, opened 2024-11-25
[OF-1440] feat: Material AssetSystem Interface

MAG-mv:
- Added asset system interface for materials
- Added tests 
- Refactored GLTFMaterial
- Added new test helper function for creating spaces


Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -251,11 +252,43 @@ class CSP_API AssetSystem : public SystemBase
 	CSP_ASYNC_RESULT_WITH_PROGRESS void
 		RegisterAssetToLODChain(const AssetCollection& AssetCollection, const Asset& Asset, int LODLevel, AssetResultCallback Callback);
 
+	/// @brief Creates a new material backed by an AssetCollection/Asset.
+	/// @param Name const csp::common::String& : The name of the new material.
+	/// @param SpaceId const csp::common::String& : The space id this material is associated with.
+	/// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
+	CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId, GLTFMaterialResultCallback Callback);
+
+	/// @brief Updates an existing materials properties.
```
MAG-OR, line 261: [Question] I tried to google it but didn't find anything - is it customary not to use apostrophes in documentation strings? (i.e. "an existing material's properties"). I have seen this several times around the code now

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -847,4 +866,303 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
...
+	// 1. Delete asset
+	auto DeleteAssetCB = [this, Callback, &Material](const NullResult& DeleteAssetResult)
+	{
+		if (DeleteAssetResult.GetResultCode() != EResultCode::Success)
+		{
+			Callback(NullResult(DeleteAssetResult.GetResultCode(), DeleteAssetResult.GetHttpResultCode()));
+			return;
+		}
+
+		// 2. Deletion asset collection
```
MAG-OR, line 990: [Nit] all your other code comments start with a verb, so this one would be "Delete asset collection".

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -0,0 +1,330 @@
...
+								  {
+									  return Name == SearchName;
+								  });
+
+
+		EXPECT_TRUE(Found != MaterialNames.end());
+	}
+
+	// Cleanup
+	DeleteMaterial(AssetSystem, CreatedMaterial1);
```
MAG-OR, line 251: [Question] would it add any value to the test to first delete one, say number 2, and check that the other two (1 and 3) are still there, and only then remove 1 and 3?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -849,22 +861,282 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
...
+				FinalResult.SetGLTFMaterial(NewMaterial);
+
+				Callback(FinalResult);
+			};
+
+			const Asset& CreatedAsset = CreateAssetResult.GetAsset();
+
+			BufferAssetDataSource AssetData;
+			AssetData.SetMimeType("application/json");
+			AssetData.Buffer	   = const_cast<char*>(MaterialJson.c_str());
```
MAG-ElliotMorris, line 918: [fix] Hesitant to sign off on const_cast, is this the standard pattern? 

This seems vulnerable to refactoring errors, as this only works coincidentally so long as `MaterialJson` remains alive in the same scope, and code-changes to `BufferAssetDataSource` may corrupt `String` when it comes to destruct.
A `move` is the correct semantic operation here, but given our types arn't there yet, i'd advise doing an explicit copy, so `Buffer` is the sole owning pointer.

Alternately, you could alter `BufferAssetDataSource` to be constructable from a `String&&`, but I suspect that's a big job given that `String` isn't move-aware. :(

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -0,0 +1,315 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris, line 2: [nit] 2024

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -0,0 +1,315 @@
...
+	auto [Result] = AWAIT_PRE(AssetSystem, CreateMaterial, RequestPredicate, Name, SpaceId);
+
+	EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+	const GLTFMaterial& Material = Result.GetGLTFMaterial();
+
+	EXPECT_EQ(Material.GetName(), Name);
+
+	OutMaterial = Result.GetGLTFMaterial();
+}
```
MAG-ElliotMorris, line 49: [nit] Spacing between each line seems a bit much, but just a style thing.

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+	// Create space to associate a material with
+	::Space Space;
+	CreateDefaultTestSpace(SpaceSystem, Space);
+
+	GLTFMaterial CreatedMaterial;
+	CreateMaterial(AssetSystem, "TestMaterial", Space.Id, CreatedMaterial);
+
+	// Cleanup
+	DeleteMaterial(AssetSystem, CreatedMaterial.GetAssetCollectionId(), CreatedMaterial.GetAssetId());
```
MAG-ElliotMorris: [Consider] I would have expected a test here, to ensure the material was actually created (and deleted)

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -98,6 +98,18 @@ void CreateSpaceWithBuffer(::SpaceSystem* SpaceSystem,
...
+{
+	// Create space
+	const char* TestSpaceName		 = "OLY-UNITTEST-SPACE-MAG";
+	const char* TestSpaceDescription = "OLY-UNITTEST-SPACEDESC-MAG";
+
+	char UniqueSpaceName[256];
+	SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+	CreateSpace(SpaceSystem, UniqueSpaceName, TestSpaceDescription, SpaceAttributes::Private, nullptr, nullptr, nullptr, OutSpace);
+}
```
MAG-ElliotMorris, line 111: [Consider][Strong] I don't see why you wouldn't create the space object in this method and return it, rather than use a `out` param.
You could even move the system setup and whatnot into here, shorten all the above tests.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: I'm no good word person, but I don't believe plurals get apostrophes unless they're possessive.

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -252,20 +252,31 @@ class CSP_API AssetSystem : public SystemBase
...
 
-	// TODO: comment
+	/// @brief Creates a new material backed by an AssetCollection/Asset.
+	/// @param Name const csp::common::String& : The name of the new material.
+	/// @param SpaceId const csp::common::String& : The space id this material is associated with.
+	/// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
 	CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId, GLTFMaterialResultCallback Callback);
 
-	// TODO: comment
+	/// @brief Updates an existing materials properties.
```
MAG-ElliotMorris, line 261: [Consider] The documentation here could use a bit more specificity, especially this one, I'm not sure you could figure it out without glancing at source. What happens if the material doesn't exist, and how does the system determine whether the new material you're passing in matches an old one or not?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -37,6 +37,9 @@ namespace chs = services::generated::prototypeservice;
 constexpr int DEFAULT_SKIP_NUMBER		= 0;
 constexpr int DEFAULT_RESULT_MAX_NUMBER = 100;
 
+static const char* MATERIAL_ASSET_COLLECTION_NAME_PREFIX = "ASSET_COLLECTION_MATERIAL_";
+static const char* MATERIAL_ASSET_NAME_PREFIX			 = "ASSET_MATERIAL_";
+static const char* MATERIAL_FILE_NAME_PREFIX			 = "ASSET_MATERIAL_FILE_";
```
MAG-ElliotMorris, line 44: [Consider] Why static? Moving these into the anonymous namespace below seems a better option. (I'm assuming you're worried about linkage?)

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -905,9 +909,8 @@ void AssetSystem::CreateMaterial(const csp::common::String& Name, const csp::com
 				Callback(FinalResult);
 			};
 
-			const csp::common::String MaterialFileName = MATERIAL_FILE_NAME_PREFIX + SpaceId + "_" + Name;
+			const csp::common::String MaterialFileName = MATERIAL_FILE_NAME_PREFIX + SpaceId + "_" + Name + ".json";
```
MAG-ElliotMorris, line 912: [Consider] Constructing these material names almost seems worth utility methods at this point, especially now that we're hard encoding file format inline in this method. It's important logic and the decision point should be centralised.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1068,14 +1074,22 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
...
 			for (size_t i = 0; i < Assets.Size(); ++i)
 			{
 				DownloadAssetData(Assets[i], DownloadMaterialCallback);
 			}
+
+			// We have to wait so this function doesn't exit and destroy "AssetsDownloaded"
+			while (!Finished)
+			{
+				std::this_thread::sleep_for(std::chrono::milliseconds {10});
+			}
```
MAG-ElliotMorris, line 1092: [fix] Don't busy wait, use a proper async primitive if you can.

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -304,7 +319,7 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, GetInvalidMaterialTest)
 
 	// Attempt to get an invalid material
 	GLTFMaterial FoundMaterial;
-	GetMaterial(AssetSystem, "InvalidAssetCollectionId", "InvalidAssetId", FoundMaterial);
+	GetMaterial(AssetSystem, "InvalidAssetCollectionId", "InvalidAssetId", FoundMaterial, csp::systems::EResultCode::Failed);
```
MAG-ElliotMorris, line 322: [Nit] Maybe `ExpectGetMaterial` or something that conveys that this is also doing asserts?

MAG-ElliotMorris (changes requested):
👍

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: I don't think plurals get apostrophes unless they're possessive.

MAG-OR: correct. however, "an" is not plural :-)

MAG-mv: I don't think it's customary  I just can't spell :(

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: I can add a separate test for deletion

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -849,22 +861,282 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void
 
 void AssetSystem::CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId, GLTFMaterialResultCallback Callback)
 {
+	// 1. Create asset collection
+	auto CreateAssetCollectionCB = [this, &Name, Callback](const AssetCollectionResult& CreateAssetCollectionResult)
```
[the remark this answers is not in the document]
MAG-mv: I have considered revamping the way we do these calls. I would rather find a better solution and refactor all of our calls after.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: I believe this is something we can updated automatically

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: We do this in other places, unfortunately. I wonder if buffer can be const. I will see what I can do

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Yep, unamed namespace is better.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: > correct. however, "an" is not plural :-)
🤦 

This is also actually possesive, because it's properties that belong to a material. I'm a stupid.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: Agreed. This is a pattern we follow with all of our tests. We should change this when we refactor

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Agreed

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Updated to future/promise

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: done

Tests/src/InternalTests/MaterialTests.cpp
```diff
@@ -234,11 +234,11 @@ CSP_INTERNAL_TEST(CSPEngine, MaterialTests, MaterialJsonSerializationTest)
 	EXPECT_EQ(DeserializedMaterial.GetEmissiveTexture().GetUVScale(), TestEmissiveTextureUVScale);
 	EXPECT_EQ(DeserializedMaterial.GetEmissiveTexture().GetTexCoord(), TestEmissiveTextureTexCoord);
 
-	EXPECT_EQ(Material.GetBaseColorTexture().IsSet(), true);
-	EXPECT_EQ(Material.GetMetallicRoughnessTexture().IsSet(), true);
-	EXPECT_EQ(Material.GetNormalTexture().IsSet(), true);
-	EXPECT_EQ(Material.GetOcclusionTexture().IsSet(), true);
-	EXPECT_EQ(Material.GetEmissiveTexture().IsSet(), true);
```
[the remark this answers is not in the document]
MAG-mv: I think it's useful testing these, because you wouldnt notice if this us failing until to serialize/deserialize a second time

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: i decided to copy to a string first. Making the variable const was breaking the wrapper gen

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-mv: Updated

Library/src/Systems/Assets/AssetSystem.cpp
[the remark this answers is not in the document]
MAG-ElliotMorris: Fair enough for the moment 🤷

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: Thanks for doing that.

[Comment] Just a note though, the comment says that the copy prevents const_cast, which it does, but I'd argue the main virtue of the copy is preventing two objects having owning handles to the same data.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-ElliotMorris: As in it will get updated automatically, or we could in theory update it automatically? Seems odd to intentionally commit the wrong year. Not blocking the PR on this though.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: 👍

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-ElliotMorris: Wish it was easier to make little issues based off stuff like this, I'm the first to admit the hurdle to doing it often leads to little stuff like this being lost forever and never changing.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: I think there is a script for this. I will speak to @MAG-AdamThorn again

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-mv: I believe this is done is the GetMaterial and the new DeleteMaterial tests

MAG-ElliotMorris: Ah of course, naming heh.

PR #549, opened 2024-11-26
Fix/OB-3725 when renaming a hotspot group an unexpected create sequence event is received

MAG-ChristopherAtkinson:
Updated Update and Rename sequence to use the new endpoint, and updated test due to change in expected behaviour.


Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -103,7 +103,26 @@ void SequenceSystem::UpdateSequence(const String& SequenceKey,
 									const csp::common::Map<csp::common::String, csp::common::String>& MetaData,
 									SequenceResultCallback Callback)
 {
-	CreateSequence(SequenceKey, ReferenceType, ReferenceId, Items, MetaData, Callback);
+	if (!ValidateKey(SequenceKey))
+	{
+		CSP_LOG_FORMAT(csp::systems::LogLevel::Error,
```
MAG-OR, line 108: [Question] why not use CSP_LOG_ERROR_FORMAT?

MAG-ChristopherAtkinson: This was based on the format of the related code, so no particular reason so happy to update for readability.

MAG-OR: nah I was just curious. thanks!

MAG-ElliotMorris (approved):
LGTM

PR #551, opened 2024-11-27
[OF-1473] feat: Change URL to Multiplayer Service

MAG-OR:
- Update domain name for Multiplayer Service in CSP initialisation
- Update WebSocketClientTests to use the new hostname


Tests/src/InternalTests/WebSocketClientTests.cpp
```diff
@@ -40,7 +40,7 @@ CSP_INTERNAL_TEST(CSPEngine, WebSocketClientTests, SignalRClientStartStopTest)
 	LogIn(UserSystem, UserId);
 
 	// Start
-	auto* WebSocket = WebSocketStart("wss://ogs-odev-internal.magnoboard.com/mag-multiplayer/hubs/v1/multiplayer?id=");
+	auto* WebSocket = WebSocketStart("wss://ogs-multiplayer-odev-internal.magnoboard.com/mag-multiplayer/hubs/v1/multiplayer?id=");
```
jeromagnopus, line 43: nitpick: I know they're just tests, but I would still argue against any hardcoded URLs in code.

Library/src/CSPFoundation.cpp
```diff
@@ -255,10 +255,15 @@ bool CSPFoundation::Initialise(const csp::common::String& EndpointRootURI, const
...
 	const std::string PrototypeServiceURI	= RootURI + "/mag-prototype";
 	const std::string SpatialDataServiceURI = RootURI + "/mag-spatialdata";
-	const std::string MultiplayerServiceURI = RootURI + "/mag-multiplayer/hubs/v1/multiplayer";
 	const std::string AggregationServiceURI = RootURI + "/oly-aggregation";
 	const std::string TrackingServiceURI	= RootURI + "/mag-tracking";
 
+	const std::string Hostname = "ogs", MultiplayerHostname = Hostname + "-multiplayer";
+	const std::string RootMultiplayerURI
+		= RootURI.substr(0, RootURI.find(Hostname)) + MultiplayerHostname + RootURI.substr(RootURI.find(Hostname) + Hostname.length());
+	const std::string MultiplayerServiceURI = RootMultiplayerURI + "/mag-multiplayer/hubs/v1/multiplayer";
```
MAG-ElliotMorris, line 264: [Consider] It's pretty hard to follow what this is doing without understanding what the intended constructed string is, a comment or two describing the string transformation could help.
I wonder if it's necessary to go to the effort to transform this from the `RootURI`, could we not directly encode the new multiplayer URI and void some of this transformation?

MAG-ElliotMorris (approved):
Approved with optional comment.

Library/src/CSPFoundation.cpp
MAG-OR: Sure I'll add comments. 

I constructed it from RootURI because it is not always the same. It is passed as an argument in the Foundation initialisation. The production one is ogs.magnopus-something, but there are also others like ogs-internal.magnopus-something. 
I made the assumption that ogs-multiplayer-internal was the right order, thinking if it wasn't the CI would tell me, but now I am not so sure we have the test coverage for that.

I reached out to CHS and the person to ask was off yesterday, back on Monday

MAG-ElliotMorris: Ah right, sorry, I had thought this was in the tests, not the core lib itself.
Aye ... it is rather difficult to be certain that this is fully correct. I couldn't say for sure if this URI construction was always right, or is just right in the case of the particular services CHS has stood up right now.

MAG-OR: Agreed, I'll ask that person on Monday for guidance

MAG-OR: Comment request fixed in 18eb922b567bae6e56fc48f7b657eb7cf5257a3d

Tests/src/InternalTests/WebSocketClientTests.cpp
raynopus: This shouldn't work. The URL is pointing to the old Dev environment which doesn't exist anymore. Can we take this opportunity to change the URL to `wss://ogs-multiplayer-internal.magnopus-dev.cloud/mag-multiplayer/hubs/v1/multiplayer`?

MAG-OR: Sure thing. Does "-multiplayer" always come before other strings like "-internal"?

MAG-OR: awesome, thank you for confirming

MAG-ElliotMorris (approved):

MAG-ChristopherAtkinson (approved):

MAG-mv (approved):

PR #552, opened 2024-11-28
[OF-1408] Replace Variant pattern for Avatar Info settings

MAG-OR:
Remove the ability to set the Identifier Type to string or int using a Variant, so that the Identifier Type is now always a string.

### DIFF
--- a/Library/include/CSP/Systems/Settings/SettingsSystem.h
+++ b/Library/include/CSP/Systems/Settings/SettingsSystem.h
@@ -130,7 +130,7 @@ public:
 	/// @brief Retrieves the Avatar Portrait information associated with the space
 	/// If the user of the Avatar portrait associated with it the result callback will be successful, the HTTP res code will be ResponseNotFound
 	/// and the Uri field inside the UriResult will be empty
-    /// @param InUserID const csp::common::String : The id of the user we are seelomg to retrieve the portrait for
+	/// @param InUserID const csp::common::String : The id of the user we are seelomg to retrieve the portrait for
 	/// @param Callback UriResultCallback : callback when asynchronous task finishes
 	CSP_ASYNC_RESULT void GetAvatarPortrait(const csp::common::String InUserID, UriResultCallback Callback);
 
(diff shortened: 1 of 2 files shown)


Library/include/CSP/Systems/Settings/SettingsSystem.h
```diff
@@ -130,7 +130,7 @@ class CSP_API SettingsSystem : public SystemBase
 	/// @brief Retrieves the Avatar Portrait information associated with the space
 	/// If the user of the Avatar portrait associated with it the result callback will be successful, the HTTP res code will be ResponseNotFound
 	/// and the Uri field inside the UriResult will be empty
-    /// @param InUserID const csp::common::String : The id of the user we are seelomg to retrieve the portrait for
+	/// @param InUserID const csp::common::String : The id of the user we are seelomg to retrieve the portrait for
```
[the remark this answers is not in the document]
MAG-OR: Fixed in 35bf52ddcb990bd9cc74c6c60123971e5811d7cd

MAG-AdamThorn (approved):

Library/src/Systems/Settings/SettingsSystem.cpp
```diff
@@ -943,10 +926,9 @@ void SettingsSystem::GetAvatarInfo(AvatarInfoResultCallback Callback)
 			return;
 		}
 
-		const auto& type		   = Json["type"];
-		const auto& identifierType = Json["identifierType"];
+		const auto& type = Json["type"];
 
-		if (!type.IsInt() || !identifierType.IsInt())
```
MAG-ElliotMorris, line 949: [Consider] I wonder if we should keep the identifierType check here, ie, assert that it's a string. I know it must always be one from now on, but I'm wondering about old data, (presuming that these json documents are persisted somewhere). Even with that aside, feels like decent practice to guarentee the types of data coming out of untyped json.

MAG-ElliotMorris (approved):

Library/src/Systems/Settings/SettingsSystem.cpp
MAG-OR: In ed4038b65734569d0604bc55e697ac9f69707415, I have re-instated the identifierType field in the json, hard-coded it to string in all new json files, and added a check to make sure it is a string when loading json files. This will now generate an error for older json files that still contain an int type. (if they exist)

MAG-OR: Following our discussion with @MAG-mv, I have made it silently convert any old-style integers to string. I have also added a test case that ensures that ints are treated gracefully, should any of them by lying around. I have removed identifierType, we no longer set it.

MAG-ElliotMorris (approved):
Seems right to me. Would have liked to see a bit more forceful instructions in the deprecation adaption block, letting the future implementor know that they're almost certainly safe to and should remove this in like ... 6 months, but it's all good.

MAG-AdamThorn (approved):

PR #553, opened 2024-11-28
[OF-1475] fix: Updated Reparent Event Timing

MAG-mv:
- Reparent event now happens before the EntitySystem::DestroyEntity callback so clients can remove children before deletion
- Updated tests to ensure ordering


Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -419,6 +409,25 @@ void SpaceEntitySystem::DestroyEntity(SpaceEntity* Entity, CallbackHandler Callb
 
 	CSP_DELETE(Keys);
 
+	csp::common::Array<ComponentUpdateInfo> Info;
+
+	RootHierarchyEntities.RemoveItem(Entity);
+
+	// Manually process the parent updates locally
+	// We want this callback to fire before the deletion so clients can react to children first
```
MAG-AdamThorn, line 417: I think this comment needs to be updated as now happens after.

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -419,6 +409,25 @@ void SpaceEntitySystem::DestroyEntity(SpaceEntity* Entity, CallbackHandler Callb
...
+	auto ChildrenToUpdate = Children;
+
+	for (size_t i = 0; i < ChildrenToUpdate.Size(); ++i)
+	{
+		ChildrenToUpdate[i]->ParentId = nullptr;
```
MAG-OR, line 422: [question] Is there a reason for not using RemoveParentEntity? (not a leading question at all, when I saw that I just thought there may be a function that does it already and I found that one)

MAG-ElliotMorris (approved):
LGTM

Library/src/Multiplayer/SpaceEntitySystem.cpp
MAG-mv: RemoveParentEntity is used when going through the patch process, as it also sets ShouldUpdateParent. We don't need to do that here, as we're updating manually

Library/src/Multiplayer/SpaceEntitySystem.cpp
MAG-mv: It still happens before. It's just outside of the callback now

MAG-OR (approved):
well done for adding tests

MAG-AdamThorn (approved):

PR #557, opened 2024-12-03
[OF-1308] feat: Expose new OnRequestToLeaveScopes

MAG-OR:
Expose OnRequestToLeaveScopes to clients and pass on any errors reported by CHS


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -627,6 +634,46 @@ void MultiplayerConnection::ResetScopes(ErrorCodeCallbackHandler Callback)
 	Connection->Invoke("ResetScopes", Params, LocalCallback);
 }
 
+void MultiplayerConnection::LeaveScopes(ErrorCodeCallbackHandler Callback)
```
MAG-OR, line 637: created this on the same model as SetScopes and ResetScopes when I realised that ExitSpace was not firing the LeaveScopes event (is this needed???)

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -627,6 +634,46 @@ void MultiplayerConnection::ResetScopes(ErrorCodeCallbackHandler Callback)
...
+			INVOKE_IF_NOT_NULL(Callback, Error);
+
+			return;
+		}
+
+		INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
+	};
+
+	std::vector<signalr::value> ScopesVec;
+	ScopesVec.push_back("MyScope1");
```
MAG-OR, line 660: how do I populate these?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -627,6 +634,46 @@ void MultiplayerConnection::ResetScopes(ErrorCodeCallbackHandler Callback)
...
+	signalr::value ScopesReason = "MyScopeReason";
+
+	std::vector<signalr::value> ParamsVec;
+	ParamsVec.push_back(ScopesVec);
+	ParamsVec.push_back(ScopesReason);
+
+	signalr::value Params = signalr::value(std::move(ParamsVec));
+
+	// ???
+	// Connection->Invoke("LeaveScopes", Params, LocalCallback);
```
MAG-OR, line 671: none of these exist. "Method not found", "Error: unknown".  (?)

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -1945,4 +1963,41 @@ void SpaceSystem::DuplicateSpace(const String& SpaceId,
...
+// create new id spacesystem leavescopes (DONE)
+// duplicate onrequesttodisconnct (DONE)
+// add a test in space system. enter a space, bind to the event, exit the space, add an await. (DONE??? fails. event not called)
+
+void SpaceSystem::BindOnRequestToLeaveScopes()
+{
+	Connection->On("OnRequestToLeaveScopes",
+				   [this](const signalr::value& Params)
+				   {
+					   auto Scopes				= Params.as_array()[0].as_array();
```
MAG-OR, line 1980: This is the callback that I am trying (and failing) to call from the test

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -2853,4 +2853,75 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, DuplicateSpaceTest)
...
+		LeaveScopesCallbackCalled = true;
+	};
+
+	Connection->SetLeaveScopesCallback(LeaveScopesCallback);
+
+	// Enter space (i.e. enter scopes)
+	auto [EnterSpace1Result] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space1.Id);
+
+	// Exit space (i.e. leave scopes)
+	auto [ExitSpaceResult] = AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
```
MAG-OR, line 2910: I added the ExitSpace await, but the LeaveScopes callback was not called, so then I experimented with the below (re-enter the same space without exiting first, enter a different space without exiting the first one first), but failed to receive the LeaveScopes callback. Then I realised that SetScopes and ResetScopes and the likes actually "Invoke" commands, so I tried to do the same thing (see another comment), but am unsure what method to invoke.

- Do I need to explicitly invoke LeaveScopes inside ExitSpace on the same model as ResetScopes, if so which method do I have to invoke?
- If not, how do I trigger the LeaveScopes callback, i.e. how can I "attempt to use the same account in a new tab/client" in the test?

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: Shooting in the dark here, but I wonder if because the hub method has 2 params on it (https://github.com/magnopus/Magnopus.Services/blob/071cfd4b7553d2314a09017456c607152f9f0a5a/Source/Magnopus.Service.Multiplayer.Contracts/HubClients/IMultiplayerHubClient.cs#L47) it might fail to find this?

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-OR: Following [this conversation with CHS](https://magnopus.slack.com/archives/C02T07E8P2B/p1734365278798399), it turns out that the event is not currently being fired at all. It will be fired, after [this PR](https://github.com/magnopus/terraform-env-olympus2-development/pull/104) is merged, when:
- a user is removed from a Space 
- a space is deleted
- you connect to any space but haven't disconnected from previous sessions

MAG-ElliotMorris:
@MAG-OR Can this draft be closed?

PR #558, opened 2024-12-04
[OF-1468] Multiplayer Test Runner.

MAG-ElliotMorris:
Add a multiplayer test runner, a command line test framework designed for cross process testing, with the intent to be launchable as multiple parallel instances to test multiple users in the same space.

## Project Layout
- `src` Primary source file directory for internal code
  - `InternalUnitTests` GTest tests. Not just called "tests" because runnable tests also exist
  - `RunnableTests` The test cases that this framework supports and is designed to invoke. Only one example there for now
- `thirdparty` Third party dependencies, just header only at the moment for simplicity. CLI11 and UUID_v4
- `include` External headers. Intended to be included by managing processes to help in the management.

## Build System
The project exists in a `MultiplayerTestRunner` directory in the solution. It is linked against the CSP DLL, and as such uses the `USING_CSP_DLL` define. This is set automatically in premake. The same old copy-dll-next-to-binary solution is used to make them available.
Googletest is linked from the solution wide `Thirdparty` directory. A preprocessor flag `RUN_MULTIPLAYER_RUNNER_TESTS` is set to branch to google test invocation or not. This flag is enabled by default on debug builds.

## Description
The concept here is that the runner is a framework for writing lots of different test stubs (the stuff in `RunnableTests`), and launching it via a CLI. A timeout is provided, so the runner does not hang indefinitely, although that's only a soft guarantee as the content of the runnable tests is opaque.
### Termination
The runner uses RAII types to try and ensure cleanup. However this will not be possible with process termination, and process termination is an inevitability when managing this in an automated environment.
The `space` field in the CLI is provided for this, allowing managers to create their own space, and thus take responsibility for cleanup in that way. This is also the mechanism by which multiple users in the same space will be supported.
In any case, leaving spaces around and users logged in isn't fatal, it's just messy.
### Process Descriptors
This is the biggest swing, and a solution I'm not 100% sure is going to work, although I don't see why it wouldn't
The concept here is that the application prints to `stdout` a set of defined phrases, which can be observed from calling processes consuming the `stdout` stream. These can then be used to make sure test timings are good, and reduce or eliminate flakiness. The important one is the `READY_FOR_ASSERTIONS` one, which is printed after the runnable test returns.
### Room for Improvement
- The main thing on my mind is passing arguments to Runnable Tests. I'm not sure duplicated explicit tests will be tenable long term. Not that tricky a change to make if it becomes neccesary though.
- On a smaller note because our result types are not movable, we capture promises by reference in this implementation, which is _fine_ but not best practice. Moving these things into off-thread callbacks instead rules out potential lifetime and synchronisation issues. 
- Many of the tests require a credentials file, which is unfortunate because the application dosen't need one to run. It'd be nice if we had some sort of default credentials we could pull from somewhere rather than asking the dev to know to provide one.
- It might be good to have a small suite of basic, common actions in this application, even if they're not stuff we particularly need to do automated tests over in this manner. Would make it a nice general validation tool.

> [!CAUTION]
>  Please do not review `thirdparty/CLI11.hpp`, `thirdparty/endianness.h` or `thirdparty/uuid_v4.h`, these are header only libraries copied into the project.

### Screenshots!
image
image


MultiplayerTestRunner/premake5.lua
```diff
@@ -0,0 +1,70 @@
...
+		-- Link to CSP in place
+        dependson {"ConnectedSpacesPlatform"}
+        libdirs {"%{wks.location}/Library/Binaries/%{cfg.platform}/%{cfg.buildcfg}"}
+		links {"ConnectedSpacesPlatform"}
+		defines {"USING_CSP_DLL"}
+			   
+		 -- Conditionally link google test, not the standard _d stuff so we need to do it per config
+	   filter "configurations:*Debug*"
+		  defines { "RUN_MULTIPLAYER_RUNNER_TESTS" } -- Run tests by default in debug configs
+		  links { "gtestd_md", "gtest_maind_md" } -- Debug versions
```
MAG-ElliotMorris, line 45: [fix] gtest_main not needed, you invoke gtest directly from your main.

premake5.lua
```diff
@@ -4,6 +4,8 @@ include "premake5_helpers.lua"
 include "Tests/premake5.lua"
 include "Tests/CSharp/premake5.lua"
 include "Tests/Multiplayer/premake5.lua"
+include "MultiplayerTestRunner/premake5.lua"
+include "MultiplayerTestManager/premake5.lua"
```
MAG-ElliotMorris, line 8: [fix] Don't leak test manager stuff in this PR

MultiplayerTestRunner/src/Main.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <thread>
+
+#ifdef RUN_MULTIPLAYER_RUNNER_TESTS
+	#include <gtest/gtest.h>
+#endif
+
+/********************************
+ * MULTIPLAYER TEST RUNNER
+ * This is a CLI application designed to launch specified tests on independent processes (as CSP currently supports only one user per process), in
+ * order to support testing multiple clients on multiple processes interacting in the same room.
```
MAG-ElliotMorris, line 21: [nit] "... in the same space"

MultiplayerTestRunner/src/Main.cpp
```diff
@@ -0,0 +1,115 @@
...
+ * When used in an automated testing context, it is
+ * designed to communicate back with the calling process via stdout process descriptors (see `include/ProcessDescriptors.h`), such that the calling
+ * process can reliably call assertions with the correct timings.
+ * The CLI itself provides some flexibility for use in a more direct manner as an ad-hoc test application.
+ *
+ * Dependencies :
+ *	- CSP. From the ConnectedSpacePlatform Solution. Linked.
```
MAG-ElliotMorris, line 28: [nit] Tab/Space indentation

MultiplayerTestRunner/src/Main.cpp
```diff
@@ -0,0 +1,115 @@
+#include "../include/ErrorCodes.h"
```
MAG-ElliotMorris, line 1: [fix] You've forgotten copyright blocks literally everywhere ...

MultiplayerTestRunner/include/ErrorCodes.h
```diff
@@ -0,0 +1,18 @@
...
+{
+constexpr const int SUCCESS							= 0;
+constexpr const int COULD_NOT_FIND_CREDENTIALS_FILE = 100;
+constexpr const int MALFORMED_CREDENTIALS_FILE		= 101;
+constexpr const int FAILED_TO_LOGIN					= 102;
+constexpr const int INVALID_TEST_SPECIFIER			= 103;
+constexpr const int OTHER_EXCEPTION					= 104; // For any exception that isn't thrown explicitly by the Multiplayer Test Runner
+constexpr const int FAILED_TO_CREATE_SPACE			= 105;
+constexpr const int FAILED_TO_ENTER_SPACE			= 106;
+constexpr const int CLI_PARSE_ERROR					= 107;
```
MAG-ElliotMorris, line 17: [Comment] 😱 We really need to fix the clang autoformatting my word...

MultiplayerTestRunner/src/Utils.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+	return TestAccountCredentials {DefaultLoginEmail.c_str(),
+								   DefaultLoginPassword.c_str(),
+								   AlternativeLoginEmail.c_str(),
+								   AlternativeLoginPassword.c_str(),
+								   SuperUserLoginEmail.c_str(),
+								   SuperUserLoginPassword.c_str()};
+}
+
+bool RequestPredicate(const csp::systems::ResultBase& Result)
```
MAG-ElliotMorris, line 54: [Consider] Is this used anywhere anymore since the migration to futures? You might be able to delete it.

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -0,0 +1,107 @@
...
+	App.add_option("-o,--timeout", TimeoutInSeconds, "How long until the process self-terminates, in seconds. If not set, defaults to 30");
+	App.add_option("-c,--endpoint", Endpoint, "Cloud services endpoint. If not set, defaults to `https://ogs-internal.magnopus-dev.cloud`");
+
+	try
+	{
+		App.parse(argc, argv);
+	}
+	catch (const CLI::Success& e)
+	{
+		// App.exit() deals with error string printing. It's an odd pattern, an exception being a "Success".
```
MAG-ElliotMorris, line 94: [fix] "help string printing" not error string printing.

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -0,0 +1,107 @@
...
+		App.exit(e);
+		throw Utils::ExceptionWithCode {MultiplayerTestRunner::ErrorCodes::SUCCESS, ""};
+	}
+	catch (const CLI::ParseError& e)
+	{
+		throw Utils::ExceptionWithCode {MultiplayerTestRunner::ErrorCodes::CLI_PARSE_ERROR, e.what()};
+	}
+
+
+	return ValidateInvocationArgs(LoginEmail, LoginPassword, TestIdentifier, Endpoint, SpaceId, TimeoutInSeconds);
```
MAG-ElliotMorris, line 104: [nit] Double newline

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
```diff
@@ -0,0 +1,152 @@
+#include "../../include/ErrorCodes.h"
+#include "../../include/TestIdentifiers.h"
+#include "../CLIArgs.h"
+
+#include <gtest/gtest.h>
+
+
+TEST(CLITest, AllArgsBasic)
```
MAG-ElliotMorris, line 8: [nit] Double newline

MultiplayerTestRunner/src/InternalUnitTests/Test_SpaceRAII.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <CSP/Common/String.h>
+#include <CSP/Systems/Spaces/SpaceSystem.h>
+#include <CSP/Systems/SystemsManager.h>
+#include <gtest/gtest.h>
+#include <optional>
+
+namespace
+{
+/* Some tests only run if there's a credentials file */
+std::optional<Utils::TestAccountCredentials> CredentialsFromFile()
```
MAG-ElliotMorris, line 16: [consider] Maybe this should be a utils method, it's duplicated more than twice.

MultiplayerTestRunner/src/CLIArgs.h
```diff
@@ -0,0 +1,39 @@
...
+	int TimeoutInSeconds;
+	// Validated when we try to enter the space. Should not be null after that point.
+	// A bit of a hack, see main.cpp (at time of writing) for when this is actually set.
+	std::optional<std::string> SpaceId;
+};
+ASSERT_MOVE_CAPABLE(RunnerSettings);
+
+/*
+ * Builds the CLI for the Multiplayer Test Runner.
+ * Call this first in main to get your CLI.
```
MAG-ElliotMorris, line 34: [fix] This _isn't_ called in main anymore. Bad advice. Remove

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -0,0 +1,107 @@
...
+/*
+ * Take the raw input from the CLI, validate, and convert to structured data.
+ * If Login or Password not provided, this will attempt to read `test_account_creds.txt`, and throw if it cannot.
+ * TestIdentifier is the only mandatory arg, this will validate that the string provided is an actual identifier.
+ * For optional arguments, if empty data is provided, will populate them with defaults.
+ * SpaceId will not be defaulted, and will be left empty if empty, in the expectation a space is about to be created and the value set externally.
+ */
+CLIArgs::RunnerSettings ValidateInvocationArgs(std::optional<std::string> LoginEmail,
+											   std::optional<std::string> LoginPassword,
+											   std::string TestIdentifier,
```
MAG-OR, line 22: [question] is it odd to have the only non-optional parameter in third place rather than first?

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+	using namespace MultiplayerTestRunner::TestIdentifiers;
+	std::string TestID		= TestIdentifierToString(TestIdentifier::CREATE_AVATAR);
+	std::vector<char*> args = {"MultiplayerTestRunner", "--test", TestID.data(), "--password", "password123"};
+
+	try
+	{
+		CLIArgs::RunnerSettings Settings = CLIArgs::ProcessCLI(args.size(), args.data());
+		// If there is a credentials file, we'll now have a login email
+		EXPECT_TRUE(Settings.LoginEmailAndPassword.first.length() > 0);
```
MAG-OR, line 109: [question] I'm not sure I understand this test. a password is passed as argument, but not an email, so you look for the email in the credentials file, and you use the password that was passed on the command line?
do we really want to do that? will it do that, if, e.g. the email was passed but with a typo (--emoil or similar)?

would it be better to completely decorrelate (email/password) and (creds file)? i.e. if no email and no password, then creds file. if email but no password: error, user forgot password, and vice versa.

I'm genuinely confused as to why you would use the email of one and the password of the other (it may just be me being thick)

MultiplayerTestRunner/src/Main.cpp
```diff
@@ -0,0 +1,115 @@
...
+ * By this point, the client should be logged in and inside a space, hence tests need not concern themselves with space creation and cleanup.
+ * This function blocks until the timeout has elapsed.
+ */
+void RunTest(CLIArgs::RunnerSettings Settings, std::chrono::steady_clock::time_point ProgramStartTime)
+{
+	using namespace MultiplayerTestRunner::TestIdentifiers;
+	switch (Settings.TestIdentifier)
+	{
+		case TestIdentifier::CREATE_AVATAR:
+			CreateAvatar::RunTest();
```
MAG-OR, line 45: [question] is this requirement to manually add new tests in the switch maintainable long term? 
"I added a test but it won't run!!!! - Have you added it in the switch? - Doh!"

("yes" is an acceptable answer)

MultiplayerTestRunner/src/RunnableTests/CreateAvatar.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <CSP/Multiplayer/SpaceEntitySystem.h>
+#include <CSP/Multiplayer/SpaceTransform.h>
+#include <CSP/Systems/Spaces/SpaceSystem.h>
+#include <CSP/Systems/SystemsManager.h>
+#include <future>
+
+namespace CreateAvatar
+{
+
+void RunTest()
```
MAG-OR, line 12: [nit/question] I'm probably nitpicking, but since this is the "multiplayer" test client, would it make sense for the very first test we have, which kind of acts as a proof of concept, to use multiple (2) players?

MultiplayerTestRunner/src/SpaceRAII.h
```diff
@@ -0,0 +1,35 @@
...
+#include <optional>
+#include <string>
+
+/*
+ * RAII container object to facilitate automatically cleaning up a space when leaving scope
+ * This involves 2 actions, leaving the space, and destroying the space.
+ * This type will create a new random space upon construction if a SpaceId is not set.
+ * If a SpaceID is set, this type only joins and leaves the room, not destroying it, as it is assumed to already exist.
+ */
+class SpaceRAII
```
MAG-mv, line 14: [consider] Name could be "ScopedSpace"?

MultiplayerTestRunner/include/TestIdentifiers.h
```diff
@@ -0,0 +1,64 @@
...
+#include <string>
+#include <unordered_map>
+
+namespace MultiplayerTestRunner::TestIdentifiers
+{
+/*
+ * The test runner works by passing a specific test identifier in an a command line arg.
+ * These things have to be strings, so this file serves to encode which tests are available,
+ * and to give a nice way to get at the correct strings.
+ * Absolutey drives me mad that there's no good compile time solution for thees conversion implementations.
```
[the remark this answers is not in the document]
MAG-mv: Looks like you already fixed

MultiplayerTestRunner/src/InternalUnitTests/Test_SpaceRAII.cpp
MAG-ElliotMorris: Na, I don't think having test specific methods in Utils is that great. There's already a little bit of that sneaking in with public/private access :(

MultiplayerTestRunner/src/CLIArgs.h
MAG-ElliotMorris: Wait! Yes it is! Please try harder when doing code reviews @MAG-ElliotMorris

MultiplayerTestRunner/src/InternalUnitTests/Test_SpaceRAII.cpp
MAG-OR: I just love this. (sorry)

premake5.lua
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/b8f88a1d5867389f340ffbd7135818ed4b4fd7f5

MultiplayerTestRunner/src/Main.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/f731d167c1b5afe99b618af098c2c83797ee78d0

MultiplayerTestRunner/src/Utils.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/b9fd2380203022080eb0e51d1a5d29cd78874eae

MultiplayerTestRunner/src/CLIArgs.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/8cf7129466935156ecb5834ec16f144a37a1822b

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/548d87dff3761867c3ef4001343d393114f079d5

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
MAG-ElliotMorris: The idea is that if either a password or a username is not passed, it will use _both_ credentials from the credentials file, so you never use one from the CLI and one from the file, it's all or nothing.
You are right that it's a bit odd though, just from an interface perspective, not surprised it confused you. What do you think, should we make it so that you need to pass both email and password arguments or else you just get an error, rather than an automatic fallback to the credentials file?

MultiplayerTestRunner/src/Main.cpp
MAG-ElliotMorris: "Yes"

:P

Seriously though I don't think it's worth engineering a solution to this sort of thing if and until the number of tests become unmanageable. If (big if :P) we've our warnings configured correctly, the build will error about non-exhaustive switch statements.

MultiplayerTestRunner/src/RunnableTests/CreateAvatar.cpp
MAG-ElliotMorris: It does! Sort of...

Recall that this is the runner component of the test client. This whole program is meant to be invoked multiple times, simultaneously. To test more than one player, you launch `MultiplayerTestRunner.exe` twice, then check if 2 avatars have been created.

MultiplayerTestRunner/src/SpaceRAII.h
MAG-ElliotMorris: Because RAII is such a terrible name for the concept? It is at that, but I think I prefer it here, just because `Scope` can be interpreted in other ways, where this object very specifically exists in order to leverage its destructor to do cleanup.

MultiplayerTestRunner/include/ErrorCodes.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+/*
+ * A listing of all the known return codes the Multiplayer Test Runner may return as an error.
+ * Succesful execution returns 0, as usual.
```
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/65c81f9c886acefb7e95b3ffe3f101d1b3cc8c93

MultiplayerTestRunner/include/ProcessDescriptors.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+constexpr const char* READY_FOR_ASSERTIONS_DESCRIPTOR = "READY_FOR_ASSERTIONS";
+
+/*
+ *  Emitted when the test has logged in.
+ */
+constexpr const char* LOGGED_IN_DESCRIPTOR = "LOGGED_IN";
+
+/*
+ * Emitted when the test has logged out. Logout may not occur emitted if process is terminated.
```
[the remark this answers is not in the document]
MAG-ElliotMorris: Aye, copy paste error.
Part of this : https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/65c81f9c886acefb7e95b3ffe3f101d1b3cc8c93

MultiplayerTestRunner/include/ProcessDescriptors.h
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+ * Emitted when a new space is created. Not always emitted as existing space may be
+ * specified for use via the `--space` CLI param.
+ */
+constexpr const char* CREATED_SPACE_DESCRIPTOR = "CREATED_SPACE";
+
+/*
+ * Emitted during cleanup when a space is deleted. Will not be emitted if the `--space` CLI param was used to
+ * specify a custom space, as cleanup is not performed in that instance. Room cleanup may not occur emitted if
```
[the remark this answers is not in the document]
MAG-ElliotMorris: Aye, copy paste error.
Part of this : https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/65c81f9c886acefb7e95b3ffe3f101d1b3cc8c93

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
MAG-OR: oh right, I didn't realise it was then using the pair from the file. it does make a lot more sense. 

my worry by falling back silently on the credentials from the file is that you may make a mistake, not realise it, and assume you are using your custom credentials when you are in fact using the default ones. it may be worth printing a warning, at least?

MultiplayerTestRunner/src/Main.cpp
MAG-OR: that is completely fair enough

MAG-OR (approved):
My comments have been addressed and questions answered. About the warning for the email/password vs creds file, I am not too fussed, depends what you think and what the others think. 
Anyway thanks for implementing this, it is going to be super useful.

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+/*
+ * Take the raw input from the CLI, validate, and convert to structured data.
+ * If Login or Password not provided, this will attempt to read `test_account_creds.txt`, and throw if it cannot.
+ * TestIdentifier is the only mandatory arg, this will validate that the string provided is an actual identifier.
+ * For optional arguments, if empty data is provided, will populate them with defaults.
+ * SpaceId will not be defaulted, and will be left empty if empty, in the expectation a space is about to be created and the value set externally.
+ */
+CLIArgs::RunnerSettings ValidateInvocationArgs(std::optional<std::string> LoginEmail,
+											   std::optional<std::string> LoginPassword,
```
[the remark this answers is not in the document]
MAG-ElliotMorris: This is one of my bias's, I tend to favour value semantics for simple types even when const refs _are_ demonstrably better. It's a habit I should change really, although I could make an argument, I think i'd eventually be wrong.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/753a7a0c4800e61b633b01dacd6da20e85313204 (Email and Password are mutated in the function so remain copies)

As an aside, I really _really_ like these tables from the core standard guidelines in this regard. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information

MultiplayerTestRunner/include/TestIdentifiers.h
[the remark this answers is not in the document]
MAG-ElliotMorris: Nope. Fixed now though : https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/adc59e72b0195e7713d0df1f69f77e1a71e4d268

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/558/commits/68334c43cdfb1f5cc6c96c99ba5a2e0e7ccdb9ac

MAG-ElliotMorris:
@MAG-OR It's sort of annoying how github is configured to automatically dismiss stale reviews, I think what you did with the approve earlier was the right thing, didn't need to dismiss the whole review. :(

MultiplayerTestRunner/include/TestIdentifiers.h
```diff
@@ -25,7 +25,7 @@ namespace MultiplayerTestRunner::TestIdentifiers
  * The test runner works by passing a specific test identifier in an a command line arg.
  * These things have to be strings, so this file serves to encode which tests are available,
  * and to give a nice way to get at the correct strings.
- * Absolutey drives me mad that there's no good compile time solution for thees conversion implementations.
+ * Absolutey drives me mad that there's no good compile time solution for these conversion implementations.
```
MAG-OR, line 28: [nit] Absolutey!

sorry. just fix it in the next ticket or something

MAG-OR (approved):
Have reviewed the change of logic re email/password vs creds file. Looks good, thank you!

MultiplayerTestRunner/include/TestIdentifiers.h
MAG-ElliotMorris: Hah! I'm absolutey just leaving this typo in.

MultiplayerTestRunner/src/CLIArgs.cpp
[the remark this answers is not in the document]
MAG-mv: That is a nice table

MAG-mv (approved):

PR #563, opened 2024-12-10
[NT-0] fix: Workaround CI std::pair comparisons

MAG-ElliotMorris:
Our teamcity runners seem to be rejecting the std::pair comparison, even though they work fine locally.
Just workaround it to unblock things.


MAG-AdamThorn (approved):

MAG-OR (approved):
LGTM

PR #564, opened 2024-12-11
[OF-1435] Add script support for entity hierarchies

MAG-ChristopherAtkinson:
Clients need a way of manipulating hierarchies through scripts, so a script interface needs to be added for this feature.


Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+	auto* AvatarComponent = InUser->GetComponent(0);
+
+	EXPECT_EQ(AvatarComponent->GetComponentType(), ComponentType::AvatarData);
+
+	std::cerr << "OnUserCreated" << std::endl;
+}
+
+} // namespace
+
+#if RUN_ALL_UNIT_TESTS || RUN_SCRIPTSYSTEM_TESTS || RUN_UPDATE_SPACE_ENTITY_GLOBAL_POSITION_TEST
```
MAG-AdamThorn, line 66: [Consider] It would be great to have some comments in these tests, it is hard to follow what is going on.

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+	EntitySystem->SetEntityCreatedCallback(
+		[](SpaceEntity* Entity)
+		{
+		});
+	EntitySystem->SetScriptSystemReadyCallback(ScriptSystemReadyCallback);
+
+	const csp::common::String& UserName = "Player 1";
+	const SpaceTransform& UserTransform
+		= {csp::common::Vector3 {1.452322f, 2.34f, 3.45f}, csp::common::Vector4 {4.1f, 5.1f, 6.1f, 7.1f}, csp::common::Vector3 {1, 1, 1}};
+	AvatarState UserAvatarState				  = AvatarState::Idle;
```
MAG-AdamThorn, line 122: [Question] Why are we creating an avatar here, is that required to execute scripts, I do not believe that it is. It seems to add a bit of unnecessary complexity if it isn't required.

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+
+	auto ScriptSystemIsReady = [&ScriptSystemReady]()
+	{
+		std::cerr << "Waiting for ScriptSystemReady" << std::endl;
+		return (ScriptSystemReady == true);
+	};
+
+	EXPECT_EQ(ResponseWaiter::WaitFor(ScriptSystemIsReady, std::chrono::seconds(5)), true);
+
+	{
```
MAG-ElliotMorris, line 335: [Question] Why the additional scope here, is there some RAII thing I'm missing?

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+			EntitySystem->ProcessPendingEntityOperations();
+
+			std::cerr << "Waiting for EntityUpdatedIsReady" << std::endl;
+			return (EntityUpdated == true);
+		};
+
+		ResponseWaiter::WaitFor(EntityUpdatedIsReady, std::chrono::seconds(5));
+
+		// EXPECT_EQ(Object->GetGlobalRotation(), csp::common::Vector4::One());
+		// EXPECT_EQ(ChildObject->GetGlobalRotation(), csp::common::Vector4(2, 2, 2, 2));
```
MAG-ElliotMorris, line 396: [Fix] Remove dead code if not neccesary.

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+		auto [Object]						 = AWAIT(EntitySystem, CreateObject, ObjectName, ObjectTransform);
+
+		const csp::common::String ChildObjectName = "Child Object 1";
+		SpaceTransform ChildObjectTransform		  = {csp::common::Vector3::Zero(), csp::common::Vector4::Zero(), csp::common::Vector3::One()};
+		auto [ChildObject]						  = AWAIT(EntitySystem, CreateObject, ChildObjectName, ChildObjectTransform);
+
+		ScriptSpaceComponent* ScriptComponent = static_cast<ScriptSpaceComponent*>(ChildObject->AddComponent(ComponentType::ScriptData));
+
+
+		ChildObject->QueueUpdate();
```
MAG-ElliotMorris, line 697: [nit] Double newline

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+
+		EXPECT_NE(Object, ChildObject->GetParentEntity());
+
+		ScriptComponent->SetScriptSource(csp::common::String(ScriptText.c_str()));
+		ChildObject->GetScript()->Invoke();
+
+		csp::CSPFoundation::Tick();
+
+
+		ChildObject->QueueUpdate();
```
MAG-ElliotMorris, line 709: [nit] Double newline

MAG-ElliotMorris (changes requested):
Great bulk of tests.

Library/src/Multiplayer/Script/EntityScriptInterface.cpp
```diff
@@ -88,6 +104,59 @@ void EntityScriptInterface::SetRotation(EntityScriptInterface::Vector4 Rot)
...
+
+SpaceEntity* EntityScriptInterface::GetParentEntity() const
+{
+	if (Entity)
+	{
+		return Entity->GetParentEntity();
+	}
+
+	return nullptr;
+}
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: Yeah the ternary is good for short handing if else blocks, wrote this way initially for consistency

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+	};
+
+	auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id);
+
+	EXPECT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+	EntitySystem->SetEntityCreatedCallback(
+		[](SpaceEntity* Entity)
+		{
+		});
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: Yes and no, if they are not set you get a log in the console but doesn't prevent the test in anyway so could remove

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -206,6 +206,25 @@ class EntitySystemScriptInterface
 		return ScriptInterface;
 	}
 
+	std::vector<EntityScriptInterface*> GetRootHierarchyEntities()
+	{
+		std::vector<EntityScriptInterface*> RootHierarchyEntities;
+		if (EntitySystem)
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: Maintained the nested block over return early for consistency in this case, but would prefer in future.

MAG-mv (dismissed):

Tests/src/InternalTests/SpaceEntityTests.cpp
MAG-ChristopherAtkinson: These test are based of the existing script system test, the scoping is just to separate the main test from the require setup and destruction step such as logging in, ensuring the script system is available, etc

MAG-ElliotMorris: Just organizational then? Fair enough.

Library/src/Multiplayer/Script/EntityScriptInterface.cpp
```diff
@@ -88,6 +104,59 @@ void EntityScriptInterface::SetRotation(EntityScriptInterface::Vector4 Rot)
...
+
+	return 0;
+}
+
+void EntityScriptInterface::SetParentId(int64_t ParentId)
+{
+	if (Entity)
+	{
+		Entity->SetParentId(ParentId);
+	}
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: The parent ID should return by default in this case if not set so we could introduce error, however this should not be done at the script end but as part of the code code base, would suggest out of scope but definitely a good observation

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+
+		auto EntityUpdatedIsReady = [&EntityUpdated, &EntitySystem]()
+		{
+			EntitySystem->ProcessPendingEntityOperations();
+
+			std::cerr << "Waiting for EntityUpdatedIsReady" << std::endl;
+			return (EntityUpdated == true);
+		};
+
+		ResponseWaiter::WaitFor(EntityUpdatedIsReady, std::chrono::seconds(5));
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: Ah yeah this was a conscious choice, as it is a recuring pattern in the tests, so to maintain current practices this was included, however we should consider if we want to migrate all these implementation in future as part of a separate ticket

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -0,0 +1,1093 @@
...
+				UniqueSpaceName,
+				TestSpaceDescription,
+				csp::systems::SpaceAttributes::Private,
+				nullptr,
+				nullptr,
+				nullptr,
+				nullptr,
+				Space);
+
+	std::atomic_bool ScriptSystemReady = false;
```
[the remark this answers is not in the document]
MAG-ChristopherAtkinson: This was defined in previous script system test, however looking further it seems likely that its assumed this  a result of the threading of multiplayer system, however the callback are fired on the main thread so should be fine to remove

MAG-ElliotMorris (approved):

MAG-mv (approved):

MAG-ChristopherAtkinson:
[Update] delaying merge as it seems the newly introduce remove space entity parent and update space entity parent id test are failing on Team City, not seeing the issue locally but will debug further.

MAG-mv (approved):

MAG-ElliotMorris (approved):

PR #566, opened 2024-12-13
[OF-1408] fix: update AvatarInfoResult type

MAG-mv:
- Updated AvatarInfoResult to use a string
- updated relevant tests

### DIFF
--- a/Library/src/Systems/Settings/SettingsSystem.cpp
+++ b/Library/src/Systems/Settings/SettingsSystem.cpp
@@ -937,15 +937,13 @@ void SettingsSystem::GetAvatarInfo(AvatarInfoResultCallback Callback)
 
 		InternalResult.SetAvatarType(static_cast<AvatarType>(type.GetInt()));
 
-		if (Json.HasMember("identifierType") && Json["identifierType"].IsInt()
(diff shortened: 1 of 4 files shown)


Library/src/Systems/Settings/SettingsSystem.cpp
```diff
@@ -937,15 +937,13 @@ void SettingsSystem::GetAvatarInfo(AvatarInfoResultCallback Callback)
 
 		InternalResult.SetAvatarType(static_cast<AvatarType>(type.GetInt()));
 
-		if (Json.HasMember("identifierType") && Json["identifierType"].IsInt()
-			&& static_cast<VariantType>(Json["identifierType"].GetInt()) == VariantType::Integer)
+		if (Json.HasMember("identifierType") && static_cast<VariantType>(Json["identifierType"].GetInt()) == VariantType::Integer)
```
MAG-ChristopherAtkinson, line 940: [Question] Do we still need this now the type is explicitly csp::common::String?

MAG-mv: For Get we do, because it may have previously been stored as an int, and we need this to be backward compatible

MAG-ElliotMorris (approved):

PR #568, opened 2024-12-17
[OB-3939] fix: Passing tags as nullptr no longer resets tags

MAG-OR:
- Passing the optional tags parameter as nullptr to UpdateSpaceMetadata now leaves the tags unchanged instead of resetting them.
- Add a getter/setter for tags in SpaceMetadataResult[s]Callback
- Add test cases to ensure non-regression

### DIFF
--- a/Library/include/CSP/Systems/Spaces/Space.h
+++ b/Library/include/CSP/Systems/Spaces/Space.h
@@ -299,6 +299,7 @@ class CSP_API SpaceMetadataResult : public csp::systems::ResultBase
 
 public:
 	const csp::common::Map<csp::common::String, csp::common::String>& GetMetadata() const;
+	const csp::common::Array<csp::common::String>& GetTags() const;
 
 	CSP_NO_EXPORT
 	SpaceMetadataResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode) : csp::systems::ResultBase(ResCode, HttpResCode) {};
@@ -309,8 +310,10 @@ private:
 	SpaceMetadataResult() {};
 
 	void SetMetadata(const csp::common::Map<csp::common::String, csp::common::String>& MetadataAssetCollection);
+	void SetTags(const csp::common::Array<csp::common::String>& TagsAssetCollection);
 
 	csp::common::Map<csp::common::String, csp::common::String> Metadata;
+	csp::common::Array<csp::common::String> Tags;
 };
 
 
@@ -328,6 +331,7 @@ class CSP_API SpacesMetadataResult : public csp::systems::ResultBase
 
 public:
 	const csp::common::Map<csp::common::String, csp::common::Map<csp::common::String, csp::common::String>>& GetMetadata() const;
+	const csp::common::Map<csp::common::String, csp::common::Array<csp::common::String>>& GetTags() const;
 
 private:
 	SpacesMetadataResult(void*) {};
@@ -335,8 +339,10 @@ private:
 	SpacesMetadataResult() {};
 
 	void SetMetadata(const csp::common::Map<csp::common::String, csp::common::Map<csp::common::String, csp::common::String>>& InMetadata);
+	void SetTags(const csp::common::Map<csp::common::String, csp::common::Array<csp::common::String>>& InTags);
 
 	csp::common::Map<csp::common::String, csp::common::Map<csp::common::String, csp::common::String>> Metadata;
+	csp::common::Map<csp::common::String, csp::common::Array<csp::common::String>> Tags;
 };
 
 
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -438,8 +438,12 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
 												const Optional<Array<String>>& Tags,
 												AssetCollectionResultCallback Callback)
 {
-	auto PrototypeInfo
-		= CreatePrototypeDto(AssetCollection.SpaceId, AssetCollection.ParentId, AssetCollection.Name, NewMetadata, AssetCollection.Type, Tags);
+	auto PrototypeInfo = CreatePrototypeDto(AssetCollection.SpaceId,
+											AssetCollection.ParentId,
+											AssetCollection.Name,
+											NewMetadata,
+											AssetCollection.Type,
+											Tags.HasValue() ? Tags : AssetCollection.Tags);
 
 	services::ResponseHandlerPtr ResponseHandler
 		= PrototypeAPI->CreateHandler<AssetCollectionResultCallback, AssetCollectionResult, void, chs::PrototypeDto>(Callback, nullptr);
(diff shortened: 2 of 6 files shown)


Library/include/CSP/Systems/Spaces/Space.h
```diff
@@ -328,15 +331,18 @@ class CSP_API SpacesMetadataResult : public csp::systems::ResultBase
...
 private:
 	SpacesMetadataResult(void*) {};
 	SpacesMetadataResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode) : csp::systems::ResultBase(ResCode, HttpResCode) {};
 	SpacesMetadataResult() {};
 
 	void SetMetadata(const csp::common::Map<csp::common::String, csp::common::Map<csp::common::String, csp::common::String>>& InMetadata);
+	void SetTags(const csp::common::Map<csp::common::String, csp::common::Array<csp::common::String>>& InTags);
 
 	csp::common::Map<csp::common::String, csp::common::Map<csp::common::String, csp::common::String>> Metadata;
+	csp::common::Map<csp::common::String, csp::common::Array<csp::common::String>> Tags;
```
MAG-ElliotMorris, line 345: [Question] The tags aren't already stored in the `Metadata` object are they? I just wanna be sure we're not duplicating data and giving it a chance to get out of sync.
(I don't _think_ they are, I had a quick check of the call path.)

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -279,6 +279,24 @@ void GetSpacesMetadata(::SpaceSystem* SpaceSystem, const Array<String>& SpaceIds
 	OutMetadata = Result.GetMetadata();
 }
 
+void GetSpaceTags(::SpaceSystem* SpaceSystem, const String& SpaceId, Array<String>& OutTags)
```
MAG-ElliotMorris, line 282: [Consider] I'd strongly recommend calling these like `GetAndAssertSpaceTags`, or something with assert/test/check in the title, so people can understand the test flow from the method names.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -279,6 +279,24 @@ void GetSpacesMetadata(::SpaceSystem* SpaceSystem, const Array<String>& SpaceIds
...
+{
+	auto [Result] = AWAIT_PRE(SpaceSystem, GetSpaceMetadata, RequestPredicate, SpaceId);
+
+	EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+	OutTags = Result.GetTags();
```
MAG-ElliotMorris, line 288: [fix] Avoid out params unless necessary, prefer return.

MAG-ElliotMorris (changes requested):
Looking really good, great work. Just one style thing in the tests to [fix].

Library/include/CSP/Systems/Spaces/Space.h
MAG-OR: I don't think they are, but I would also prefer someone who is more familiar with the code to confirm

Library/src/Systems/Spaces/Space.cpp
```diff
@@ -290,11 +290,21 @@ const Map<String, String>& SpaceMetadataResult::GetMetadata() const
...
+
 void SpaceMetadataResult::SetMetadata(const Map<String, String>& InMetadata)
 {
 	Metadata = InMetadata;
 }
 
+void SpaceMetadataResult::SetTags(const Array<String>& InTags)
+{
+	Tags = InTags;
+}
```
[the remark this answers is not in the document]
MAG-OR: I agree. the first time I saw this I did a double take too

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-OR: This is essentially a copy of GetSpace[s]Metadata. would you like me to change both?

MAG-ElliotMorris: Ah I see, and I see there's an argument (a weak one imo in test code) to do it this way, to avoid a double construct throughout the test.

I'll drop this to a [Consider], although if it were me I would change both to be return base. The less mutation the better, I want as few possible reasons for the tests to go wrong as possible.

Library/include/CSP/Systems/Spaces/Space.h
MAG-ElliotMorris: I'm not sure that person exists, you should check in the debugger if there's any doubt.

MAG-OR: I had checked the debugger, they're not in there. you can see it in the test as well. however, it would help to know why there was never a getter for tags, and why we have 2 things (tags on one hand and metadata on the other) in there. having two separate things in something called SpaceMetadata feels wrong. but then again, they are updated at the same time in UpdateSpaceMetadata as well, so it's not new. I don't know what the metadata is as opposed to tags and what their relation to each other is (could tags be stored as metadata under a "tags" entry?), so it would help to talk to someone who wrote this or read a design document. I'll have a look tomorrow

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -438,8 +438,12 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
...
 												AssetCollectionResultCallback Callback)
 {
-	auto PrototypeInfo
-		= CreatePrototypeDto(AssetCollection.SpaceId, AssetCollection.ParentId, AssetCollection.Name, NewMetadata, AssetCollection.Type, Tags);
+	auto PrototypeInfo = CreatePrototypeDto(AssetCollection.SpaceId,
+											AssetCollection.ParentId,
+											AssetCollection.Name,
+											NewMetadata,
+											AssetCollection.Type,
+											Tags.HasValue() ? Tags : AssetCollection.Tags);
```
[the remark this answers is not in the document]
MAG-OR: done in ac6f3d3006b386a3f496bff7fee3023e0ead9334

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-OR: done in ac6f3d3006b386a3f496bff7fee3023e0ead9334

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-OR: done in ac6f3d3006b386a3f496bff7fee3023e0ead9334

Library/include/CSP/Systems/Spaces/Space.h
MAG-OR: found it. OF-1455

MAG-ElliotMorris: I see yeah, looks to be a different field. Thanks for confirming.

MAG-ElliotMorris (approved):

MAG-mv (approved):
Looks good!

PR #569, opened 2024-12-18
[OB-3643] fix: update callbacks

MAG-mv:
This PR is created as a reference and isn't intended to be the final fix for this issue.

When the multiplayer connection fails, the callback is returned on a separate thread, which is unsupported in wasm builds. This PR ensures that the callbacks are called on the main thread.

It also fixes issues with unbound std::functions being called.

### DIFF
--- a/Library/src/Systems/Users/UserSystem.cpp
+++ b/Library/src/Systems/Users/UserSystem.cpp
@@ -27,6 +27,8 @@
 #include "Systems/Users/Authentication.h"
 
 #include <CallHelpers.h>
+#include <future>
+#include <thread>
 
 
 namespace chs_user		  = csp::services::generated::userservice;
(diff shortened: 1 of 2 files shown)


Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -142,22 +144,35 @@ void UserSystem::Login(const csp::common::String& UserName,
...
-					Callback(LoginStateRes);
+					ConnectPromise.set_value(ErrCode);
 				};
 
 				auto* MultiplayerConnection = SystemsManager::Get().GetMultiplayerConnection();
 				MultiplayerConnection->Connect(ErrorCallback);
+
+				// Ensure this doesn't block to prevent wasm errors
+				std::async(std::launch::deferred,
+						   [&ConnectFuture, Callback, LoginStateRes]
```
MAG-ElliotMorris, line 161: [Comment] My big technical worry with this is twofold.

1. With `std::async` launch modes, I recall when they were first coming out talk about whether or not they were actually executed on a different thread being implementation defined. 
2. std::launch::deferred may still block right? It's lazily evaluated, dosen't it just execute immediately on the calling thread, since we don't attempt to defer requesting its result via a future? I'm actually confused by this, my reading of this code is that the async block does nothing. What am I missing?

MAG-ElliotMorris: My big non-technical worry is, as we've discussed, that this is a very specific solve for the failing multiplayer connection, and I'm not sure it will scale to all callbacks that need to be executed off-thread. I'm willing to grant it might though. The thing I want to avoid is multiple patterns emerging in the complicated threading space, especially when it's non-obvious from a purely c++ perspective why it's being done this way at all!

MAG-ElliotMorris: My big _other_ worry not specific to this implementation is we can't put automated regression tests in for this ... and like ... this is such a prime candidate for needing tests like that, due to the complexity of the space and the low level of observability. 🫤 😖

MAG-mv: The usage of async here is to prevent the callback from firing in that code path. Something that the client is doing in that callback and that point is causing the error about blocking the main thread. By deferring, we are preventing blocking at that point in time.

MAG-ElliotMorris: I'm wondering _how_ though. According to the [docs ](https://en.cppreference.com/w/cpp/thread/launch)for `std::launch::deferred`, 
> the task is executed on the calling thread the first time its result is requested (lazy evaluation)

Normally, I would expect this to be controlled by the future returned from std::async, something like
```cpp
auto future = std:async(std::launch::deferred, callable);
... some intermediate steps
result = future.get(); //The callback is called here.
```

I don't understand what the expectation is here when we just launch a `std::async` as deferred without any controlling future? When does the async block get executed?

PR #570, opened 2024-12-18
[OF-1503] test: Automatically generate test accounts when tests run

MAG-OR:
Use the existing CreateTestUser function to create test user accounts on the fly. Also move said function to the same place as the LogIn functions.

IIRC, @MAG-AdamThorn mentioned that test users are cleaned up automatically and regularly, therefore I have not made use of CleanupTestUser(), because only the super user has the privileges required to do so (see organization tests).

Also, because the super user credentials are required in organization tests (to clean up test users, which we can potentially get rid of, see above, but also to create an organization) and that I am not aware of a method to create such credentials on the fly, we cannot quite get rid of the credentials file just yet.

I have updated all the other C++ tests to generate test accounts as needed. This change excludes the CSharp tests and the Multiplayer Test Client.

Usage: if you need to log in only once and don't need to know the email address, use `LogInAsNewUser()`. If you need to log in, out and back in, or need to know the email address, use `Profile TestUser = CreateTestUser(); LogIn(UserSystem, TestUserId, TestUser.Email, GeneratedTestAccountPassword);`


MAG-ElliotMorris:
> IIRC, @MAG-AdamThorn mentioned that test users are cleaned up automatically and regularly, therefore I have not made use of CleanupTestUser(), because only the super user has the privileges required to do so (see organization tests).

How certain are you both of this? I can see a world where this is true, but the scale of the amount of test users we will make on CHS with this change being unacceptable to them. If you think there's any chance that this might be a surprise to them, we should check in an make sure they're okay with, i dunno, hundreds of throwaway accounts per day? Maybe thousands?

Tests/src/Main.cpp
```diff
@@ -32,7 +32,9 @@ int main(int argc, char* argv[])
 	Listeners.Append(Listener);
 #endif
 
-	LoadTestAccountCredentials();
+#if RUN_ALL_UNIT_TESTS || RUN_ORGANIZATIONSYSTEM_TESTS
+	LoadTestAccountCredentials(); // Needed as long as we cannot create superuser credentials on the fly
+#endif
```
MAG-ElliotMorris, line 37: [Comment] Good idea conditionally compiling this, at least the credentials file can be ignored _mostly_

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -2412,9 +2435,9 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, EnterSpaceAsModeratorTest)
 
 	LogOut(UserSystem);
 
-	LogIn(UserSystem, AltUserId, AlternativeLoginEmail, AlternativeLoginPassword);
+	LogIn(UserSystem, AltUserId, AltUser.Email, GeneratedTestAccountPassword);
 
-	// Note the space is now out of date and does not have the new user in it's lists
+	// Note the space is now out of date and does not have the new user in its lists
```
MAG-ElliotMorris, line 2440: [comment]👍 _nice_

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -75,12 +71,47 @@ void LoadTestAccountCredentials()
...
-	DefaultLoginPassword	 = _DefaultLoginPassword.c_str();
-	AlternativeLoginEmail	 = _AlternativeLoginEmail.c_str();
-	AlternativeLoginPassword = _AlternativeLoginPassword.c_str();
-	SuperUserLoginEmail		 = _SuperUserLoginEmail.c_str();
-	SuperUserLoginPassword	 = _SuperUserLoginPassword.c_str();
+	SuperUserLoginEmail	   = _SuperUserLoginEmail.c_str();
+	SuperUserLoginPassword = _SuperUserLoginPassword.c_str();
+}
+
+
```
MAG-ElliotMorris, line 78: [nit] Double space

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -75,12 +71,47 @@ void LoadTestAccountCredentials()
...
+							  UniqueUserName,
+							  TestDisplayName,
+							  UniqueEmail,
+							  GeneratedTestAccountPassword,
+							  false,
+							  true,
+							  nullptr,
+							  nullptr);
+
+	EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
```
MAG-ElliotMorris, line 106: [Consider] Some human friendly logging would go a long way here, folks won't know that they're baking assertions into their tests when they call `CreateTestUser`, and if this happens to fail, it'll speed up debugging flow a lot if you can see a "Failed to create temporary test user in `CreateTestUser`" message

MAG-ElliotMorris (dismissed):
Awesome! I love codebase wide changes like this, they're not so tough. 

Shame about the superuser thing, please make sure that's captured somewhere, it's not an unsolvable problem, and having a credentials file lying around for _just a few tests_ almost seems more debtful than it being the primary mechanism.

Conditionally approving based on the presumption that we're certain that CHS is okay with us making throwaway accounts at this scale. I'm a little dubious about that tbh, would be good to get a confirmation. Seems to me cleaning up after ourselves would be the preferred solution.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: found SCOPED_TRACE that appears to do just what I want (https://stackoverflow.com/questions/53875175/how-can-i-customize-google-test-failure-messages/53875784). will add.

MAG-OR: Below, an example of the error message for the first call: 
image
And the second call:
image

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: done in 332fe579e7f84a116298d7bca2545c9e4ff2c4f2

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: done in 332fe579e7f84a116298d7bca2545c9e4ff2c4f2. For what it looks like, see above (changed EXPECT_EQ to EXPECT_NE just to trigger the error messages)

MAG-OR:
> > IIRC, @MAG-AdamThorn mentioned that test users are cleaned up automatically and regularly, therefore I have not made use of CleanupTestUser(), because only the super user has the privileges required to do so (see organization tests).
> 
> How certain are you both of this? I can see a world where this is true, but the scale of the amount of test users we will leave dangling on CHS with this change being unacceptable to them. If you think there's any chance that this might be a surprise to them, we should check in an make sure they're okay with, i dunno, hundreds of throwaway accounts per day? Maybe thousands?
> 
> @jeromagnopus pinging you for visibility.

Conversation happening here: https://magnopus.slack.com/archives/C02T07E8P2B/p1736173539223249

MAG-ElliotMorris:
CHS are good with hundreds of test users per run you say? Brilliant! Thanks for verifying. 🙏

MAG-ElliotMorris (dismissed):

MAG-AdamThorn (dismissed):

MAG-ElliotMorris (dismissed):

MAG-mv (approved):

PR #571, opened 2024-12-23
[NT-0] doc: Update example tenant from FOUNDATION_HELLO_WORLD to CSP_HELLO_WORLD

MAG-ElliotMorris:
CHS got in touch asking us to stop using the FOUNDATION_HELLO_WORLD tenant.
I'd been using it to repro a bug, having started from the web example.

The foundation tenant only appears in a few of our examples.

Update to CSP_HELLO_WORLD, which CHS recommends instead now. We shouldn't have the term "Foundation" in our docs anyhow.


MAG-OR (approved):
lgtm

MAG-AdamThorn (approved):

PR #573, opened 2025-01-06
[OF-1519] fix: replicated map

MAG-mv:
* Update custom map to use supported string map
* Added tests to ensure CHS correctly supports this type

### DIFF
--- a/Library/include/CSP/Multiplayer/ReplicatedValue.h
+++ b/Library/include/CSP/Multiplayer/ReplicatedValue.h
@@ -223,7 +223,7 @@ public:
 	/// @return csp::common::Map
 	const csp::common::Map<csp::common::String, ReplicatedValue>& GetStringMap() const;
 
-	/// @brief Set a string Map value for this replicated value from a csp::common::Map, will overwrite and previous value.
+	/// @brief Set a string Map value for this replicated value from a csp::common::Map, will overwrite any previous value.
 	void SetStringMap(const csp::common::Map<csp::common::String, ReplicatedValue>& InValue);
 
 	/// @brief Get a generic default StringMap.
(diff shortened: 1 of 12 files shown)


Library/src/Multiplayer/ReplicatedValue.cpp
```diff
@@ -131,9 +132,9 @@ ReplicatedValue& ReplicatedValue::operator=(const ReplicatedValue& InValue)
 			SetVector4(InValue.GetVector4());
 			break;
 		}
-		case ReplicatedValueType::Map:
```
MAG-OR, line 134: [question] is it worth keeping the old ReplicatedValueType::Map case to throw a specific error message (e.g.: you can no longer use Map, use StringMap instead)? I  know the default case does throw an error, but to specifically say that this has been changed recently (?)

MAG-mv: This was never used outside of video textures previously, so I think it's  redundant keeping it

Library/include/CSP/Multiplayer/ReplicatedValue.h
```diff
@@ -215,19 +215,20 @@ class CSP_API ReplicatedValue
...
 	///
 	/// Use ReplicatedValue::GetReplicatedValueType to ensure type before accessing.
 	///
 	/// @return csp::common::Map
-	const csp::common::Map<ReplicatedValue, ReplicatedValue>& GetMap() const;
+	const csp::common::Map<csp::common::String, ReplicatedValue>& GetStringMap() const;
 
-	/// @brief Set a Map value for this replicated value from a csp::common::Map, will overwrite and previous value.
-	void SetMap(const csp::common::Map<ReplicatedValue, ReplicatedValue>& InValue);
+	/// @brief Set a string Map value for this replicated value from a csp::common::Map, will overwrite and previous value.
```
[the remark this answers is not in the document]
MAG-mv: Fixed in 930077d

MAG-AdamThorn (approved):

PR #574, opened 2025-01-06
[OF-1460] fix: replace client session with factory

MAG-ChristopherAtkinson:
The CHS team have dockerised the Services, which means that a clean instance of them can be spun on a local machine and then teared down after use. Current when pinging the local services the server will return a 503 due to the request being processed as HTTPS, this change resolves that by using the POCO factory to change the client session based on the protocol specified in the URI. 

This can be configured to point to the local services once set up by specifying localhost and port in the  PublicTestBase.h
PublicTestBase.h L21: `constexpr char* EndpointBaseURI = "http://localhost:8081";`


MAG-SamBirley:
Nice work! Could be neat in the future to make the endpoint root URI a command line argument to the tests executable?

MAG-ChristopherAtkinson:
> Nice work! Could be neat in the future to make the endpoint root URI a command line argument to the tests executable?

Yeah I agree, currently this work is blocking another ticket so would be worth ticketing this separately rather than tackling in this scope.

MAG-AdamThorn (approved):

MAG-ElliotMorris (approved):
Oh, this is much less of a hack than I was expecting based on the conversations. Nice one bud.

MAG-ChristopherAtkinson:
> Oh, this is much less of a hack than I was expecting based on the conversations. Nice one bud.

you should have seen the first implementation but some documentation diving later made this a nice simple change

PR #578, opened 2025-01-07
[OF-1441] fix: wasm crash

MAG-mv:
* Fixed wasm crash with blocking the main thread by replacing promise/future with a container of jobs which exists on the asset system to prevent data going out of scope
* Fixed issue with material ids not being set

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -246,7 +246,6 @@ if not Project then
                 "-sALLOW_MEMORY_GROWTH=1",                                      -- we don't know how much memory we'll need, so allow WASM to dynamically allocate more memory
                 "-sINITIAL_MEMORY=33554432",
                 "-sMAXIMUM_MEMORY=1073741824",                                  -- set an upper memory allocation bound to prevent Emscripten from trying to allocate too much memory
-                "-sPROXY_TO_PTHREAD",
                 "-sEXPORTED_RUNTIME_METHODS=[" ..
                     "'ccall'," ..
                     "'setValue'," ..
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -27,6 +27,7 @@
 
 // StringFormat needs to be here due to clashing headers
 #include "CSP/Common/StringFormat.h"
+#include "Common/UUIDGenerator.h"
 
 #include <future>
 #include <thread>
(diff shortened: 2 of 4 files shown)


MAG-AdrianMeredith:
Fix confirmed via local builds on web

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1059,55 +1060,56 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
...
+					}
+				};
 
-			for (size_t i = 0; i < Assets.Size(); ++i)
-			{
 				DownloadAssetData(Assets[i], DownloadMaterialCallback);
 			}
-
-			// We have to wait so this function doesn't exit and destroy "AssetsDownloaded"
-			CallbackFuture.wait();
```
MAG-ElliotMorris, line 1110: [Question] This block was the issue aye? Going by the comment, it doesn't seem like an altogether necessary block, `AssetsDownloaded` dosen't need to be captured by reference, or even captured at all as far as I can see. 
(Even if it was, you can deal with this sort of unclear lifetime with shared ownership). Would doing something like that have solved the issue? 

I suspect you maybe mean space materials? Moving the ownership through the callback and into the result would deal with scope, (it's _the_ reason move types are so essential in callback based systems)

MAG-ChristopherAtkinson (approved):
The changes look good to me, @MAG-ElliotMorris comments did get me thinking, and I do agree that they may be better method to resolve this, however due to the impact of delaying the feature moving forward, I will leave this to your discretion.

Library/premake5.lua
```diff
@@ -246,7 +246,6 @@ if not Project then
                 "-sALLOW_MEMORY_GROWTH=1",                                      -- we don't know how much memory we'll need, so allow WASM to dynamically allocate more memory
                 "-sINITIAL_MEMORY=33554432",
                 "-sMAXIMUM_MEMORY=1073741824",                                  -- set an upper memory allocation bound to prevent Emscripten from trying to allocate too much memory
-                "-sPROXY_TO_PTHREAD",
```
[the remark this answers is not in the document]
MAG-mv: I added this as a test on the base branch. I just removed it here

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: The issue is counting how many have been downloaded using AssetsDownloaded. This variable needed to stay alive until all downloads were complete. If this value was passed in by ref, it would go out of scope

Library/premake5.lua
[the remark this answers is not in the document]
MAG-ElliotMorris: 👍

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1059,55 +1056,58 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
...
+						  const AssetDataResult& DownloadResult)
 				{
-					Callback(GLTFMaterialsResult(DownloadResult.GetResultCode(), DownloadResult.GetHttpResultCode()));
-					return;
-				}
+					if (DownloadResult.GetResultCode() != EResultCode::Success)
+					{
+						if (DownloadResult.GetResultCode() == EResultCode::Failed)
+						{
+							(*AssetsDownloaded)++;
```
MAG-ElliotMorris, line 1076: [Question] This is the equivalent of the `if Deserialized == false` guarded increment in the prior code? (Trying to figure out if this is a refactor or a behaviour change)

MAG-ElliotMorris (approved):
Can't really speak much to the change in implementation of the function itself, don't quite see the reason to change the prior implementation outside of the lifetime issue, but i'll trust that it's a pure refactor and that you prefer how it is now.

edit: Oh wait I see in the description there was an issue with material Id's, missed that.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: Resolving in favour of new changeset, we talked offline about this and moved towards the shared-ownership thing that's currently there.

MAG-ChristopherAtkinson (approved):
The recent changes look good, approving but would be great to have @MAG-ElliotMorris comments resolved before commit.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Its an addition. If one of the materials fails, we will never get the final callback

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1060,8 +1056,8 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
 				return;
 			}
 
-			csp::common::String JobUUID				 = GenerateUUID().c_str();
-			DownloadMaterialsJobs[JobUUID].Materials = csp::common::Array<GLTFMaterial>(AssetsToDownload);
+			auto DownloadedMaterials = std::make_shared<csp::common::Array<GLTFMaterial>>(AssetsToDownload);
+			auto AssetsDownloaded	 = std::make_shared<size_t>();
```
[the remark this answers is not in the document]
MAG-mv: I'll add this on the main feature branch to prevent this PR being held up

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: Ah I see, makes sense I suppose.

PR #581, opened 2025-01-09
[NT-0] doc: updated stack graphic

MAG-SamBirley:
Updating the apps/CSP/services stack graphic with a new visual treatment.

I'd like to replace this png with the svg equivalent, but the svg uses the proxima nova font, which the site doesn't currently support. I didn't want to go down the yak shaving route for updating just this graphic, so for now I have elected for a simple update of the png.


MAG-OR (approved):
Looks great!

PR #582, opened 2025-01-09
[OF-1441] fix: elegantly fail when error

MAG-mv:
* Ensures AssetSystem::GetMarerials gracefully exits when an error occurs

### DIFF
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -1057,6 +1057,7 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
 			}
 
 			// These are shared references to prevent going out of scope between callbacks
+			// Note: The callbacks ARE called on the main thread
 			auto DownloadedMaterials = std::make_shared<csp::common::Array<GLTFMaterial>>(AssetsToDownload);
 			auto AssetsDownloaded	 = std::make_shared<size_t>();
 			auto Failed				 = std::make_shared<bool>();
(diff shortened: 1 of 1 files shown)


MAG-AdamThorn (approved):

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1066,14 +1068,20 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
...
+					{
+						return;
+					}
+
 					if (DownloadResult.GetResultCode() != EResultCode::Success)
 					{
 						if (DownloadResult.GetResultCode() == EResultCode::Failed)
 						{
-							(*AssetsDownloaded)++;
+							*Failed = true;
```
MAG-ElliotMorris, line 1084: [Consider] You're certain these arn't vulnerable to race conditions? Smells fishy to me.

MAG-mv: They always come back on the main thread. The HTTP request system polls for responses on the main thread

MAG-ElliotMorris: Do you think it's a good idea to notate that here as a comment? I do. I'd love to notate thread expectations in every callback, but in this case because of the shared ownership, I think it's extra important to the future reader.

MAG-mv: I'll add a comment

MAG-mv: Fixed in 0bbd3bfb59e9beda7069d799f811d5d91bc54ffe

MAG-ElliotMorris: Thanks :)

MAG-ElliotMorris (approved):

PR #585, opened 2025-01-13
[OF-1527] Allow test endpoint to be configurable

MAG-ChristopherAtkinson:
Initial the ticket was to expose the end point to the cli, however due to complication with google test and workflow this PR has gone for the approach using environment variable to the same effect


Tests/src/PublicTestBase.h
```diff
@@ -17,8 +17,8 @@
 
 #include <gtest/gtest.h>
 
-
-constexpr char* EndpointBaseURI = "https://ogs-internal.magnopus-dev.cloud";
+constexpr char* EndpointEnvironmentName = "MAGNOPUS_SERVICES_ENDPOINT";
+extern const char* EndpointBaseURI;
```
MAG-ElliotMorris, line 21: [Consider] I wonder if this is vulnerable to the static-initialization-order-fiasco. Would we be more comfortable with something like
```suggestion
const char* GetEndpointBaseURI() {
  return GetEnvironmentVariableOrDefault(EndpointEnvironmentName, "https://ogs-internal.magnopus-dev.cloud");
}
```

MAG-ElliotMorris (commented):
During my testing, I found that environment variables set from the ENVIRONMENT field in the VS test settings panel were not picked up when running tests individually from the test runner.

I suspect you've tested via setting a variable in your global environment. I don't really think it matters too much if devs are unable to use the environment field in this case, (because devs are just as able to edit the source code if they need to test a different endpoint,) but wanted to flag it in case you weren't aware. If global environment vars work both ways, i'm cool with it.

MAG-ChristopherAtkinson:
> During my testing, I found that environment variables set from the ENVIRONMENT field in the VS test settings panel were not picked up when running tests individually from the test runner.
> 
> I suspect you've tested via setting a variable in your global environment. I don't really think it matters too much if devs are unable to use the environment field in this case, (because devs are just as able to edit the source code if they need to test a different endpoint,) but wanted to flag it in case you weren't aware. If global environment vars work both ways, i'm cool with it.

That is odd I have tested through the main enviroment variables but can check that as well, looking at the project setting the this should be supported 

image

Tests/src/PublicTestBase.h
MAG-ChristopherAtkinson: Yeah this would be a lot nicer in general, however was attempting to keep the changes limited and have as few files as possible be effect in the change.

MAG-ChristopherAtkinson: As for static-initialization-order this shouldn't be a problem due to the use of extern deferring the definition and directly defined in the cpp file. But of course would be good to check.

MAG-ElliotMorris: Granted currently, we are likely safe by coincidence, this sort of approach is still vulnerable.

Take this example : https://onlinegdb.com/VzZqSFk82

In that example, two global variables are defined in this manner, one depending on the other. Depending on the order of compilation (undefined), the variable that depends on the prior one may be garbage, (is is during my execution.)

It's a pretty bad idea to leave these sorts of footguns in place when we can avoid them.

[Cpp core standards section.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i22-avoid-complex-initialization-of-global-objects)

MAG-AdamThorn (approved):

Tests/src/PublicTestBase.h
MAG-ChristopherAtkinson: should be resolved as part of recent commits, went with static initialization ensure we don't run into these pit falls, as you mentioned it would be best resolved these changes before they become a potential problem down the line

MAG-ElliotMorris: Thanks for doing that.

Tests/src/PublicTestBase.h
```diff
@@ -18,7 +18,21 @@
...
+constexpr char* EndpointEnvironmentName = "MAGNOPUS_SERVICES_ENDPOINT";
+
+const char* GetEnvironmentVariableOrDefault(const char* EnvironmentKey, const char* DefaultValue)
+{
+	const auto EnvironmentVariable = std::getenv(EnvironmentKey);
+	return (EnvironmentVariable) ? EnvironmentVariable : DefaultValue;
+}
+} // namespace
+
+static const char* EndpointBaseURI()
```
MAG-ElliotMorris, line 32: [fix] I think you probably mean `inline` here. 

The difference between a static and inline method in a global context is subtle, they both achieve the same thing but in different ways. 

- `static` duplicates the function into each translation unit this header is included in, hence preventing symbol collisions that way.
- `inline` tells the linker that it's fine if there are duplicate symbols so as not to violate the ODR, but still enforces only one implementation of the function, (chosen effectively at random)

You can even verify this behaviour by taking the address of the function in different translation units and seeing that with `static` they will be different but with `inline` they will be the same.

Long story short, using static over inline here will inflate binary size for no reason.

MAG-ElliotMorris (changes requested):

Tests/src/PublicTestBase.h
MAG-ChristopherAtkinson: resolved in latest commit, and yeah I agree that inline is the correct use here, thank you for catching that

MAG-ElliotMorris (approved):

PR #589, opened 2025-01-16
[OB-3785] fix: only log to console and file if there is no log callback

MAG-OR:
Only log to the console and the log file if there is no log callback set by a client.

### DIFF
--- a/Library/src/Systems/Log/LogSystem.cpp
+++ b/Library/src/Systems/Log/LogSystem.cpp
@@ -70,21 +70,23 @@ void LogSystem::LogMsg(const csp::systems::LogLevel Level, const csp::common::St
         return;
     }
 
-#if defined(CSP_WASM)
-    printf("%s\n", InMessage.c_str());
-#endif
-
-#if defined(CSP_ANDROID)
-    __android_log_print(ANDROID_LOG_VERBOSE, "CSP", InMessage.c_str());
-#endif
-
-    // Log to our Connected Spaces Platform file system.
-    LogToFile(InMessage);
-
     if (Callbacks->LogCallback != nullptr)
     {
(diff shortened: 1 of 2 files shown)


Library/src/Systems/Log/LogSystem.cpp
```diff
@@ -70,21 +70,23 @@ void LogSystem::LogMsg(const csp::systems::LogLevel Level, const csp::common::St
...
 
-    // Log to our Connected Spaces Platform file system.
-    LogToFile(InMessage);
-
-    if (Callbacks->LogCallback != nullptr)
-    {
-        // Send message to clients to display the log on the client side.
-        Callbacks->LogCallback(InMessage);
+        // Log to our Connected Spaces Platform file system.
+        LogToFile(InMessage);
```
MAG-ElliotMorris, line 89: [Consider] I don't know the answer to this, but I wonder if (suspect so really) we should LogToFile all the time, to keep the file logs complete. Conditionalizing the printf` and `android_log_print` is still correct though.

MAG-OR: I see your point, and I was on the fence. @MAG-SamBirley what do you think?

MAG-OR: Sam said he's OK as long as we've considered it. I think it is best to keep the file logs complete, so moving it back out

MAG-OR: changed in b2480cc71ce7dd34c717c467668371a977be87c7

MAG-AdamThorn (approved):

MAG-ElliotMorris (approved):

PR #593, opened 2025-01-20
[OF-1494] refac!: Remove unused serialization vector methods

MAG-ElliotMorris:
These methods are not at all called internally or in tests, and we confirmed with clients that they're not used downstream :

https://magnopus.slack.com/archives/C037VC209QT/p1737118085667259

Prior to the most recent change, these wern't even calling the correct overload to store vector data, so I'm pretty confident these can be ripped out.


MAG-OR (approved):
You provided evidence that these are not used, so I am happy

MAG-AdamThorn (approved):

PR #594, opened 2025-01-21
[OF-1494] fix: Fix SpaceTransform multiplication

MAG-ElliotMorris:
(This issue found during the warnings as errors sweep, i'm fixing it now, it's the final fix)

The existing implementation had several things wrong with it, one has to imagine it was never called anywhere, because this would be so apparent. Unfortunately, it was untested since the beginning.

Issues :
- [[nodiscard]] warning on normalize ignored, although post-rotation normalization is only needed to prevent cascading floating point errors
- The W component for glm::quat is the first argument, but for our vectors it is the last, all our quaternions were being constructed with nonsense values. (Strong typing please! Vector4 is inappropriate to store quaternions for precisely this reason)
- I don't think you can do complex compositions in this manner, separating the translation, scale and rotation components like that, but I could be wrong there.
- Untested

The prior code, even with the code issues fixed, failed to pass even the identity assumptions in the new test suite that has been written. For that reason, and due to the face that matrix compositions are more comprehensible, (at least to me, I grant that they're somewhat verbose) I've replaced the implementation with a matrix based one. It's probably slightly slower than direct quaternion work, but at least it has a chance of being correct.

In a perfect world, one would have gone further here, fixing the `Vector4` `Quaternion` mismatch issue, and establishing normalization/negative scale preconditions, but changes like that would have been breaking, which I didn't want to do. (Note, the data members of this type are public, making type-level invariants impossible, alas)

### DIFF
--- a/Library/src/Multiplayer/SpaceTransform.cpp
+++ b/Library/src/Multiplayer/SpaceTransform.cpp
@@ -17,7 +17,6 @@
 #include "CSP/Multiplayer/SpaceTransform.h"
 
 #define GLM_ENABLE_EXPERIMENTAL // glm::decompose is still technically an experimental feature, but better than a hand-rolled solution in my opinion.
-#include "glm/gtx/string_cast.hpp"
 #include <glm/gtc/quaternion.hpp>
 #include <glm/gtx/matrix_decompose.hpp>
 #include <glm/gtx/quaternion.hpp>
--- a/Library/src/Multiplayer/SpaceTransform.cpp
+++ b/Library/src/Multiplayer/SpaceTransform.cpp
@@ -16,7 +16,12 @@
 
 #include "CSP/Multiplayer/SpaceTransform.h"
 
+#define GLM_ENABLE_EXPERIMENTAL // glm::decompose is still technically an experimental feature, but better than a hand-rolled solution in my opinion.
+#include "glm/gtx/string_cast.hpp"
 #include <glm/gtc/quaternion.hpp>
+#include <glm/gtx/matrix_decompose.hpp>
+#include <glm/gtx/quaternion.hpp>
+#include <tuple>
 
 csp::multiplayer::SpaceTransform::SpaceTransform()
     : Position(0.0f, 0.0f, 0.0f)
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include "CSP/Multiplayer/SpaceTransform.h"
+#include "TestHelpers.h"
+
+#include <glm/glm.hpp>
+#include <glm/gtc/quaternion.hpp>
+
+#include "gtest/gtest.h"
+
+using namespace csp::common;
```
MAG-ElliotMorris, line 25: [fix] Unnecesary I think, probably a copy paste from the old file

Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    const auto Output = Identity * Transformation;
+    EXPECT_EQ(Output.Position, (csp::common::Vector3 { 1.0f, 0.0f, 1.0f }));
+
+    glm::vec3 Euler = glm::eulerAngles(glm::normalize(glm::quat { Output.Rotation.W, Output.Rotation.X, Output.Rotation.Y, Output.Rotation.Z }));
+
+    const float epsilon = 1e-5f; // Matrix operations involving euler conversions arn't terribly stable at the floating point level cross platform.
+    EXPECT_NEAR(Euler.x, 1.57080138f, epsilon); // 90 degrees
+    EXPECT_NEAR(Euler.y, 0.785398126f, epsilon); // 45 degrees
+    EXPECT_NEAR(Euler.z, 0.785398126f, epsilon); // 45 degrees
```
MAG-ElliotMorris, line 142: [Consider] Check if you took these from live output or not, I'm not sure these are exactly 90 & 45 when converted from radians. Would be better to use the exact values.

Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // Apply a complicated transform to the identity.
+    const csp::multiplayer::SpaceTransform Transformation { { 1.0f, 0.0f, 1.0f }, NinetyDegAroundXYAxisNonNormalizedQuat, { 2.0f, 1.0f, 4.0f } };
+
+    const auto Output = Identity * Transformation;
+    EXPECT_EQ(Output.Position, (csp::common::Vector3 { 1.0f, 0.0f, 1.0f }));
+
+    glm::vec3 Euler = glm::eulerAngles(glm::normalize(glm::quat { Output.Rotation.W, Output.Rotation.X, Output.Rotation.Y, Output.Rotation.Z }));
+
+    const float epsilon = 1e-5f; // Matrix operations involving euler conversions arn't terribly stable at the floating point level cross platform.
+    EXPECT_NEAR(Euler.x, 1.57080138f, epsilon); // 90 degrees
```
MAG-OR, line 140: [consider] might be more obvious to the reader if you wrote this as (value in deg * DEG_TO_RAD), where DEG_TO_RAD is a constant with the appropriate unit conversion

MAG-OR (approved):
thank you for the comprehensive test cases

Library/src/Multiplayer/SpaceTransform.cpp
```diff
@@ -16,7 +16,12 @@
 
 #include "CSP/Multiplayer/SpaceTransform.h"
 
+#define GLM_ENABLE_EXPERIMENTAL // glm::decompose is still technically an experimental feature, but better than a hand-rolled solution in my opinion.
+#include "glm/gtx/string_cast.hpp"
```
MAG-ElliotMorris, line 20: [fix] Unnecesary, you're not doing print debugging anymore

Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+    EXPECT_NE(SpaceTransform1, SpaceTransformIdentity);
+}
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_SPACETRANSFORM_TESTS || RUN_SPACETRANSFORM_MULTIPLICATION_IDENTITY_AXIOM_TEST
+CSP_PUBLIC_TEST(CSPEngine, SpaceTransformTests, MulitplicationIdentityAxiomTest)
+{
+    // Two identity matrices should be identity
+    const csp::multiplayer::SpaceTransform Identity;
+    const auto result = Identity * Identity;
```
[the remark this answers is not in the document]
MAG-ElliotMorris: Oh aye yeah, will [fix]

Tests/src/PublicAPITests/SpaceTransformTests.cpp
MAG-OR: note that you can use M_PI from math.h to do the conversion

MAG-SamBirley:
OKO client teams may rely on this for their entity hierarchy features, so worth checking in with them to see if this has an impact (could be a positive impact!). 

I know at least that OKO Unreal doesn't rely on the SpaceTransform `*` operator, so you don't have to worry about that one.

MAG-ElliotMorris:
@MAG-SamBirley I'm almost certain that they arn't, because the nonsense quaternions would make any transformations so obviously incorrect that it'd have been spotted. I'll post nonetheless though, I've been wrong before.

Library/src/Multiplayer/SpaceTransform.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/71ef3f4dae62398b2d78523bce3244bfe71bfcec

Tests/src/PublicAPITests/SpaceTransformTests.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/71ef3f4dae62398b2d78523bce3244bfe71bfcec

Tests/src/PublicAPITests/SpaceTransformTests.cpp
MAG-ElliotMorris: Subsumed by @MAG-OR 's comment.

Tests/src/PublicAPITests/SpaceTransformTests.cpp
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/bf199ba02c05997fe7f190e102e516c47ebf1d9e

Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+    EXPECT_EQ((Identity * Scaled), Scaled);
+    EXPECT_EQ((Scaled * Identity), Scaled);
+}
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_SPACETRANSFORM_TESTS || RUN_SPACETRANSFORM_MULTIPLICATION_TRS_TEST
+CSP_PUBLIC_TEST(CSPEngine, SpaceTransformTests, MulitplicationTRSTest)
+{
+    const csp::multiplayer::SpaceTransform Identity;
+    // An angle-axis rotation equivilent to a 90 degree rotation along the (1,1,0) axis.
```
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/f08c2890d9cfcca8546700d20172a17ef3b89667

Tests/src/PublicAPITests/SpaceTransformTests.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/c83fad887ff265ee3c73a0de61d2e45f040493a5

.. I realize i've left the now pretty redundant comments ... but they're not doing any harm.

Tests/src/PublicAPITests/SpaceTransformTests.cpp
```diff
@@ -0,0 +1,146 @@
...
+    const auto result = Identity * Identity;
+    EXPECT_EQ((Identity * Identity), Identity);
+}
+#endif
+
+#if RUN_ALL_UNIT_TESTS || RUN_SPACETRANSFORM_TESTS || RUN_SPACETRANSFORM_MULTIPLICATION_IDENTITY_TRANSFORM_TEST
+CSP_PUBLIC_TEST(CSPEngine, SpaceTransformTests, MulitplicationIdentityTransformTest)
+{
+    const csp::multiplayer::SpaceTransform Identity;
+    const csp::common::Vector4 NinetyDegAroundXNormalizedQuat { 0.7071068f, 0.0f, 0.0f, 0.7071068f };
```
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/594/commits/ce993bd2432bf6b821f8e220ecb3bd4012be00a0

Put it in each function out of a general aversion to cross-test fixtures when it can be avoided.

PR #596, opened 2025-01-23
[OB-4013] Unable to add hotspots to an existing hotspot group

MAG-ChristopherAtkinson:
Some background information regarding this issue, this was initial introduced as part of https://github.com/magnopus-opensource/connected-spaces-platform/pull/549 

The endpoint in question was returning success on the test because a creation is still valid but functionally it failed to find the hotspot. 

~~However this now causes a new problem, while testing I was seeing duplicate hotspot groups when using an older space, it likely we will see this issue on existing spaces and was due to unexpected creation from the endpoint. We are likely to want to resolve this behaviour as part of this ticket if possible, would be interested if there are any suggestions? I am planning to added step to reproduce and a video demonstrating the issue.~~

[Update] While trying to create a video to demonstrate the issues, it seems to have vanished into the ether, I am going to assume was likely due to the cache be invalidated either on the web client or local web client. anyway I will add a note to the ticket to ensure that this is check when going through QA.


MAG-OR:
[question] is there a way to update the test that failed to catch the problem, so that it catches the problem?

Library/src/Systems/Sequence/SequenceSystem.cpp
```diff
@@ -107,7 +107,7 @@ void SequenceSystem::UpdateSequence(const String& SequenceKey, const String& Ref
         = SequenceAPI->CreateHandler<SequenceResultCallback, SequenceResult, void, chs::SequenceDto>(Callback, nullptr);
 
     static_cast<chs::SequenceApi*>(SequenceAPI)
-        ->apiV1SequencesPut(SequenceKey, // NewKey
+        ->apiV1SequencesPut(csp::common::Encode::URI(SequenceKey), // NewKey
```
MAG-ChristopherAtkinson, line 110: We current do not have clear consideration for what should be encoded when interacting with CHS, this was likely missed as part of the initial PR due to the endpoint still return a valid 200 response however, the target of the update was incorrect. Encoding the key will now identify the correct target and return a valid 200 response for a sufficient update. 

I tested locally with the changes and am no longer seeing issues but adding additional information here for clarity.

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #597, opened 2025-01-23
[NT-0] fix: Disable erroring warnings on android builds

MAG-ElliotMorris:
Looks like android builds arn't ran on the CI, so this was missed during warnings as errors sweep. Another reason to make sure we're doing all our checks (builds at the very least!) before merge.

Had to suppress a fair amount of legitimate warnings to get this through quickly, there's probably 1 or 2 days worth of effort to actually squash them, which we should spend. I'll make sure to capture it.

Because of the rush, havn't performed my usual level of commit isolation. Please do actually look through the changeset to ensure there's no danger. I don't think there's any risk, but we all know how blind implementors get to their own changes.

Android run : https://uk-teamcity.magnopus.global/buildConfiguration/Olympus_OlympusFoundation_Android_ReleaseDLL_2/250409?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true
(Although I think this will appear in the checks panel now that it's been run explicitly against this PR)


Library/src/Multiplayer/Election/ClientProxy.h
```diff
@@ -115,13 +115,9 @@ class ClientProxy
     ClientId HighestResponseId;
     EventId Eid;
 
-    ElectionEvent PendingEvent;
```
MAG-AdamThorn, line 118: Is this event not still required? I couldn't see that it had been moved elsewhere.

MAG-ElliotMorris: It's never referenced at all, so has no effect.

MAG-ChristopherAtkinson (approved):
I appreciate the rush on this, approving for now as the changes look good given the context.

However, as you mentioned there are some areas we should revisit, my largest concern out of these would be the pragma directive, it raises more questions then answers. as well as the lambda capture, can look into these further at a later date

MAG-ElliotMorris:
The lambda capture should just be dealt with, it's bad practice all over the codebase.
The pragma suppression is actually fine, it's all thirdparty signalR. Having to put it in the main config is annoying, a premake issue similar to how we have to do explicit suppressions for xcode. It's already suppressed in the signalR premake, but warnings as errors (incorrectly) overrides the suppression at a project level.

MAG-AdamThorn (approved):

PR #598, opened 2025-01-23
[OF-1385] feat: Annotations number of replies

MAG-mv:
- Added AssetSystem::GetAssetCollectionCount
- Added ConversationSpaceComponent::GetNumberOfReplies
- Added tests to verify new functions


Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -156,6 +156,22 @@ class CSP_API AssetSystem : public SystemBase
 														const csp::common::Optional<csp::common::Array<csp::common::String>>& Tags,
 														AssetCollectionResultCallback Callback);
 
+	/// @brief Retrieves the number asset collections based on the specified search criteria.
+	/// @param ParentId csp::common::String : optional asset collection parent id to get asset collections associated with it
+	/// @param Names csp::common::Optional<csp::common::Array<csp::common::String>> : optional array of strings representing asset
```
MAG-AdamThorn, line 161: [Question] Should the Name property description say "optional array of Asset Collection names".

MAG-AdamThorn (approved):

Library/include/CSP/Multiplayer/Components/ConversationSpaceComponent.h
```diff
@@ -52,6 +52,42 @@ enum class ConversationPropertyKeys
...
+/// @brief Data class used to contain information for GetNumberOfReplies.
+class CSP_API NumberOfRepliesResult : public csp::systems::ResultBase
+{
+	/** @cond DO_NOT_DOCUMENT */
+	friend class ConversationSpaceComponent;
+
+	CSP_START_IGNORE
+	template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+	CSP_END_IGNORE
+	/** @endcond */
```
MAG-ElliotMorris, line 65: [Comment] My instinct is to make a case to find a way to do this without friendship, because friend is one of the most nightmarish things to work around during any refactoring. However, I see this is the common pattern throughout the codebase.

Library/include/CSP/Systems/Assets/AssetCollection.h
```diff
@@ -189,7 +189,34 @@ class CSP_API AssetCollectionsResult : public csp::systems::ResultBase
 	uint64_t ResultTotalCount = 0;
 };
 
+/// @ingroup Asset System
+class CSP_API AssetCollectionCountResult : public csp::systems::ResultBase
```
MAG-ElliotMorris, line 193: [Question] Perhaps a theme for the future, you think there's any legs in centralizing these result objects? The two here seem like they could both be `CountResult` or something similar.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -731,10 +731,21 @@ const csp::common::Vector3& ConversationSpaceComponent::GetConversationCameraPos
...
 }
 
-const int64_t ConversationSpaceComponent::GetNumberOfReplies() const
+void ConversationSpaceComponent::GetNumberOfReplies(NumberOfRepliesResultCallback Callback)
 {
-	// TODO: Implement getNumberOfReplies - OF-1385
-	return 0;
+	auto* AssetSystem = csp::systems::SystemsManager::Get().GetAssetSystem();
+
+	auto GetMessageCountCallback = [Callback, this](const csp::systems::AssetCollectionCountResult& GetMessageResult)
```
MAG-ElliotMorris, line 738: [consider] Unnecesary `this` capture?

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -731,10 +731,21 @@ const csp::common::Vector3& ConversationSpaceComponent::GetConversationCameraPos
...
+	{
+		NumberOfRepliesResult Result(GetMessageResult);
+		Result.SetCount(GetMessageResult.GetCount());
+		Callback(Result);
+	};
+
+	auto ConversationId													  = GetConversationId();
```
MAG-ElliotMorris, line 745: [nit] `const auto`

Library/src/Systems/Assets/AssetCollection.cpp
```diff
@@ -112,10 +112,10 @@ void PrototypeDtoToAssetCollection(const chs::PrototypeDto& Dto, csp::systems::A
 		AssetCollection.IsUnique = Dto.GetHighlander();
 	}
 
-    if (Dto.HasOrganizationId())
+	if (Dto.HasOrganizationId())
 	{
 		AssetCollection.OrganizationId = Dto.GetOrganizationId();
-    }
+	}
```
MAG-ElliotMorris, line 118: [fix] Your editor is still set to put tabs in, or this file is old and needs to be reformatted.

Library/src/Systems/Assets/AssetCollection.cpp
```diff
@@ -266,4 +266,20 @@ void AssetCollectionsResult::FillResultTotalCount(const csp::common::String& Jso
 		}
 	}
 }
+uint64_t AssetCollectionCountResult::GetCount() const
+{
+	return Count;
```
MAG-ElliotMorris, line 271: [fix] Tabs here as well, in fact probably in all your new code. Be sure to run a reformat. Transitional teething issues, such a bummer.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -488,6 +487,63 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
...
+										  const csp::common::Optional<csp::common::Array<EAssetCollectionType>>& Types,
+										  const csp::common::Optional<csp::common::Array<csp::common::String>>& Tags,
+										  const csp::common::Optional<csp::common::Array<csp::common::String>>& SpaceIds,
+										  csp::systems::AssetCollectionCountResultCallback Callback)
+{
+	std::optional<std::vector<String>> PrototypeIds	  = Convert(Ids);
+	std::optional<String> ParentPrototypeId			  = Convert(ParentId);
+	std::optional<std::vector<String>> PrototypeNames = Convert(Names);
+
+
```
MAG-ElliotMorris, line 502: [nit] Double newline

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -488,6 +487,63 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
...
 
+void AssetSystem::GetAssetCollectionCount(const csp::common::Optional<csp::common::Array<csp::common::String>>& Ids,
+										  const csp::common::Optional<csp::common::String>& ParentId,
+										  const csp::common::Optional<csp::common::Array<csp::common::String>>& Names,
+										  const csp::common::Optional<csp::common::Array<EAssetCollectionType>>& Types,
+										  const csp::common::Optional<csp::common::Array<csp::common::String>>& Tags,
+										  const csp::common::Optional<csp::common::Array<csp::common::String>>& SpaceIds,
+										  csp::systems::AssetCollectionCountResultCallback Callback)
+{
+	std::optional<std::vector<String>> PrototypeIds	  = Convert(Ids);
```
MAG-ElliotMorris, line 498: [comment] `Convert` is a funny function. Conversion operators on the interop types would be much less opaque. Ach well, for the future.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -488,6 +487,63 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
...
+	if (Types.HasValue())
+	{
+		std::vector<String> Vals;
+
+		for (size_t i = 0; i < Types->Size(); ++i)
+		{
+			Vals.push_back(ConvertAssetCollectionTypeToString(Types->operator[](i)));
+		}
+
+		PrototypeTypes = std::move(Vals);
```
MAG-ElliotMorris, line 514: [Consider] This whole block could be :
```suggestion
		std::transform(Types.value.cbegin(), Types.value.cend(), std::back_inserter(PrototypeTypes.value), [](const EAssetCollectionType& type) { return ConvertAssetCollectionTypeToString(type); });
```

Although ... bet this dosen't work because our interop types don't have properly conforming iterators. :(

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -488,6 +487,63 @@ void AssetSystem::UpdateAssetCollectionMetadata(const AssetCollection& AssetColl
...
+
+	std::optional<std::vector<String>> PrototypeTypes;
+
+	if (Types.HasValue())
+	{
+		std::vector<String> Vals;
+
+		for (size_t i = 0; i < Types->Size(); ++i)
+		{
+			Vals.push_back(ConvertAssetCollectionTypeToString(Types->operator[](i)));
```
MAG-ElliotMorris, line 511: [Consider] Why not push directly into `PrototypeTypes`?

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -296,6 +296,23 @@ void UpdateAssetCollectionMetadata(csp::systems::AssetSystem* AssetSystem,
...
+							 const csp::common::Optional<csp::common::Array<csp::common::String>>& Tags,
+							 const csp::common::Optional<csp::common::Array<csp::common::String>>& SpaceIds,
+							 uint64_t& Count)
+{
+
+	auto [Result] = AWAIT_PRE(AssetSystem, GetAssetCollectionCount, RequestPredicate, Ids, ParentId, Names, Types, Tags, SpaceIds);
+
+	EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+	Count = Result.GetCount();
```
MAG-ElliotMorris, line 313: [nit] Does this need so many newlines?

Library/include/CSP/Systems/Assets/AssetCollection.h
MAG-mv: One of them contains request deserialization logic. I agree we could change this in some way, though

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-mv: Fixed in 299ce784b8a2698cb8d825c424ceca4fd679b5bb

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Agreed. I would like to change these functions

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Fixed in b758af7857e056069a9528f918c2ed6eafb827c0

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: The array is optional, so would require a dereference each loop

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-mv: YES

Library/src/Systems/Assets/AssetCollection.cpp
MAG-mv: I believe this was because I hadnt pulled main into this branch yet

Library/include/CSP/Multiplayer/Components/ConversationSpaceComponent.h
MAG-mv: Yep. I would love that

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Yep :(

MAG-mv: We can update this with your recent change eventually

MAG-OR:
[Consider] I only know this because of OB-3917. Now that this has changed, do we still need to expose number of replies as a script property? (i.e. do we need to keep [this](https://github.com/magnopus-opensource/connected-spaces-platform/blob/39f4b32ff351a63ca2d3d202a1d200203d620ef5/Library/src/Multiplayer/Script/ComponentBinding/ConversationSpaceComponentScriptInterface.cpp#L36) and [that](https://github.com/magnopus-opensource/connected-spaces-platform/blob/39f4b32ff351a63ca2d3d202a1d200203d620ef5/Library/src/Multiplayer/Script/ComponentBinding/ConversationSpaceComponentScriptInterface.h#L41) and associated property setting code?)

MAG-mv:
@MAG-OR These were taken out in another pr on my annotations branch

Library/src/Systems/Assets/AssetCollection.cpp
MAG-ElliotMorris: Assuming this is fixed, i know you can't link rebase stuff to show this, if it was from not having merged yet.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: Oh the iterator change, yeah would be nice. (Resolve yourself if you're not planning to address this now)

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: I don't think this matters really, but quick-bench isn't working and I it's not worth going to any more effort than that to prove it. 🤷

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-ElliotMorris: FINE THEN

PR #599, opened 2025-01-23
[OB-4047] fix: call connection callback on connection established

MAG-OR:
- Call the connection callback when the connection succeeds
- Add a test to make sure the connection callback is called
- Remove the need for the connection to be established when registering system callbacks
- Add a test to make sure the callbacks are set successfully even before the connection is established
- Based on feedback from the Web team, update the callback ListenNetworkEvent function to let callbacks be overwritten
without having to deregister first. The System Callbacks can already be overwritten without deregistering
first (via the SetCallbackXXX function), but the System itself (in the System ListenNetworkEvent function) cannot.
- Make other minor improvements related to logging errors and unneeded macros


MAG-AdamThorn (approved):

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -42,33 +42,33 @@ EventBus::EventBus(MultiplayerConnection* InMultiplayerConnection)
...
-        std::string ErrorMessage = "Error: Multiplayer connection is not available.";
-        CSP_LOG_ERROR_FORMAT("%s\n", ErrorMessage.c_str());
-        return;
-    }
-
     if (!System)
     {
         std::string ErrorMessage = "Error: Expected non-null system.";
-        CSP_LOG_ERROR_FORMAT("%s\n", ErrorMessage.c_str());
+        CSP_LOG_ERROR_FORMAT("%s", ErrorMessage.c_str());
```
MAG-ElliotMorris, line 48: [comment] It's the most trivial example in the world, but that's possibly why it's a good one.
I hope we eventually get to the point where even this is considered a change worth capturing in release notes. The more I think about it, the more autogenerating these notes seem innapropriate, otherwise you'd have to make new PR's for little stuff like this which isn't practicable. 

I suppose the other route there is enforcing better commit isolation and not squashing everything, treat the commits as the release notes, but i'm less sure that'll scale.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -42,33 +42,33 @@ EventBus::EventBus(MultiplayerConnection* InMultiplayerConnection)
...
         }
     }
 
     if (!SystemsNetworkEventMap.empty() && SystemsNetworkEventMap.find(EventName) != SystemsNetworkEventMap.end()
         && SystemsNetworkEventMap[EventName])
     {
-        CSP_LOG_ERROR_FORMAT("Error: there is already a system registered for %s.\n", EventName.c_str());
+        if (SystemsNetworkEventMap[EventName] != System)
+        {
+            CSP_LOG_ERROR_FORMAT("Error: there is already a system registered for %s. Deregister it first.", EventName.c_str());
```
MAG-ElliotMorris, line 66: [Question] Deregister it first before what? This is in `ListenNetworkEvent`, does this mean Deregister it before listening to network events? Struggling to follow the logic of the conditionals around this. Can you tell me what falling into each of these blocks means?

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -77,36 +77,30 @@ void EventBus::ListenNetworkEvent(const csp::common::String& EventName, csp::sys
...
 {
-    if (MultiplayerConnectionInst->Connection == nullptr || !MultiplayerConnectionInst->Connected)
+    if (!Callback)
     {
-        std::string ErrorMessage = "Error: Multiplayer connection is not available.";
-        CSP_LOG_ERROR_FORMAT("%s\n", ErrorMessage.c_str());
+        std::string ErrorMessage = "Error: Expected non-null callback.";
+        CSP_LOG_ERROR_FORMAT("%s", ErrorMessage.c_str());
         return;
     }
```
MAG-ElliotMorris, line 85: [Question] How confident are you that this multiplayer connection isn't actually required for some other unknown reason at this point? This feels like an unmanaged untested change.

[Comment] This dosen't mean I'm against taking it out, I understand if we have to take the risk here, but it makes me nervous.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -77,36 +77,30 @@ void EventBus::ListenNetworkEvent(const csp::common::String& EventName, csp::sys
...
 
-    if (!Callback)
+    if (!SystemsNetworkEventMap.empty() && SystemsNetworkEventMap.find(EventName) != SystemsNetworkEventMap.end()
+        && SystemsNetworkEventMap[EventName])
     {
-        std::string ErrorMessage = "Error: Expected non-null callback.";
-        CSP_LOG_ERROR_FORMAT("%s\n", ErrorMessage.c_str());
+        CSP_LOG_ERROR_FORMAT(
+            "Error: there is already a system registered for %s. Deregister the system before registering a callback.", EventName.c_str());
```
MAG-ElliotMorris, line 91: [Question] System registered for event? I guess I'm confused, do we not register callbacks to events, not systems?

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -77,36 +77,30 @@ void EventBus::ListenNetworkEvent(const csp::common::String& EventName, csp::sys
...
     }
 
     if (!CallbacksNetworkEventMap.empty() && CallbacksNetworkEventMap.find(EventName) != CallbacksNetworkEventMap.end())
     {
         if (CallbacksNetworkEventMap[EventName])
         {
-            CSP_LOG_ERROR_FORMAT("Error: there is already a callback registered for %s.\n", EventName.c_str());
-            return;
+            // We cannot compare callbacks, so we can't know whether it is the same callback that is already set. Therefore, we always update it
+            CSP_LOG_FORMAT(csp::systems::LogLevel::VeryVerbose, "The callback set for %s was overwritten with a new callback.", EventName.c_str());
```
MAG-ElliotMorris, line 100: [comment] Great info, appropriate log level. (Although I might consider this worth being at regular `Verbose`, but 🤷 )

Library/src/Multiplayer/EventBus.cpp
MAG-ElliotMorris: Ah, I see later this is guarded more appropriately in `StartEventMessageListening`. If that guards against this code path being called without the MultiplayerSystem, then that makes some sense.

Library/src/Multiplayer/EventBus.cpp
```diff
@@ -146,7 +146,7 @@ void EventBus::StartEventMessageListening()
         if (CallbacksNetworkEventMap.find(EventType) == CallbacksNetworkEventMap.end()
             && SystemsNetworkEventMap.find(EventType) == SystemsNetworkEventMap.end())
         {
-            CSP_LOG_MSG(csp::systems::LogLevel::Log, "Event was not found in event map.\n");
+            CSP_LOG_FORMAT(csp::systems::LogLevel::VeryVerbose, "Event %s is no longer registered to, discarding...", EventType.c_str());
```
MAG-ElliotMorris, line 149: [Consider] Some of these seem important enough to just be regular `Verbose`

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1136,29 +1136,35 @@ void AssetSystem::GetMaterial(const csp::common::String& AssetCollectionId, cons
...
+    if (!EventBusPtr)
+    {
+        CSP_LOG_ERROR_MSG("Error: Failed to register AssetSystem. EventBus must be instantiated in the MultiplayerConnection first.");
+        return;
+    }
+
+    if (!MaterialChangedCallback && !AssetDetailBlobChangedCallback)
     {
-        RegisterSystemCallback();
+        return;
```
MAG-ElliotMorris, line 1163: [Consider] Worth logging that we're bailing out early without registering in this case?

Library/src/Multiplayer/EventBus.cpp
MAG-OR: this is part of the bug, as one of the complaints from Web was the extra blank line. so I fixed them all.

Library/src/Multiplayer/EventBus.cpp
MAG-OR: this message appears if you try to register a system (or a callback) for an event when there is already a system registered. so before registering the system or callback, you need to first deregister the system that is already registered. we allo only one system per event

Library/src/Multiplayer/EventBus.cpp
MAG-OR: It is later guarded, yes. I have run all the system tests through the debugger to make sure that this error message was never caught. it is not. and that includes the new test that checks that you can set callbacks before the connection is available. I am sure as far as our tests go, but this bug has already proven that our tests only go so far. this is why I asked Adrian to confirm that this fix works. he did, but obviously he could not test _everything_.

Library/src/Multiplayer/EventBus.cpp
MAG-OR: in the new event bus implementation, there are 2 ways to register to events, one through systems and one through callbacks directly (like ClientElectionManager for example). We only allow one system or callback to be registered for an event at any one time. Systems handle their own callbacks, while the event bus handles the non-system callbacks

Library/src/Systems/Assets/AssetSystem.cpp
MAG-OR: fixed in 1f190d8ab28bebd71b44293764e3bbeec71d654c

Library/src/Multiplayer/EventBus.cpp
MAG-ElliotMorris: Right, okay, thanks for the explanation.

Library/src/Multiplayer/EventBus.cpp
MAG-ElliotMorris: Yeah, you've done all you can, ... still afraid of the risk but I suppose the fix must flow.

Library/src/Multiplayer/EventBus.cpp
MAG-ElliotMorris: Alright I think I can trace the conditionals now. I guess this makes sense. Tbh, part of my problem here was not figuring that `ListenNetworkEvent` was an explicit registration function, it dosen't sound like one from its name.

MAG-ElliotMorris (approved):
Signing off despite some fear around how this interacts with the vagueness of our multiplayer connection flow. It's been tested and checked as much as possible in the current setup.

PR #600, opened 2025-01-24
[NT-0] fix: add connected callback

MAG-mv:
- Added connection callback in StartListening callback to ensure its called on the main thread in wasm
- Added test

### DIFF
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -476,7 +476,7 @@ void MultiplayerConnection::StartListening(ErrorCodeCallbackHandler Callback)
             return;
         }
 
-        INVOKE_IF_NOT_NULL(ConnectionCallback, "Connected");
+        INVOKE_IF_NOT_NULL(ConnectionCallback, "Success");
         INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
     };
 
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -466,7 +466,7 @@ void MultiplayerConnection::StartListening(ErrorCodeCallbackHandler Callback)
         return;
     }
 
-    std::function<void(signalr::value, std::exception_ptr)> LocalCallback = [Callback](signalr::value Result, std::exception_ptr Except)
+    std::function<void(signalr::value, std::exception_ptr)> LocalCallback = [Callback, this](signalr::value Result, std::exception_ptr Except)
     {
         if (Except != nullptr)
         {
@@ -476,6 +476,7 @@ void MultiplayerConnection::StartListening(ErrorCodeCallbackHandler Callback)
             return;
         }
 
+        INVOKE_IF_NOT_NULL(ConnectionCallback, "Connected");
         INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
     };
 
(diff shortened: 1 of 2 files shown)


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -466,7 +466,7 @@ void MultiplayerConnection::StartListening(ErrorCodeCallbackHandler Callback)
         return;
     }
 
-    std::function<void(signalr::value, std::exception_ptr)> LocalCallback = [Callback](signalr::value Result, std::exception_ptr Except)
+    std::function<void(signalr::value, std::exception_ptr)> LocalCallback = [Callback, this](signalr::value Result, std::exception_ptr Except)
```
MAG-OR, line 469: [consider] is `this` required?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -476,6 +476,7 @@ void MultiplayerConnection::StartListening(ErrorCodeCallbackHandler Callback)
             return;
         }
 
+        INVOKE_IF_NOT_NULL(ConnectionCallback, "Connected");
```
MAG-OR, line 479: [consider] the description of SetConnectionCallback claims that the string is Success/Failure, not Connected/Disconnected: `/// @param Callback ConnectionCallbackHandler : The callback for connection, contains a string with a status of connection (Success, Failure etc).`

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-mv: I need it to access the ConnectedCallback property in multiplayer connection from within the lambda

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-mv: Good catch

MAG-OR: I had already written a test, btw :-) https://github.com/magnopus-opensource/connected-spaces-platform/pull/599/files#diff-db685b047997965f8f8aab090b91a4fb86fc6c5abb1883cfed53eeaf651f0614

MAG-mv: Fixed in [27c5dbd](https://github.com/magnopus-opensource/connected-spaces-platform/pull/600/commits/27c5dbde4fcf15334b5976eadfda62a5c583b3c7)

Tools/WrapperGenerator/chevron
MAG-ChristopherAtkinson, line 1: [Question] was this intended to part of the commit?

Library/src/Multiplayer/MultiplayerConnection.cpp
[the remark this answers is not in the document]
MAG-mv: I just spoke to @MAG-ElliotMorris. This was due to the this pointer not being used. It is now used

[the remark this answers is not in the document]
MAG-ChristopherAtkinson: That is great thought it was best to check, happy for this to be resolved.

MAG-OR (approved):
approving with the expectation that @MAG-ChristopherAtkinson 's question about chevron being included on purpose in this commit will be answered prior to merging

Tools/WrapperGenerator/chevron
MAG-mv: Removed in [727cec8](https://github.com/magnopus-opensource/connected-spaces-platform/pull/600/commits/727cec822dd0dd06b5c05498e07b73d760598a6c)

MAG-ChristopherAtkinson (approved):

MAG-AdamThorn (approved):

MAG-ElliotMorris:
We're changing where and how the connection callback is sent, these are exactly the sort of changes that we need to document when we do them. I don't know how to do that properly right now with the way our release notes are generated, but it would be good to at least ensure the documentation for connectionCallback is specific on this.

PR #608, opened 2025-01-29
[NT-0] feat!: Add metadata and tag parameters to material creation

MAG-DanielMurdolo:
- Add parameters for metadata and tags to the create material method to support being able to filter them out from other types of assets
- Update the relevant tests

### DIFF
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -244,8 +244,12 @@ public:
     /// @brief Creates a new material backed by an AssetCollection/Asset.
     /// @param Name const csp::common::String& : The name of the new material.
     /// @param SpaceId const csp::common::String& : The space id this material is associated with.
+    /// @param Metadata csp::common::Map<csp::common::String, csp::common::String>& : The metadata to be associated with the created asset colection.
+    /// @param AssetTags csp::common::Array<csp::common::String>& : Tags to be associated with the created asset collection.
     /// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
-    CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId, GLTFMaterialResultCallback Callback);
+    CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId,
+        csp::common::Map<csp::common::String, csp::common::String> Metadata, const csp::common::Array<csp::common::String> AssetTags,
+        GLTFMaterialResultCallback Callback);
 
     /// @brief Updates an existing material's properties.
     /// The material should be retrieved through GetMaterials or GetMaterial.
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -244,8 +244,8 @@ public:
     /// @brief Creates a new material backed by an AssetCollection/Asset.
     /// @param Name const csp::common::String& : The name of the new material.
     /// @param SpaceId const csp::common::String& : The space id this material is associated with.
-    /// @param Metadata csp::common::Map<csp::common::String, csp::common::String>& : The metadata to be associated with the created asset colection.
-    /// @param AssetTags csp::common::Array<csp::common::String>& : Tags to be associated with the created asset collection.
+    /// @param Metadata csp::common::Map<csp::common::String, csp::common::String>& : The metadata to be associated with the created material.
+    /// @param AssetTags csp::common::Array<csp::common::String>& : Tags to be associated with the created material.
     /// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
     CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId,
         csp::common::Map<csp::common::String, csp::common::String> Metadata, const csp::common::Array<csp::common::String> AssetTags,
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -248,7 +248,8 @@ public:
     /// @param AssetTags csp::common::Array<csp::common::String>& : Tags to be associated with the created material.
     /// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
     CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId,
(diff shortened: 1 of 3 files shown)


Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -896,8 +896,10 @@ void AssetSystem::CreateMaterial(const csp::common::String& Name, const csp::com
     };
 
     const csp::common::String MaterialCollectionName = CreateUniqueMaterialAssetCollectionName(Name, SpaceId);
+    Array<String> AssetTag = { "material" };
+    csp::common::Map<csp::common::String, csp::common::String> Metadata = { { "AssetType", "Material" } };
```
MAG-ElliotMorris, line 900: [Question] Is there a reason for the Tag being lower-cased but the Metadata value being upper?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -896,8 +896,10 @@ void AssetSystem::CreateMaterial(const csp::common::String& Name, const csp::com
     };
 
     const csp::common::String MaterialCollectionName = CreateUniqueMaterialAssetCollectionName(Name, SpaceId);
+    Array<String> AssetTag = { "material" };
```
MAG-mv, line 899: [Nit] this could be static const

MAG-mv (approved):

Library/src/Systems/Assets/AssetSystem.cpp
MAG-DanielMurdolo: The paradigm we established with CHS for tags has been to use lower case, as set with the space template work.

The asset type is a legacy thing and all the existing ones are sentence case.

Library/src/Systems/Assets/AssetSystem.cpp
MAG-DanielMurdolo: Done

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -244,8 +244,12 @@ class CSP_API AssetSystem : public SystemBase
     /// @brief Creates a new material backed by an AssetCollection/Asset.
     /// @param Name const csp::common::String& : The name of the new material.
     /// @param SpaceId const csp::common::String& : The space id this material is associated with.
+    /// @param Metadata csp::common::Map<csp::common::String, csp::common::String>& : The metadata to be associated with the created asset colection.
```
MAG-mv, line 247: [Consider] We want to hide that materials are using asset collections. I think we should change "asset collections" to materials

MAG-mv (changes requested):

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -244,8 +244,12 @@ class CSP_API AssetSystem : public SystemBase
...
+    /// @param AssetTags csp::common::Array<csp::common::String>& : Tags to be associated with the created asset collection.
     /// @param Callback GLTFMaterialResultCallback : Callback when asynchronous task finishes.
-    CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId, GLTFMaterialResultCallback Callback);
+    CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId,
+        csp::common::Map<csp::common::String, csp::common::String> Metadata, const csp::common::Array<csp::common::String> AssetTags,
+        GLTFMaterialResultCallback Callback);
```
MAG-ElliotMorris, line 252: [fix] In all interfaces, you should be using const-ref to prevent copying.

```suggestion
    CSP_ASYNC_RESULT void CreateMaterial(const csp::common::String& Name, const csp::common::String& SpaceId,
        const csp::common::Map<csp::common::String, csp::common::String>& Metadata, const csp::
        common::Array<csp::common::String>& AssetTags,
        GLTFMaterialResultCallback Callback);
```

MAG-ElliotMorris (changes requested):
Just a c++ semantics thing.

We really should add tests for the new functionality, but given the need for a publish _today_, I guess what's one more piece of untested functionality eh? 😔 I suspect we'll log an issue to track this.

MAG-ElliotMorris:
Error from CI: 
> C:\teamcity\us-hw-win-01-01\work\86601c535a079422\Tools\WrapperGenerator\Output\CSharp\Systems\Assets\AssetSystem.cs(269,13): error CS0246: The type or namespace name 'Map' could not be found (are you missing a using directive or an assembly reference?) [C:\teamcity\us-hw-win-01-01\work\86601c535a079422\Library\CSharpWrapper\CSharpWrapper.csproj]

Weird, we use maps everywhere why is it different here?

MAG-ElliotMorris:
I tested after changing to const-ref parameters and the csharp generator accepted it, seems we don't support value-params at all for these structures.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-DanielMurdolo: Rapid C++ lesson. What does adding the & do?

MAG-ElliotMorris: It means that rather than copying the value, the value will be passed by reference, (ie just a pointer)
You won't need to update any other code other than the function declaration and definition, since the semantics of a value param (no &) and a reference param are identical.

MAG-DanielMurdolo:
What's the estimated amount of work for adding a test or tests @MAG-ElliotMorris?

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: You'll note that in this function, the parameters are passed down into `CreateAssetCollection` and then later into `CreatePrototypeDto`, both of which take `const &` parameters. This way, a copy is never made and we use the same memory the entire chain.

MAG-ElliotMorris:
I wouldn't imagine this PR will get merged and released today if we add a test. Already 50/50 on that as it is.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: (It also allows our csharp generators to work, since they only account for const-ref params for complex types ... apparently)

MAG-DanielMurdolo: Thanks

MAG-ElliotMorris:
Just pushed the over side of the const-ref change, you need to have them in both .h and .cpp files. Should be fine now.

MAG-ElliotMorris (approved):

MAG-mv (approved):

PR #609, opened 2025-01-29
[OF-1191] refac: actually delete component when calling DeleteComponent

MAG-OR:
- Add Delete to the ComponentType enum and set it to 56 to match CHS
- Change SerialisePatch to use ComponentType::Delete instead of Invalid so that the component gets actually deleted instead of invalidated
- Add test case


MAG-mv (approved):

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1203,6 +1203,133 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ObjectRemoveComponentTest)
...
+        EntitiesCreated = true;
+        EXPECT_TRUE(Success);
+    };
+
+    EntitySystem->SetInitialEntitiesRetrievedCallback(EntitiesReadyCallback);
+
+    auto [EnterResult2] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id);
+    EXPECT_EQ(EnterResult2.GetResultCode(), csp::systems::EResultCode::Success);
+
+    // Wait 2 seconds for the CHS database to update
```
MAG-ChristopherAtkinson, line 1305: [Fix] Comment is out of date, could exclude the time e.g. "Wait for the CHS database to update"

Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -69,7 +69,8 @@ enum class ComponentType
     GaussianSplat,
     Text,
     Hotspot,
-    CinematicCamera
+    CinematicCamera,
+    Delete = 56
```
MAG-ChristopherAtkinson, line 73: [Question] I will assume that 56 has significant here, CHS expecting this value?

MAG-ChristopherAtkinson (commented):
I am glad to see the introduction of test for this behaviour I have added some comments but happy with the overall changes.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1203,6 +1203,133 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ObjectRemoveComponentTest)
...
+    std::this_thread::sleep_for(std::chrono::seconds(4));
```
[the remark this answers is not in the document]
MAG-OR: yes we absolutely need to discuss this, as it DOES NOT WORK. it was the fix suggested by CHS in the slack that @mag-mv linked. I started with 2 seconds, which is the duration of the delay in CHS, didn't work so I doubled it in case it was just lagging, and the test still does not pass. so we need to find another solution

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-OR: yes

Tests/src/PublicAPITests/MultiPlayerTests.cpp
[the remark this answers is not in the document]
MAG-mv: 3 Seconds works fine on my test. There may be another issue here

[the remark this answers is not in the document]
MAG-OR: aha! thanks for confirming, that's helpful. I'll look into it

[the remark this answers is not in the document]
MAG-ElliotMorris: [Comment] Just a side note from intuition, even if you get this working at 3-4 seconds, I would double this _again_ just to be safe. I feel it's within the realms of conceivability that dropped packets could still delay an otherwise working TCP exchange by 2ish seconds, and flaky tests due to non-addressable network are nightmares.

Remember also that these tests are not _strictly_ (although I grant, at the moment in practice they are) just for us. External users may wish to run them when evaluating quality during the decision to adopt the platform or not.

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-ElliotMorris: [Question] Is it possible to link (as a code comment) to where this requirement is documented, even if just a slack chat?
[Consider] I don't think this change breaks anything, but as you know, manually specifying this value will cause all subsequent values to start counting from 56, maybe it's worth explicitly defining all the values here and now, to lock them down, and guard against potential careless reorderings in the future? (Ie, assign the first one = 0, the next one 1, etc).

Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -42,34 +42,38 @@ class ComponentScriptInterface;
...
+    Fog = 21,
+    ECommerce = 22,
+    FiducialMarker = 23,
+    GaussianSplat = 24,
+    Text = 25,
+    Hotspot = 26,
+    CinematicCamera = 27,
+    // spare values
+    Delete
+    = 56 // the value 56 is a requirement that comes from MessagePack, see:
```
MAG-OR, line 75: it is the clang formatting that forces the split to a new line.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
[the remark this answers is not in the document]
MAG-OR: so, for avoidance of doubt, I have changed it to 8, but what solved it is that I moved it between the exit and the re-enter

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-OR: changed

Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -42,34 +42,38 @@ class ComponentScriptInterface;
...
+    ECommerce = 22,
+    FiducialMarker = 23,
+    GaussianSplat = 24,
+    Text = 25,
+    Hotspot = 26,
+    CinematicCamera = 27,
+    // spare values
+    Delete = 56
+    // the value 56 is a requirement that comes from MessagePack - see:
+    // https://github.com/magnopus/Magnopus.Services/blob/8a376c4454c5c082794bceb940f3f79b8ee0d1db/Source/Magnopus.Service.Multiplayer.MessagePack/ItemComponentDataResolver.cs#L86
```
MAG-ChristopherAtkinson, line 76: [Fix] Link to non public assessable Source file, realistically we shouldn't included changes such as these within this repo.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1203,6 +1203,133 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ObjectRemoveComponentTest)
...
+
+    EXPECT_EQ(RealComponents.Size(), 1);
+    EXPECT_TRUE(RealComponents.HasKey(KeepKey));
+    EXPECT_FALSE(RealComponents.HasKey(DeleteKey));
+
+    // Exit space and enter again, making sure the entities have been created
+    auto [ExitSpaceResult] = AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
+
+    // Wait a few seconds for the CHS database to update
+    std::this_thread::sleep_for(std::chrono::seconds(8));
```
MAG-ChristopherAtkinson, line 1293: [Comment] Mentioning for completeness but would like to avoid introducing waits in future, in this case I understand this is needed.

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-OR: This was done as per https://github.com/magnopus-opensource/connected-spaces-platform/pull/609#discussion_r1948787635

MAG-ElliotMorris (approved):

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-ChristopherAtkinson: Thanks @MAG-ElliotMorris that makes a lot of sense, I wouldn't worry too much about internal structures in this case, it would be very minor but do appreciate the consideration and if we can we should avoid it in future. 

As far as documentation break and accessibility, I was just meaning that any link should also be accessible at the same level of the source code or more, and if it can be helped should be avoided. I tend to think about in the same vein as C++ access specifiers. (if that make sense)

MAG-OR: we agreed to just remove it and unblock the PR

MAG-ElliotMorris (approved):

PR #612, opened 2025-01-30
[NT-0] doc: Remove refs to old develop branch

MAG-AdamThorn:
Remove references to the old develop branch from our documentation.


Library/docs/source/manual/documentation/multiplayer_state_replication.md
```diff
@@ -65,11 +65,11 @@ A View Component is essentially a core property, present for any every Space Ent
...
+ View components are represented by _specific_ keys in the Component map, which are 1024 reserved keys at the end of the component map. These keys can be seen in [SpaceEntityKeys.h](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/src/Multiplayer/SpaceEntityKeys.h).
 
 ### CSP Components
 
-CSP components are the components that inherit from [ComponentBase](https://github.com/magnopus-opensource/connected-spaces-platform/blob/develop/Library/include/CSP/Multiplayer/ComponentBase.h) (such as `StaticModelSpaceComponent`), and they utilise the component keys starting from zero and incrementing as new components are added to an entity.
+CSP components are the components that inherit from [ComponentBase](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/include/CSP/Multiplayer/ComponentBase.h) (such as `StaticModelSpaceComponent`), and they utilise the component keys starting from zero and incrementing as new components are added to an entity.
```
MAG-SamBirley, line 72: [Suggestion] If you would like to make this line more invariant to future URL changes (or even source files changes), you could consider using the ReStructedText directives we're employing elsewhere to reference classes in the documentation, and then rely on the docs pipeline to figure out the appropriate URL to resolve that to.

There's an example [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/docs/source/learn/spaces/events.md?plain=1#L6).

In this case, the change I think would be:

```
    ```eval_rst
    CSP components are the components that inherit from :class:`csp::Multiplayer::ComponentBase` (such as `StaticModelSpaceComponent`), and they utilise the component keys starting from zero and incrementing as new components are added to an entity.
    ```
```

Library/docs/source/manual/documentation/branching_strategy.md
```diff
@@ -19,7 +17,7 @@ Akin to _Feature Branches_ in GitFlow, these cover any change needed to the code
 If the work being done is based on a Jira ticket, the ticket ID should be included in the branch name. If there is no ticket for the work, you may use `NT-0` in place of the ticket ID.
 
 > `<work type>/XX-YY_Title`
-> * work type - A label denoting what kind of work this branch covers. Should be one of either `fix`, `feature`, `doc`, or `misc`. `misc` branches cover refactoring work, style changes/fixes, and other work that does not fit in the aforementioned labels.
+> * work type - A label denoting the commit category. We have adopted the Conventional Commits specification, and so this label should be one of the following: `fix`, `feat`, `build`, `chore`, `ci`, `docs`, `style`, `refac`, `perf` or `test`.
```
MAG-OR, line 20: [comment] this will only be true once `.githooks/commit-msg` and `.githooks/commit-template.txt ` are updated accordingly. it may be best not to preemptively update the docs, but I'll leave it up to you

Library/docs/source/manual/documentation/branching_strategy.md
```diff
@@ -28,54 +26,37 @@ Example branch names:
...
-3. Submit a PR to the `develop` branch once work is complete.
+3. Submit a PR to the `main` branch once work is complete.
 4. Await for PR build checks to validate the change is stable
 5. Gain approval from two members of the Connected Spaces platform team
-6. Merge to `develop`
-
-### Making a new release version
-
-1. Once `develop` has been validated as stable and complete, a PR is made to merge into `Main`.
-2. Once merged into `Main`, a tag is applied to mark the version, and builds are made of the associated packages.
```
MAG-OR, line 81: [comment] I think this line (item 2) could be item 7 in the new list (with `main` instead of `Main`)

.github/workflows/staging-newBranch-restrict.yml
MAG-OR, line 1: [question] is this PR a good opportunity to also delete `.github/workflows/main-pr-restrict.yml`?

MAG-OR (approved):

Library/docs/source/manual/documentation/branching_strategy.md
MAG-AdamThorn: Agreed I have made this change.

.github/workflows/staging-newBranch-restrict.yml
MAG-AdamThorn: Yes I think it is. I have removed it.

Library/docs/source/manual/documentation/multiplayer_state_replication.md
MAG-AdamThorn: Agreed done. I did not update the link to the SpaceEntityKeys.h file above this though as I could not find an obvious way to do it (it is not a class, just a file with constexpr in a namespace). I also updated the links in ScriptSystem.md to use ReStructuredText Directives.

MAG-SamBirley: LGTM 👍

MAG-mv (approved):

PR #613, opened 2025-01-30
[OF-1388] feat: Conversation assets are now deleted when component is deleted

MAG-mv:
- Implemented OnLocalDelete for ConversationSpaceComponent to call DeleteConversation when the component is deleted
- Fixed issue with DeleteConversation not removing the entire conversation
- Added test to verify this

### DIFF
--- a/Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
+++ b/Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
@@ -162,21 +162,12 @@ void ConversationSpaceComponent::DeleteConversation(csp::systems::NullResultCall
             return;
         }
 
-        const csp::systems::NullResultCallback DeleteMessagesCallback
-            = [Callback, ConversationId, this](const csp::systems::NullResult& DeleteMessagesResult)
+        const csp::systems::NullResultCallback NullResultCallback
+            = [Callback, ConversationId, this](const csp::systems::NullResult& NullResultCallbackResult)
         {
-            if (DeleteMessagesResult.GetResultCode() == csp::systems::EResultCode::Failed)
-            {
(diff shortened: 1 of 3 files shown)


Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -162,21 +162,12 @@ void ConversationSpaceComponent::DeleteConversation(csp::systems::NullResultCall
             return;
         }
 
-        const csp::systems::NullResultCallback DeleteMessagesCallback
-            = [Callback, ConversationId, this](const csp::systems::NullResult& DeleteMessagesResult)
+        const csp::systems::NullResultCallback NullResultCallback
+            = [Callback, ConversationId, this](const csp::systems::NullResult& NullResultCallbackResult)
```
MAG-ElliotMorris, line 166: [Question] Why this rename? Doesn't it make the name less descriptive?

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -763,4 +748,10 @@ void ConversationSpaceComponent::DeleteMessages(
     AssetSystem->DeleteMultipleAssetCollections(Messages, Callback);
 }
 
+void ConversationSpaceComponent::OnLocalDelete()
+{
+    const auto Callback = [](const NullResult& Result) {};
+    DeleteConversation(Callback);
```
MAG-ElliotMorris, line 754: [Nit] Either `std::move` or define the Callback directly in the param list to prevent a copy

MAG-ElliotMorris (approved):

MAG-AdamThorn (approved):

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-mv: It was moved. Turns out one of these callbacks wasnt even being called

PR #615, opened 2025-02-11
[OF-1415] feat: change AssetPlatform to DeviceID in HTTP header

MAG-OR:
Change AssetPlatform to DeviceID in HTTP header.

This makes the WebClientUserAgentTest fail, as we no longer record the TESTS_CLIENT_SKU. I still need to figure out if it's expected and I can just delete the test, or if it should still be recorded another way.


Library/src/Multiplayer/SignalR/EmscriptenSignalRClient/EmscriptenSignalRClient.cpp
```diff
@@ -184,7 +184,7 @@ std::string CSPWebSocketClientEmscripten::GetWebSocketConnectURL(const std::stri
         std::string WebSocketEndpoint = InitialUrl.substr(0, QueryParamPos);
 
         WebSocketConnectURL = WebSocketEndpoint + "?access_token=" + std::string(csp::web::HttpAuth::GetAccessToken().c_str())
-            + "&X-AssetPlatform=" + std::string(csp::CSPFoundation::GetClientUserAgentString().c_str());
+            + "&X-DeviceUDID=" + std::string(csp::CSPFoundation::GetDeviceId().c_str());
```
MAG-OR, line 187: are we positive that this URL (WebSocketConnectURL) needs to be updated? Jerome wasn't too sure, and I can't find tests to become more confident about this change. this is purely related to emscripten, i.e. WASM.

MAG-OR, line 187: I asked on the channel but I'll ask again here: this changes a URL, but only for WASM. Jerome wasn't too sure what this did but he thought this might need to change as well as the other SignalRConnection. Does anyone know either way?

MAG-ElliotMorris: I doubt anyone on the team is going to know without investigation. It's a shame that the Emscripten client isn't symmetric to the non-emscripten one. 

This is called in `CSPWebSocketClientEmscripten::Start`, so that at least indicates that this is the initial connection as the ticket requests, but that's not a huge amount to go on.

At the end of the day, if the requirement from CHS is "Send us X-DeviceUDID and not X_AssetPlatform", then this does seem like a correct change. You can verify this by running in emscripten and non-emscripten with a packet sniffer active an inspecting the headers yourself.

MAG-ElliotMorris:
[Question] Why is the `X-AssetPlatform` in `HttpPayload` still there? There must be _somewhere_ where these headers and what they mean are documented...

MAG-ElliotMorris:
Ah, great! (Still wish there was a big table documenting what the CHS expectations are with these ... _grumble grumble_)

MAG-mv (commented):
Looks good, bnut we need to figure out if the Emscripten change is needed

Library/src/Multiplayer/SignalR/EmscriptenSignalRClient/EmscriptenSignalRClient.cpp
MAG-OR: I have run the Basic Web Example locally and I have inspected the header of the first multiplayer packet. the X-DeviceUDID parameter is indeed there:
image

MAG-ElliotMorris (approved):
I think confirming that we're sending the header in initialization in both desktop and web modalities is the best we can do right now given the requirements. 

P.S: This is a small PR so it's not too important, but would have been nice to avoid all the merge commits.

MAG-OR:
> I think confirming that we're sending the header in initialization in both desktop and web modalities is the best we can do right now given the requirements.
> 
> P.S: This is a small PR so it's not too important, but would have been nice to avoid all the merge commits.

These merge commits happen every time "Update branch" is clicked on the PR.

Library/src/Multiplayer/SignalR/EmscriptenSignalRClient/EmscriptenSignalRClient.cpp
MAG-ElliotMorris: Oh hold on, (I've just approved this but am noticing now.) This isn't a header, it's part of the URL. I guess I didn't look closely.

Argh. 😖 I don't understand, no one understands. I still think make the change and hope tbh, what else are we to do?

MAG-ElliotMorris:
You can hit the dropdown and update via a rebase, _or_ do it manually yourself.

MAG-mv (approved):

PR #616, opened 2025-02-11
[OF-1512] refac: In-Thread Continuations, EnterSpace refactor.

MAG-ElliotMorris:
Refactor EnterSpace to remove use of nested callbacks.
This is done via integrating [Async++](https://github.com/Amanieu/asyncplusplus), an async library that provides `.then()` continuations, allowing us to program linearly. The hope is that this will establish a pattern for more comprehensible flows, where we can reason about the higher level data path, rather than scrabbling through umpteen layers of nested callback branches.

Coincidentally fixes this bug, as well as a few other latent bugs. : https://magnopus.atlassian.net/browse/OB-4016
Also provides a vector of motion for this issue: https://magnopus.atlassian.net/browse/OF-1557 

## What's the point?
Take a look at the old [EnterSpace](https://github.com/magnopus-opensource/connected-spaces-platform/blob/17870cfb5de18189d571016a882ace1264bd8323/Library/src/Systems/Spaces/SpaceSystem.cpp#L90) method. The idea is to be able to express that in a linear, pipeline style flow. It looks like this now :
```cpp
void SpaceSystem::EnterSpace(const String& SpaceId, NullResultCallback Callback)
{
    CSP_LOG_MSG(csp::systems::LogLevel::Log, "SpaceSystem::EnterSpace");
    constexpr auto GENERIC_UNKNOWN_ERROR_CODE = 500;

    GetSpace(SpaceId)
        .then(async::inline_scheduler(),
            csp::common::continuations::AssertRequestSuccessOrErrorFromResult<SpaceResult>(Callback,
                "SpaceSystem::EnterSpace, successfully discovered space.",
                "Logged in user does not have permission to discover this space. Failed to enter space.", {}, {}, {}))
        .then(async::inline_scheduler(), AddUserToSpaceIfNeccesary(Callback, *this))
        .then(async::inline_scheduler(),
            csp::common::continuations::AssertRequestSuccessOrErrorFromResult<SpaceResult>(Callback,
                "SpaceSystem::EnterSpace, successfully added user to space (if not already added).",
                "Failed to Enter Space. AddUserToSpace returned unexpected failure.", {}, {}, {}))
        .then(async::inline_scheduler(), FireEnterSpaceEvent(CurrentSpace))
        .then(async::inline_scheduler(), RefreshMultiplayerScopes())
        .then(async::inline_scheduler(),
            csp::common::continuations::AssertRequestSuccessOrErrorFromErrorCode(Callback,
                "SpaceSystem: EnterSpace, successfully refreshed multiplayer scopes", EResultCode::Failed, GENERIC_UNKNOWN_ERROR_CODE,
                ERequestFailureReason::Unknown, csp::systems::LogLevel::Error))
        .then(async::inline_scheduler(), csp::common::continuations::ReportSuccess(Callback, "Successfully entered space."))
        .then(
            async::inline_scheduler(), csp::common::continuations::InvokeIfExceptionInChain([&CurrentSpace = CurrentSpace]() { CurrentSpace = {}; }));
}
```
> [!NOTE]  
> See how the majority of this uses functions from the csp::common::continuations namespace. These are general purpose continuations. They're structural, and common to most flows. With this, we separate our structural mechanisms from our actual business logic, giving benefits to not only comprehensibility, but also testability and self-similarity. You'll see in this PR how we can unit-test via mocking individual steps of this flow in ways that give me increased confidence.

## Why not something more standard?
It's because of WASM. Normally when exposing a synchronous interface around services you wait and block until you get a result. (Arguably, this is a bad pattern because it introduces unnecessary delay, but it's far and away the simplest.)
However, this, as well as the other "normal" option of firing-and-forgetting a handler on another thread, are not available to us due to WASM not allowing main thread blocking, and not allowing cross-thread callbacks.

This approach is, essentially, what we were doing before. The code will generate out to become nested callbacks more or less. This just allows us to structure things much nicer, and thread the mechanisms deeper into our task code without having to engineer a whole framework. (We would have basically ended up reinventing this lib, ... in fact, Matt and I got halfway there during iteration.)

## How to review.
I strongly encourage going commit by commit on this one, it's layed out to try and remove all the cruft before getting to the main affair.

> [!NOTE]  
> The first commit is the googletest integration done in a separate PR, you don't need to review that.
> The second commit is the async++ integration, which should be reviewed, but you don't need to review the library code.

## Testing
By testing, I mostly mean WASM testing. What WASM testing has been done? I have verified EnterSpace functions via the examples, but not gone further than that. I _will_ confirm that an actual OKO build functions, joining a few spaces, before merging this. It's the shame this is the best we can do for now.

This PR also includes an extension to our existing test framework to allow parameterized testing, which I used to setup acceptable regression tests prior to this refactor. I caught a fair few latent bugs doing this. (My favourite way to catch bugs, way cheaper than waiting for them to be reported.)

## Drawbacks
- Due to it not currently being easy to test private functionality, the non-general continuations are not unit tested, (although they are pretty well tested via the functional tests. There's an argument that this is good enough given that these continuations are intended only for EnterSpace)

- Generator limitations force us to adopt a pattern where we expose internal duplicates for the callback methods `GetSpace`, `AddUserToSpace`, in this PR. This is a shame, and the biggest issue right now imo.

- I never figured out how to set the default scheduler to inline, though I didn't try too hard. It'd be good to do, cause forgetting that will cause issues that only manifest in WASM.

- It's unfortunate that we can't mock internal methods. Note how we test `LogErrorAndCancelContinuation` in the unit tests, but then essentially test it again in subsequent runs because we can't inject a mock method for it to assert it was called, so we have to test its actual behaviour. This is an advanced topic that we may not need to focus on now, but something to think on.

- The non-modularity of CSP forces us to make what should be Internal tests regular Public tests. Quite unfortunate.

> [!NOTE]  
> Did you know `.then()` was a proposed extension to std::future that got dropped when the concurrency TS was [dropped ](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1445r0.pdf) for C++20? Not because it was a bad idea, but because other thinking (coroutines I presume) had overtaken the concurrency TS as a whole. 🤷


Library/include/CSP/Systems/WebService.h
```diff
@@ -79,6 +79,7 @@ enum class ERequestFailureReason
     UserShopifyLimitReached,
     UserTokenRefreshFailed,
     InvalidSequenceKey,
+    RequestTimedOut,
```
MAG-ElliotMorris, line 82: [Fix] Remove, artifact from the old blocking refactor attempt.

Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -352,6 +366,14 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
 
     void GetSpaceGeoLocationInternal(const csp::common::String& SpaceId, SpaceGeoLocationResultCallback Callback);
 
+    // EnterSpace Continuations
+    auto AddUserToSpaceIfNeccesary(NullResultCallback Callback, SpaceSystem& SpaceSystem);
+    auto FireEnterSpaceEvent(Space& OutCurrentSpace);
+    auto RefreshMultiplayerScopes();
+
+    void RefreshMultiplayerConnectionToEnactScopeChange(csp::common::String SpaceId,
```
MAG-ElliotMorris, line 374: [Consider] A comment would be nice here. It's not super clear by the name.

Library/src/Common/Continuations.h
```diff
@@ -0,0 +1,166 @@
...
+template <typename ResultT>
+inline auto AssertRequestSuccessOrErrorFromResult(NullResultCallback Callback, std::string SuccessMsg, std::string ErrorMsg,
+    std::optional<EResultCode> ResultCode, std::optional<uint16_t> HttpResultCode, std::optional<ERequestFailureReason> FailureReason,
+    csp::systems::LogLevel LogLevel = csp::systems::LogLevel::Log)
+{
+    return [Callback, SuccessMsg = std::move(SuccessMsg), ErrorMsg = std::move(ErrorMsg), ResultCode, HttpResultCode, FailureReason, LogLevel](
+               const ResultT& Result)
+    {
+        if (Result.GetResultCode() != EResultCode::Success)
+        {
```
MAG-ElliotMorris, line 86: [Consider] Here and below. Marking these blocks with `//Error case` and `//Success case` will make scanning easier.

Library/src/Common/Continuations.h
```diff
@@ -0,0 +1,166 @@
...
+inline auto AssertRequestSuccessOrErrorFromErrorCode(NullResultCallback Callback, std::string SuccessMsg, std::optional<EResultCode> ResultCode,
+    std::optional<uint16_t> HttpResultCode, std::optional<ERequestFailureReason> FailureReason,
+    csp::systems::LogLevel LogLevel = csp::systems::LogLevel::Log)
+{
+    return [Callback, SuccessMsg = std::move(SuccessMsg), ResultCode, HttpResultCode, FailureReason, LogLevel](
+               const std::optional<csp::multiplayer::ErrorCode>& ErrorCode)
+    {
+        if (ErrorCode.has_value())
+        {
+            constexpr auto GENERIC_UNKNOWN_ERROR_CODE = 500;
```
MAG-ElliotMorris, line 115: [Comment] It'd be nice to have a central HTTP code source for this sort of stuff. Maybe one for later. (Maybe there is one already?)

Library/src/Common/Continuations.h
```diff
@@ -0,0 +1,166 @@
...
+        NullResult SuccessResult(EResultCode::Success, SUCCESS_HTTP_CODE, ERequestFailureReason::None);
+        INVOKE_IF_NOT_NULL(Callback, SuccessResult);
+    };
+}
+
+/*
+ * Intended to be placed at the end of an async++ continuation chain.
+ * If the chain throws an unhandled exception, this will attempt to unwrap it,
+ * and call a passed in callable, (probably a state-reset or cleanup function of some sort).
+ * The callable is perfectly forwarded, ownership will be taken if appropriate.
```
MAG-ElliotMorris, line 145: [nit] "if appropriate" is a bit of a weasel phrase. What you mean here is "if given", although do you even need to mention that? Perfect forwarding is a standard concept. (I miss when these were called Universal References, it's a cooler name)

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -87,222 +91,183 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient)
...
-                                INVOKE_IF_NOT_NULL(Callback, MakeInvalid<NullResult>());
-                                return;
-                            }
-                            else
-                            {
-                                CSP_LOG_MSG(csp::systems::LogLevel::Verbose, "SetScopes was called successfully");
-                            }
+        auto UserAddedToSpaceChainStartEvent = std::make_shared<async::event_task<SpaceResult>>();
+        auto UserAddedToSpaceChainContinuation = UserAddedToSpaceChainStartEvent->get_task();
+        // AddUserToSpace does not give a callback (feels like it should...) if the user is already added to the space.
```
MAG-ElliotMorris, line 181: [Comment] Should probably go to the effort of replicating this in a CHS only reproduction and logging it. I'd almost call this a bug, and makes our nice continuation chain messier than it has to be.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -87,222 +91,183 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient)
...
+ * ** EnterSpace Flow **
+ * GetSpace
+ * AssertRequestSuccessOrError (GetSpace Validation)
+ * AddUserToSpaceIfNeccesary
+ * AssertRequestSuccessOrError (AddUserToSpace Validation)
+ * FireEnterSpaceEvent
+ * RefreshMultiplayerScopes
+ * AssertRequestSuccessOrErrorFromErrorCode (RefreshMultiplayerScopes Validation)
+ * ReportSuccess
+ * ResetJoinStateIfChainHasThrown
```
MAG-ElliotMorris, line 245: [nit] This method has changed name

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
```diff
@@ -0,0 +1,300 @@
...
+    EResultCode ResultCode = EResultCode::Success;
+    uint16_t HttpResultCode = 200;
+    ERequestFailureReason FailureReason = ERequestFailureReason::None;
+
+    NullResult ExpectedResult(ResultCode, HttpResultCode, FailureReason);
+
+    ::testing::MockFunction<void(const NullResult& Result)> MockResultCallback;
+    // Expect that the callback is called with the result constructed as expected.
+    EXPECT_CALL(MockResultCallback, Call(ExpectedResult)).Times(1);
+    // Expect that we log the error message
```
MAG-ElliotMorris, line 69: [fix] Success message.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/04279e94b85b0bbe99106acd8771b8c3ca517ecd

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/81c9e3cc19f1fe746544694bf35b36f4bc02cfec

Library/src/Common/Continuations.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/81c9e3cc19f1fe746544694bf35b36f4bc02cfec

Library/src/Common/Continuations.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/81c9e3cc19f1fe746544694bf35b36f4bc02cfec

Library/src/Common/Continuations.h
```diff
@@ -0,0 +1,166 @@
...
+ * limitations under the License.
+ */
+
+/*
+ * General purpose utility continuations for use in async++ based task chaining.
+ * These exist for a couple of reasons:
+ *  - Raw nested callbacks were becoming unreadable and causing bugs
+ *  - Via using common continuations, we can move towards standardised error handling
+ * If you're wondering why we go to the effort of integrate an async lib rather than
+ * simply using a standard blocking future/promise interface, it's because of WASM.
```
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/81c9e3cc19f1fe746544694bf35b36f4bc02cfec

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/81c9e3cc19f1fe746544694bf35b36f4bc02cfec

Library/include/CSP/Systems/WebService.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/84a0c4ea2468a95ece8952d9329efa69c0a0640e

Library/include/CSP/Common/Array.h
```diff
@@ -130,6 +130,15 @@ template <typename T> class CSP_API Array
     /// @return const T*
     CSP_NO_EXPORT const T* Data() const { return ObjectArray; }
 
+    // Iterators
+    CSP_NO_EXPORT T* begin() { return Data(); }
+    CSP_NO_EXPORT const T* begin() const { return Data(); }
+    CSP_NO_EXPORT const T* cbegin() const { return Data(); }
```
MAG-OR, line 136: [question] I am very likely missing something obvious, but what is the difference between this and the previous line, apart from the c? are both needed?

MAG-ElliotMorris: It's merely historical convention, not great, but best to conform imo. To allow a typing that enforces non-mutation, rather than relying on the const context of the calling function by using `begin` both ways. Some/much generic code uses `cbegin` so expects it to be there.

You still want the `begin() const` one though, otherwise you couldn't use `begin` in const methods which again, would cause some generic code to balk.

(I'm not 100% on all of this, I could be proven wrong)

MAG-ElliotMorris: Note here, the MSVC STL vector implementation does this : https://github.com/microsoft/STL/blob/d43d49a3fdfa9d4fdbafd82dc517250b93269c94/stl/inc/vector#L1832

MAG-ElliotMorris:
Working in OKO. See here the highlighted text is a new debug string from this change. 
image

Library/include/CSP/Common/Array.h
MAG-OR: thanks!

MAG-mv (approved):

MAG-AdamThorn (approved):

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -87,222 +91,183 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient)
...
+ * FireEnterSpaceEvent
+ * RefreshMultiplayerScopes
+ * AssertRequestSuccessOrErrorFromErrorCode (RefreshMultiplayerScopes Validation)
+ * ReportSuccess
+ * InvokeIfExceptionInChain (Handle any errors from the above Assert methods in chain, resets state)
+ */
+void SpaceSystem::EnterSpace(const String& SpaceId, NullResultCallback Callback)
+{
+    CSP_LOG_MSG(csp::systems::LogLevel::Log, "SpaceSystem::EnterSpace");
+    constexpr auto GENERIC_UNKNOWN_ERROR_CODE = 500;
```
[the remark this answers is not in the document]
MAG-ElliotMorris: I wish i'd known about this going in. 👍 

https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/4511907aef6e4e72496d325cd36fbc49154aa200
and
https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/fff06951a442542dc87faedaefb85015c06375bf

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -87,222 +91,183 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient)
...
-                });
-        }
+auto SpaceSystem::RefreshMultiplayerScopes()
+{
+    return [this](const SpaceResult& SpaceResult)
+    {
+        CSP_LOG_MSG(csp::systems::LogLevel::Log, "SpaceSystem::RefreshMultiplayerScopes");
+
+        /* Refresh the multiplayer connection to force the scopes to change */
+        /* This is wrapping an as yet non-refactored method that uses nested callbacks, hence the event, and shared pointer for lifetime */
```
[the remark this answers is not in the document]
MAG-ElliotMorris: I think this reads correctly? Rewrote it to try to be clearer anyway.
https://github.com/magnopus-opensource/connected-spaces-platform/pull/616/commits/8236b20251caaa7fc11fbaa1b177d900a4525bf0

Did you mean a spelling mistake or something? Can't quite spot what you're getting at.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -87,222 +91,183 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient)
...
-
-            CSP_LOG_MSG(csp::systems::LogLevel::Log, " EnterSuccess:");
-            CSP_LOG_MSG(csp::systems::LogLevel::Log, std::to_string(EnterSuccess).c_str());
+                            CSP_LOG_MSG(csp::systems::LogLevel::Log, " MultiplayerConnection->StartListening success");
 
-            // If the user is not the owner check are they a moderator
-            if (!EnterSuccess)
-            {
-                EnterSuccess = systems::SpaceSystemHelpers::IdCheck(UserId, RefreshedSpace.ModeratorIds);
+                            // TODO: Support getting errors from RetrieveAllEntities
```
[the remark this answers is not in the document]
MAG-ElliotMorris: This is pre-existing moved code, so I'm not quite sure what this TODO is referring to personally. I wouldn't want to log a ticket without the context.

Library/src/Systems/Spaces/SpaceSystem.cpp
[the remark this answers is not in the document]
MAG-mv: Reading over it again, it did make sense. However, new comment is clearer

MAG-mv (approved):

PR #617, opened 2025-02-11
[NT-0] test: Add googlemock, update googletest

MAG-ElliotMorris:
Add googlemock to thirdparty, as well as updating googletest, as they are interdependent.

Our googletest/googlemoc libs are directly commited, so these artifacts are from local builds. This is not a good workflow, we should endevour to remove (and purge) these libs from the repo ... at some point. For sure this stops anyone not on windows running our tests, and I think it might even stop people on different versions of VS running them. :(

This will allow mock-based testing, probably the best way to do actual unit-testing in our ecosystem.

(Stemmed from this larger change, but is independent and useful, so individual PR : https://github.com/magnopus-opensource/connected-spaces-platform/pull/616)


Tests/premake5.lua
```diff
@@ -24,6 +24,7 @@ if not Tests then
         externalincludedirs { 
             "%{prj.location}/src",
             "%{wks.location}/ThirdParty/googletest/include",
+			"%{wks.location}/ThirdParty/googlemock/include",
```
MAG-OR, line 27: [consider] how come the indentation is so weird?

MAG-OR (dismissed):

MAG-AdamThorn (approved):

Tests/premake5.lua
MAG-ElliotMorris: It's not on my screen, tabs and spaces again, (you can set github tabs width to make this go away if you like)

Clang format dosen't apply to non source files, so we'll still have this problem here. Would be nice to deal with it, but .lua files are less of a concern to me. Will [fix] nonetheless.

MAG-OR (approved):
thanks for addressing!

Tests/premake5.lua
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/617/commits/91d08eceb357e9dbe270ce1e7793a06db649380a

MAG-mv (approved):

PR #618, opened 2025-02-12
[OB-3959] refac: Update delete sequence to use appropriate endpoint

MAG-ChristopherAtkinson:
The delete sequence has been updated to a single api call, this has been marked as refac, however, this is related to [OB-3959] and should help reduce the delay in a delete sequence.

[OB-3959]: https://magnopus.atlassian.net/browse/OB-3959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -629,47 +629,10 @@ void SpaceSystem::DeleteSpace(const csp::common::String& SpaceId, NullResultCall
 {
     CSP_PROFILE_SCOPED();
 
-    const String InGroupId = SpaceId;
-
-    // Delete space metadata AssetCollection first, as users without super-user will not be able to do so after the space is deleted
```
MAG-OR, line 634: [question] can you confirm whether the space metadata and thumbnail deletion is still done with the new endpoint?

MAG-OR (commented):
It would be nice if a test showing that the slowness problem is solved was added. Is this doable?

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ChristopherAtkinson: This should be the case according to the swagger documentation: 

> Locates the Spaces by their unique groupIds represented here as spaceIds and the associated objects that belong to it and deletes it from the data store. This Function will search for items related to a spaceId even if the spaceId itself has already been deleted.

MAG-OR (approved):

PR #621, opened 2025-02-20
[OF-1523] feat: Add getter for ids of users who accepted invites

MAG-OR:
- Create getter for the IDs of users who accepted an invite. It works on the same principle as GetPendingInvites, except that it retrieves the user ID, not the email address.
- Improve pending invites test and add a comment
- Add a test for the accepted invites
- Update csp-services to the latest version so it contains the accepted endpoint wrapper

### DIFF
--- a/Tests/src/PublicAPITests/SpaceSystemTests.cpp
+++ b/Tests/src/PublicAPITests/SpaceSystemTests.cpp
@@ -1849,7 +1849,15 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, GetPendingUserInvitesTest)
     char UniqueSpaceName[256];
     SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
 
-    const char* TestUserEmail = "testnopus.pokemon@magnopus.com";
+    // It is not possible to record pending invites and see them decrement as they are accepted,
+    // because the invites are sent by email and have to be accepted by clicking a link in them.
+    // The test suite does not have the capability to open emails and click links in them.
+    // The workaround is to test each separately.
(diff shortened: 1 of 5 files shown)


Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -1882,6 +1893,90 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, GetPendingUserInvitesTest)
...
+    auto* SpaceSystem = SystemsManager.GetSpaceSystem();
+
+    const char* TestSpaceName = "OLY-UNITTEST-SPACE-REWIND";
+    const char* TestSpaceDescription = "OLY-UNITTEST-SPACEDESC-REWIND";
+
+    char UniqueSpaceName[256];
+    SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+    const char* TestEmailLinkUrl = "https://dev.magnoverse.space/";
+    const char* TestSignupUrl = "https://dev.magnoverse.space/";
```
MAG-ChristopherAtkinson, line 1912: [Question] should this be using EndpointBaseURI() as just from looking can imagine this would cause some issue with local services.

MAG-ChristopherAtkinson (approved):
The changes look good to me, added a question just in regards to local services. Glad to see the introduction of additional test for alternative execution path.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-OR: this address is used in several test cases in this test suite and local services are not part of the scope for this story. these will need to be addressed in a dedicated story.

MAG-mv (approved):

PR #624, opened 2025-02-26
[OB-3917] fix! : change all int_64t exposed script parameters to int_32t

MAG-OR:
This is because our current version of quickjs does not support BigInt (64 bits ints), which led to an error if using BigInt in a script on Web, or an unsupported conversion warning if using a Number.

Now that we expose 32 bits ints, the web scripts no longer show an unsupported conversion warning when using a Number.

Breaking: This change requires the web client team to update their script interface to match the new types.

### DIFF
--- a/Library/src/Multiplayer/Script/ComponentBinding/AvatarSpaceComponentScriptInterface.cpp
+++ b/Library/src/Multiplayer/Script/ComponentBinding/AvatarSpaceComponentScriptInterface.cpp
@@ -33,14 +33,14 @@ DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, AvatarId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, UserId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, AgoraUserId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, CustomAvatarUrl);
-DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::AvatarState, int64_t, State);
-DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, int64_t, int64_t, AvatarMeshIndex);
+DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::AvatarState, int32_t, State);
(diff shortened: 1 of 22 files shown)


Library/src/Multiplayer/Script/ComponentBinding/AvatarSpaceComponentScriptInterface.cpp
```diff
@@ -33,14 +33,14 @@ DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, AvatarId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, UserId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, AgoraUserId);
 DEFINE_SCRIPT_PROPERTY_STRING(AvatarSpaceComponent, CustomAvatarUrl);
-DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::AvatarState, int64_t, State);
-DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, int64_t, int64_t, AvatarMeshIndex);
+DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::AvatarState, int32_t, State);
+DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, int64_t, int32_t, AvatarMeshIndex);
```
MAG-ElliotMorris, line 37: [consider] This casts back to `uint32_t` anyway in `AvatarSpaceComponent::SetAvatarMeshIndex`. Can you  change the sig of the methods to `int32_t` and avoid the (double) redundant casting?

MAG-ElliotMorris (approved):
Lookin' good. Just one consider and it's not a big deal.

MAG-ChristopherAtkinson (approved):
Great work, approved.

Library/src/Multiplayer/Script/ComponentBinding/AvatarSpaceComponentScriptInterface.cpp
MAG-OR: That's a very interesting question, and something I looked at for the parameter this bug was originally about (playbackState), back when I thought that my change hadn't worked (i.e. before I realised I needed a change on the web side as well, for the warning to be resolved).
If I were to change the setters to use 32 bits instead of using 64 bits and casting afterwards, I would have to implement a 32 bits integer in ReplicatedValues, alongside the 64 bits one. That means a new ReplicatedValueType for 32 bits ints, a new set of setter/getter, and adding the 32 bits int in every switch on ReplicatedValueType. It seemed to me that it was a lot of effort for little return, but happy to be convinced otherwise.

PR #630, opened 2025-03-03
[OF-1568] feat: Populate timestamps internally

MAG-mv:
- Added UTC timestamp to internal DateTime class
- Created/Edited timestamps are now added to conversations internally
- Tests have been added

### DIFF
--- a/Library/src/Systems/Conversation/ConversationSystemInternal.cpp
+++ b/Library/src/Systems/Conversation/ConversationSystemInternal.cpp
@@ -9,6 +9,7 @@
 #include "Multiplayer/EventSerialisation.h"
 
 #include "CallHelpers.h"
+#include "Common/DateTime.h"
 #include "Debug/Logging.h"
 #include "Systems/ResultHelpers.h"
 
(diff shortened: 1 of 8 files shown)


Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -66,9 +67,12 @@ void ConversationSystemInternal::CreateConversation(const common::String& Messag
     const common::String& UserId = UserSystem->GetLoginState().UserId;
     const common::String& SpaceId = CurrentSpace.Id;
 
+    const common::DateTime CurrentTime = common::DateTime::UtcTimeNow();
+    const common::String CurrentTimeString = CurrentTime.GetUtcString();
```
MAG-ElliotMorris, line 71: [Consider] Do you think we should do this closer to the point of set? This is all done prior to callbacks which could take time. It's an extremely minor point, just wondering if in the future there will ever be comparisons between conversation timestamps and asset collection creation timestamps, or multiplayer event send timestamps, or anything like that? Probably just tilting at windmills.

MAG-ElliotMorris (commented):
[Question] Have you considered/tried this with spaces with older conversations? Smells like there could be a backwards compatibility thing, though I'm not sure. (Didn't think too hard about it)

MAG-OR (approved):
Looks great to me

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-mv: I did consider this. The problem is, if I did it after the creation callback, I would then have to do another api call to update the metadata with the new time value.

MAG-mv:
> [Question] Have you considered/tried this with spaces with older conversations? Smells like there could be a backwards compatibility thing, though I'm not sure. (Didn't think too hard about it)

@MAG-ElliotMorris We dont care about backwards compatibilty for this feature

PR #632, opened 2025-03-05
[OF-1557] chore: Update CSP-Services to v2.0-d20250224.rbdaf75117.Rel…

MAG-ElliotMorris:
…ease

Update services to get new tag api on GroupDTO.


MAG-ElliotMorris:
Ah, this change is actually breaking, didn't realize. Closing PR, will have to be done alongside CSP library changes.

PR #633, opened 2025-03-05
[OF-1537] chore: remove organization system, tests, and creds file

MAG-OR:
Remove the Organization System
Remove the Organization tests
Remove the need for the credentials file to run the test suite Remove the logic to load the contents of the credentials file in the test suite


Library/src/Systems/Spaces/Space.cpp
```diff
@@ -111,11 +111,6 @@ void GroupDtoToSpace(const chs_users::GroupDto& Dto, csp::systems::Space& Space)
             Space.BannedUserIds[i] = users[i];
         }
     }
-
-    if (Dto.HasOrganizationId())
```
MAG-OR, line 115: checked with Sam, and we're leaving the Organization ID field in the DTO. we're just not using it in CSP.

Library/docs/source/learn/overview/mcs.md
```diff
@@ -24,9 +24,9 @@ User Accounts & Management handles user authentication, authorization, and profi
 
 Security & Access Control is responsible for protecting data and ensuring that only authorized users can access certain parts of the platform. This component uses various security measures, such as encryption and permissions, to safeguard sensitive information and maintain privacy.
 
-**3\. Groups, Organizations, and Social Connections**
+**3\. Groups and Social Connections**
 
-This component facilitates social interactions and organizational structures within CSP. Users can form groups, join organizations, and connect with others to collaborate and share experiences. It helps build a community and enhances user engagement.
```
MAG-ElliotMorris, line 29: [Comment] Great work remembering to update the guide-doc, we often forget to check.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -638,7 +638,7 @@ void SpaceSystem::GetSpacesByAttributes(const Optional<bool>& InIsDiscoverable,
         std::nullopt, // AutoModerator
         RequiresInvite, // RequiresInvite
         IsArchived, // Archived
-        std::nullopt, // OrganizationIds
```
MAG-ElliotMorris, line 641: [Nit] These comments are here because it's hard to keep track of the giant number or arguments. I think it saying `OrganizationIds (no longer used)` would be superior.

MAG-ElliotMorris (approved):

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-OR: done in 7e062cc5e36c2977189ebc3599a9b1def0497a57

MAG-mv (approved):

PR #634, opened 2025-03-07
[OF-1572] chore: remove credentials file support from MultiPlayerTestRunner

MAG-OR:
- Make email and password compulsory arguments
- Remove support for the credentials file backup mechanism
- Create temporary test accounts before logging in

### DIFF
--- a/MultiplayerTestRunner/include/ErrorCodes.h
+++ b/MultiplayerTestRunner/include/ErrorCodes.h
@@ -24,7 +24,7 @@ namespace MultiplayerTestRunner::ErrorCodes
 {
 constexpr const int SUCCESS = 0;
 constexpr const int FAILED_TO_CREATE_USER = 100;
-constexpr const int UNUSED_ERROR_CODE = 101;
+constexpr const int INVALID_LOGIN_PASSWORD = 101;
 constexpr const int FAILED_TO_LOGIN = 102;
 constexpr const int INVALID_TEST_SPECIFIER = 103;
 constexpr const int OTHER_EXCEPTION = 104; // For any exception that isn't thrown explicitly by the Multiplayer Test Runner
--- a/MultiplayerTestRunner/include/ErrorCodes.h
+++ b/MultiplayerTestRunner/include/ErrorCodes.h
@@ -23,8 +23,8 @@
 namespace MultiplayerTestRunner::ErrorCodes
 {
 constexpr const int SUCCESS = 0;
-constexpr const int COULD_NOT_FIND_CREDENTIALS_FILE = 100;
-constexpr const int MALFORMED_CREDENTIALS_FILE = 101;
+constexpr const int FAILED_TO_CREATE_USER = 100;
+constexpr const int UNUSED_ERROR_CODE = 101;
 constexpr const int FAILED_TO_LOGIN = 102;
 constexpr const int INVALID_TEST_SPECIFIER = 103;
 constexpr const int OTHER_EXCEPTION = 104; // For any exception that isn't thrown explicitly by the Multiplayer Test Runner
--- a/MultiplayerTestRunner/src/CLIArgs.cpp
+++ b/MultiplayerTestRunner/src/CLIArgs.cpp
@@ -33,18 +33,11 @@ constexpr int DEFAULT_TIMEOUT_IN_SECONDS = 30;
  * For optional arguments, if empty data is provided, will populate them with defaults.
  * SpaceId will not be defaulted, and will be left empty if empty, in the expectation a space is about to be created and the value set externally.
  */
-CLIArgs::RunnerSettings ValidateInvocationArgs(const std::string& TestIdentifier, std::optional<std::string> LoginEmail,
-    std::optional<std::string> LoginPassword, const std::optional<std::string>& Endpoint, const std::optional<std::string>& SpaceId,
-    const std::optional<int>& TimeoutInSeconds)
+CLIArgs::RunnerSettings ValidateInvocationArgs(const std::string& TestIdentifier, std::string LoginEmail, std::string LoginPassword,
+    const std::optional<std::string>& Endpoint, const std::optional<std::string>& SpaceId, const std::optional<int>& TimeoutInSeconds)
 {
-    if (!LoginEmail.has_value() || !LoginPassword.has_value())
-    {
-        // If only one of the email/password pair has been provided, error out
-        throw Utils::ExceptionWithCode(MultiplayerTestRunner::ErrorCodes::CLI_PARSE_ERROR, "Both email and password must be provided together.");
-    }
-
     CLIArgs::RunnerSettings settings;
-    // Test identifiers need to be valid, and are not optional. An incorrect one is grounds to abort.
+    // Test identifier, emails and passwords need to be valid, and are not optional. An incorrect one is grounds to abort.
     try
     {
         settings.TestIdentifier = MultiplayerTestRunner::TestIdentifiers::StringToTestIdentifier(TestIdentifier);
@@ -54,7 +47,15 @@ CLIArgs::RunnerSettings ValidateInvocationArgs(const std::string& TestIdentifier
         // The reason this rethrow is here is an annoying quirk of `StringToTestIdentifier` being a public method.
         throw Utils::ExceptionWithCode(MultiplayerTestRunner::ErrorCodes::INVALID_TEST_SPECIFIER, exception.what());
     }
-    settings.LoginEmailAndPassword = std::make_pair(LoginEmail.value(), LoginPassword.value());
+    try
+    {
+        settings.LoginEmailAndPassword = std::make_pair(LoginEmail, LoginPassword);
+    }
+    catch (std::exception& exception)
+    {
+        // If the email/password pair has not been provided, error out
+        throw Utils::ExceptionWithCode(MultiplayerTestRunner::ErrorCodes::INVALID_LOGIN_PASSWORD, exception.what());
+    }
     settings.Endpoint = Endpoint.value_or(DEFAULT_TEST_ENDPOINT);
     settings.TimeoutInSeconds = TimeoutInSeconds.value_or(DEFAULT_TIMEOUT_IN_SECONDS);
     settings.SpaceId = SpaceId; // This value is not defaulted, as an empty value means a space is about to be created, and this value will be set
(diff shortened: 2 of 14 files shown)


MultiplayerTestRunner/src/Utils.h
```diff
@@ -35,6 +37,22 @@ namespace common
 
 namespace Utils
 {
+
+#if defined CSP_WINDOWS
+#define SPRINTF sprintf_s
+#else
+#define SPRINTF sprintf
+#endif
```
MAG-ElliotMorris, line 45: [Consider] Could you assemble strings without `sprintf` (use the + operator, or streams, or even bring in `libfmt`) rather than bringing this nasty cross platform switch over to the test runner?

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -85,17 +76,13 @@ RunnerSettings ProcessCLI(int argc, char* argv[])
...
 
     App.add_option("-t,--test", TestIdentifier, "The test to run. See `include/TestIdentifiers.h` for available options.")->required();
-    App.add_option("-e,--email", LoginEmail,
-        "Login email for the test CHS account. If not set, the application will attempt to source this from a `test_account_creds.txt` "
-        "next to the binary.");
-    App.add_option("-p,--password", LoginPassword,
-        "Password for the test CHS account. If not set, the application will attempt to source this from a `test_account_creds.txt` next "
-        "to the binary.");
+    App.add_option("-e,--email", LoginEmail, "Login email for the test CHS account.`")->required();
+    App.add_option("-p,--password", LoginPassword, "Password for the test CHS account.")->required();
```
MAG-ElliotMorris, line 85: [Fix] This isn't documented well enough considering the account creation behaviour. We're at least sometimes creating accounts, does this expect pre-existing accounts, or is it going to take the emails and create accounts (in which case, you also should state what happens if the accounts already exist/cannot be created)

MultiplayerTestRunner/src/LoginRAII.cpp
```diff
@@ -59,11 +59,16 @@ csp::common::String LogIn(csp::systems::UserSystem& UserSystem, const csp::commo
...
-LoginRAII::LoginRAII(const std::string& AccountLoginEmail, const std::string& AccountPassword)
+LoginRAII::LoginRAII(const std::string& AccountLoginEmail, const std::string& AccountPassword, bool CreateAccount /* = true */)
 {
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto& UserSystem = *SystemsManager.GetUserSystem();
 
+    if (CreateAccount)
+    {
+        Utils::CreateTestUser(AccountLoginEmail, AccountPassword);
+    }
```
MAG-ElliotMorris, line 70: [Fix] If i'm reading this correctly this is the default?

Whilst i'm fine with this being an additional option, I don't think CSP tests should ever have the MultiplayerTestRunner create their accounts for them, simply because if we've learned anything, it's that off-process logic is hard to debug, especially over teamcity logs.

I would do 1 of 2 things

1. (better for stability & simplicity) Do not have the MultiplayerTestRunner handle creating accounts at all, make it mandatory they are passed in
2. (better for this being a utility) Provide a way in the CLI to flag that you intend to create an account, not login, but have our internal tests always create accounts externally.

I prefer option 1, because it removes a potential error return that may popup in our tests and cause us to have to debug through, but either would be acceptable.

MultiplayerTestRunner/src/LoginRAII.cpp
```diff
@@ -59,11 +59,16 @@ csp::common::String LogIn(csp::systems::UserSystem& UserSystem, const csp::commo
 }
 } // namespace
 
-LoginRAII::LoginRAII(const std::string& AccountLoginEmail, const std::string& AccountPassword)
+LoginRAII::LoginRAII(const std::string& AccountLoginEmail, const std::string& AccountPassword, bool CreateAccount /* = true */)
```
MAG-ElliotMorris, line 62: [Fix] Needs an internal unit test testing both the `true` and `false` behaviour of `CreateAccount`.

MAG-AdamThorn, line 62: [Consider] Handle create account separately so you do need to pass in an optional CreateAccount param and have the LoginRAII method also handle account creation.

MultiplayerTestRunner/include/ErrorCodes.h
```diff
@@ -23,8 +23,8 @@
 namespace MultiplayerTestRunner::ErrorCodes
 {
 constexpr const int SUCCESS = 0;
-constexpr const int COULD_NOT_FIND_CREDENTIALS_FILE = 100;
-constexpr const int MALFORMED_CREDENTIALS_FILE = 101;
+constexpr const int FAILED_TO_CREATE_USER = 100;
+constexpr const int UNUSED_ERROR_CODE = 101;
```
MAG-ElliotMorris, line 27: [nit] I'd leave it as the old name, but comment it out 
```suggestion
// constexpr const int MALFORMED_CREDENTIALS_FILE = 101; --Credentials file no longer used
```

MultiplayerTestRunner/src/CLIArgs.cpp
```diff
@@ -28,28 +28,19 @@ constexpr int DEFAULT_TIMEOUT_IN_SECONDS = 30;
...
-        LoginPassword = credentials.DefaultLoginPassword;
-    }
-    else if (!LoginEmail.has_value() || !LoginPassword.has_value())
-    {
-        // If only one of the email/password pair has been provided, error out entirely, it's probably a mistake
-        throw Utils::ExceptionWithCode(MultiplayerTestRunner::ErrorCodes::CLI_PARSE_ERROR,
-            "Both email and password must be provided together. Missing one likely indicates a mistake. Omit both if you "
-            "wish to use the credentials file.");
+        // If only one of the email/password pair has been provided, error out
+        throw Utils::ExceptionWithCode(MultiplayerTestRunner::ErrorCodes::CLI_PARSE_ERROR, "Both email and password must be provided together.");
```
MAG-ElliotMorris, line 43: [fix] We should update the interface types to match the new source-of-data.
Both `LoginEmail` and `LoginPassword` are mandatory now, so should not be optionals, and do not need to be validated, (as CLI11 does it for us)

MultiplayerTestRunner/src/Utils.h
```diff
@@ -35,6 +37,22 @@ namespace common
...
+
+std::string GetUniqueString();
+
+const char GeneratedTestAccountEmailFormat[] = "testnopus.pokemon+%s@magnopus.com";
+const char GeneratedTestAccountPassword[] = "3R{d2}3C<x[J7=jU";
+
+csp::systems::Profile CreateTestUser(std::string UniqueEmail, std::string Password, bool AgeVerified = true,
+    csp::systems::EResultCode ExpectedResultCode = csp::systems::EResultCode::Success,
+    csp::systems::ERequestFailureReason ExpectedResultFailureCode = csp::systems::ERequestFailureReason::None);
```
MAG-ElliotMorris, line 54: [Nit] Use the default arg values, rather than spelling them all out.

MAG-ElliotMorris (changes requested):

MultiplayerTestRunner/include/ErrorCodes.h
MAG-OR: I think I ended up using it, so not actioning for now. will do if it turns out I don't need it

MultiplayerTestRunner/src/CLIArgs.cpp
MAG-OR: yeah I realised I removed one pair of optional keywords and forgot about 10 or more others elsewhere, and indeed this part disappears. fix to come

MultiplayerTestRunner/src/CLIArgs.cpp
MAG-OR: Updating to: `Login email for the test CHS account. The account must already exist. If it does not exist, errors out.`
taking into account the design change mentioned below.
I'm still checking a few things before I'm ready to commit, so feel free to propose a better phrasing

MultiplayerTestRunner/src/LoginRAII.cpp
MAG-OR: this comment is outdated following the redesign mentioned below (the boolean no longer exists)

MultiplayerTestRunner/src/LoginRAII.cpp
MAG-OR: As discussed, we are going for 1. Option 2 may be considered later on, if we decide to use it as a tool. for now, there isn't really a use-case

MultiplayerTestRunner/src/Utils.h
MAG-OR: addressed in 8ae555e14ff1d227b985b99e2fb9736cbe1da3d9

MultiplayerTestRunner/src/CLIArgs.cpp
MAG-OR: addressed in 57df9a92e98fc9a533ef025faa66902ee4d94831

MultiplayerTestRunner/src/LoginRAII.cpp
MAG-OR: addressed in 8ae555e14ff1d227b985b99e2fb9736cbe1da3d9

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
```diff
@@ -105,19 +103,15 @@ TEST(CLITest, WhenNoEmailThenError)
     }
     catch (const Utils::ExceptionWithCode& Exception)
     {
-        // If there's no credentials file, we'll throw an error.
         EXPECT_EQ(Exception.ErrorCode, MultiplayerTestRunner::ErrorCodes::CLI_PARSE_ERROR);
-        EXPECT_EQ(std::string(Exception.what()),
-            std::string("Both email and password must be provided together. Missing one likely indicates a mistake. Omit both if you "
-                        "wish to use the credentials file."));
+        EXPECT_EQ(std::string(Exception.what()), std::string("Both email and password must be provided together."));
```
MAG-ElliotMorris, line 107: [Question] Do these still pass? From my reading of the code, this error is no longer emitted (because we don't need to do the validation anymore because these are mandatory)

If i'm right, we should [fix] this by removing the `WhenNoEmail` and `WhenNoPassword` and replacing them with tests similar to `TestIdentifierRequired`

Note: These are pure internal unit tests for development of the runner, and don't run from the main CSP test suite, you need to define `RUN_MULTIPLAYER_RUNNER_TESTS` and launch the test runner project to run these. (Maybe they should be invoked on CI though, come to think of it.)

MAG-ElliotMorris:
All the changes look good to me, just one thing with the internal tests given the new arrangement of mandatory CLI args.

MultiplayerTestRunner/src/InternalUnitTests/Test_CLI.cpp
MAG-OR: yes of course it fails. will address

MAG-OR: image

Fixed in 7a585b8bba3c2540b08f50f0fc07b08c18c5c922.

MAG-ElliotMorris (approved):

MAG-ElliotMorris:
> oh I assumed all the test macros were removed

The reason the MultiplayerTestRunner still has a macro is because it serves as its own test project, rather than having a separate test project. Needs to be a way to tell it if it's running in test mode or not.

MAG-ChristopherAtkinson (approved):
Changes look good to me, glad to see credentials as mandatory arguments.

PR #636, opened 2025-03-11
[OF-1576] feat: Prevent clients from deleting others messages

MAG-mv:
UpdateConversation, DeleteMessage, and UpdateMessage will now fail if the clients attempting these operations aren't the client who created them.

### DIFF
--- a/Library/src/Systems/Conversation/ConversationSystemInternal.cpp
+++ b/Library/src/Systems/Conversation/ConversationSystemInternal.cpp
@@ -46,6 +46,25 @@ namespace
 
         return true;
     }
+
+    bool EnsureUserHasPermission(const common::String& UserId, const common::String& ConversationUserId, bool IsConversation)
+    {
+        if (UserId != ConversationUserId)
+        {
+            if (IsConversation)
+            {
+                CSP_LOG_ERROR_MSG("User does not have permission to modify this conversation.");
+            }
+            else
+            {
+                CSP_LOG_ERROR_MSG("User does not have permission to modify this message.");
+            }
+
+            return false;
+        }
+
+        return true;
+    }
 }
 
 ConversationSystemInternal::ConversationSystemInternal(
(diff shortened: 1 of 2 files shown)


Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -46,6 +46,25 @@ namespace
...
+        if (UserId != ConversationUserId)
+        {
+            if (IsConversation)
+            {
+                CSP_LOG_ERROR_MSG("User does not have permission to modify this conversation.");
+            }
+            else
+            {
+                CSP_LOG_ERROR_MSG("User does not have permission to modify this message.");
+            }
```
MAG-ElliotMorris, line 61: [Comment] This is the sort of stuff I wish we had `libfmt` for
```suggestion
           CSP_LOG_ERROR_MSG(libfmt::fmt("User does not have permission to modify this {}.", IsConversation ? "conversation" : "message"));
```

Doable without it, but more clumsy and often creates unnecessary copies.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -113,39 +132,54 @@ void ConversationSystemInternal::CreateConversation(const common::String& Messag
...
         };
 
-        this->AssetSystem->DeleteAssetCollection(ConversationAssetCollection, DeleteAssetCollectionCallback);
+        multiplayer::MessageInfo MessageInfo(ConversationId, true, "", "", "", "", "");
+        SendConversationEvent(multiplayer::ConversationEventType::DeleteConversation, MessageInfo, EventBus, SignalRCallback);
     };
 
-    multiplayer::MessageInfo MessageInfo(ConversationId, true, "", "", "", "", "");
-    SendConversationEvent(multiplayer::ConversationEventType::DeleteConversation, MessageInfo, EventBus, SignalRCallback);
+    AssetSystem->GetAssetCollectionById(ConversationId, GetConversationCallback);
```
MAG-ElliotMorris, line 182: [Comment] I understand the desire to just do this quickly and get it out of the way, and I agree ... would have been lovely as a continuation though.

Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
```diff
@@ -914,4 +915,204 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationTests, ConversationComponentSecondClientE
...
+            MessageId = Result.GetMessageInfo().MessageId;
+        }
+
+        // Logout
+        auto [ExitSpaceResult] = AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
+        LogOut(UserSystem);
+    }
+
+    // Ensure component data has been written to database by chs before entering the space again
+    std::this_thread::sleep_for(7s);
```
MAG-ElliotMorris, line 995: [Question] 
Aw man ... is this still really a thing, this _sucks_.
Is CHS aware of this do you know? I'd want to log a ticket if not.

MAG-ElliotMorris (approved):

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -113,39 +132,54 @@ void ConversationSystemInternal::CreateConversation(const common::String& Messag
...
-
-            INVOKE_IF_NOT_NULL(Callback, MakeInvalid<NullResult>());
             return;
         }
 
-        // 2. Delete the asset colleciton associated with this conversation
-        AssetCollection ConversationAssetCollection;
-        ConversationAssetCollection.Id = ConversationId;
+        // Ensure client has correct permissions to delete the conversation
+        multiplayer::MessageInfo Info
```
MAG-AdamThorn, line 144: [Question] Where is this `Info` object being used? You seem to be missing the call to `EnsureUserHasPermission`.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-AdamThorn: I agree, do we have time?

Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
```diff
@@ -914,4 +915,204 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationTests, ConversationComponentSecondClientE
...
+        EntitySystem->ProcessPendingEntityOperations();
+
+        // Create conversation
+        {
+            auto [Result] = AWAIT(ConversationComponent, CreateConversation, "TestMessage");
+
+            EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+            EXPECT_TRUE(Result.GetValue() != "");
+
+            ConversationId = Result.GetValue();
```
MAG-AdamThorn, line 975: {Question} Is this here for debugging? We don't use this `ConversationId` value.

Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
MAG-mv: They are. I dont think its something theyre planning to fix

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-mv: This would be a nice to have

MAG-AdamThorn (changes requested):

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-mv: The continuations wasnt merged when i started this feature, which is why I didnt do it then

Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
MAG-ElliotMorris: We should re-raise it at some point. They don't need to fix it as much as give us a way to know when the data has been written.

MAG-mv: That would be great, although they may argue against this because of additional bandwidth. I guess we could bind to an event for specific scenarious

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-mv: Fixed in [789c0d6](https://github.com/magnopus-opensource/connected-spaces-platform/pull/636/commits/789c0d6b34211925ce1bcb70e92fe81f59cdd770)

Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
MAG-mv: Fixed in: [75e4536](https://github.com/magnopus-opensource/connected-spaces-platform/pull/636/commits/75e453607b83be3c4da75e3021125e9ebcd45f58)

MAG-AdamThorn (approved):

PR #637, opened 2025-03-11
[OF-1577] feat!: Add tag controls to Space, removing them from the metadata prototype.

MAG-ElliotMorris:
Tags are now on spaces directly, rather than bundled with the metadata, as they were before. 

This seems like a more natural design to me.

The consequences of this that you'll see in this PR:
- Any place where metadata was set, tags used to also be set ... they are no longer
- Tags are fundamental elements of a space now, much like the name and attributes
- Tags can be gotten from the `Space` object, and updated with `UpdateSpace`

Updating the services is also done in this PR, as it was a breaking change and needed code changes around it.

### DIFF
--- a/Library/include/CSP/Common/Array.h
+++ b/Library/include/CSP/Common/Array.h
@@ -24,6 +24,7 @@
 #include <initializer_list>
 #include <stdexcept>
 #include <utility>
+#include <vector>
 
 #ifndef CSP_DISABLE_OVERFLOW_CHECKING
 #ifdef _MSC_VER
@@ -68,20 +69,6 @@ public:
         }
     }
 
-    /// @brief Constructs an array from a buffer.
-    /// @param Buffer const T* : Pointer to the beginning of the buffer
-    /// @param Size size_t : Number of elements in the buffer
-    CSP_NO_EXPORT Array(const T* Buffer, size_t Size)
-        : ArraySize(0)
-        , ObjectArray(nullptr)
-    {
-        if (Buffer != nullptr && Size > 0)
-        {
-            AllocArray(Size);
-            memcpy(ObjectArray, Buffer, Size * sizeof(T));
-        }
-    }
-
     /// @brief Copy constructor.
     /// @param Other const Array<T>& Other
     CSP_NO_EXPORT Array(const Array<T>& Other)
@@ -97,6 +84,24 @@ public:
             for (size_t i = 0; i < ArraySize; i++)
             {
                 ObjectArray[i] = Other.ObjectArray[i];
+            }
+        }
+    }
+
+    /// @brief Constructs an array from std::vector.
+    /// @param Vector std::Array : Elements to construct the array from. Elements are copied.
+    /// @pre T must be copyable.
+    CSP_NO_EXPORT Array(const std::vector<T>& Vector)
+        : ArraySize(0)
+        , ObjectArray(nullptr)
+    {
+        if (Vector.size() > 0)
+        {
+            AllocArray(Vector.size());
+
+            for (size_t i = 0; i < Vector.size(); ++i)
+            {
+                ObjectArray[i] = Vector[i];
             }
         }
     }
@@ -216,6 +221,10 @@ public:
 
         return std::move(Result);
     }
+
+    /// @brief Returns a copy of this Array as a std::vector
+    /// @return List<T>
+    CSP_NO_EXPORT std::vector<T> ToStdVector() const { return std::vector<T>(begin(), end()); }
 
 private:
     /// @brief Allocates memory for the array.
--- a/Library/include/CSP/Common/Array.h
+++ b/Library/include/CSP/Common/Array.h
@@ -24,7 +24,6 @@
 #include <initializer_list>
 #include <stdexcept>
 #include <utility>
-#include <vector>
 
 #ifndef CSP_DISABLE_OVERFLOW_CHECKING
 #ifdef _MSC_VER
@@ -84,24 +83,6 @@ public:
             for (size_t i = 0; i < ArraySize; i++)
             {
                 ObjectArray[i] = Other.ObjectArray[i];
-            }
-        }
-    }
-
-    /// @brief Constructs an array from std::vector.
-    /// @param Vector std::Array : Elements to construct the array from. Elements are copied.
-    /// @pre T must be copyable.
-    CSP_NO_EXPORT Array(const std::vector<T>& Vector)
-        : ArraySize(0)
-        , ObjectArray(nullptr)
-    {
-        if (Vector.size() > 0)
-        {
-            AllocArray(Vector.size());
-
-            for (size_t i = 0; i < Vector.size(); ++i)
-            {
-                ObjectArray[i] = Vector[i];
             }
         }
     }
@@ -221,10 +202,6 @@ public:
 
         return std::move(Result);
     }
-
-    /// @brief Returns a copy of this Array as a std::vector
-    /// @return List<T>
-    CSP_NO_EXPORT std::vector<T> ToStdVector() const { return std::vector<T>(begin(), end()); }
 
 private:
     /// @brief Allocates memory for the array.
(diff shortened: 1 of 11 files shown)


Library/include/CSP/Common/Array.h
```diff
@@ -68,20 +69,6 @@ template <typename T> class CSP_API Array
         }
     }
 
-    /// @brief Constructs an array from a buffer.
-    /// @param Buffer const T* : Pointer to the beginning of the buffer
-    /// @param Size size_t : Number of elements in the buffer
-    CSP_NO_EXPORT Array(const T* Buffer, size_t Size)
```
MAG-ElliotMorris, line 74: [Comment] Deleted due to not being compatible with types in its own namespace, and in a very dangerous way. (Shallow copies, double deletes). Wasn't used anywhere anyhow.

Library/include/CSP/Common/Array.h
```diff
@@ -101,6 +88,24 @@ template <typename T> class CSP_API Array
         }
     }
 
+    /// @brief Constructs an array from std::vector.
+    /// @param Vector std::Array : Elements to construct the array from. Elements are copied.
```
MAG-ElliotMorris, line 92: [fix] 
```suggestion
    /// @param Vector std::vector<T>: Elements to construct the array from. Elements are copied.
```

Library/src/Systems/Spaces/Space.cpp
```diff
@@ -79,37 +84,24 @@ void GroupDtoToSpace(const chs_users::GroupDto& Dto, csp::systems::Space& Space)
...
     if (Dto.HasModerators())
     {
-        auto& Moderators = Dto.GetModerators();
-        Space.ModeratorIds = Array<String>(Moderators.size());
-
-        for (int i = 0; i < Moderators.size(); ++i)
-        {
-            Space.ModeratorIds[i] = Moderators[i];
-        }
+        Space.ModeratorIds = Array<String> { Dto.GetModerators() };
```
MAG-AdamThorn, line 99: [Comment] I love that you have moved this to the Array class.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -963,7 +983,7 @@ void SpaceSystem::UpdateSpaceMetadata(
         };
 
         const auto& AssetCollection = Result.GetAssetCollection();
-        AssetSystem->UpdateAssetCollectionMetadata(AssetCollection, NewMetadata, Tags, UpdateAssetCollCallback);
+        AssetSystem->UpdateAssetCollectionMetadata(AssetCollection, NewMetadata, {}, UpdateAssetCollCallback);
```
MAG-AdamThorn, line 986: [Question] Is there a reason we are not removing the optional Tags array argument from the AssetSystem::UpdateAssetCollectionMetadata() method?

MAG-AdamThorn: I presume we still want to support tags on AssetCollections as a feature.

MAG-ElliotMorris: Yeah, that decision came from a discussion from Jerome. He felt that AssetCollections were a perfectly good place to have both tags and metadata on. 

I agree though, the concept does remain confused here, since this still inherits the old "metadata = metadata + tags" thing which is odd. Although I fear if we remove it we'll end up with a "Add tags back to assetcollections!" ticket.

Maybe the AssetCollection api should mirror the space one treating metadata and tags differently, even though that would be a pretty large pessimisation?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -624,6 +638,10 @@ void SpaceSystem::GetSpacesByAttributes(const Optional<bool>& InIsDiscoverable,
         CSP_LOG_WARN_FORMAT("Provided value `%i` for ResultsMax exceeded max value and was reduced to `%i`.", *InResultsMax, MAX_SPACES_RESULTS);
     }
 
+    auto Tags = MustContainTags.HasValue() ? MustContainTags->ToStdVector() : std::optional<std::vector<csp::common::String>>(std::nullopt);
```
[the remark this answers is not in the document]
MAG-ElliotMorris: No it can't, it gets confused as to whether the nullopt is to construct vector or to construct optional, annoying.

Library/include/CSP/Common/Array.h
```diff
@@ -217,6 +222,10 @@ template <typename T> class CSP_API Array
         return std::move(Result);
     }
 
+    /// @brief Returns a copy of this Array as a std::vector
+    /// @return List<T>
+    CSP_NO_EXPORT std::vector<T> ToStdVector() const { return std::vector<T>(begin(), end()); }
```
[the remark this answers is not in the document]
MAG-ElliotMorris: Removed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/637/commits/fb27c8eebf9658f85f2a19f875282e89f41d1fa0

Library/include/CSP/Common/Array.h
```diff
@@ -101,6 +88,24 @@ template <typename T> class CSP_API Array
         }
     }
 
+    /// @brief Constructs an array from std::vector.
+    /// @param Vector std::Array : Elements to construct the array from. Elements are copied.
+    /// @pre T must be copyable.
+    CSP_NO_EXPORT Array(const std::vector<T>& Vector)
```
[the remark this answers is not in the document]
MAG-ElliotMorris: Removed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/637/commits/fb27c8eebf9658f85f2a19f875282e89f41d1fa0

I'm on the fence about this, but will yield to you since you seem to care more about it (that sounds glib but it isn't, when it's near 50/50, who cares wins grants great harmony!)

Some observations however
- As we agree on, the naming of these methods is a little unfortunate, you sort of don't know what you're getting.
- The non uniform interface is unfortunate, would be nice to be able to expect to be passing/getting optionals all the time, or not, with this sort of pattern.
- Discoverability is a problem, looking on the types is much more natural than having to go searching for includes. I think this shows with how our DTO code wasn't using them.
- I agree that the lack of dependency introduced into `Array` is a very good thing. we could still achieve that via forward declaration, but not in as a robust way as these templated methods.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -1052,7 +1035,109 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, GetPaginatedPrivateSpacesTest)
...
+    String UserId;
+    LogInAsNewTestUser(UserSystem, UserId);
+
+    // Create test spaces with tags
+    std::string TestSpaceName = "OLY-UNITTEST-SPACE-REWIND";
+    const char* TestSpaceDescription = "OLY-UNITTEST-SPACEDESC-REWIND";
+
+    std::vector<std::pair<std::string, csp::common::Array<csp::common::String>>> SpacesWithTags
+        = { { TestSpaceName + "Space1", { "tag1" } }, { TestSpaceName + "Space2", { "tag2" } }, { TestSpaceName + "Space3", { "tag1", "tag2" } } };
+    std::array<csp::common::String, 3> SpaceIds;
```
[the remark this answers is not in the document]
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/637/commits/49d4f7d112a8f97c22ceb589f614078a3e535a06

Library/include/CSP/Common/Array.h
MAG-ElliotMorris: No longer relevent, https://github.com/magnopus-opensource/connected-spaces-platform/pull/637/commits/fb27c8eebf9658f85f2a19f875282e89f41d1fa0

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: Something for us to discuss further as a team I think.

MAG-AdamThorn (approved):

MAG-mv (approved):

PR #642, opened 2025-03-14
[OF-1525] Remove ifdefs from unit tests.

MAG-ElliotMorris:
Remove all conditional compilation from the tests project.
Making this change seems to have no significant compile time implications, nor any immediate evident effects on builds on CI.

### VS Test Explorer
Continues to look and work exactly as before as if you had `RUN_ALL_UNIT_TESTS` defined. You don't need to go messing around in the preprocessor settings each time you regenerate your project now.

image

### Command line
Can pass a `gtest_filter=FILTER_PATTERN` arg to run specific tests

image

## CI
GTest runs all tests by default with no args passed, which is what we were doing anyway.

You can now use the `GTEST_FILTER` environment variable directly in the teamcity CI, to rerun problematic or investigative tests without having to run the full suite, or even do any rebuilding. (To be fair you could probably have done this beforehand, it was just a little muddy with the conditional compilation)

image

image

**Can anyone think of any reason we wouldn't do this? I was expecting this to have more evident problems. There must have be a reason it was done in the first place?**

### DIFF
--- a/Tests/SpaceComponentTests.cpp
+++ b/Tests/SpaceComponentTests.cpp
@@ -7,7 +7,6 @@
 
 using namespace oly_multiplayer;
 
-#if RUN_ALL_UNIT_TESTS || RUN_REPLICATEDVALUE_TESTS
 CSP_PUBLIC_TEST(OlympusEngine, SpaceComponentTests, ExternalLinkComponentProperties)
 {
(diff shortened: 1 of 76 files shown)


MAG-mv (approved):
So much nicer!

MAG-AdamThorn (approved):

PR #643, opened 2025-03-14
[NT-0] test: Hail Mary attempt to fix avatar test

MAG-ElliotMorris:
So, i never connected the flaky avatar test to the fact that it was the MultiplayerTestRunner one I wrote. If I had, I would have jumped on this earlier to satisfy my stupid pride.

I would say about 15-20% of the time this fails on CI, never locally though, ran it for 15 minutes in a loop.

Unfortunately, I don't really know why it's failing. It's possible a CHS processing delay may cause it, but I've assumed up until now receiving a result is like an ACID transaction, when perhaps it more means "Request acknowledged?" (Documentation for this sort of stuff exists? It might. CHS feels like such a black box at the moment 😖 )

In any case, let's try this. I'll rerun the functional tests on the CI 4 or 5 times from this PR to see if it stays stable, and if so put it in. Fingers crossed eh 🤞

### DIFF
--- a/MultiplayerTestRunner/src/RunnableTests/CreateAvatar.cpp
+++ b/MultiplayerTestRunner/src/RunnableTests/CreateAvatar.cpp
@@ -16,6 +16,7 @@
 
 #include "CreateAvatar.h"
 
+#include "uuid_v4.h"
 #include <CSP/Multiplayer/SpaceEntitySystem.h>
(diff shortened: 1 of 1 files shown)


MAG-ElliotMorris:
That's 2 successes ... come on no whammies.

MAG-AdamThorn (approved):

MAG-ChristopherAtkinson (approved):

PR #648, opened 2025-03-19
[NT-0] chore: put all tests under CSPEngine or DISABLED_CSPEngine

MAG-OR:
Put all the tests under CSPEngine or DISABLED_CSPEngine.
Before:
image

After:
image

### DIFF
--- a/Tests/src/PublicAPITests/CancellationTokenTests.cpp
+++ b/Tests/src/PublicAPITests/CancellationTokenTests.cpp
@@ -21,9 +21,6 @@
 
 #include "gtest/gtest.h"
 
-namespace
-{
-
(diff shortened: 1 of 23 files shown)


MAG-ElliotMorris (approved):
👍

MAG-ElliotMorris:
[Comment[ Something interesting I found out when doing data driven tests.
The VS test runner only supports up to 3 levels of heirarchy for gtest, as far as I can tell. That's why the data driven SpaceSystem Tests don't live under `SpaceSystemTests` in the view, they can't be visualised that way.

If we, in the future, got rid of this top level `CSPEngine` and just had our tests directly there, it would open up the flexibility to do that. Having the top level engine seems a bit redundant now.

MAG-mv (approved):

PR #649, opened 2025-03-23
[OF-1575] Enable Local CHS, Remove hard-coded URIs, migrate multiplayer URI handling.

MAG-ElliotMorris:
## Summary
**tl;dr: Local CHS works now, but a couple of tests have (probably timing related) issues that still need to be shaken out. Full test suite runs are 4-5 minutes long.**

Do the CSP work to allow local CHS to function. 

This changeset does a fair amount more than what was specified in the ticket. There had as of yet been no thought put into the issues with the multiplayer signalR connection. It was never going to just work, so the changes to get that going have been done here too.

This change involves, broadly, 3 things.
- ✔️ Remove all hard-coded URI references to the cloud services throughout the codebase.
- ✔️ Revamp our URI handling for the multiplayer signalR connection
- 🟡 Fix any tests that are demonstrating new issues from the faster cloud services response. 

The first two are complete in this PR. On the tests, I have fixed the easy tests, but the ones that weren't obvious have been left. The ticket only wanted the first item to be complete, but if we kept working like that, we'd never get local CHS over the line. Nonetheless, I'm troubled not to be able to deal with these, but I wanted to get this up for review sooner rather than later, considering how I need to start AOI implementation.

> [!IMPORTANT]  
> **Local CHS still will not function on WASM**, as the URI parsing changes have been done for the POCO client _only_, (as this is what our tests use). I didn't want to blow the scope of this change even further beyond it has been already, especially considering my upcoming time off. You'll probably tell from the tone of the changes that I think this area needs some TLC, but we shouldn't forget to do this, leaving WASM/Non-WASM with a difference in capability will only come back to bite us, when we add WASM testing especially.

## Tests
See here, a manually run local CHS run on teamcity, the whole suite runs, only 2 tests fail :
https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_TestsLocalChsInstance/284828

However, this isn't the whole story, as a local run consistently shows these failures also:

- ❌  [  FAILED  ] CSPEngine.PointOfInterestSystemTests.GetPOIInsideCircularAreaTest- 
- ❌  [ FAILED  ] CSPEngine.PointOfInterestSystemTests.QuerySpacePOITest
- ❌  [  FAILED  ] CSPEngine.UserSystemTests.GetAgoraUserTokenTest

Additionally, about 20% of the time, have observed this error running the full suite locally. :
```
[ RUN      ] CSPEngine.MaterialTests.GetMultipleMaterialsTest
mimalloc: error: allocation request is too large (16131858542891098112 bytes)
mimalloc: assertion failed: at "C:\dev\connected-spaces-platform\ThirdParty\mimalloc\src\segment.c":544, _mi_segment_thread_collect
  assertion: "tld->pages_reset.first == NULL"
```
16131858542891098112 bytes is 16 thousand petabyes, probably an underflow or something. 🤔 

@MAG-AdamThorn @MAG-SamBirley I would like to progress shaking out the remaining tests as soon as reasonably possible, as the local CHS story needs to be closed. I think that's a good story for anyone else on the team though. That's why I'm getting this PR up perhaps a bit earlier than I normally would, want to move fast.

> [!NOTE]  
> I have tested the OKO-WASM build locally to make sure you can still enter spaces, as this does technically touch CSPFoundation so there's a risk there. Crazy how you can still test this stuff even when there's breaking changes (the tag stuff). Dynamic languages are wild.

> [!NOTE]  
> These tests are so fast now that the sleeps we put in are actually significant, each 7 second sleep is 2.3% of total execution time. I also half-suspect our console logging might be a slowdown worth considering too. It's great to be at 5 minutes, but we should never be satisfied. I want everything, including instant test runs.

## Action items
- Log work for getting the multiplayer system working with local CHS on WASM.
- Log work for shaking out the errors in the remaining tests well enough to enable local CHS on teamcity.

### DIFF
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -107,6 +107,11 @@ public:
     /// @return const EndpointURIs& : The EndpointURIs class with current endpoint data
     static const EndpointURIs& GetEndpoints();
 
+    /// @brief Create an EndpointsURIs object containing URIs to the various services needed by CSP.
+    /// @param EndpointRootURI URI to the root of the cloud services.
+    /// @return EndpointURIs class with deduced endpoint URIs.
+    static EndpointURIs CreateEndpointsFromRoot(const csp::common::String& EndpointRootURI);
+
     /// @brief Sets a class containing all relevant Client info currently set for Foundation.
     /// Used internally to generate ClientUserAgentString.
     /// @param The Client Info class with current Client Info data
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -108,7 +108,7 @@ public:
     static const EndpointURIs& GetEndpoints();
 
     /// @brief Create an EndpointsURIs object containing URIs to the various services needed by CSP.
-    /// @param EndpointRootURI URI to the root of the cloud services.
+    /// @param EndpointRootURI csp::common::String: URI to the root of the cloud services.
     /// @return EndpointURIs class with deduced endpoint URIs.
     static EndpointURIs CreateEndpointsFromRoot(const csp::common::String& EndpointRootURI);
 
--- a/Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
+++ b/Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
@@ -24,6 +24,7 @@
 #include <Poco/Net/HTTPResponse.h>
 #include <Poco/Net/HTTPSClientSession.h>
 #include <Poco/Net/NetException.h>
+#include <Poco/URI.h>
 #include <chrono>
 #include <stdexcept>
 #include <thread>
@@ -48,28 +49,48 @@ CSPWebSocketClientPOCO::~CSPWebSocketClientPOCO()
     Stop(nullptr);
 }
 
+CSPWebSocketClientPOCO::ParsedURIInfo CSPWebSocketClientPOCO::ParseMultiplayerServiceUriEndPoint(const std::string& MultiplayerServiceUriEndpoint)
+{
+    Poco::URI EndpointURI { MultiplayerServiceUriEndpoint.c_str() };
+
+    // It's common folk may provide a "localhost:port/path/path" sort of string, with omits the protocol, just make sure one's there.
+    if ((EndpointURI.getScheme() != "https") && (EndpointURI.getScheme() != "http"))
+    {
+        throw std::runtime_error("CSPWebSocketclientPOCO::ParsedURIInfo, Expected `https` or `http` scheme, found neither.");
+    }
+
+    CSPWebSocketClientPOCO::ParsedURIInfo Out;
+    Out.Endpoint = EndpointURI.toString();
+    Out.Protocol = EndpointURI.getScheme();
+    Out.Domain = EndpointURI.getHost();
+    Out.Path = EndpointURI.getPath();
+    Out.Port = EndpointURI.getPort();
+
+    return Out;
+}
+
 void CSPWebSocketClientPOCO::Start(const std::string& Url, CallbackHandler Callback)
 {
     CSP_PROFILE_SCOPED();
 
-    std::string endpoint = csp::CSPFoundation::GetEndpoints().MultiplayerServiceURI.c_str();
-    auto index = endpoint.find(':');
-    std::string protocol = endpoint.substr(0, index);
-    index += 3; // "://"
-    auto endIndex = endpoint.find('/', index);
-    std::string domain = endpoint.substr(index, endIndex - index);
-    index = endIndex;
-    std::string path = endpoint.substr(index);
+    CSPWebSocketClientPOCO::ParsedURIInfo ParsedEndpoint
+        = ParseMultiplayerServiceUriEndPoint(csp::CSPFoundation::GetEndpoints().MultiplayerServiceURI.c_str());
+
+    auto domain = ParsedEndpoint.Domain;
+    auto protocol = ParsedEndpoint.Protocol;
+    auto path = ParsedEndpoint.Path;
+    auto endpoint = ParsedEndpoint.Endpoint;
+    auto port = ParsedEndpoint.Port;
 
     Poco::Net::HTTPClientSession* cs;
 
     if (protocol == "https")
     {
-        cs = CSP_NEW Poco::Net::HTTPSClientSession(domain, 443);
+        cs = CSP_NEW Poco::Net::HTTPSClientSession(domain, port);
     }
     else
     {
-        cs = CSP_NEW Poco::Net::HTTPClientSession(domain, 80);
+        cs = CSP_NEW Poco::Net::HTTPClientSession(domain, port);
     }
 
     Poco::Net::HTTPRequest request(Poco::Net::HTTPRequest::HTTP_GET, path, Poco::Net::HTTPMessage::HTTP_1_1);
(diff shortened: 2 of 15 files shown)


Tests/src/PublicAPITests/ComponentTests/CustomComponentTests.cpp
```diff
@@ -222,7 +223,7 @@ CSP_PUBLIC_TEST(CSPEngine, CustomTests, CustomComponentTest)
             TestTime = std::chrono::duration_cast<std::chrono::seconds>(Current - Start).count();
         }
 
-        EXPECT_TRUE(GotAllEntities);
+        ASSERT_TRUE(GotAllEntities);
```
MAG-OR, line 226: [question] what is the reason for changing EXPECT_TRUE to ASSERT_TRUE? just curious really

Tests/src/PublicAPITests/ComponentTests/StaticModelComponentTests.cpp
```diff
@@ -276,6 +276,8 @@ CSP_PUBLIC_TEST(CSPEngine, StaticModelTests, StaticModelComponentEnterSpaceTest)
         auto [ExitSpaceResult] = AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
     }
 
+    std::this_thread::sleep_for(std::chrono::seconds(7));
```
MAG-OR, line 279: [comment] i don't know if it's worth putting a comment as to why we have to wait

Tests/src/PublicAPITests/MultiplayerTestRunnerProcessTests.cpp
```diff
@@ -64,10 +65,15 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTestRunnerProcessTests, ArgTest)
 
 CSP_PUBLIC_TEST(CSPEngine, MultiplayerTestRunnerProcessTests, FutureTest)
 {
+
+    auto TestUser = CreateTestUser();
+
```
MAG-OR, line 70: [question] it may be best if I merge my PR first, then you would already have these changes, wouldn't you?

MAG-OR (approved):
amazing work.

Tests/src/PublicAPITests/ComponentTests/StaticModelComponentTests.cpp
MAG-AdamThorn: [comment] I see the same delay added to static and animated model tests along with the following comment:
// Ensure component data has been written to database by chs before entering the space again
// This is due to an enforced 2 second chs database write delay
std::this_thread::sleep_for(7s);

We have our CHS/CSP sync this afternoon so I can ask for more information about this delay. But in the meantime I am not sure why we are waiting for 7s specifically based on the comment.

Library/include/CSP/CSPFoundation.h
```diff
@@ -107,6 +107,11 @@ class CSP_API CSPFoundation
     /// @return const EndpointURIs& : The EndpointURIs class with current endpoint data
     static const EndpointURIs& GetEndpoints();
 
+    /// @brief Create an EndpointsURIs object containing URIs to the various services needed by CSP.
+    /// @param EndpointRootURI URI to the root of the cloud services.
```
MAG-mv, line 111: [fix] param should follow our doc standards: * @param ParamName csp::namespace::ParamT : Description of param. If the params are

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
```diff
@@ -48,28 +49,48 @@ CSPWebSocketClientPOCO::~CSPWebSocketClientPOCO()
...
 }
 
+CSPWebSocketClientPOCO::ParsedURIInfo CSPWebSocketClientPOCO::ParseMultiplayerServiceUriEndPoint(const std::string& MultiplayerServiceUriEndpoint)
+{
+    Poco::URI EndpointURI { MultiplayerServiceUriEndpoint.c_str() };
+
+    // It's common folk may provide a "localhost:port/path/path" sort of string, with omits the protocol, just make sure one's there.
+    if ((EndpointURI.getScheme() != "https") && (EndpointURI.getScheme() != "http"))
+    {
+        throw std::runtime_error("CSPWebSocketclientPOCO::ParsedURIInfo, Expected `https` or `http` scheme, found neither.");
```
MAG-mv, line 59: [question] We don't usually throw exceptions that clients need to handle. Should we handle this a different way?

MAG-mv (changes requested):

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-ElliotMorris: What would our canonical error handling strategy be for something like this? A log and empty struct return? Seems dangerous to me.

Tests/src/PublicAPITests/ComponentTests/CustomComponentTests.cpp
MAG-ElliotMorris: EXPECT conditions continue even if there's a failure. This test makes no sense to continue if this isn't true, which came up, so I changed this. It's minor, just means the error you get most immediately is the actual problem, rather than having to read up in the logs.

Tests/src/PublicAPITests/ComponentTests/StaticModelComponentTests.cpp
MAG-ElliotMorris: 7s is down to me pushing this number up in other areas of the codebase, since 2s is the advertised CHS value, but that isn't the full time we have to wait, because we have to account for possible network latency and dropped packets. The value is this high because I've been pushing us to prioritize test stability more, and if we _have_ to sleep, I want to be darn sure that we never get flakiness in tests due to too low a time.

Library/include/CSP/CSPFoundation.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/649/commits/48f86ce93fb82ba750361189908a5781b5e74ae1

Tests/src/PublicAPITests/MultiplayerTestRunnerProcessTests.cpp
MAG-ElliotMorris: It was, ty. Rebased your changes in.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-ElliotMorris: Ah no, i see what it is, there's a callback in `Start` I should populate as a failure. will [fix]

MAG-ElliotMorris: Have extended the `try/catch` block in `Start` to use the callback population already there. Also moved to `unique_ptr` for the memory in that block, to handle releasing memory in the case of an exception.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/649/commits/0206a9df5315461b84d9ca7b56a2b1655b522a91

MAG-mv: [quesiton] Do we want to tets that this error is logging correctly?

MAG-ElliotMorris: No. It would be idea, but this is the reason this method is extracted, so I can at least test the primary behaviour.

Testing the actual callback would require either:
- Mutating environment variables during test execution, perhaps a workable idea, but having to store the prior value and reset it makes designing the RAII type not immediately trivial
- Replumbing our connection flow to make these URI arguments actually matter, so you can test `Start` directly
- Adding some sort of `OverrideEndpoints` method. My favourite, but also not something I want to do as part of this change.

MAG-mv (approved):

PR #650, opened 2025-03-26
[OF-1447] feat: add ability to post to service proxy directly

MAG-OR:
Expose ServiceProxyPost and allow user to pass their own token info parameters. 
Add a new test and check error messages are not emitted in the log for both the new ServiceProxyPost test and the old AgoraUserToken tests

### DIFF
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -213,6 +213,11 @@ public:
     /// @param Callback UserTokenResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void GetAgoraUserToken(const AgoraUserTokenParams& Params, StringResultCallback Callback);
 
+    /// @brief Post Service Proxy to perform specified operation of specified service
+    /// @param Params TokenInfoParams : Params to specify service, operation, set help and parameters
+    /// @param Callback UserTokenResultCallback : callback to call when a response is received
+    CSP_ASYNC_RESULT void PostServiceProxy(const TokenInfoParams& Params, StringResultCallback Callback);
+
     /// @brief Re-send user verification email
     /// @param InEmail csp::common::String : User's email address
     /// @param InRedirectUrl csp::common::Optional<csp::common::String> : URL to redirect user to after they have registered
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -209,13 +209,13 @@ public:
     CSP_ASYNC_RESULT void Ping(NullResultCallback Callback);
 
     /// @brief Retrieve User token from Agora
-    /// @param Params AgoraUserTokenParams : Params to configure the User token
-    /// @param Callback UserTokenResultCallback : callback to call when a response is received
+    /// @param Params const AgoraUserTokenParams& : Params to configure the User token
+    /// @param Callback StringResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void GetAgoraUserToken(const AgoraUserTokenParams& Params, StringResultCallback Callback);
 
     /// @brief Post Service Proxy to perform specified operation of specified service
-    /// @param Params TokenInfoParams : Params to specify service, operation, set help and parameters
-    /// @param Callback UserTokenResultCallback : callback to call when a response is received
+    /// @param Params const TokenInfoParams& : Params to specify service, operation, set help and parameters
+    /// @param Callback StringResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void PostServiceProxy(const TokenInfoParams& Params, StringResultCallback Callback);
 
     /// @brief Re-send user verification email
--- a/Library/src/Systems/Users/UserSystem.cpp
+++ b/Library/src/Systems/Users/UserSystem.cpp
@@ -25,6 +25,7 @@
 #include "Multiplayer/ErrorCodeStrings.h"
 #include "Multiplayer/EventSerialisation.h"
 #include "Services/AggregationService/Api.h"
+#include "Services/AggregationService/Dto.h"
 #include "Services/UserService/Api.h"
 #include "Systems/ResultHelpers.h"
 #include "Systems/Users/Authentication.h"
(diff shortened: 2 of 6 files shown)


Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -711,6 +712,29 @@ void UserSystem::GetAgoraUserToken(const AgoraUserTokenParams& Params, StringRes
...
 }
 
+void UserSystem::PostServiceProxy(const TokenInfoParams& Params, StringResultCallback Callback)
+{
+    auto TokenInfo = std::make_shared<chs_aggregation::ServiceRequest>();
+    TokenInfo->SetServiceName(Params.ServiceName);
+    TokenInfo->SetOperationName(Params.OperationName);
+    TokenInfo->SetHelp(Params.SetHelp);
+    std::map<csp::common::String, csp::common::String> ParamsMap;
+    auto* Keys = Params.Parameters.Keys();
```
MAG-mv, line 722: [fix] Keys need to be deleted, or it will cause a memory leak.  It may be worth using a unique_ptr here and creating a custom deleter. We have done this somewhere else in the codebase

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -711,6 +712,29 @@ void UserSystem::GetAgoraUserToken(const AgoraUserTokenParams& Params, StringRes
...
+    for (auto idx = 0; idx < Keys->Size(); ++idx)
```
MAG-mv, line 723: [fix] We already have functionality in Convert.h to convert between csp and std types. Also, the map should be deleted when using map.Keys()

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -985,6 +996,81 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, GetAgoraUserTokenTest)
...
+
+CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, PostServiceProxyTest)
+{
+    SetRandSeed();
+
+    auto& SystemsManager = csp::systems::SystemsManager::Get();
+    auto* UserSystem = SystemsManager.GetUserSystem();
+    auto* SpaceSystem = SystemsManager.GetSpaceSystem();
+
+    const char* TestSpaceName = "OLY-UNITTEST-SPACE-REWIND";
```
MAG-mv, line 1021: [fix] "OLY" should be replaced with "CSP" and "REWIND" should be replaced with "MAG" as these are old concepts. We try to do this in all new tests

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -213,6 +213,11 @@ class CSP_API UserSystem : public SystemBase
     /// @param Callback UserTokenResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void GetAgoraUserToken(const AgoraUserTokenParams& Params, StringResultCallback Callback);
```
MAG-mv, line 214: [fix] UserTokenResultCallback is incorrect, this should be StringResultCallback

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -213,6 +213,11 @@ class CSP_API UserSystem : public SystemBase
...
 
+    /// @brief Post Service Proxy to perform specified operation of specified service
+    /// @param Params TokenInfoParams : Params to specify service, operation, set help and parameters
+    /// @param Callback UserTokenResultCallback : callback to call when a response is received
+    CSP_ASYNC_RESULT void PostServiceProxy(const TokenInfoParams& Params, StringResultCallback Callback);
```
MAG-mv, line 219: [fix] Same as above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -213,6 +213,11 @@ class CSP_API UserSystem : public SystemBase
     /// @param Callback UserTokenResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void GetAgoraUserToken(const AgoraUserTokenParams& Params, StringResultCallback Callback);
 
+    /// @brief Post Service Proxy to perform specified operation of specified service
+    /// @param Params TokenInfoParams : Params to specify service, operation, set help and parameters
```
MAG-mv, line 217: [fix] Can we also include const and & as part of the type in docs? Its probably worth reading our new "Doc Vision" doc

MAG-mv (changes requested):

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: fixed in d84489de295b6e48c725ed91a6464306cad6096d

Library/include/CSP/Systems/Users/UserSystem.h
MAG-OR: fixed in d84489de295b6e48c725ed91a6464306cad6096d

Library/include/CSP/Systems/Users/UserSystem.h
MAG-OR: fixed in d84489de295b6e48c725ed91a6464306cad6096d

Library/src/Systems/Users/UserSystem.cpp
MAG-OR: ah, I was looking for one in Map.h itself and couldn't find one. thanks! fixed in 03dcfb8ad6445457259bf2cfc0e20c29bcd18e21

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -985,6 +996,81 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, GetAgoraUserTokenTest)
 
     EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
     EXPECT_FALSE(Result.GetValue().IsEmpty());
+    WaitForCallback(LogConfirmed1);
+    EXPECT_FALSE(LogConfirmed1);
+    WaitForCallback(LogConfirmed1);
```
MAG-ChristopherAtkinson, line 1001: [fix] Looks like this is meant to be LogConfirmed2

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -985,6 +996,81 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, GetAgoraUserTokenTest)
...
+    Params.Parameters["shareScreen"] = "false";
+
+    // Get agora token through PostServiceProxy
+    auto [Result] = AWAIT_PRE(UserSystem, PostServiceProxy, RequestPredicate, Params);
+
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+    EXPECT_FALSE(Result.GetValue().IsEmpty());
+    WaitForCallback(LogConfirmed1);
+    EXPECT_FALSE(LogConfirmed1);
+    WaitForCallback(LogConfirmed1);
```
MAG-ChristopherAtkinson, line 1070: [fix] Looks like this is meant to be LogConfirmed2

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: oh oh. well spotted, thanks

MAG-OR: fixed in c7705114359f8b458df70a93652ce8a84a30f570

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-OR: fixed in c7705114359f8b458df70a93652ce8a84a30f570

MAG-ChristopherAtkinson (approved):

PR #651, opened 2025-03-31
[OF-1585] refac: Connect Continuations

MAG-ElliotMorris:
> [!TIP]
>This benefits from a commit-by-commit review

Refactor the `Connect` function to continuations. 

Several refactorings are done as part of this work:
 - Create a mocking pattern via `SignalRClient`, and show the first use of full type mocks in our tests.
 - Use that pattern to test untriggerable network errors through the `Connect` flow.
 - Slightly evolve the error handling for the Connection Object, to at least try to give some context
 - Rename existing continuations utilities to `HTTP`, to better represent their difference from Multiplayer results.
 
 ## Discussion
 Three points jump out to me having done this work, for future consideration.
 - Our conversion from SignalR errors to our own error reporting mechanism is rather confused.
 - There are four callbacks in play, which is too many and very confusing. I think we should reduce this.
   - The regular callback passed to the connect function (Called the ErrorCallback sometimes, but not treated consistently as that)
   - The network interrupted callback (connected to the SignalR `set_disconnected` method)
   - The disconnection callback, called when there's an error during connection, or when the connection shuts down. Almost always called redundantly with the regular callback.
   - The connection callback, called when connection is a success
- The use of mimalloc and CSP_NEW in our codebase makes taking ownership of pointers from external to CSP fragile, how are users to know that they have to be constructed with CSP_NEW?

### DIFF
--- a/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
+++ b/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
@@ -32,6 +32,13 @@ CSP_START_IGNORE
 class CSPEngine_MultiplayerTests_SignalRConnectionTest_Test;
 CSP_END_IGNORE
 
+namespace async
+{
+CSP_START_IGNORE
+template <typename T> class task;
+CSP_END_IGNORE
+}
+
 namespace csp::systems
 {
 
@@ -177,23 +184,25 @@ private:
 
     typedef std::function<void(std::exception_ptr)> ExceptionCallbackHandler;
 
-    /// @brief Start the connection and register to start receiving updates from the server.
-    /// Connect should be called after LogIn and before EnterSpace.
-    /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
-    void Connect(ErrorCodeCallbackHandler Callback);
+    void Start(ExceptionCallbackHandler Callback) const;
+
+    //<void> template tags not supported in wrapper generator. (Why we're even wrapping private methods is a mystery to me though)
+    CSP_START_IGNORE
+    async::task<void> Start() const;
+    /* Connect Continuations */
+    // Delete the entity specified by the EntityId. std::numeric_limits<uint64_t>::max() means ALL_ENTITIES_ID, and deletes everything.s
+    auto DeleteEntities(uint64_t EntityId) const;
+    // Get the client ID and return it (does not set it locally)
+    auto RequestClientId();
+    // Invoke "StartListening" on already created Connection
+    std::function<async::task<void>()> StartListening();
+    CSP_END_IGNORE
 
     /// @brief End the multiplayer connection.
     /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
     void Disconnect(ErrorCodeCallbackHandler Callback);
-
-    void Start(ExceptionCallbackHandler Callback) const;
     void Stop(ExceptionCallbackHandler Callback) const;
-
-    void StartListening(ErrorCodeCallbackHandler Callback);
     void StopListening(ErrorCodeCallbackHandler Callback);
-
-    void InternalDeleteEntity(uint64_t EntityId, ErrorCodeCallbackHandler Callback) const;
-    void DeleteOwnedEntities(ErrorCodeCallbackHandler Callback);
 
     /// @brief Subscribes the connected user to the specified space's scope.
     /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
@@ -202,8 +211,6 @@ private:
     /// @brief Clears the connected user's subscription to their current set of scopes.
     /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
     void ResetScopes(ErrorCodeCallbackHandler Callback);
-
-    void RequestClientId(ErrorCodeCallbackHandler Callback);
 
     void DisconnectWithReason(const csp::common::String& Reason, ErrorCodeCallbackHandler Callback);
 
--- a/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
+++ b/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
@@ -190,7 +190,7 @@ private:
     CSP_START_IGNORE
     async::task<void> Start() const;
     /* Connect Continuations */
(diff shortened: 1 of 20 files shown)


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -56,29 +59,40 @@ namespace csp::multiplayer
...
     }
     catch (const std::exception& e)
     {
-        ErrorMessage = e.what();
-
-        CSP_LOG_ERROR_FORMAT("%s\n", ErrorMessage.c_str());
+        return ParseMultiplayerError(e);
     }
 
+    return { ErrorCode::Unknown, "MultiplayerConnection::ParseMultiplayerErrorFromExceptionPtr, Unexpectedly no exception was thrown." };
```
MAG-OR, line 74: [question] can this actually happen?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -181,75 +372,47 @@ void MultiplayerConnection::Connect(ErrorCodeCallbackHandler Callback)
...
-                return;
-            }
-
-            Connected = true;
-
-            DeleteOwnedEntities(
-                [this, Callback](ErrorCode Error)
+                // Success
+                INVOKE_IF_NOT_NULL(ConnectionCallback, "Successfully connected to SignalR hub.");
+                INVOKE_IF_NOT_NULL(Callback, ErrorCode::None); // ????
```
MAG-OR, line 406: [question] ?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -134,23 +134,27 @@ void SpaceSystem::RefreshMultiplayerConnectionToEnactScopeChange(
...
-
-                            // TODO: Support getting errors from RetrieveAllEntities
-                            csp::systems::SystemsManager::Get().GetSpaceEntitySystem()->RetrieveAllEntities();
-
-                            // Success!
-                            RefreshMultiplayerContinuationEvent->set({});
-                        });
+                                CSP_LOG_MSG(csp::systems::LogLevel::Log, " MultiplayerConnection->StartListening success");
+
+                                // TODO: Support getting errors from RetrieveAllEntities
```
MAG-OR, line 143: [question] is this todo meant to be done as part of the current story, or a future one? If so, might be best to put the story number in there?

Tests/src/Mocks/SignalRConnectionMock.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+#pragma once
+
+#include "../../Library/src/Multiplayer/SignalR/ISignalRConnection.h"
+#include <gmock/gmock.h>
+
+class SignalRConnectionMock : public csp::multiplayer::ISignalRConnection
+{
+public:
+    MOCK_METHOD(void, Start, (std::function<void(std::exception_ptr)>), (override));
```
MAG-OR, line 24: [comment] well done for implementing mocks.

MAG-OR (approved):

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: I don't know, you'll note I didn't write this comment, this code has merely moved.

MAG-OR: ah ok

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: Probably not, but technically exception_ptrs can be null so an inappropriate call to this method could trigger this path.

Nonetheless, a default path is required for this try/catch pattern anyhow.

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: Ah yeah, i'll remove the comment. I was just confused during dev about how we're calling what is sometimes called the "Error callback" in the success case. Still am a bit befuddled, but am not changing the behaviour.

MAG-OR: would "return callback"/"return code" be more appropriate? could be success (if no error) or various errors. anyway, this is probably more of a philosophical conversation than anything else, and this story is likely not the place to do it

MAG-ElliotMorris: Yeah dunno, I might log an issue to think about this, it is chunky enough to warrant a ticket.

MAG-ChristopherAtkinson (approved):
Changes look good to me and feedback has been address from other reviewers, this was really helpful.

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: Logged : https://magnopus.atlassian.net/browse/OF-1610

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -177,24 +184,26 @@ class CSP_API MultiplayerConnection
...
-    /// Connect should be called after LogIn and before EnterSpace.
-    /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
-    void Connect(ErrorCodeCallbackHandler Callback);
+    void Start(ExceptionCallbackHandler Callback) const;
+
+    //<void> template tags not supported in wrapper generator. (Why we're even wrapping private methods is a mystery to me though)
+    CSP_START_IGNORE
+    async::task<void> Start() const;
+    /* Connect Continuations */
+    // Delete the entity specified by the EntityId. std::numeric_limits<uint64_t>::max() means ALL_ENTITIES_ID, and deletes everything.s
```
MAG-AdamThorn, line 193: [fix] trailing 's' at the end of the comment line.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/651/commits/67800345fe8b9d11333515952fa411a21f62db6f

PR #653, opened 2025-04-03
[OF-1515] build: put all WASM build artefacts in one place

MAG-OR:
Previous artefact locations:
- Tools/WrapperGenerator/Output/TypeScript
- Library/Binaries/wasm/[Debug|Release]

New artefact locations:
Library\Binaries\package\wasm\connected-spaces-platform.web

Contains:
- WASM Debug and Release build folders
- package.json
- README.md
- connectedspacesplatform.d.ts, connectedspacesplatform.js, connectedspacesplatform.js.map, connectedspacesplatform.ts, ConnectedSpacesPlatform_WASM.d.ts

The Emscripten build scripts now call the python utilities that we use on teamcity to generate the README and the Web package. They do NOT publish the package to npm.
I have also added some much needed error handling.

I have tested a local build on a local Web server -- it worked without errors.
I have also tested `CSP Publish Feature`: it needed one small change of path, and then succeeded. I have downloaded the Web package and tested it with a local Web server -- it worked without errors. (see: [here](https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_PublishCspFeature_2/285584?buildTab=artifacts#%2FOF-1515-Centralized-Web-Packaging%2FWeb%2FWeb_OF-1515-Centralized-Web-Packaging_285584.zip)
N.B.: the same path change will have to be applied to `CSP Publish` after this is merged.
image


Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
```diff
@@ -32,10 +34,23 @@ exit /b 1
...
 docker run -w /src -v %cd%:/src --rm emscripten/emsdk:%emsdk_version% emmake make -j 8 config=%~1_wasm
-cd Tools/Emscripten
\ No newline at end of file
+if %ERRORLEVEL% NEQ 0 (goto Error)
+xcopy /s /e Library\Binaries\WASM\%~1\ Tools\WrapperGenerator\Output\TypeScript\connected-spaces-platform.web\%~1\
+if %ERRORLEVEL% NEQ 0 (goto Error) else (goto Success)
+
+:Error
+cd Tools/Emscripten
+echo ERROR: Build failed. Have you started the docker engine?
```
MAG-OR, line 50: I don't know how many times you guys have launched this the first time without starting docker, and sat there wondering why it failed. But I need this.

Tools/Emscripten/EmscriptenFullBuildAndConfigure.sh
```diff
@@ -27,5 +27,29 @@ fi
 
 python Tools/Emscripten/ReplaceComSpec.py
 docker run -w /src -v `pwd`:/src --rm emscripten/emsdk:$emsdk_version emmake make config="$1"_wasm clean
+if [ $? -ne 0 ]
```
MAG-OR, line 30: The shell scripts were "kind of" tested. As in, I ran these from git bash and added an echo in front of the actual commands (docker, python, etc.) because they are not supported (yeah because I'm on Windows and faking a bash shell by using git bash)

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
```diff
@@ -32,10 +34,23 @@ exit /b 1
...
-:End
+:Wasm
 python Tools/Emscripten/ReplaceComSpec.py
 docker run -w /src -v %cd%:/src --rm emscripten/emsdk:%emsdk_version% emmake make config=%~1_wasm clean
+if %ERRORLEVEL% NEQ 0 (goto Error)
 docker run -w /src -v %cd%:/src --rm emscripten/emsdk:%emsdk_version% emmake make -j 8 config=%~1_wasm
-cd Tools/Emscripten
\ No newline at end of file
+if %ERRORLEVEL% NEQ 0 (goto Error)
+xcopy /s /e Library\Binaries\WASM\%~1\ Tools\WrapperGenerator\Output\TypeScript\connected-spaces-platform.web\%~1\
```
MAG-OR, line 45: This is the meaningful change. after the build, we're copying the build folder in the output folder

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
```diff
@@ -1,3 +1,5 @@
+@echo off
```
MAG-OR, line 1: Without this, Windows thinks it is a good idea to echo every single command before executing it, which I think is overkill. Happy to have a healthy debate about this.

UnitTesting/tests/RunNodeTests.py
MAG-OR, line 1: So... according to the source code, this is run as part of a GitHub action workflow.  Except, GitHub shows that this workflow has actually never been run. I was the first one to launch it on my branch, and it failed to even start: https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/14243934101 (rather unsurprisingly). 
Should we remove the file and the workflow instead?

teamcity/BuildNPMWebPackage.py
MAG-OR, line 1: I might need support to test this on team city. I think this may actually publish a package, and I would rather avoid doing this.

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
MAG-OR, line 45: This is the meaningful change. after the build, we're copying the build folder in the output folder

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
MAG-OR, line 50: I don't know how many times you guys have launched this the first time without starting docker, and sat there wondering why it failed. But I need this.

teamcity/BuildNPMWebPackage.py
MAG-AdamThorn: @MAG-OR You can test this by running the **Publish CSP Feature** build config and selecting your _build branch_ from the drop down under the **Changes** tab. That will run the _BuildNPMWebPackage.py_ script, but the build step for this config passes in an optional parameter which stops it from publishing the package to npmjs. Once completed you can download your built artefact to confirm that it is correct.

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
MAG-ElliotMorris: I've done this loads of times, this is great.

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
```diff
@@ -32,10 +34,23 @@ exit /b 1
...
 :ArgOk
 "modules/premake/bin/release/premake5" gmake2 --generate_wasm
-goto End
+goto Wasm
 
-:End
+:Wasm
 python Tools/Emscripten/ReplaceComSpec.py
 docker run -w /src -v %cd%:/src --rm emscripten/emsdk:%emsdk_version% emmake make config=%~1_wasm clean
+if %ERRORLEVEL% NEQ 0 (goto Error)
```
MAG-ElliotMorris, line 42: [Comment] 👍

Tools/Emscripten/EmscriptenFullBuildAndConfigure.sh
MAG-ElliotMorris: I think that's good enough. I guess these exist for folks using macs? Might be an idea to get someone to test in a mac shell ... but 🤷

Tools/WrapperGenerator/TypeScriptWrapperGenerator.py
```diff
@@ -23,7 +23,7 @@
 class TypeScriptWrapperGenerator:
     __TEMPLATE_DIRECTORY = config["template_directory"] + "TypeScript/"
     __PARTIALS_DIRECTORY = __TEMPLATE_DIRECTORY + "Partials/"
-    __OUTPUT_DIRECTORY = config["output_directory"] + "TypeScript/"
+    __OUTPUT_DIRECTORY = config["output_directory"] + "TypeScript/connected-spaces-platform.web/"
```
MAG-ElliotMorris, line 26: [Comment] Aye, I think I prefer this, so it is directly copyable into the `node_modules` folder for the web client. It's probably not strictly standard, but convenient for us.

UnitTesting/tests/RunNodeTests.py
MAG-ElliotMorris: Yes I think so, although @MAG-AdamThorn might have more context on that.

MAG-ElliotMorris (approved):

MAG-ChristopherAtkinson (commented):
[Consideration] The changes look good, I am concerned about the effective changes in the CI pipeline, maybe worth speaking with @mag-aidangustard ahead of approval for any changes required for publish

UnitTesting/tests/RunNodeTests.py
MAG-AdamThorn: Yes I think we can remove this

Library/premake5.lua
```diff
@@ -352,10 +352,10 @@ if not Project then
             filter {}
         end
 
-        -- Debug/Release config settings
-        filter "configurations:*Debug*"
+        -- Debug/Release/debug/release config settings
+        filter "configurations:*ebug*"
```
MAG-OR, line 356: happy to add something akin to [dD] and [rR] of this is not clear for everyone

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
```diff
@@ -24,18 +26,36 @@ for /f "delims=" %%x in (emsdk_version.txt) do (
...
+if "%~1"=="" (goto ArgNone) else (goto ArgCheck)
 
-:ArgError
+:ArgNone
 echo No configuration name provided
 exit /b 1
 
+:ArgCheck
+for %%a in ("debug" "release") do if "%~1"==%%a (goto ArgOk)
+echo Unsupported configuration "%~1". Supported configurations are "debug" and "release". Please try again.
```
MAG-OR, line 37: this was doing my head in: having to wait until after the wrapper generator is finished and for the wasm build command to fail because I made a typo in the target name. now it tells you off straight away. tested for both batch scripts and both bash scripts.

believe it or not, cmd does not have the concept of "or" inside an if condition, and the workaround is a for loop.

premake5.lua
```diff
@@ -20,7 +20,7 @@ solution( "ConnectedSpacesPlatform" )
     if CSP.HasCommandLineArgument("DLLOnly") then
         configurations { "DebugDLL", "ReleaseDLL", "DebugDLL-CSharp", "ReleaseDLL-CSharp" }
     elseif CSP.IsWebAssemblyGeneration() then
-        configurations { "Debug", "Release" }
+        configurations { "debug", "release" }
```
MAG-OR, line 23: for avoidance of doubt, this was done so that the folders we create match the actual wasm targets. this is not strictly necessary because windows doesn't care about case, but it is neater (and probably avoids potential mismatches on Unix). testing on windows shows everything works as expected with the change.

teamcity/BuildNPMWebPackage.py
MAG-OR: this is outdated

Library/premake5.lua
MAG-OR: this is outdated

premake5.lua
MAG-OR: this is outdated

Tools/Emscripten/EmscriptenFullBuildAndConfigure.bat
MAG-ElliotMorris: These are the sorts of incidental improvements we need to be making habitually. Nice one.

MAG-ElliotMorris (approved):
Lookin good!

On a procedural note, it would have been nice to have gone to the effort to get all the "undo all unnecesary changes" into one commit, or probably even better just rebased the original changes out of existence entirely.

When there's significant changes like this, a "Changes since last review" header in the description might be helpful too? You have written what changed in there which is good, it's just a tad hard to parse specifically what the new parts of the approach are from the decription, as a returning reviewer.

Minor points though, good job figuring out the right way to do this.

PR #659, opened 2025-04-09
[OF-1598] Add support for Alpha Video Materials

MAG-AdamThorn:
Previously we only supported a single GLTF material type.
Support has now been added for a second material type: AlphaVideoMaterial.

### DIFF
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -57,8 +57,8 @@ public:
     const TextureInfo& GetColorTexture() const;
 
     /// @brief Sets the doubleSided property which specifies whether the material is double sided.
-    /// @details When this value is false, back - face culling is enabled, i.e., only front-facing triangles are rendered.
-    /// When this value is true, back - face culling is disabled and double sided lighting is enabled.
+    /// @details When this value is false, back-face culling is enabled, i.e., only front-facing triangles are rendered.
+    /// When this value is true, back-face culling is disabled and double sided lighting is enabled.
     /// @param DoubleSided bool : The double sided value.
     void SetDoubleSided(bool DoubleSided);
 
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -33,121 +33,109 @@ class JsonDeserializer;
 void ToJson(csp::json::JsonSerializer& Serializer, const csp::systems::AlphaVideoMaterial& Obj);
 void FromJson(const csp::json::JsonDeserializer& Deserializer, csp::systems::AlphaVideoMaterial& Obj);
 
-namespace csp::services
-{
-
-class ApiResponseBase;
-
-CSP_START_IGNORE
-template <typename T, typename U, typename V, typename W> class ApiResponseHandler;
-CSP_END_IGNORE
-
-} // namespace csp::services
-
 namespace csp::systems
 {
 
-
 /// @ingroup Asset System
-/// @brief Data class which represents a GLTF material.
+/// @brief Data class which represents a Alpha Video material.
 class CSP_API AlphaVideoMaterial : public Material
 {
 public:
-    /// @brief Sets the color texture. The first three components (RGB) MUST be encoded with the sRGB transfer function.
-    /// They specify the color of the material.
+    /// @brief Sets the color texture.
+    /// @details The first three components (RGB) MUST be encoded with the sRGB transfer function. They specify the color of the material.
     /// If the fourth component (A) is present, it represents the linear alpha coverage of the material.
     /// Otherwise, the alpha coverage is equal to 1.0.
     /// The material.blendMode property defines how the alpha is blended when alphaMode is set to BLEND.
     /// The material.readAlphaFromChannel property defines which color channel to read the alpha values from.
     /// The stored texels MUST NOT be premultiplied.
     /// When undefined, the texture MUST be sampled as having 1.0 in all components.
-    /// @param Texture const TextureInfo&
+    /// @param Texture const TextureInfo& : The material color texture.
     void SetColorTexture(const TextureInfo& Texture);
 
-    /// @brief Gets the color texture
-    /// @return const TextureInfo&
+    /// @brief Gets the color texture.
+    /// @return const TextureInfo& : The material color texture.
     const TextureInfo& GetColorTexture() const;
 
     /// @brief Sets the doubleSided property which specifies whether the material is double sided.
-    /// When this value is false, back - face culling is enabled, i.e., only front-facing triangles are rendered.
+    /// @details When this value is false, back - face culling is enabled, i.e., only front-facing triangles are rendered.
     /// When this value is true, back - face culling is disabled and double sided lighting is enabled.
-    /// @param DoubleSided bool
+    /// @param DoubleSided bool : The double sided value.
     void SetDoubleSided(bool DoubleSided);
 
-    /// @brief Gets the double sided value
-    /// @return bool
+    /// @brief Gets the double sided value.
+    /// @return bool : The double sided value.
     bool GetDoubleSided() const;
 
     /// @brief Sets whether the material is emissive, if not material should be lit by the scene lighting.
-    /// @param isEmissive bool
-    void SetIsEmissive(bool isEmissive);
+    /// @param IsEmissive bool : The emissive value.
+    void SetIsEmissive(bool IsEmissive);
 
     /// @brief Gets whether the material is emissive, if not material should be lit by the scene lighting.
     /// @return bool
     bool GetIsEmissive() const;
 
     /// @brief Sets the color channel to read the alpha values from.
-    /// @param channel EColorChannel
-    void SetReadAlphaFromChannel(EColorChannel channel);
+    /// @param ColorChannel EColorChannel : The color channel to read the alpha values from.
+    void SetReadAlphaFromChannel(EColorChannel ColorChannel);
 
-    /// @brief Gets the color channel to read the alpha values from
-    /// @return EColorChannel
+    /// @brief Gets the color channel to read the alpha values from.
+    /// @return EColorChannel : The color channel to read the alpha values from.
     EColorChannel GetReadAlphaFromChannel() const;
 
-    /// Sets how to alpha value is blended
-    /// @param Mode EAlphaMode
+    /// @brief Sets the mode to use for alpha blending.
+    /// @param Mode EBlendMode : The mode to use for alpha blending.
     void SetBlendMode(EBlendMode Mode);
 
-    /// Gets how to alpha value is blended
-    /// @return EBlendMode
+    /// @brief Gets the alpha blend mode.
+    /// @return EBlendMode : The alpha blend mode.
     EBlendMode GetBlendMode() const;
 
-    /// Sets how to alpha value is blended
-    /// @param Mode EAlphaMode
-    void SetFresnelFactor(float factor);
+    /// @brief Sets the fresnel factor.
+    /// @param Factor float : The fresnel factor.
+    void SetFresnelFactor(float Factor);
 
-    /// Gets how to alpha value is blended
-    /// @return EFresnelFactor
+    /// @brief Gets the fresnel factor.
+    /// @return float : The fresnel factor.
     float GetFresnelFactor() const;
 
-    /// SetTint
-    /// @param factor float
-    void SetTint(const csp::common::Vector3& factor);
+    /// @brief Set the Material tint.
+    /// @param Tint const csp::common::Vector3& : The tint value.
+    void SetTint(const csp::common::Vector3& Tint);
 
-    /// GetTint
-    /// @return float
+    /// @brief Get the Material tint.
+    /// @return const csp::common::Vector3& GetTint() : The tint value.
     const csp::common::Vector3& GetTint() const;
 
-    /// SetAlphaFactor
-    /// @param factor float
-    void SetAlphaFactor(float factor);
+    /// @brief Set the alpha factor.
+    /// @param Factor float : The alpha factor.
+    void SetAlphaFactor(float Factor);
 
-    /// GetAlphaFactor
-    /// @return float
+    /// @brief Get the alpha factor.
+    /// @return float : The alpha factor.
     float GetAlphaFactor() const;
 
-    /// SetEmissiveIntensity
-    /// @param factor float
-    void SetEmissiveIntensity(float factor);
+    /// @brief Set the emissive intensity.
+    /// @param Intensity float : The emissive intensity.
+    void SetEmissiveIntensity(float Intensity);
 
-    /// GetEmissiveIntensity
-    /// @return float
+    /// @brief Get the emissive intensity.
+    /// @return float : The emissive intensity.
     float GetEmissiveIntensity() const;
 
-    /// SetAlphaMask
-    /// @param mask float
-    void SetAlphaMask(float mask);
+    /// @brief Set the alpha mask.
+    /// @param Mask float : The alpha mask.
+    void SetAlphaMask(float Mask);
 
-    /// GetAlphaMask
-    /// @return float
+    /// @brief Get the alpha mask.
+    /// @return float : The alpha mask.
     float GetAlphaMask() const;
 
-    /// @brief Constructor which links the material to an asset
+    /// @brief Constructor which links the material to an asset.
     /// @param Name const csp::common::String& : The name of the material.
-    /// @param AssetCollectionId const csp::common::String& : The asset collection where the material info is stored
-    /// @param AssetId const csp::common::String& : The asset where the material info is stored
-    AlphaVideoMaterial(const csp::common::String& Name, const csp::common::String& AssetCollectionId, const csp::common::String& AssetId);
-    
+    /// @param MaterialCollectionId const csp::common::String& : The asset collection which holds the associated material asset.
+    /// @param MaterialId const csp::common::String& : The asset where the material info is stored.
+    AlphaVideoMaterial(const csp::common::String& Name, const csp::common::String& MaterialCollectionId, const csp::common::String& MaterialId);
+
     virtual ~AlphaVideoMaterial() = default;
     AlphaVideoMaterial();
 
@@ -158,7 +146,7 @@ private:
     EColorChannel ReadAlphaFromChannel;
     EBlendMode BlendMode;
     float FresnelFactor;
-    csp::common::Vector3 Tint;
+    csp::common::Vector3 MaterialTint;
     float AlphaFactor;
     float EmissiveIntensity;
     float AlphaMask;
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Magnopus LLC
+ * Copyright 2025 Magnopus LLC
 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -37,7 +37,7 @@ namespace csp::systems
 {
 
 /// @ingroup Asset System
-/// @brief Data class which represents a Alpha Video material.
+/// @brief Data class which represents an Alpha Video material.
 class CSP_API AlphaVideoMaterial : public Material
 {
 public:
@@ -48,8 +48,8 @@ public:
     /// The material.blendMode property defines how the alpha is blended when alphaMode is set to BLEND.
     /// The material.readAlphaFromChannel property defines which color channel to read the alpha values from.
     /// The stored texels MUST NOT be premultiplied.
-    /// When undefined, the texture MUST be sampled as having 1.0 in all components.
     /// @param Texture const TextureInfo& : The material color texture.
+    /// @pre If the fourth component (A) is undefined, the texture MUST be sampled as having 1.0 in all components.
     void SetColorTexture(const TextureInfo& Texture);
 
     /// @brief Gets the color texture.
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -141,7 +141,7 @@ public:
 
 private:
     TextureInfo ColorTexture;
-    bool DoubleSided;
+    bool IsDoubleSided;
     bool IsEmissive;
     EColorChannel ReadAlphaFromChannel;
     EBlendMode BlendMode;
--- a/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
+++ b/Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2023 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "CSP/Common/Array.h"
+#include "CSP/Systems/WebService.h"
+#include "Material.h"
+
+namespace csp::systems
+{
+class AlphaVideoMaterial;
+}
+
+namespace csp::json
+{
+class JsonSerializer;
+class JsonDeserializer;
+} // namespace csp::json
+
+void ToJson(csp::json::JsonSerializer& Serializer, const csp::systems::AlphaVideoMaterial& Obj);
+void FromJson(const csp::json::JsonDeserializer& Deserializer, csp::systems::AlphaVideoMaterial& Obj);
+
+namespace csp::services
+{
+
+class ApiResponseBase;
+
+CSP_START_IGNORE
+template <typename T, typename U, typename V, typename W> class ApiResponseHandler;
+CSP_END_IGNORE
+
+} // namespace csp::services
+
+namespace csp::systems
+{
+
+
+/// @ingroup Asset System
+/// @brief Data class which represents a GLTF material.
+class CSP_API AlphaVideoMaterial : public Material
+{
+public:
+    /// @brief Sets the color texture. The first three components (RGB) MUST be encoded with the sRGB transfer function.
+    /// They specify the color of the material.
+    /// If the fourth component (A) is present, it represents the linear alpha coverage of the material.
+    /// Otherwise, the alpha coverage is equal to 1.0.
+    /// The material.blendMode property defines how the alpha is blended when alphaMode is set to BLEND.
+    /// The material.readAlphaFromChannel property defines which color channel to read the alpha values from.
+    /// The stored texels MUST NOT be premultiplied.
+    /// When undefined, the texture MUST be sampled as having 1.0 in all components.
+    /// @param Texture const TextureInfo&
+    void SetColorTexture(const TextureInfo& Texture);
+
+    /// @brief Gets the color texture
+    /// @return const TextureInfo&
+    const TextureInfo& GetColorTexture() const;
+
+    /// @brief Sets the doubleSided property which specifies whether the material is double sided.
+    /// When this value is false, back - face culling is enabled, i.e., only front-facing triangles are rendered.
+    /// When this value is true, back - face culling is disabled and double sided lighting is enabled.
+    /// @param DoubleSided bool
+    void SetDoubleSided(bool DoubleSided);
+
+    /// @brief Gets the double sided value
+    /// @return bool
+    bool GetDoubleSided() const;
+
+    /// @brief Sets whether the material is emissive, if not material should be lit by the scene lighting.
+    /// @param isEmissive bool
+    void SetIsEmissive(bool isEmissive);
+
+    /// @brief Gets whether the material is emissive, if not material should be lit by the scene lighting.
+    /// @return bool
+    bool GetIsEmissive() const;
+
+    /// @brief Sets the color channel to read the alpha values from.
+    /// @param channel EColorChannel
+    void SetReadAlphaFromChannel(EColorChannel channel);
+
+    /// @brief Gets the color channel to read the alpha values from
+    /// @return EColorChannel
+    EColorChannel GetReadAlphaFromChannel() const;
+
+    /// Sets how to alpha value is blended
+    /// @param Mode EAlphaMode
+    void SetBlendMode(EBlendMode Mode);
+
+    /// Gets how to alpha value is blended
+    /// @return EBlendMode
+    EBlendMode GetBlendMode() const;
+
+    /// Sets how to alpha value is blended
+    /// @param Mode EAlphaMode
+    void SetFresnelFactor(float factor);
+
+    /// Gets how to alpha value is blended
+    /// @return EFresnelFactor
+    float GetFresnelFactor() const;
+
+    /// SetTint
+    /// @param factor float
+    void SetTint(const csp::common::Vector3& factor);
+
+    /// GetTint
+    /// @return float
+    const csp::common::Vector3& GetTint() const;
+
+    /// SetAlphaFactor
+    /// @param factor float
+    void SetAlphaFactor(float factor);
+
+    /// GetAlphaFactor
+    /// @return float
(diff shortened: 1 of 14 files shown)


Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
```diff
@@ -0,0 +1,158 @@
...
+namespace csp::systems
+{
+
+/// @ingroup Asset System
+/// @brief Data class which represents a Alpha Video material.
+class CSP_API AlphaVideoMaterial : public Material
+{
+public:
+    /// @brief Sets the color texture.
+    /// @details The first three components (RGB) MUST be encoded with the sRGB transfer function. They specify the color of the material.
```
MAG-ElliotMorris, line 45: [consider] Is it possible to point more strongly to what function we mean, as this is public API. A doxygen link would be ideal. Maybe also consider documenting this as a `@pre` precondition.

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
```diff
@@ -0,0 +1,158 @@
...
+    /// If the fourth component (A) is present, it represents the linear alpha coverage of the material.
+    /// Otherwise, the alpha coverage is equal to 1.0.
+    /// The material.blendMode property defines how the alpha is blended when alphaMode is set to BLEND.
+    /// The material.readAlphaFromChannel property defines which color channel to read the alpha values from.
+    /// The stored texels MUST NOT be premultiplied.
+    /// When undefined, the texture MUST be sampled as having 1.0 in all components.
```
MAG-ElliotMorris, line 51: [Consider] Also possible `@pre` 's

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
+/// @brief Enum representing the shader type of a material.
 enum class EShaderType
 {
-    Standard
+    Standard = 0,
+    AlphaVideo = 1
 };
 
-/// @ingroup Asset System
+/// @brief Defines how to alpha value is interpreted.
```
MAG-ElliotMorris, line 54: [fix] Here and below, grammar.

Also "How the alpha value of a material is interpreted" would be better if they are to remain in the `systems` namespace (ie, without context that these types relate to materials)

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
     /// @param Name const csp::common::String& : The name of the material.
-    /// @param MaterialCollectionId const csp::common::String& : The asset collection where the material info is stored
-    /// @param MaterialId const csp::common::String& : The asset where the material info is stored
+    /// @param MaterialCollectionId const csp::common::String& : The asset collection which holds the associated material asset.
+    /// @param MaterialId const csp::common::String& : The asset where the material info is stored.
     Material(const csp::common::String& Name, const csp::common::String& MaterialCollectionId, const csp::common::String& MaterialId);
-    Material() = default;
 
-protected:
+    /// @brief Constructor which links a versioned material of a given type to an asset.
```
MAG-ElliotMorris, line 113: [Nit] What does this mean in terms of the object this constructor is creating? Can we frame this more as `Creates a Material that is X`? Motivation being that constructors create obects, whereas this description makes it feel more like a free function that does some sort of data join.

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
+
+    CSP_START_IGNORE
+    template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+    CSP_END_IGNORE
+    /** @endcond */
+
+public:
+    /// @brief Retreives an Array of Materials from the result.
+    /// @return const Array<Material*>* : Returns a const pointer to an Array of Material class pointers. The caller should take ownership of the
+    /// pointer.
```
MAG-ElliotMorris, line 196: [Comment] _Which_ pointers? This isn't a criticism of this particular doc (in fact, this doc is better than most because it does specify ownership, hooray!), it's more a criticism of our Array types themselves, there's two categories of ownership here, the pointer holding the `Array` and all the pointers in the array that hold the objects themselves. 

We gotta change this sort of thing at some point soon, it's an untenably difficult API surface.

Library/src/Json/JsonSerializer.h
```diff
@@ -125,22 +127,38 @@ class JsonDeserializer
...
-    /// This will return true if the given key exists
-    /// @param Key const char* : The key to check
-    /// @return bool : Returns true if the key exists in this object
+    /// @brief Should be called within custom FromJson function.
+    /// @details This will safely deserialize a member with the given key by first checking that the member exists.
+    /// If the member is another custom type, this was internally call FromJson on that type.
+    /// @param Key const char* : The key which references this member.
+    /// @param Val T& : The member to deserialize to.
+    /// @return bool : Returns true if the key exists in this object and the member has been deserialized.
+    template <typename T> bool SafeDeserializeMember(const char* Key, T& Val) const
```
MAG-ElliotMorris, line 148: [Comment] 👍 I'm a fan of this, good doc too.
[Consider] renaming `Val` -> `OutVal`, or even better, returning an optional type instead.

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
```diff
@@ -0,0 +1,168 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris, line 2: [nit] 2025

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -93,6 +95,36 @@ String ConvertAssetTypeToString(systems::EAssetType AssetType)
     }
 }
 
+String ConvertShaderTypeToString(systems::EShaderType ShaderType)
+{
+    if (ShaderType == systems::EShaderType::AlphaVideo)
```
MAG-ElliotMorris, line 100: [Consider] Here and elsewhere, using switch statements when enumerating all cases of an enum like this. The benefit being that with high enough compiler warnings, it will tell you when all the cases arn't exhausted, preventing bugs when a new enum is added but we forget about these methods.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -165,6 +197,99 @@ csp::common::String CreateUniqueMaterialFileName(const csp::common::String& Name
 namespace csp::systems
 {
 
+void InstantiateMaterialOfType(csp::systems::EShaderType ShaderType, const csp::common::String& Name, const csp::common::String& AssetCollectionId,
+    const csp::common::String& AssetId, Material** OutMaterial)
```
MAG-ElliotMorris, line 201: [consider] Here and below. Pointer to pointer is a confusing interface, could you return a pointer instead?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -165,6 +197,99 @@ csp::common::String CreateUniqueMaterialFileName(const csp::common::String& Name
...
+}
+
+bool DeserializeIntoMaterialOfType(const char* MaterialData, csp::systems::EShaderType ShaderType, Material** OutMaterial)
+{
+    bool Deserialized = false;
+
+    if (ShaderType == EShaderType::AlphaVideo)
+    {
+        AlphaVideoMaterial* NewAlphaVideoMaterial = static_cast<AlphaVideoMaterial*>(*OutMaterial);
+        Deserialized = csp::json::JsonDeserializer::Deserialize(MaterialData, *NewAlphaVideoMaterial);
```
MAG-ElliotMorris, line 226: [nit] You could probably get away without the casting by doing `Deserialize<MaterialType>(...)`, although I guess relying on deduction from the cast is effectively the same thing 🤷

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -165,6 +197,99 @@ csp::common::String CreateUniqueMaterialFileName(const csp::common::String& Name
...
+    {
+        assert(false && "Unable to serialize material. Unsupported Shader Type.");
+    }
+}
+
+EShaderType GetShaderTypeFromMaterialCollection(const csp::systems::AssetCollection& AssetCollection)
+{
+    // Please note: Older Spaces built when only GLTF materials were supported will not have stored
+    // shader type metadata in the Material Asset Collection. We therefore default to a standard shader type.
+    EShaderType ShaderType = EShaderType::Standard;
```
MAG-ElliotMorris, line 263: [comment] 👍 great context more of this

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -35,50 +37,81 @@ bool RequestPredicate(const csp::systems::ResultBase& Result) { return Result.Ge
 
 } // namespace
 
-void CreateMaterial(AssetSystem* AssetSystem, const csp::common::String& Name, const csp::common::String& SpaceId,
-    const csp::common::Map<csp::common::String, csp::common::String>& Metadata, const csp::common::Array<csp::common::String>& AssetTags,
-    GLTFMaterial& OutMaterial)
+void CreateMaterial(AssetSystem* AssetSystem, const csp::common::String& Name, const csp::systems::EShaderType shaderType,
```
MAG-ElliotMorris, line 40: [Consider] Is there a need for some new tests exercising error cases as well as happy paths? Would seem like there's probably some new errors introduced by the addition of branching on material types, like when the user provides the wrong material enum to an interface.

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
-    Material() = default;
 
-protected:
+    /// @brief Constructor which links a versioned material of a given type to an asset.
+    /// @param Name const csp::common::String& : The name of the material.
+    /// @param MaterialCollectionId const csp::common::String& : The asset collection which holds the associated material asset.
+    /// @param MaterialId const csp::common::String& : The asset where the material info is stored.
+    /// @param InType EShaderType : The material shader type.
+    /// @param InVersion const int : The material version.
+    Material(const csp::common::String& Name, const csp::common::String& MaterialCollectionId, const csp::common::String& MaterialId,
```
MAG-ElliotMorris, line 119: [consider] this In naming pattern is problematic imo, InVersion can very easily be read (as I did) as inversion, like the colors are inverted or something? (As well as also being the mirror to the common OutVariable pattern, indicating an interface that takes memory ownership)

We should consider going to _ or something, or even better, adopting the m_memberVar pattern. I think there's justification as it does effect our public API, as evidenced here.

Maybe one for discussion in a broader context.

MAG-ElliotMorris (changes requested):
[fix] Remove `mac-build` binary file. (Can't comment on it directly ... thanks github)

I know there were extenuating circumstances given that we had to adopt prior work, however it's a shame that this PR isn't reviewable as a story, alas 🤷.

In the future, a callout box in the description saying that the PR should be reviewed as a whole as it undoes work from prior commits might be helpful, if we can't massage the history well enough. Although in that case, one might argue you may as well squash everything into a single commit to remove the possibility of mis-review. 🤔

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -165,6 +197,99 @@ csp::common::String CreateUniqueMaterialFileName(const csp::common::String& Name
...
+        Deserialized = csp::json::JsonDeserializer::Deserialize(MaterialData, *NewAlphaVideoMaterial);
+    }
+    else if (ShaderType == EShaderType::Standard)
+    {
+        GLTFMaterial* NewGLTFMaterial = static_cast<GLTFMaterial*>(*OutMaterial);
+        Deserialized = csp::json::JsonDeserializer::Deserialize(MaterialData, *NewGLTFMaterial);
+    }
+    else
+    {
+        assert(false && "Unable to deserialize material. Unsupported Shader Type.");
```
MAG-mv, line 235: [question] Should these asserts be logs? If this can happen due to incorrect CSP versions, I would say this shouldnt be an assert

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -834,47 +834,55 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void AssetSystem::RegisterAssetToLODChain(
...
-                    Callback(GLTFMaterialResult(UploadResult.GetResultCode(), UploadResult.GetHttpResultCode()));
+                    Callback(MaterialResult(UploadResult.GetResultCode(), UploadResult.GetHttpResultCode()));
                     return;
                 }
 
                 // 4. Return created material
-                GLTFMaterialResult FinalResult(UploadResult.GetResultCode(), UploadResult.GetHttpResultCode());
-                FinalResult.SetGLTFMaterial(NewMaterial);
+                MaterialResult FinalResult(UploadResult.GetResultCode(), UploadResult.GetHttpResultCode());
+                FinalResult.SetMaterial(*NewlyCreatedMaterial);
```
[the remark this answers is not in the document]
MAG-mv: DId we decide to get the MaterialResult to delete the pointers? If not, we should ass a post condition to our docs to mention they should delete

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -246,37 +247,36 @@ class CSP_API AssetSystem : public SystemBase
...
     /// @brief Deletes a given material.
     /// The material should be retrieved through GetMaterials or GetMaterial.
     /// @param Material const GLTFMaterial& : The material to delete
     /// @param Callback NullResultCallback : Callback when asynchronous task finishes.
-    CSP_ASYNC_RESULT void DeleteMaterial(const GLTFMaterial& Material, NullResultCallback Callback);
+    CSP_ASYNC_RESULT void DeleteMaterial(const Material& Material, NullResultCallback Callback);
 
     /// @brief Gets all materials associated with the given space.
     /// @param SpaceId const csp::common::String& : The space id the material is associated with.
     /// @param Callback GLTFMaterialsResultCallback : Callback when asynchronous task finishes.
```
MAG-ME, line 271: MaterialsResultCallback

Library/src/Systems/Assets/AssetSystem.cpp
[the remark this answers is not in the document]
MAG-ElliotMorris: That was our decision, my comment here was misinterpreting the fact that `SetMaterial` takes a reference, but the result type _does_ still store a pointer, so our contract still holds, as shaky as it is to be manually asking users to take ownership.

[Consider] having the setter take a pointer though, references generally imply not taking ownership.

MAG-AdamThorn:
Removed mac build binary file.

[bb87233](https://github.com/magnopus-opensource/connected-spaces-platform/commit/bb872330761ab55e08349812c6ef3c162555bf82)

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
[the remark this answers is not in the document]
MAG-AdamThorn: Fixed:

[926fa67](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/926fa6736dad2c303e4e6914818a0cf1ce06c54c)

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
+        EShaderType InType, const int InVersion);
+
+    virtual ~Material() = default;
+
+    Material() = default;
     csp::common::String Name;
```
MAG-ME, line 125: Is there a need for these to be public. Seems risky especially as MaterialsResult::GetMaterial returns a pointer to a material which could then me used to modify the material

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
```diff
@@ -0,0 +1,168 @@
...
+    Deserializer.SafeDeserializeMember(AlphaVideoMaterialProperties::AlphaFactor, Obj.AlphaFactor);
+    Deserializer.SafeDeserializeMember(AlphaVideoMaterialProperties::AlphaMask, Obj.AlphaMask);
+}
+
+namespace csp::systems
+{
+
+AlphaVideoMaterial::AlphaVideoMaterial(
+    const csp::common::String& Name, const csp::common::String& AssetCollectionId, const csp::common::String& AssetId)
+    : Material(Name, AssetCollectionId, AssetId, EShaderType::AlphaVideo, 1)
```
MAG-ME, line 108: Using numeric literals means the reader needs to understand the parameters to the constructor. Consider using something named to explain the parameter's use

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
```diff
@@ -0,0 +1,170 @@
...
+    /// @param Texture const TextureInfo&
+    void SetColorTexture(const TextureInfo& Texture);
+
+    /// @brief Gets the color texture
+    /// @return const TextureInfo&
+    const TextureInfo& GetColorTexture() const;
+
+    /// @brief Sets the doubleSided property which specifies whether the material is double sided.
+    /// When this value is false, back - face culling is enabled, i.e., only front-facing triangles are rendered.
+    /// When this value is true, back - face culling is disabled and double sided lighting is enabled.
```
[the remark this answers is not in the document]
MAG-AdamThorn: Fixed in: [a547d85](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/a547d85935d8ef24f55dd414bfaf1320fc1ceec6)

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -16,48 +16,209 @@
...
+
+/// @ingroup Asset System.
 /// @brief Base class for a material.
 class CSP_API Material
 {
 public:
-    /// @brief Gets the user-defined name of the material
-    /// @return csp::common::String&
+    /// @brief Gets the user-defined name of the material.
+    ///  @return const common::String& : Returns the material name.
```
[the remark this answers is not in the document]
MAG-AdamThorn: Fixed in: [a547d85](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/a547d85935d8ef24f55dd414bfaf1320fc1ceec6)

Library/include/CSP/Systems/Assets/Material.h
MAG-AdamThorn: To be honest this was a miss from me. I removed other instances of arguments being prefixed with 'in', but missed the Version argument here and a Type argument elsewhere in the class. I have renamed the argument to be just 'Version' in this instance to make it clearer. I agree that a broader conversation with the team would be of value.

I have addressed this here:  [48ce73f](https://github.com/magnopus-opensource/connected-spaces-platform/commit/48ce73f290d8595903f671d5d2af722773042c18)

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
```diff
@@ -0,0 +1,186 @@
...
+    SafeDeserialize(Deserializer, AlphaVideoMaterialProperties::EmissiveIntensity, Obj.EmissiveIntensity);
+    SafeDeserialize(Deserializer, AlphaVideoMaterialProperties::AlphaFactor, Obj.AlphaFactor);
+    SafeDeserialize(Deserializer, AlphaVideoMaterialProperties::AlphaMask, Obj.AlphaMask);
+}
+
+namespace csp::systems
+{
+
+AlphaVideoMaterial::AlphaVideoMaterial(const csp::common::String& Name, const csp::common::String& AssetCollectionId, const csp::common::String& AssetId)
+    : Material(Name, AssetCollectionId, AssetId, EShaderType::AlphaVideo, 1)
```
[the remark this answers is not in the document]
MAG-AdamThorn: It is my understanding that this material version is for internal versioning only. It will allow us to make changes to materials in the future and to gate their use depending on which versions of a material are supported by a client application for example. I will confirm this though.

Library/include/CSP/Systems/Assets/Material.h
```diff
@@ -18,16 +18,66 @@
...
+
+class ApiResponseBase;
+
+CSP_START_IGNORE
+template <typename T, typename U, typename V, typename W> class ApiResponseHandler;
+CSP_END_IGNORE
+
+} // namespace csp::services
+
+namespace csp::systems
```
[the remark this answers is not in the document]
MAG-AdamThorn: I attempted to move the materials into a csp::systems::material namespace as suggested as I think that is the right approach. Unfortunately we currently hardcode the namespace mappings in the C# Wrapper Generator (as seen [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/81e9f44cc7eecb5726a0997e7dd983d986394f33/Tools/WrapperGenerator/CSharpWrapperGenerator.py#L37)), so this did not work. I will capture that as something that needs to be addressed.

I have therefore updated inline documentation to specify intended usage. Fix: [72957ae](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/72957ae3b4fd7b27f39ef08fa86e410ca42f2267)

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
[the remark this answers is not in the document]
MAG-DanielMurdolo: As the person who didn't write this spec I'd have to defer to the tech docs. It does outline the version there: https://github.com/magnopus/oko-tech-docs/blob/main/features/video-texture-custom-materials/alpha-texture.md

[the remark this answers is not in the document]
MAG-DanielMurdolo: But what Adam said sounds like how I'd interpret it

Library/src/Systems/Assets/AssetSystem.cpp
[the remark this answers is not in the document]
MAG-AdamThorn: Agreed. I have updated the SetMaterial method to take a pointer and the documentation for this and the SetMaterials method to state that the result objects are taking ownership of the pointers passed to them.

Fix: [dc3c606](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/dc3c6067a89c4c10c6b12599bc905c038d557c67)

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: All good

Library/src/Systems/Assets/GLTFMaterial.cpp
```diff
@@ -18,131 +18,105 @@
...
+        GLTFMaterialProperties::EmissiveFactor, csp::common::Array<float> { Obj.EmissiveFactor.X, Obj.EmissiveFactor.Y, Obj.EmissiveFactor.Z });
 
-    // Roughness Factor
-    Serializer.SerializeMember("roughnessFactor", Obj.RoughnessFactor);
-
-    // Emissive Factor
-    Serializer.SerializeMember("emissiveFactor", csp::common::Array<float> { Obj.EmissiveFactor.X, Obj.EmissiveFactor.Y, Obj.EmissiveFactor.Z });
-
-    // Textures
     if (Obj.BaseColorTexture.IsSet())
```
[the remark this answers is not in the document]
MAG-AdamThorn: These have not been removed.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-AdamThorn: @MAG-ME I am not clear what this comment means. This method takes a **NullResultCallback**, not a MaterialsResultCallback.

MAG-AdamThorn:
@MAG-mv I could not reply to your comment [above](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659#discussion_r2037530914) so I am replying here.

After speaking to the Adrian it was decided not to have the MaterialResult dtor clean up the pointer as they destroy the result object after parsing it. I have added inline documentation to the MaterialResult and MaterialsResult GetMaterial/s methods. Example show below:

/// @brief Retrieves the Material from the result.
/// @return const Material* : Returns a pointer to the Material object. The caller should take ownership of the pointer.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ME: Apologies, I was referring to the documentation `@param Material const GLTFMaterial& : The material to delete` when the parameter is of type `Material`

Library/src/Systems/Assets/AssetSystem.cpp
MAG-AdamThorn: Agreed. Fixed: [8bf1747](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/8bf17475d93d1762d22fa03d80ccf13e381e057a)

Library/src/Systems/Assets/AssetSystem.cpp
MAG-AdamThorn: This Method has been updated to return a std::pair based on one of your other comments. The specific change you suggested here is not possible however.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-AdamThorn: Ah right I see it now.

Fixed: [9de0739](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/9de0739989b37d8dd5268967303b41d73dd4ad1d)

Library/include/CSP/Systems/Assets/Material.h
MAG-AdamThorn: Good spot! Fixed: [eb98518](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/eb98518403119e4bc6f163df2738776899f70e03)

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
MAG-AdamThorn: Fixed: [b6ea34c](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/b6ea34c39da2850a9907b4401c7bc9f85f1c1f49)

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -165,6 +197,99 @@ csp::common::String CreateUniqueMaterialFileName(const csp::common::String& Name
...
+    }
+
+    return Deserialized;
+}
+
+void SerializeMaterialOfType(EShaderType ShaderType, const Material& Material, csp::common::String& OutMaterialJson)
+{
+    if (ShaderType == EShaderType::AlphaVideo)
+    {
+        const AlphaVideoMaterial* NewAlphaVideoMaterial = static_cast<const AlphaVideoMaterial*>(&Material);
```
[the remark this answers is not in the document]
MAG-AdamThorn: Agreed. I have updated the method to take the material by ptr as internally it is being cast to a derived type.

Fix: [737b56d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/737b56d116b2edd45cd39b517c27aba286af77cf)

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-AdamThorn: Agreed, I have added three new additional tests:
- CreateIncorrectMaterialTypeTest
- GetMaterialsWithIncorrectShaderTypeTest
- GetMaterialWithIncorrectShaderTypeTest

Also added in dynamic_casts to ensure casts to derived material types are valid. Added guard to avoid having to turn on RTTI for WASM.

[dd8ea90](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/dd8ea90cd51b1375af18e085196cbe5d5fafa1d1)

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1054,13 +1077,24 @@ void AssetSystem::GetMaterials(const csp::common::String& SpaceId, GLTFMaterials
                         return;
                     }
 
-                    (*DownloadedMaterials)[i] = FoundMaterial;
+                    if (FoundBaseMaterial.GetShaderType() == EShaderType::Standard)
+                    {
+                    GLTFMaterial* FoundMaterial = CSP_NEW GLTFMaterial("", AssetCollectionId, AssetId);
```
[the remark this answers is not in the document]
MAG-AdamThorn: This logic was updated by one of the prior commits.

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
MAG-AdamThorn: The sRGB transfer function is a gamma correction curve that converts image file color values for display. This is not a function in CSP that they must call, but rather it is a non-linear mapping that is automatically applied when they save their image as a png for example.

Library/include/CSP/Systems/Assets/Material.h
MAG-AdamThorn: Agreed.

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
```diff
@@ -0,0 +1,158 @@
...
+} // namespace csp::json
+
+void ToJson(csp::json::JsonSerializer& Serializer, const csp::systems::AlphaVideoMaterial& Obj);
+void FromJson(const csp::json::JsonDeserializer& Deserializer, csp::systems::AlphaVideoMaterial& Obj);
+
+namespace csp::systems
+{
+
+/// @ingroup Asset System
+/// @brief Data class which represents a Alpha Video material.
```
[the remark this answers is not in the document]
MAG-AdamThorn: Updated to 'an' though left the casing as is. That feels correct to me.

Fix: [2877b60](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/2877b601ad1339890e1343a81d2c06ff62567b0b)

Library/include/CSP/Systems/Assets/AlphaVideoMaterial.h
MAG-AdamThorn: Updated this line to make it slightly clearer and moved it to be a precondition.

Fix: [2877b60](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/2877b601ad1339890e1343a81d2c06ff62567b0b)

Library/include/CSP/Systems/Assets/Material.h
MAG-AdamThorn: Fixed the first point re the grammar. Sorry but I did not understand your second point.

Fix: [b03be98](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/b03be984db880820ddfd77d36db8f05e2c824c9f)

Library/include/CSP/Systems/Assets/Material.h
MAG-AdamThorn: Fix: [b03be98](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/b03be984db880820ddfd77d36db8f05e2c824c9f)

Library/src/Systems/Assets/AlphaVideoMaterial.cpp
MAG-AdamThorn: Fixed this and the copyright date for a few other source files.

Fix: [0cebd3a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/0cebd3ae99d57cc3a181e0ea8007ecee5281859c)

Library/src/Systems/Assets/AssetSystem.cpp
MAG-AdamThorn: Excellent point.

Updated: [c1e945a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/c1e945aec3cdee52ebf648c79bdde79793dc6b15)

Library/include/CSP/Systems/Assets/Material.h
MAG-ElliotMorris: Seems you did, as you change addresses it by directly referring to materials in the doc :P

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: [fix] (I've taken over this PR so i'll be fixing it), use an optional, not a pair here for the return with a success result. Clearer semantics.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-ElliotMorris: [fix] Avoid using out params when an optional return is viable.

Library/src/Json/JsonSerializer.h
MAG-ElliotMorris: I see every other method in this file doesn't do the `Out` thing, so leaving it for consistency.

Tests/src/PublicAPITests/MaterialTests.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/f4e6d88b987f9c887547d1693a48a38129d8bcf6

Library/src/Systems/Assets/AssetSystem.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/659/commits/f4e6d88b987f9c887547d1693a48a38129d8bcf6

PR #660, opened 2025-04-10
[OF-1550] chore: Raise warning level to 4

MAG-OR:
Note that a few warnings had to be suppressed in externals or in the
wrapper generator:
- "unreachable code" (4702) was suppressed in quickjs
- "'const' qualifier on return type has no effect" (5266) was suppressed
  in the wrapper generator for the MacOS and WASM builds
- "missing field initializer" was suppressed in the wrapper generator for
  the MacOS and WASM builds

Given its size, it is a good idea to review this PR commit by commit. The first commit contains the build changes, and every subsequent commit contains all the fixes for a given warning. The very last commit is actually a bug fix: removing the const char* serializer/deserializer because it never worked and the test gave a false positive.

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -173,9 +173,8 @@ if not Project then
 			buildoptions {
 				"-Wno-error=deprecated-declarations", --Don't error on deprecation warnings, this is because we use Uri a lot in our services generated code, which has deprecation warnings for some unused but still generated endpoints.
 				"-Wno-braced-scalar-init", -- Don't warn against doing stuff like `return {0}`, which we do in the interop output.
-				"-Wno-missing-field-initializers", -- Don't warn against missing field initializers (because of the wrapper generator)
-				"-Wno-error=unused-lambda-capture", --This shouldn't be disabled, we just had to rush to unblock android builds. Take all the this captures out and remove.
-				"-Wno-unknown-pragmas", --Also not the greatest. This is to try and suppress a signalR warning, even though the signalR project dosen't emit warnings (I think this error is a bit unique cause of preprocessor stuff)
+				"-Wno-missing-field-initializers", -- Don't warn against missing field initializers, e.g. uninitialized fields in structs (because of the wrapper generator)
+				"-Wno-unknown-pragmas", --Not the greatest. This is to try and suppress a signalR warning, even though the signalR project dosen't emit warnings (I think this error is a bit unique cause of preprocessor stuff)
 				"-Wno-error=nonportable-include-path", --Include paths dont match file structure. Should get around to fixing
 				"-Wno-error=format-security" --In logging, __android_log_print(ANDROID_LOG_VERBOSE, "CSP", InMessage.c_str()) is unnaceptable for some reason.
             }
@@ -258,8 +257,8 @@ if not Project then
                 "-fwasm-exceptions",    -- enable native wasm exceptions
                 "-Wno-error=deprecated-declarations", --Don't error on deprecation warnings, this is because we use Uri a lot in our services generated code, which has deprecation warnings for some unused but still generated endpoints.
                 "-Wno-braced-scalar-init", -- Don't warn against doing stuff like `return {0}`, which we do in the interop output.
-                "-Wno-missing-field-initializers", -- Don't warn against missing field initializers (because of the wrapper generator)
-                "-Wno-ignored-qualifiers" -- Don't warn against ignored qualifiers
+                "-Wno-missing-field-initializers", -- Don't warn against missing field initializers, e.g. uninitialized fields in structs (because of the wrapper generator)
+                "-Wno-ignored-qualifiers" -- Don't warn against ignored qualifiers, e.g. "const qualifier on return type has no effect" (because of the wrapper generator)
             }
 
             linkoptions { 
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -173,6 +173,7 @@ if not Project then
 			buildoptions {
 				"-Wno-error=deprecated-declarations", --Don't error on deprecation warnings, this is because we use Uri a lot in our services generated code, which has deprecation warnings for some unused but still generated endpoints.
 				"-Wno-braced-scalar-init", -- Don't warn against doing stuff like `return {0}`, which we do in the interop output.
+				"-Wno-missing-field-initializers", -- Don't warn against missing field initializers (because of the wrapper generator)
 				"-Wno-error=unused-lambda-capture", --This shouldn't be disabled, we just had to rush to unblock android builds. Take all the this captures out and remove.
 				"-Wno-unknown-pragmas", --Also not the greatest. This is to try and suppress a signalR warning, even though the signalR project dosen't emit warnings (I think this error is a bit unique cause of preprocessor stuff)
 				"-Wno-error=nonportable-include-path", --Include paths dont match file structure. Should get around to fixing
@@ -215,7 +216,7 @@ if not Project then
 			-- Could not manage to get xcode to co-operate in any other less specific manner of setting the flags.
 			-- These disables are to do with warnings in generated code that we should get around to dealing with.
 			xcodebuildsettings {
-				["WARNING_CFLAGS"] = "-Wno-error=deprecated-declarations -Wno-braced-scalar-init"
+				["WARNING_CFLAGS"] = "-Wextra -Wno-error=deprecated-declarations -Wno-braced-scalar-init -Wno-missing-field-initializers -Wno-ignored-qualifiers"
 			}
 
             links { 
@@ -236,7 +237,7 @@ if not Project then
 			-- Could not manage to get xcode to co-operate in any other less specific manner of setting the flags.
 			-- These disables are to do with warnings in generated code that we should get around to dealing with.
 			xcodebuildsettings {
-				["WARNING_CFLAGS"] = "-Wno-error=deprecated-declarations -Wno-braced-scalar-init"
+				["WARNING_CFLAGS"] = "-Wextra -Wno-error=deprecated-declarations -Wno-braced-scalar-init -Wno-missing-field-initializers -Wno-ignored-qualifiers"
 			}
 
             links {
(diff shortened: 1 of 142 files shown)


Tests/src/PublicAPITests/HotspotSequenceTests.cpp
```diff
@@ -243,6 +239,8 @@ CSP_PUBLIC_TEST(CSPEngine, HotspotSequenceTests, GetHotspotGroupTest)
     auto* SpaceSystem = SystemsManager.GetSpaceSystem();
     auto* HotspotSystem = SystemsManager.GetHotspotSequenceSystem();
 
+    const char* TestSpaceName = "CSP-UNITTEST-SPACE-MAG";
```
MAG-OR, line 242: since globals aren't used in the vast majority of the tests, I've removed them wherever they were redefined, and replaced the few former global uses with local variables where needed.

Tests/src/PublicAPITests/LogSystemTests.cpp
```diff
@@ -622,6 +622,7 @@ CSP_INTERNAL_TEST(CSPEngine, LogSystemTests, ProfileTest)
             std::cout << InMessage << std::endl;
         });
 
+#if CSP_PROFILING_ENABLED
```
MAG-OR, line 625: moved this so it doesn't complain below that some variables are unused. it's not obvious but there are ifdefs inside the macros

Library/src/Common/DateTime.cpp
```diff
@@ -88,8 +88,21 @@ DateTime::DateTime(const csp::common::String& DateString)
...
+        Second, // tm_sec
+        Minute, // tm_min
+        Hour, // tm_hour
+        Day, // tm_mday
+        Month - 1, // tm_mon
+        Year - 1900, // tm_year
+        0, // tm_wday
+        0, // tm_yday
+        -1, // tm_isdst
+#ifndef CSP_WINDOWS
```
MAG-OR, line 101: to state the obvious: the last 2 are unix. the macos build was complaining about them. all the builds were complaining about the other 3.

MAG-mv (changes requested):
Looks good overall! A couple of issues that need fixing

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -147,7 +147,7 @@ class EntitySystemScriptInterface
             for (size_t i = 0; i < EntitySystem->GetNumEntities(); ++i)
             {
                 const SpaceEntity* Entity = EntitySystem->GetEntityByIndex(i);
-                if (Entity->GetId() == EntityId)
+                if (static_cast<int32_t>(Entity->GetId()) == EntityId)
```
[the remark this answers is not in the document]
MAG-OR: yeah I have no idea why it says int32. probably a copy paste error that got lost. fixed in 91968aabd610889fbee630bcf3c2c762f2f1e924

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -170,7 +170,7 @@ class EntitySystemScriptInterface
             for (size_t i = 0; i < EntitySystem->GetNumEntities(); ++i)
             {
                 SpaceEntity* Entity = EntitySystem->GetEntityByIndex(i);
-                if (Entity->GetId() == EntityId)
+                if (static_cast<int32_t>(Entity->GetId()) == EntityId)
```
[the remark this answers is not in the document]
MAG-OR: fixed in 91968aabd610889fbee630bcf3c2c762f2f1e924

MultiplayerTestRunner/src/RunnableTests/CreateAvatar.cpp
```diff
@@ -59,8 +56,6 @@ void RunTest()
     EntitySystem.CreateAvatar(UserName, UserTransform, UserAvatarState, UserAvatarId, UserAvatarPlayMode,
         [&ResultPromise](csp::multiplayer::SpaceEntity* Result) { ResultPromise.set_value(Result); });
 
-    csp::multiplayer::SpaceEntity* Avatar = ResultFuture.get();
```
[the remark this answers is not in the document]
MAG-OR: good catch, fixed in 91968aabd610889fbee630bcf3c2c762f2f1e924

MAG-mv (approved):

Tests/src/PublicAPITests/LogSystemTests.cpp
MAG-ElliotMorris: [Consider] A comment in the code saying this.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -70,8 +70,6 @@ std::pair<ErrorCode, std::string> MultiplayerConnection::ParseMultiplayerErrorFr
     {
         return ParseMultiplayerError(e);
     }
-
-    return { ErrorCode::Unknown, "MultiplayerConnection::ParseMultiplayerErrorFromExceptionPtr, Unexpectedly no exception was thrown." };
```
MAG-ElliotMorris, line 74: [fix] I think this is a bug, one I implemented. I'm going to ask you to fix it here if that's okay.

I was under the impression when I implemented this function that `rethrow_exception` would not throw if the exception pointer is empty. _However_, I read the docs just now and it actually states that the exception pointer must not be null before you call it.

This is a latent bug, luckily, I don't think this functions can be called with a null exception pointer at the moment, but we should fix it anyway, to :

```cpp
try
{
  if (Exception) 
  {
    rethrow_exception(Exception);
  }
  else
  {
    return { ErrorCode::Unknown, "MultiplayerConnection::ParseMultiplayerErrorFromExceptionPtr, Unexpectedly no exception was thrown." };
  }
}
{
  return ParseMultiplayerError(e);
}
 ```
I've grepped the codebase and we don't seem to do an unguarded rethrow anywhere else outside of the tests.

File: Tests/src/PublicAPITests/SystemResultTests.cpp [hunk 8]
```diff
@@ -129,8 +129,9 @@ CSP_PUBLIC_TEST(CSPEngine, SystemResultTests, BaseResultTest)
     ResponseReceiver Receiver;
 
     // Synthesise a response to feed to ApiResponseBase
-    csp::web::HttpRequest MyTestRequest = csp::web::HttpRequest(WebClient, csp::web::ERequestVerb::GET, csp::web::Uri(),
-        csp::web::HttpPayload(MyTestPayload), &Receiver, csp::common::CancellationToken::Dummy());
+    csp::web::HttpPayload* MyHttpPayload = CSP_NEW csp::web::HttpPayload(MyTestPayload);
+    csp::web::HttpRequest MyTestRequest = csp::web::HttpRequest(
+        WebClient, csp::web::ERequestVerb::GET, csp::web::Uri(), *MyHttpPayload, &Receiver, csp::common::CancellationToken::Dummy());
```
MAG-ElliotMorris, line 134: [Consider] Presumably this error was about how this is a reference to a temporary?

Can you construct `MyHttpPayload` as a value rather than a pointer, and provide it directly to the `HttpRequest` constructor, it takes it by ref anyhow?

MAG-ElliotMorris (changes requested):
🥇 
Perhaps the most beautiful changeset I have ever seen :P

One [fix], which is me selfishly asking you to fix my mistake.

Tests/src/PublicAPITests/LogSystemTests.cpp
MAG-OR: I don't know what I was on when I made the note to write this in the PR later. The unused variable is TestValue, there is no mystery with hidden macros. what painstakingly fixing warnings will do to you...

Library/premake5.lua
```diff
@@ -254,9 +255,11 @@ if not Project then
...
                 "-pthread",             -- enable threading
-                "-fwasm-exceptions",     -- enable native wasm exceptions
-				"-Wno-error=deprecated-declarations", --Don't error on deprecation warnings, this is because we use Uri a lot in our services generated code, which has deprecation warnings for some unused but still generated endpoints.
-				"-Wno-braced-scalar-init" -- Don't warn against doing stuff like `return {0}`, which we do in the interop output.
+                "-fwasm-exceptions",    -- enable native wasm exceptions
+                "-Wno-error=deprecated-declarations", --Don't error on deprecation warnings, this is because we use Uri a lot in our services generated code, which has deprecation warnings for some unused but still generated endpoints.
+                "-Wno-braced-scalar-init", -- Don't warn against doing stuff like `return {0}`, which we do in the interop output.
+                "-Wno-missing-field-initializers", -- Don't warn against missing field initializers (because of the wrapper generator)
+                "-Wno-ignored-qualifiers" -- Don't warn against ignored qualifiers
```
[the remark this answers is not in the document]
MAG-OR: fixed in 890ce4bb009b99b77abc4c8f9ae13a548c92881d

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-OR: fixed in 0942880886b4b0b44d333f9c406ec17c0d1e08f9

Tests/src/PublicAPITests/SystemResultTests.cpp
MAG-OR: I think I had tried various combinations, but I'll check again. Isn't "construct as value and pass directly" what was there in the first place?

MAG-OR: defined as a value and then used in 8ffb392a7c3d54829c08a847c894bec03a73c567

MAG-AdamThorn (approved):

Library/src/Multiplayer/SpaceTransform.cpp
```diff
@@ -78,17 +78,18 @@ csp::multiplayer::SpaceTransform csp::multiplayer::SpaceTransform::operator*(con
     const auto TRS2 = T2 * R2 * S2;
     const auto ComposedTRS = TRS1 * TRS2;
 
-    glm::vec3 Scale;
-    glm::quat Rotation;
+    glm::vec3 TransformScale;
```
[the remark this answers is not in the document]
MAG-OR: the changes here are solely for the purpose of fixing the "variable name hides class member" warnings. the others don't have the warning, hence they are unchanged. this PR is big enough that we should avoid mixing in changes that are not warning fixes, otherwise it will grow exponentially.

Tests/src/PublicAPITests/SystemResultTests.cpp
MAG-ElliotMorris: > Isn't "construct as value and pass directly" what was there in the first place?


Yes an no, you had a temporary, an R-value, (or a PrValue or xValue or whatever the damn spec calls that specific thing). This is an L-value now. :D 

If you didn't know, the longer name for `&` is L-value reference, just like how `&&` is R-value reference.

note: I've always assumed the L and R mean left and right, as in

```
Type myType = Type()
//myType is on the left, an L-value
//Type() is on the right, an R-value
```

But i don't know if that's why they're actually called that.

MAG-ElliotMorris (approved):

PR #662, opened 2025-04-11
[NT-0] fix: Correct script property declaration types for avatar component HandRotation and HeadRotation properties

MAG-ThomasGreenhalgh:
The avatar component `HandRotation` and `HeadRotation` were declared with the wrong types in the component script interface header. This PR updates the declaration to match the definition and component property types.


MAG-ElliotMorris (approved):
Hey, nice spot, thanks for just fixing this.

Did a quick run through and can see all the other Hand/HeadRotation types are Vector4's, so this seems rather cut and dry.

Annoying none of our tests saw this, I suspect we don't test our component bindings all that well if at all :(

MAG-OR (approved):

PR #663, opened 2025-04-14
[NT-0] refac: extend continuation to support templates and default scheduler

MAG-ChristopherAtkinson:
Added changes to support templated type in continuations as well as default scheduler based of off @MAG-mv changes on OF-1386-annotations

### DIFF
--- a/Library/src/Common/Continuations.h
+++ b/Library/src/Common/Continuations.h
@@ -39,6 +39,16 @@
 
 #pragma once
 
+#define LIBASYNC_CUSTOM_DEFAULT_SCHEDULER 1;
+
+// Declare async::default_scheduler with csp_scheduler
+class csp_scheduler;
+
+namespace async
+{
+csp_scheduler& default_scheduler();
+}
+
 #include "CSP/Systems/Log/LogSystem.h"
 #include "CSP/Systems/SystemsResult.h"
 #include "CSP/Systems/WebService.h"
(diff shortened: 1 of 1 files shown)


Library/src/Common/Continuations.h
```diff
@@ -174,6 +206,17 @@ template <typename Callable> inline auto InvokeIfExceptionInChain(Callable&& Inv
         }
     };
 }
+
+/* Stores the result in a shared pointer for access outside of continuation */
+template <typename ResultT> inline auto GetResultFromContinuation(std::shared_ptr<ResultT>& Ptr)
```
MAG-ElliotMorris, line 211: [Comment] Approving because I can see why this folks feel this is neccesary right now, but I hope we come up with a pattern to make this unnecesary. There isn't any reason we shouldn't be able to pass aggregate types through the continuation chain. Using this creates unfortunate coupling.

MAG-ElliotMorris (approved):
LGTM

MAG-ElliotMorris:
As an aside, it probably would have been better if we could have cherry picked the initial commits so that git would understand the history, but tbh ... even given a little friction I probably would have just made new commits too, I get it.

Library/src/Common/Continuations.h
MAG-mv: Agreed. There are better ways to do this, it would just me we will have to redesign our helper functions

PR #670, opened 2025-04-25
[NT-0] refac: Remove unnecesary dynamic memory from SpaceEntity

MAG-ElliotMorris:
This was done in another ticket, as at one point I thought SpaceEntity needed to be movable, but it doesn't. Nonetheless, it would be a shame to throw away this work. Value semantics are better understandable, and we were new'ing and deleting for no real reason, (which is a minor performance pessimisation generally)

This makes mutexes In SpaceEntity mutable, which they already implicitly were anyway because we were using the pointer indirection to break const.  Better this way imo where it's explicit.

> [!NOTE]  
> The change from pointer to reference for `GetScript` makes no wrapper generator changes for either TS or C#, (confirmed via diff)


Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -413,14 +410,12 @@ class CSP_API SpaceEntity
...
     uint16_t NextComponentId;
 
-    csp::multiplayer::EntityScript* Script;
-    csp::multiplayer::EntityScriptInterface* ScriptInterface;
-
-    std::mutex* EntityLock;
-    std::mutex* ComponentsLock;
-    std::mutex* PropertiesLock;
+    EntityScript Script;
+    std::unique_ptr<csp::multiplayer::EntityScriptInterface> ScriptInterface;
```
MAG-ElliotMorris, line 414: [Comment] Unique_ptr due to the need for polymorphism on this type.

MAG-OR (approved):
lgtm

MAG-AdamThorn (approved):

PR #671, opened 2025-04-25
[OF-1586] refac: CreateAvatar Continuations

MAG-ElliotMorris:
Refactor CreateAvatar to be linear.

This adds continuation capability to `SignalRConnection::Invoke`, which is the core of this ticket.

As this is an internal change and we're interested mostly in how it handles being interrupted, testing is performed on a continuation by continuation basis via mocks, mocking failures in the `SignalRConnection`.

A new continuation utility is added to unwrap types that come out of `Invoke` (and other SignalR methods I suspect). This utility introduces a `std::conditional` pattern using `if constexpr` to optionally forward or not forward the input arg to subsequent continuations.

You may be surprised by the double construct of the avatar. This _is_ unfortunate, but is explained in a large comment in the file.

> [!Warning]
> Do not review the [first commit](https://github.com/magnopus-opensource/connected-spaces-platform/commit/69d1a12072208e1e9f25bf102e71e1ff010c4065). This has been split into a seperate PR [here.](https://github.com/magnopus-opensource/connected-spaces-platform/pull/670) This PR is only slightly dependent on it, so if it gets rejected it'll be fine.

> [!Note]
> I have confirmed manually that a web build of this allows a user to enter an OKO web space and spawn in as an avatar (both creator an non-creator modes, not that it makes a difference)

### DIFF
--- a/Library/src/Multiplayer/SignalR/SignalRConnection.cpp
+++ b/Library/src/Multiplayer/SignalR/SignalRConnection.cpp
@@ -17,6 +17,7 @@
 
 #include "Debug/Logging.h"
 #include "SignalRClient.h"
+#include <memory>
 
 #if ENABLE_SIGNALR_LOGGING
 #include <iostream>
@@ -106,15 +107,22 @@ void SignalRConnection::On(const std::string& EventName, const SignalRConnection
     Connection.on(EventName, Handler);
 }
 
-void SignalRConnection::Invoke(
+async::task<std::tuple<signalr::value, std::exception_ptr>> SignalRConnection::Invoke(
     const std::string& MethodName, const signalr::value& Arguments, std::function<void(const signalr::value&, std::exception_ptr)> Callback)
 {
     CSP_PROFILE_SCOPED();
 
+    auto InvokeOnCompleteEvent = std::make_shared<async::event_task<std::tuple<signalr::value, std::exception_ptr>>>();
+    auto InvokeOnCompleteContinuation = InvokeOnCompleteEvent->get_task();
+
     std::function<void(const signalr::value&, std::exception_ptr)> InvocationCallback
-        = [Callback, this](const signalr::value& Value, std::exception_ptr ExceptionPtr)
+        = [Callback, this, CompleteContinuation = std::move(InvokeOnCompleteEvent)](const signalr::value& Value, std::exception_ptr ExceptionPtr)
     {
-        Callback(Value, ExceptionPtr);
+        if (Callback)
+        {
+            Callback(Value, ExceptionPtr);
+        }
+        CompleteContinuation->set(std::make_tuple(Value, ExceptionPtr));
 
         PendingInvocations--;
         if (PendingStopCallback && PendingInvocations == 0)
@@ -133,6 +141,7 @@ void SignalRConnection::Invoke(
 
     PendingInvocations++;
     Connection.invoke(MethodName, Arguments, InvocationCallback);
+    return InvokeOnCompleteContinuation;
 }
 
 void SignalRConnection::Send(const std::string& MethodName, const signalr::value& Arguments, std::function<void(std::exception_ptr)> Callback)
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -126,6 +126,18 @@ SpaceEntity::SpaceEntity(SpaceEntitySystem* InEntitySystem)
 {
 }
 
+SpaceEntity::SpaceEntity(SpaceEntitySystem* EntitySystem, uint64_t Id, const csp::common::String& Name,
+    const csp::multiplayer::SpaceTransform& Transform, uint64_t OwnerId, bool IsTransferable, bool IsPersistant)
+    : SpaceEntity(EntitySystem)
+{
+    this->Id = Id;
+    this->Name = Name;
+    this->Transform = Transform;
+    this->OwnerId = OwnerId;
+    this->IsTransferable = IsTransferable;
+    this->IsPersistant = IsPersistant;
+}
+
 SpaceEntity::~SpaceEntity()
 {
     auto& Keys = *Components.Keys();
@@ -389,6 +401,10 @@ void SpaceEntity::MarkForUpdate()
 #else
         EntitySystem->QueueEntityUpdate(this);
 #endif
+    }
+    else
+    {
+        CSP_LOG_MSG(csp::systems::LogLevel::Warning, "Space Entity not marked for update, no local EntitySystem found.");
     }
 }
 
--- a/Library/src/Multiplayer/SpaceEntitySystem.cpp
+++ b/Library/src/Multiplayer/SpaceEntitySystem.cpp
@@ -378,7 +378,7 @@ void SpaceEntitySystem::CreateAvatar(const csp::common::String& InName, const Sp
     async::when_all(GetAvatarNetworkIdChain, SerializeAndSendChain)
         .then(CreateNewLocalAvatar(InName, InSpaceTransform, InAvatarId, InState, InAvatarPlayMode, Callback))
         .then(csp::common::continuations::InvokeIfExceptionInChain(
-            [Callback, this](const std::exception& Except)
+            [Callback](const std::exception& Except)
             {
                 CSP_LOG_FORMAT(csp::systems::LogLevel::Error, "Failed to create Avatar. Exception: %s", Except.what());
                 Callback(nullptr);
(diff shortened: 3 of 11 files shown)


Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -388,6 +400,10 @@ void SpaceEntity::MarkForUpdate()
         EntitySystem->QueueEntityUpdate(this);
 #endif
     }
+    else
+    {
+        CSP_LOG_MSG(csp::systems::LogLevel::Warning, "Space Entity not marked for update, no local EntitySystem found.");
```
MAG-ElliotMorris, line 405: [Comment] Took me a while to track down that in the first iteration, I hadn't injected the EntitySystem, so adding a log for that case as it doesn't seem like it's ever a good thing.

.githooks/post-checkout
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks').\n"; exit 2; }
+git lfs post-checkout "$@"
```
MAG-ElliotMorris, line 3: [fix] Remove all githook changes

MAG-ElliotMorris: Force pushed to do this as no one had reviewed yet.

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
```diff
@@ -0,0 +1,314 @@
...
+        = { csp::common::Vector3 { 1.452322f, 2.34f, 3.45f }, csp::common::Vector4 { 4.1f, 5.1f, 6.1f, 7.1f }, csp::common::Vector3 { 1, 1, 1 } };
+
+    async::spawn(async::inline_scheduler(), []() { return uint64_t(55); }) // This continuation takes the ID as its input
+        .then(async::inline_scheduler(),
+            SpaceEntitySystem->SendNewAvatarObjectMessage("Username", UserTransform, "AvatarId", AvatarState::Idle, AvatarPlayMode::Default))
+        .then(async::inline_scheduler(),
+            [](std::tuple<const signalr::value&, std::exception_ptr> Results)
+            {
+                auto [Result, Exception] = Results;
+                EXPECT_EQ(Result.as_bool(), true);
```
MAG-AdamThorn, line 206: [Question] Are we not expecting an error in this test?

MAG-AdamThorn (approved):

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
MAG-ElliotMorris: Hey, great catch! This was bad copy-paste code from a success case.

_However_, it's interesting that this still passed. What's happening here is that the `as_bool` is throwing, since the type of the nonsense value populated is a string. What async++ does with exceptions is consume them and pack them into the task objects. The takeaway here is that if you don't have an exception handler at the end of your continuation chains, it's possible they could just appear to be working if they throw.

I've added a handler to all of the functions I've added here for safety, although in the future i'll think twice about using chains for assertions in tests, since they can just be skipped. (To be fair, we have similar issues where we check stuff in raw callbacks)

https://github.com/magnopus-opensource/connected-spaces-platform/pull/671/commits/5809351b68dcb0f7197d37879003349a57fc9a27

MAG-AdamThorn (approved):

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    // SignalR populates a result and not an exception
+    EXPECT_CALL(*SignalRMock, Invoke("GenerateObjectIds", ::testing::_, ::testing::_))
+        .WillOnce(
+            [](const std::string& /**/, const signalr::value& /**/,
+                std::function<void(const signalr::value&, std::exception_ptr)> /**/) -> async::task<std::tuple<signalr::value, std::exception_ptr>>
+            {
+                // For some reason the ID value has to be an array :/
+                std::vector<signalr::value> ids;
+                ids.emplace_back(signalr::value(uint64_t(55)));
```
MAG-mv, line 76: [Question] Where is the 55 coming from?

Library/src/Multiplayer/SpaceEntity.cpp
MAG-ElliotMorris: I didn't realize that.

The thing that happened here is that when I wrote my new method to construct the SpaceEntity, I just forgot to pass the EntitySystem, so it was silently not queuing the entity update.

My intent here was to make that programming error less easy to make in the future, but you're saying there's a case where _should_ just silently do nothing? _Bit_ of a smell imo, but I'll revert this log if that's the case.

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -246,88 +250,139 @@ void SpaceEntitySystem::Shutdown()
...
+         * pretty much all ownership consideration from clients.
+         * To do this, we would want to create our own CSP pointer type that is backed by a shared_ptr, so we can provide that to
+         * clients via the callback, and have shared ownership of this sort of thing across the DLL/SO boundary.
+         *
+         * Note also however, that we don't double fetch the network ID, which is the main cost of constructing these things anyhow.
+         * (Stricter interface segregation for our serializers would also have solved this problem, but only in the local sense)
+         */
+        std::unique_ptr<csp::multiplayer::SpaceEntity, SpaceEntityDeleter> NewAvatar
+            = BuildNewAvatar(*csp::systems::SystemsManager::Get().GetUserSystem(), *this, NetworkId, Name, Transform,
+                MultiplayerConnectionInst->GetClientId(), false, false, AvatarId, AvatarState, AvatarPlayMode);
```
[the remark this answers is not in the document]
MAG-ElliotMorris: 👍

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -246,88 +250,139 @@ void SpaceEntitySystem::Shutdown()
...
+
+        std::scoped_lock EntitiesLocker(*EntitiesLock);
+        // Release to vague ownership. Entities is the "true" owner I _think_
```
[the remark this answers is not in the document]
MAG-ElliotMorris: I've expanded the comment to write exactly that. : https://github.com/magnopus-opensource/connected-spaces-platform/pull/671/commits/7dfad40f5cebd68973d91db4ab6e0d22a9000e99

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
MAG-ElliotMorris: It's just an arbitrary number, irrelevant.

Library/src/Multiplayer/SignalR/SignalRConnection.cpp
```diff
@@ -106,15 +107,22 @@ void SignalRConnection::On(const std::string& EventName, const SignalRConnection
     Connection.on(EventName, Handler);
 }
 
-void SignalRConnection::Invoke(
+async::task<std::tuple<signalr::value, std::exception_ptr>> SignalRConnection::Invoke(
```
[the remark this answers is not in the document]
MAG-ElliotMorris: 🤔 Yeah, I get where you're coming from, but I think it must be like this for the moment for compatibility. Changing this before all uses of `Invoke` are converted to continuations would create a lot of shims, and put a lot of code pathways under change, and thus risk, that arn't really related to this PR.

It's similar to what we do in `CreateHandler`, remember how we pass the empty lambda sometimes?

What do you think?

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: I believe its currently used for tested, so we can change a space entity when its not attached to a system

Library/src/Multiplayer/SignalR/SignalRConnection.cpp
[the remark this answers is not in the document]
MAG-mv: All good

Library/src/Multiplayer/SpaceEntitySystem.cpp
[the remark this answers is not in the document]
MAG-mv: Nice!

MAG-mv (approved):

PR #673, opened 2025-04-28
[NT-0] test: API key now required for requests

MAG-AdamThorn:
Added missing API key to the header for the failing requests.

Payload.AddHeader(CSP_TEXT("x-api-key"), CSP_TEXT("reqres-free-v1"));

We use an external service (https://reqres.in/) for testing against a real API, and there requirements seem to have changed.


MAG-ElliotMorris:
[Question] Do you have the reason why this is suddenly required now? Some MCS change?

MAG-ElliotMorris (approved):

MAG-mv (approved):

PR #674, opened 2025-04-29
[OF-1624] fix: Add missing auth header to web client test

MAG-ElliotMorris:
We literally just missed this due to a case of collective blindness.

EDIT: I've also added api keys to all other methods that use this service, as the get query seemed to not be consistent on whether it needed it or not, which may happen for other queries down the line.


MAG-OR (approved):

MAG-AdamThorn (approved):

MAG-SamBirley:
Looking good 🎉

image

PR #676, opened 2025-04-29
[NT-0] fix: Seed true randomness in username/email generation

MAG-ElliotMorris:
We were getting some [name collision errors from MCS](https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_FunctionalTests/287383?showLog=287383_3581_103.3634.3638&logFilter=debug&logView=flowAware) in some of our tests. It may or may not be because of us not setting an actual random seed here, but it's a good idea to do anyway, and low risk.


MAG-ElliotMorris:
It does this by default anyway...

PR #677, opened 2025-04-30
[OW-2017] refac: Add emissiveIntensity support to GLTFMaterial

MAG-DanielMurdolo:
With the addition of Alpha Video Material support, it was identified that having the emissive intensity property on a GLTF Material is also desired.

This change adds the property for use in the clients.


Library/include/CSP/Systems/Assets/GLTFMaterial.h
```diff
@@ -106,6 +106,14 @@ class CSP_API GLTFMaterial : public Material
     /// @return csp::common::Vector3& : The emissive factor.
     const csp::common::Vector3& GetEmissiveFactor() const;
 
+    /// @brief Set the emissive intensity.
+    /// @param Intensity float : The emissive intensity.
+    void SetEmissiveIntensity(float Intensity);
```
MAG-AdrianMeredith, line 111: We should call it emissive strength
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md

MAG-AdrianMeredith: (we can still calling emissive intensity in the ui if we want it to be like the other shader)

MAG-AdrianMeredith: quite possibly but only if its done now

MAG-DanielMurdolo: Considering we've just released the alpha video materials with emissiveIntensity, if we change it in AlphaVideoMaterial.h/.cpp now I assume it'd be a breaking change?

MAG-DanielMurdolo:
> We should up the version number on the shader for completeness

This?

constexpr int InitialMaterialVersion = 1;

Library/include/CSP/Systems/Assets/GLTFMaterial.h
MAG-DanielMurdolo: I've renamed it in GLTFMaterial

MAG-ElliotMorris (changes requested):
I wonder if you might thread this property through `MaterialUnitTests` before merging? The pattern is pretty simple.

https://github.com/magnopus-opensource/connected-spaces-platform/blob/269e33e270576f2fb3b5223fff46044d5ff89221/Tests/src/InternalTests/MaterialUnitTests.cpp#L44

Library/src/Systems/Assets/GLTFMaterial.cpp
```diff
@@ -145,6 +150,10 @@ void GLTFMaterial::SetEmissiveFactor(const csp::common::Vector3& Factor) { Emiss
 
 const csp::common::Vector3& GLTFMaterial::GetEmissiveFactor() const { return EmissiveFactor; }
 
+void GLTFMaterial::SetEmissiveStrength(float Strength) { EmissiveStrength = Strength; }
+
+float GLTFMaterial::GetEmissiveStrength() const { return EmissiveStrength; }
+
 void GLTFMaterial::SetBaseColorTexture(const TextureInfo& Texture) { BaseColorTexture = Texture; }
 
 const TextureInfo& GLTFMaterial::GetBaseColorTexture() const { return BaseColorTexture; }
```
MAG-mv, line 159: [Fix] Can we update the following tests for these calls:

- GLTFMaterialConstructorTest
- GLTFMaterialSetterTest
- GLTFMaterialJsonSerializationTest

MAG-mv (changes requested):

Library/src/Systems/Assets/GLTFMaterial.cpp
MAG-DanielMurdolo: Done

MAG-DanielMurdolo:
> I wonder if you might thread this property through `MaterialUnitTests` before merging? The pattern is pretty simple.
> 
> https://github.com/magnopus-opensource/connected-spaces-platform/blob/269e33e270576f2fb3b5223fff46044d5ff89221/Tests/src/InternalTests/MaterialUnitTests.cpp#L44
> 
> Other than that LGTM though. Thanks for doing this.

Done

Library/include/CSP/Systems/Assets/GLTFMaterial.h
MAG-DanielMurdolo: @MAG-AdrianMeredith I figured we'd leave it in AlphaVideoMaterial now to avoid breaking things

MAG-AdamThorn (approved):

MAG-mv (approved):

PR #680, opened 2025-05-02
Research/remove allocators

MAG-ElliotMorris:
Test PR please ignore


MAG-ElliotMorris:
I've used the wrong issue number for all the commits :( will make sure to change them over at point of squash.

MAG-ElliotMorris:
image
🤩

Library/include/CSP/Common/List.h
```diff
@@ -267,12 +267,23 @@ template <typename T> class CSP_API List
...
-        ObjectPtr->~T();
+        if constexpr (std::is_pointer<T>::value)
+        {
+            // I dont think the list owns any pointers that are in it. That seems weird, because it _will_ destruct value types
+            // To maintain exactly the same behaviour, do the destruct behaviour for pointers. Use regular RAII cleanup for value types below during
+            // the shift.
 
+            // delete ObjectArray[Index]; <-- What we should be doing ... (although actually, we should just have this type backed by std::list)
+            T* ObjectPtr = &ObjectArray[Index];
+            ObjectPtr->~T();
```
MAG-ElliotMorris, line 278: [Consider] Write a test to demonstrate this behaviour.

MAG-ElliotMorris:
We have decided to do a feature publish (once the CI is fixed) and run an special case manual test across our platforms with this one.

This is out of an abundance of caution, not because we suspect there are any true issues with this change, but it _is_ so fundamental I can't disagree with the approach.

Library/include/CSP/Common/List.h
```diff
@@ -267,12 +267,23 @@ template <typename T> class CSP_API List
...
+        }
+
+        // Shift everything left
+        for (size_t i = Index; i < CurrentSize - 1; ++i)
+        {
+            ObjectArray[i] = ObjectArray[i + 1];
+        }
         --CurrentSize;
-        auto Remaining = CurrentSize - Index;
-        std::memmove(ObjectArray + Index, ObjectArray + (Index + 1), sizeof(T) * Remaining);
```
MAG-mv, line 275: [Question] Why cant we use memmove anymore?

Library/src/Common/StlString.h
```diff
@@ -15,13 +15,11 @@
...
 
-#include "Memory/StlAllocator.h"
-
 #include <string>
 
 namespace csp
 {
 
-using StlString = std::basic_string<char, std::char_traits<char>, csp::memory::StlAllocator<char>>;
+using StlString = std::basic_string<char, std::char_traits<char>>;
```
MAG-mv, line 23: [Question] isnt this just std::String now?

Library/include/CSP/Common/Map.h
```diff
@@ -37,34 +36,27 @@ template <typename TKey, typename TValue> class CSP_API Map
...
+        Container = new MapType(*Other.Container); // Copy construct
     }
 
     /// @brief Move constructor.
     /// @param Other Map<TKey, TValue>&&
     CSP_NO_EXPORT Map(Map<TKey, TValue>&& Other)
     {
-        Container = (MapType*)csp::memory::DllAlloc(sizeof(MapType));
-        new (Container) MapType(*Other.Container);
+        Container = new MapType(std::move(*Other.Container)); // Copy construct
```
MAG-AdamThorn, line 52: [fix] Comment here should say // Copy move construct

Library/include/CSP/Common/Map.h
```diff
@@ -106,45 +94,20 @@ template <typename TKey, typename TValue> class CSP_API Map
...
 
+        delete Container;
+        Container = new MapType(*Other.Container); // Copy construct
         return *this;
     }
 
-    /// @brief Move assignment.
-    /// @param Other Map<TKey, TValue>&&
-    /// @return Map<TKey, TValue>&
     CSP_NO_EXPORT Map<TKey, TValue>& operator=(Map<TKey, TValue>&& Other)
```
MAG-AdamThorn, line 104: [Question] Why are you removing the comments for this method?

MAG-AdamThorn (changes requested):

Library/include/CSP/Common/List.h
MAG-ElliotMorris: It's because it bypasses destruction behaviour. I was trying to be _super_ clear about what was happening in this method, because the manual destructor call is surprising. This way, we document that this list has a bug where pointer memory is destructed but not deallocated, but the value type path is exactly what you'd expect ,destructors called when item goes out of scope.

Library/src/Common/StlString.h
MAG-ElliotMorris: Yeah it is, I didn't want to remove this concept because of risk of public api change, and I still don't. I could just change to `using StlString = std::string`, although I don't see much of a difference.

Probably _should_ log something to actually remove this entirely once this goes in though, much less dangerous in isolation, pretty nice small task.

Library/include/CSP/Common/Map.h
MAG-ElliotMorris: Just removed the comments as they're pointless anyhow. https://github.com/magnopus-opensource/connected-spaces-platform/pull/680/commits/f1d8ba79ffddcc2a3c0b2c1fbde628539ee4c1d0

Library/include/CSP/Common/List.h
MAG-ElliotMorris: Considered, answer : _Naaaaaa_

Library/src/Common/StlString.h
MAG-ElliotMorris: Logged : https://magnopus.atlassian.net/browse/OF-1667

MAG-ElliotMorris:
Not sure why the live-tests are waiting for status, there's a passing build here, maybe just a consequence of this happening during a changeover. https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_FunctionalTests/289153?logView=linear

Wanted to wait for live service tests to pass for this, but they do, as do our new primary local tests, so merging.

PR #682, opened 2025-05-08
[OF-1510] Replace old serialization with new method

MAG-mv:
- Final PR for replacing old entity serialization with the new method. 
- Commit history looks weird, as I branched off of a branch that has now been merged into main
- Added MCSComponentPacker/Unpaker to automatically handle type conversions


Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -0,0 +1,231 @@
...
+    const auto& String = std::get<std::string>(ComponentData.GetValue());
+    Value = String.c_str();
+}
+
+void MCSComponentUnpacker::ReadValue(const mcs::ItemComponentData& ComponentData, ReplicatedValue& Value)
+{
+    std::visit([&Value](const auto& ValueType) { ReplicatedValueFromType(ValueType, Value); }, ComponentData.GetValue());
+}
+
+void MCSComponentUnpacker::ReplicatedValueFromType(const std::vector<float>& Type, ReplicatedValue& Value)
```
MAG-OR, line 65: [comment] should the function name start with a verb? (get? read?)

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+        Value = ReplicatedValue { csp::common::Vector4 { Type[0], Type[1], Type[2], Type[3] } };
+    }
+    else
+    {
+        throw std::runtime_error("Unsupported float array.");
+    }
+}
+
+void MCSComponentUnpacker::ReplicatedValueFromType(uint64_t, ReplicatedValue&) { throw std::runtime_error("Unsupported"); }
```
MAG-OR, line 85: [question] likely a silly question -- why is this unsupported?

Library/src/Multiplayer/MCSComponentPacker.h
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    ReadValue(Key, Value);
+    return true;
+}
+
+template <typename T> inline void MCSComponentUnpacker::ReadValue(uint16_t Key, T& Value)
+{
+    const mcs::ItemComponentData& ComponentData = Components[Key];
+    ReadValue(ComponentData, Value);
```
MAG-OR, line 119: [question] is this function really that useful, when we have TryReadValue? you could use the first one all the time? (is it significantly longer/using more resources?)

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1440,88 +1448,236 @@ mcs::ObjectPatch SpaceEntity::CreateObjectPatch()
...
+    for (size_t i = 0; i < TransientDeletionComponentIds.Size(); ++i)
+    {
+        DeletionComponent.Id = TransientDeletionComponentIds[i];
+        ComponentPacker.WriteValue(DeletionComponent.Id, &DeletionComponent);
+    }
 
-    // Manually write the component type, as this isnt stored in the component properties.
-    // This is currently the ONLY value that uses a uint64 types as a key for some reason. The rest use int64.
-    ComponentProperties[COMPONENT_KEY_COMPONENTTYPE] = { static_cast<uint64_t>(Component->GetComponentType()) };
+    // 4. Create the object patch using the reqired properties and our created components.
```
MAG-OR, line 1460: [nit] required

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1487,90 +1494,60 @@ void SpaceEntitySystem::CreateObjectInternal(const csp::common::String& InName,
...
             {
-                if (Entities[i]->GetId() == EntityID)
-                {
-                    EntityFound = true;
-                    Entities[i]->ShouldUpdateParent = ShouldUpdateParent;
-                    Entities[i]->ParentId = ParentId;
-                    Entities[i]->DeserialiseFromPatch(Deserialiser);
-                    Entities[i]->OwnerId = OwnerID;
-                }
+                Entities[i]->FromObjectPatch(Patch);
```
MAG-OR, line 1542: [comment] there will be merge conflicts between this PR and mine. I expect them to be relatively minor.

MAG-OR (approved):

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1373,147 +967,314 @@ void SpaceEntity::ResolveParentChildRelationship()
...
         {
-            auto* Component = DirtyComponents[(*Keys)[i]].Component;
-            ComponentData[(*Keys)[i]] = CreateItemComponentData(Component);
+            auto* Component = DirtyComponents[Key].Component;
+            ComponentPacker.WriteValue(Key, Component);
         }
         else
         {
-            assert(DirtyComponents[(*Keys)[i]].Component != nullptr && "DirtyComponent given a null component!");
+            assert(DirtyComponents[Key].Component != nullptr && "DirtyComponent given a null component!");
```
MAG-AdamThorn, line 995: [fix] This assert check is the wrong way round. It should be asserting if it IS a nullptr.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1373,147 +967,314 @@ void SpaceEntity::ResolveParentChildRelationship()
...
             {
-                auto* Component = DirtyComponents[(*Keys)[i]].Component;
-                ComponentData[(*Keys)[i]] = CreateItemComponentData(Component);
+                auto* Component = DirtyComponents[Key].Component;
+                ComponentPacker.WriteValue(Key, Component);
             }
             else
             {
-                assert(DirtyComponents[(*Keys)[i]].Component != nullptr && "DirtyComponent given a null component!");
+                assert(DirtyComponents[Key].Component != nullptr && "DirtyComponent given a null component!");
```
MAG-AdamThorn, line 1035: [fix] Same as above, it should be checking if it is == nullptr

MAG-AdamThorn (changes requested):

modules/mimalloc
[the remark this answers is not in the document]
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/682/commits/904e13d479754120b37af0591c5b20f2a9c0a215

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+namespace csp::multiplayer
+{
+
+MCSComponentUnpacker::MCSComponentUnpacker(const std::map<uint16_t, mcs::ItemComponentData>& Components)
+    : Components { Components }
+{
+}
+
+uint64_t MCSComponentUnpacker::GetRealComponentsCount() const
```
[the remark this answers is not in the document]
MAG-mv: MCS components are used to store CSP view components and CSP runtime components. In this context. "real" means CSP runtime components. I copied this name from elsewhere but maybe RuntimeComponents is better?

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -0,0 +1,230 @@
...
+        if (Key < COMPONENT_KEY_END_COMPONENTS)
+        {
+            ++ComponentCount;
+        }
+    }
+
+    return ComponentCount;
+}
+
+void MCSComponentUnpacker::ReadValue(const mcs::ItemComponentData& ComponentData, uint64_t& Value)
```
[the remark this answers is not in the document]
MAG-mv: This is a pattern we follow with the SignalR serializer and the json serializer, so I think its good imo. And yes, this is an internal class.

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1476,90 +1481,60 @@ void SpaceEntitySystem::CreateObjectInternal(const csp::common::String& InName,
...
     {
-        const uint64_t EntityID = Deserialiser.ReadUInt64();
-        const uint64_t OwnerID = Deserialiser.ReadUInt64();
-        const bool Destroy = Deserialiser.ReadBool();
-        bool ShouldUpdateParent = false;
-        csp::common::Optional<uint64_t> ParentId = nullptr;
-
-        if (Deserialiser.NextValueIsArray())
+        // This is an entity deletion.
+        // Deletion
```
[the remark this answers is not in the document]
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/682/commits/30be4dc1c0de92cb92fbe8e422fdfa097e78fcbd

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: Thats correct. Asserts fire is the boolean is false. In this case, if DirtyComponents[Key].Component != nullptr (a valid value), that will be true, otherwise false

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: Same as above

Library/src/Multiplayer/MCSComponentPacker.cpp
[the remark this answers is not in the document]
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/682/commits/2d0c4762e9d60a8adf65aba823dc7111449f5b67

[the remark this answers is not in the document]
MAG-AdamThorn: I prefer GetRuntimeComponentsCount over GetRealRuntimeComponentsCount, but happy with which ever you choose.

Library/src/Multiplayer/SpaceEntity.cpp
MAG-AdamThorn: Gotcha but is an assert required here. If it is not nullptr it will go into the 'if' block, otherwise it will go into 'else'. We are asserting something we have already checked.

MAG-mv: Good point. I have removed from the else block: https://github.com/magnopus-opensource/connected-spaces-platform/pull/682/commits/55026bd8f18f8eb06f6c15868d66657f32ec2e20

MAG-AdamThorn (approved):

PR #685, opened 2025-05-13
[OF-1608] Disable the tests that are problematic on local CHS

MAG-OR:
there are two parts to this PR. it is debatable whether the second part should stay in this PR (if it doesn't, it should go in the ticket that will fix the conversation tests). the obvious argument in favour of keeping this change in is that local CHS is ready to use in PR checks.

Part 1: disable the problematic tests. there are 8 of them. More info about the failures in the story. Will also add dedicated new stories, see links in the story.

Part 2: fix crash in local CHS (a conversation test was calling the hard-coded live endpoint, leading to a crash). this is the part that arguably should be moved to the conversation story.

### DIFF
--- a/Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
+++ b/Tests/src/PublicAPITests/ComponentTests/ConversationComponentTests.cpp
@@ -1224,6 +1224,7 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationTests, ConversationComponentSecondClientE
               .SetSpaceId(Space.Id.c_str())
               .SetLoginEmail(TestUser.Email.c_str())
               .SetPassword(GeneratedTestAccountPassword)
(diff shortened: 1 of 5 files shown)


MAG-ChristopherAtkinson (approved):
The changes look good, once we move over to full local chs testing on CI we will be validating new test coming in from as part of the PR process but this was a good catch.

MAG-OR:
image
Local CHS run on my local machine showing that all the enabled tests pass, and listing the newly disabled tests. In that run, the transient tests happened to both pass.

https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_TestsLocalChsInstance/288570?buildTab=overview
image
Successful local CHS run on the CI. Note that the CI does not run disabled tests (as expected) and does not list them either, but note that the number of disabled tests matches.

MAG-ElliotMorris (approved):

PR #686, opened 2025-05-15
[OF-1633] chore: remove friends from SpaceEntity et al.

MAG-OR:
Remove friends from SpaceEntity, SpaceEntitySystem, EventBus and MultiplayerConnection.

Create new getters and setters where private members were used directly, and make necessary private methods public (using CSP_NO_EXPORT, to avoid unnecessary exports).

Add docstrings on every new or changed function.

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -114,6 +114,8 @@ class CSP_API SpaceEntity
 {
     CSP_START_IGNORE
     /** @cond DO_NOT_DOCUMENT */
+    friend class SpaceEntitySystem;
+
 #ifdef CSP_TESTS
     friend class ::CSPEngine_MultiplayerTests_LockPrerequisitesTest_Test;
 #endif
@@ -408,10 +410,6 @@ public:
     /// @return SpaceEntity*
     CSP_NO_EXPORT SpaceEntity* GetParent();
 
-    /// @brief Setter for the parent entity
-    /// @param InParent SpaceEntity : the parent entity to set
-    CSP_NO_EXPORT void SetParent(SpaceEntity* InParent);
-
     /// @brief Getter for the parent id
     /// @return csp::common::Optional<uint64_t>
     CSP_NO_EXPORT csp::common::Optional<uint64_t> GetParentId();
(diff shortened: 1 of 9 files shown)


Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -52,7 +52,6 @@ CSP_END_IGNORE
 
 namespace csp::multiplayer
 {
-class SpaceEntitySystem;
 class EntityScriptInterface;
```
MAG-OR, line 55: [comment] this one is needed for the wasm build

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -410,6 +403,111 @@ class CSP_API SpaceEntity
...
+    /// @return csp::common::List<uint16_t>
+    CSP_NO_EXPORT csp::common::List<uint16_t> GetTransientDeletionComponentIds();
+
+    /// @brief Getter for the ShouldUpdateParent boolean
+    /// @return bool
+    CSP_NO_EXPORT bool GetShouldUpdateParent();
+
+    /// @brief Setter for the ShouldUpdateParent boolean
+    /// @param Boolean bool : the boolean value to set ShouldUpdateParent to
+    CSP_NO_EXPORT void SetShouldUpdateParent(const bool Boolean);
```
MAG-mv, line 447: [Question] I am a bit concerned that we are exposing this underlying functionality, even if it is only to the Unreal client, and any other clients using raw C++. Would it be detrimental to this ticket to at least temporarily keep the friendship between SpaceEntity <-> SpaceEntitySystem? @MAG-ElliotMorris

MAG-ElliotMorris: I see your point, and yes, this specific friendship could _probably_ continue to exist, although it will make any necessary refactoring trickier.

Why is this one in particular worse than other exposures? There's plenty of CSP_NO_EXPORT's around that would be catastrophic if unreal decided to actually call them.

MAG-mv: I think the SpaceEntity just caught my eye, as I have worked on a lot of these private functions, so I can see how detrimental making these public could be.

MAG-ElliotMorris (approved):
👍 I know Matt has concerns on this exposure which we should discuss, but from my side this is exactly what I wanted.

As a side note, I'm quite pleased with how this brings the SpaceEntity refactor into stark focus. Architecturally the goal can now be to reduce this public interface down to something that can actually be used correctly and safely, when before that was completely hidden and unknown.

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: The issue is we are open-source, so we don't know who else is using our library. If this were just an issue with Unreal, I wouldnt have much of an issue.

MAG-ElliotMorris: Capturing offline discussion, Matt is going to make a list of the methods that are particularly dangerous to expose, and for-sure don't need to go across the `IRealtimeEngine` interface, and we'll revert back to hiding behind friendship for just those.

We're also going to try and push the priority of this symbol visibility issue higher, since it's a real problem that we've been hiding behind friendship for too long, and there are options, even quick ones, to deal with it.

MAG-mv: After our discussion, I believe we should friend the following SpaceEntity functions temporarily:

SpaceEntity:

void SetParent(SpaceEntity* InParent)
void SetTimeOfLastPatch(std::chrono::milliseconds NewTime)
void SetOwnerId(const uint64_t InOwnerId)
void RemoveChildEntities()
void RemoveParentId()
void MarkForUpdate()
void ClaimScriptOwnership()
bool InternalSetSelectionStateOfEntity(const bool SelectedState, uint64_t ClientID)

There are other issues in SpaceEntitySystem. However, I don't think we can address these yet, as they are used by other systems, which will be in different libraries.

I will create another ticket to try and address these issues for the future.

MAG-OR: We agreed to leave MarkForUpdate and ClaimScriptOwnership as public because they are used elsewhere. The rest is done in ab148231abf7e3938dedc1a65e5cb7b5a1671d57

MAG-mv (approved):

PR #687, opened 2025-05-16
[NT-0] fix: removed redundant removal of thumbnail and metadata

MAG-ChristopherAtkinson:
On the CI we currently have two flaky test introduced by 021dffc5bbae6e298a5aac8a77197160a81c9b22

On review of the change it seems that behaviour has changed from a linear deletion process to non blocking async workflow. This resulted in a race condition where we are deleting the same resources, in the event these are executed in the wrong order would result in the tests crashing. 

The DELETE /api/v1/spaces/{SpaceId} endpoint will delete the thumbnail and metadata, which make the previous calls redundant, and removal prevent the race condition as only a single action will attempt to delete a resource. 

```
DELETE /api/v1/spaces/{SpaceId}
Locates the Space by its unique groupId represented here as spaceId and the associated objects that belong to it
and deletes it from the data store. This Function will search for items related to a spaceId even if the spaceId itself has already been deleted.
```

### DIFF
--- a/Library/include/CSP/Systems/Spaces/SpaceSystem.h
+++ b/Library/include/CSP/Systems/Spaces/SpaceSystem.h
@@ -151,7 +151,7 @@ public:
         const csp::common::Optional<csp::common::String>& Description, const csp::common::Optional<SpaceAttributes>& Type,
         const csp::common::Optional<csp::common::Array<csp::common::String>>& Tags, BasicSpaceResultCallback Callback);
 
-    /// @brief Deletes a given space and the corresponding UserService group.
+    /// @brief Deletes a given space and the associated objects that belong to it. Including UserService group, Metadata, and Thumbnail.
     /// @param Space Space : space to delete
     /// @param Callback NullResultCallback : callback when asynchronous task finishes
     CSP_ASYNC_RESULT void DeleteSpace(const csp::common::String& SpaceId, NullResultCallback Callback);
--- a/Library/src/Systems/Spaces/SpaceSystem.cpp
+++ b/Library/src/Systems/Spaces/SpaceSystem.cpp
@@ -531,14 +531,10 @@ void SpaceSystem::CreateSpace(const String& Name, const String& Description, Spa
         .then(csp::common::continuations::InvokeIfExceptionInChain(
             [this, CurrentSpaceResult, Callback](const std::exception& /*Except*/)
             {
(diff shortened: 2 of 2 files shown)


MAG-OR (approved):
thanks for jumping on this

MAG-ElliotMorris:
[Question]
> On review of the change it seems that behaviour has changed from a linear deletion process to one that can spawn detached threads.

Can you point out exactly where? I wasn't aware we were detaching any threads as part of these refactors, the `inline_scheduler` async runner shouldn't detach or use a threadpool.

MAG-ChristopherAtkinson:
> [Question]
> 
> > On review of the change it seems that behaviour has changed from a linear deletion process to one that can spawn detached threads.
> 
> Can you point out exactly where? I wasn't aware we were detaching any threads as part of these refactors, the `inline_scheduler` async runner shouldn't detach or use a threadpool. Important to know this if it is happening as it may invalidate some of our assumptions.

My mistake I assumed due to the race condition that the `ApiResponseHandler` was threaded, but checked and seems it single threading but non blocking, which would still be same issue with the race condition. 

In this case we attempt to delete the thumbnail resource both directly and through the delete space endpoint, in the event CHS resolves delete space first, then remove thumbnail will assume assets are available in the asset collection. 

Will update the description to remove the incorrect assumption.

MAG-AdamThorn:
[Consider] This strikes me as a good opportunity to update the comments for the SpaceSystem::DeleteSpace method to make it clear that it cleans up associated objects (metadata and thumbnail) to avoid mistakes in the future.

MAG-ChristopherAtkinson:
> [Consider] This strikes me as a good opportunity to update the comments for the SpaceSystem::DeleteSpace method to make it clear that it cleans up associated objects (metadata and thumbnail) to avoid mistakes in the future.

Great suggestion, have added some addition information to the comment: [9d445de](https://github.com/magnopus-opensource/connected-spaces-platform/pull/687/commits/9d445de18bafd8b8f7f4a377d73b39f108e1cc4d)

jeromagnopus:
> @raynopus added to the PR to confirm that `DELETE /api/v1/spaces/{SpaceId}` endpoint deletes both thumbnail / metadata and any other associated asset as the description in the swagger documentation states.

Space Deletion from a CHS POV deletes all prototypes associated to the SpaceId.

PR #689, opened 2025-05-21
[NT-0] fix: fix MaterialEventTest and MaterialAssetEventTest 

MAG-OR:
Fix MaterialEventTest and MaterialAssetEventTest by shamelessly reusing the fix for https://github.com/magnopus-opensource/connected-spaces-platform/pull/688.

Note that this PR has been re-purposed, as when I opened it I only had a fix for the crash in MaterialEventTest (not for the test failure), but now I have a real fix, which happens to also fix a second test, MaterialAssetEventTest.

Therefore, if I haven't squashed yet, please ignore the first commit, as it has the crash fix and not the fix fix.

Reasons why I have reasonable grounds to think that the transient failure is fixed:
- have 1000 out of 1000 successful runs of each of the 2 aforementioned tests, on local CHS, on my local machine. Note that before the fix (this morning), it wouldn't go past 5 or 6 runs at most before encountering an error.
image
- have 1 successful local CHS CI job that was green the first time around: https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_TestsLocalChsInstance/289172
- have 1 successful live CHS CI job that was green the first time around: https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_FunctionalTests/289185

### DIFF
--- a/Tests/src/PublicAPITests/MaterialTests.cpp
+++ b/Tests/src/PublicAPITests/MaterialTests.cpp
@@ -727,24 +727,15 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, MaterialEventTest)
     {
         bool CallbackCalled = false;
 
-        auto CB = [&CallbackCalled, &CreatedGLTFMaterial](const csp::multiplayer::MaterialChangedParams& Params)
-        {
-            const auto Start = std::chrono::steady_clock::now();
-            constexpr std::chrono::seconds Timeout = std::chrono::seconds(10);
-            while (!CreatedGLTFMaterial)
-            {
-                if ((std::chrono::steady_clock::now() - Start) >= Timeout)
-                {
-                    // Timeout reached
-                    FAIL() << "Busy wait timeout reached waiting for Material creation.";
-                }
-            }
+        csp::common::String MaterialCollectionId;
+        csp::common::String MaterialId;
 
-            EXPECT_EQ(Params.MaterialCollectionId, CreatedGLTFMaterial->GetMaterialCollectionId());
-            EXPECT_EQ(Params.MaterialId, CreatedGLTFMaterial->GetMaterialId());
+        auto CB = [&CallbackCalled, &MaterialCollectionId, &MaterialId](const csp::multiplayer::MaterialChangedParams& Params)
+        {
+            MaterialCollectionId = Params.MaterialCollectionId;
+            MaterialId = Params.MaterialId;
 
             EXPECT_EQ(Params.ChangeType, csp::multiplayer::EAssetChangeType::Created);
-
             CallbackCalled = true;
         };
 
@@ -756,8 +747,11 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, MaterialEventTest)
         CreatedGLTFMaterial = static_cast<GLTFMaterial*>(CreatedMaterial);
 
         WaitForCallback(CallbackCalled);
-
         EXPECT_TRUE(CallbackCalled);
+
+        // Do the check here where we know it exists
+        EXPECT_EQ(MaterialCollectionId, CreatedGLTFMaterial->GetMaterialCollectionId());
+        EXPECT_EQ(MaterialId, CreatedGLTFMaterial->GetMaterialId());
     }
 
     // Update material and listen for event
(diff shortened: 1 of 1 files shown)


Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -730,12 +730,15 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, MaterialEventTest)
 
         auto CB = [&CallbackCalled, &CreatedGLTFMaterial](const csp::multiplayer::MaterialChangedParams& Params)
         {
-            EXPECT_EQ(Params.MaterialCollectionId, CreatedGLTFMaterial->GetMaterialCollectionId());
-            EXPECT_EQ(Params.MaterialId, CreatedGLTFMaterial->GetMaterialId());
+            if (CreatedGLTFMaterial)
```
MAG-mv, line 733: [Question] The are multiple callbacks in this test that are checking properties on the CreatedGLTFMaterial. Why does putting an if statement around this one test prevent the crash?

MAG-OR: the crash was caused by CreatedGLTFMaterial->GetMaterialCollectionId() when CreatedGLTFMaterial was null

MAG-mv: There are multiple callbacks in the test calling CreatedGLTFMaterial->GetMaterialCollectionId(). Does this mean CreatedGLTFMaterial is only null in one of these callbacks?

MAG-OR: I'm unsure. I guess this is the first callback that gets called so it catches the first error/crash. 
It may that by the time the other callbacks are called, the material consistently exists, but this first callback sometimes expects it too early. 

I am still running loads or re-runs to see if any of the later callbacks ever crash (so far it does not look like it)

MAG-OR: yeah I think there is a race condition there. 
you create CreatedGLTFMaterial as a nullptr, then you create the callback to check it's value and set the callback,
then you create the material as a Material, and only then you cast it to CreatedGLTFMaterial. so if CreateMaterial triggers to callback but CreatedMaterial has not been assigned to CreatedGLTFMaterial yet, crash.

MAG-OR: and looking at the other disabled test, MaterialAssetEventTest, it's the same thing

MAG-OR: I have a fix. (this is basically the same as https://github.com/magnopus-opensource/connected-spaces-platform/pull/688)
so far so good:
image

MAG-OR: Fixed in 53fa1b67ecbb897c2a969ca9965ece94b2100ecf. On local CHS on  my local machine, I am now at 592 successful test runs and counting.

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -730,6 +730,17 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, MaterialEventTest)
 
         auto CB = [&CallbackCalled, &CreatedGLTFMaterial](const csp::multiplayer::MaterialChangedParams& Params)
         {
+            const auto Start = std::chrono::steady_clock::now();
+            constexpr std::chrono::seconds Timeout = std::chrono::seconds(10);
```
MAG-mv, line 734: [Question] Can we use WaitForCallback instead?

Tests/src/PublicAPITests/MaterialTests.cpp
```diff
@@ -730,6 +730,17 @@ CSP_PUBLIC_TEST(CSPEngine, MaterialTests, MaterialEventTest)
 
         auto CB = [&CallbackCalled, &CreatedGLTFMaterial](const csp::multiplayer::MaterialChangedParams& Params)
         {
+            const auto Start = std::chrono::steady_clock::now();
```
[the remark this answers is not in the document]
MAG-mv: Example:

String MaterialCollectionId;
String MaterialId;

 auto CB = [&CallbackCalled, &MaterialCollectionId, &MaterialId](const csp::multiplayer::MaterialChangedParams& Params)
 {
     MaterialCollectionId = Params.MaterialCollectionId;
     MaterialId = Params.MaterialId;

     EXPECT_EQ(Params.ChangeType, csp::multiplayer::EAssetChangeType::Created);
     CallbackCalled = true;
 };
 
 WaitForCallback(CallbackCalled);
 
 EXPECT_EQ(MaterialCollectionId, CreatedGLTFMaterial->GetMaterialCollectionId());
 EXPECT_EQ(MaterialId , CreatedGLTFMaterial->GetMaterialId());

MAG-ElliotMorris (approved):

Tests/src/PublicAPITests/MaterialTests.cpp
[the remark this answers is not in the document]
MAG-OR: changed to your fix in ff6e7b4c561852d32b4c58820235fea69571fce4
I am re-running the same tests that I included in the description of the PR.
Local CHS CI run, green the first time: https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_TestsLocalChsInstance/289270
Live CHS CI run, green the first time:
https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_FunctionalTests/289261?buildTab=overview
And 1000 successful local CHS runs of the 2 tests on my local machine:
image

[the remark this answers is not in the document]
MAG-mv: Looks good!

MAG-mv (approved):

PR #691, opened 2025-05-22
[OB-4307] fix: remove no script component error handling

MAG-ChristopherAtkinson:
From Investigation: it seems the code in written in a manner that assumes the entities which are passed are assumed to contain a script component. However, this simply checks all components in the space resulting in an incorrect assumption. 

In a traditional ECS system the assumption would be true, but is not the case for CSP.  Looking at the timeline this was introduced in 20/06/2023 by MAG-pw, only receiving formatting changes.

The PR simplify Removes the assumption and reformatting the code to reflect this is all entities  and not a subset.

### DIFF
--- a/Library/src/Multiplayer/Script/EntityScript.cpp
+++ b/Library/src/Multiplayer/Script/EntityScript.cpp
@@ -26,7 +26,6 @@
 namespace csp::multiplayer
 {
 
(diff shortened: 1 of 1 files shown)


MAG-ChristopherAtkinson:
Apologies accidently grabbed the wrong ticket information, renamed branch which has trigger the PR to close.

PR #693, opened 2025-05-22
[OF-1653] Puppeteer HTML tests

MAG-ElliotMorris:
Add puppeteer based tests that run WASM code directly in a browser.

Adds two tests, one that tests login, and one that tests the cross-thread callback error, mostly as proofs of concepts.

Pretty chuffed with how flexible this is, we've already discussed how we can pretty much use this framework verbatim for the idea of runnable examples, and would also be very simple to pull in the head of the web client and run a true integration test there, which is likely what we'll do when and if web develop invokable unit tests.

Whilst this is clearly missing an enter space test (great low hanging fruit thing), running these tests, in browser, on CI is going to practically remove the burden on CSP implementors to check against web client integration for WASM builds, (which, lets be honest, we often neglected to do at all.)

For more detailed implementation information, you can see the readme file.

> [!NOTE]  
> @mag-aidangustard is looking into ways to install `caddy` on the teamcity runners, and even to have the launch of the webserver done in a more teamcity friendly way, as the `.cmd` file provided _is_ a little fragile in a CI setting. Shouldn't stop this PR going ahead though. Whatever the outcome, I don't want the single-click test runs in the local workflow to go away.


WasmTesting/csp-wasm.test.ts
```diff
@@ -1,29 +1,44 @@
...
-  await ready(USE_DEBUG_CSP);
+  return initializeCSP(USE_DEBUG_CSP); //gotta return the promise or tests wont automatically await
 });
 
-test('Initialize CSP', () => {
-    CSPFoundation.initialise(CHS_ENDPOINT_ROOT, TENANT);
-    assert.equal(CSPFoundation.getIsInitialised(), true);
-    assert.equal(CSPFoundation.getTenant(), TENANT);
+test ('Login', async() => {
+  console.log("Got to login");
```
MAG-ElliotMorris, line 18: [fix] remove

WasmTesting/csp-wasm.test.ts
```diff
@@ -1,29 +1,44 @@
 import './pretend-to-be-a-browser'
+import  {CreateTestUser, LaunchTestPage, TEST_ACCOUNT_PASSWORD} from './testhelpers'
 
 import { test } from 'uvu';
 import * as assert from 'uvu/assert';
-import { CSPFoundation, ready, Systems } from 'connected-spaces-platform.web';
-
-let CHS_ENDPOINT_ROOT = "https://ogs-internal.magnopus-dev.cloud";
-let TENANT = "OKO_TESTS";
+import { ClientUserAgent, CSPFoundation, ready, Systems } from 'connected-spaces-platform.web';
```
MAG-ElliotMorris, line 6: [fix] Don't need clientUserAgent here anymore

WasmTesting/shared/csp-initializer.js
```diff
@@ -0,0 +1,37 @@
...
+      throw new Error("CSPFoundation failed to initialize");
+    }
+
+    const userAgent = ClientUserAgent.create();
+    userAgent.cSPVersion = CSPFoundation.getBuildID();
+    userAgent.clientOS = TESTS_CLIENT_OS;
+    userAgent.clientSKU = TESTS_CLIENT_SKU;
+    userAgent.clientVersion = CSPFoundation.getBuildID();
+    userAgent.clientEnvironment = "ODev";
+    userAgent.cHSEnvironment = "oDev";
```
MAG-ElliotMorris, line 29: [Question] I copied this from the C++ tests, but is this right, seems odd to have both `ODev` and `oDev`

WasmTesting/testhelpers.ts
```diff
@@ -0,0 +1,77 @@
+import { v4 as uuidv4 } from 'uuid';
+import { Systems } from 'connected-spaces-platform.web';
+import * as assert from 'uvu/assert';
+import puppeteer, { LaunchOptions } from 'puppeteer';
+import { pathToFileURL } from 'url';
```
MAG-ElliotMorris, line 5: [fix] Not needed now that we're not doing file:// protocol

WasmTesting/testhelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+    waitUntil: Array<'load' | 'domcontentloaded' | 'networkidle0' | 'networkidle2'> = ['load', 'networkidle0'], //Default will almost always be fine
+  ): Promise<PageTestResult> {
+    const browser = await puppeteer.launch(launchOptions);
+    const page = await browser.newPage();
+    const errors: Error[] = [];
+    const consoleMessages: string[] = [];
+  
+    page.on('pageerror', e => errors.push(e));
+    page.on('console', msg => consoleMessages.push(msg.text()));
```
MAG-ElliotMorris, line 61: [consider] Oh ... so I guess `const` just means nothing in JS land? These should probably be `let` I guess.

MAG-ElliotMorris: Guess this is just a misunderstanding of what const does.

WasmTesting/shared/csp-initializer.js
MAG-ChristopherAtkinson: hmm I am unsure but looking at the inline doc seems this was done with reason, may be worth reaching out to MCS to confirm

MAG-ElliotMorris: I did, they don't know either. 🤷 Not the end of the world.

WasmTesting/shared/csp-initializer.js
```diff
@@ -0,0 +1,37 @@
+import { ready, CSPFoundation, ClientUserAgent} from '../node_modules/connected-spaces-platform.web/connectedspacesplatform.js';
+
+let CHS_ENDPOINT_ROOT = "https://ogs-internal.magnopus-dev.cloud";
```
MAG-OR, line 3: [question] likely a silly question. will we ever want to run this on local CHS?

WasmTesting/testhelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    if (userCredentials != null && userCredentials.email && userCredentials.password) {
+        htmlPath = `${htmlPath}?useDebugCsp=${encodeURIComponent(useDebugCSP)}&email=${encodeURIComponent(userCredentials.email)}&password=${encodeURIComponent(userCredentials.password)}`;
+    }
```
MAG-OR, line 65: [question] would having an error message if the credentials are empty or the email or password is missing for any reason be useful?

MAG-OR (approved):

WasmTesting/shared/csp-initializer.js
MAG-ElliotMorris: Likely eventually, but no immediate plans. We'd probably have to do the WASM webclient fixes before we did that, as local CHS is only supported in poco right now.

WasmTesting/testhelpers.ts
```diff
@@ -0,0 +1,77 @@
+import { v4 as uuidv4 } from 'uuid';
+import { Systems } from 'connected-spaces-platform.web';
+import * as assert from 'uvu/assert';
+import puppeteer, { LaunchOptions } from 'puppeteer';
+
+export const TEST_ACCOUNT_PASSWORD = "3R{d2}3C<x[J7=jU";
```
MAG-ChristopherAtkinson, line 6: [Question] can we just configure information such as this through .env file?

MAG-ChristopherAtkinson (approved):
The changes look good to me

WasmTesting/testhelpers.ts
MAG-ElliotMorris: Going to go ahead and leave this, can't really decide one way or the other.

PR #694, opened 2025-05-23
[NT-0] fix!: fix IsPersistant typo

MAG-OR:
It is IsPersistent.

This typo has been present, and ignored, in the last 3 or 4 PRs that touched space entity. I got tired of seeing it.

Unfortunately, ObjectMessage::GetIsPersistent() is public (defined in MCSTypes.h), so I marked the change as breaking...


MAG-ElliotMorris (approved):
Love to see it!

The spelling of an argument doesn't make a change breaking though, it's only the position in the arg list that really matters. A function type is represented as `returnT(ArgT1, ArgT2, etc)`, names are irrelevant.

MAG-ChristopherAtkinson (approved):
Took me a minute to realise it was not spelled correctly.

PR #695, opened 2025-05-27
[NT-0] fix: fix mistake in the commit hook that prevents pushing breaking fixes

MAG-OR:
That's all there is to it. This mistake means I cannot amend the commit message in https://github.com/magnopus-opensource/connected-spaces-platform/pull/694 to mark it as breaking and therefore cannot merge it.


MAG-ElliotMorris (approved):
I would just do `--no-verify` for the moment if this is blocking you. Nice one fixing stuff that's broken as you come across it.

MAG-OR:
> I would just do `--no-verify` for the moment if this is blocking you. Nice one fixing stuff that's broken as you come across it.

took some gymnastics because it was a reword but it worked. thanks!

MAG-AdamThorn (approved):

PR #696, opened 2025-05-28
[OF-1639] feat: Guard against component operations for locked Entities

MAG-AdamThorn:
Ensure that Components cannot be added to or removed from locked Entities.

- Add an IsModifiable check to both the AddComponent and RemoveComponent methods.
- Log an error if the user attempts to perform either operation on a locked Entity.
- Add tests to validate both paths.


MAG-ChristopherAtkinson (approved):
Changes look good, and good to see mocking being used.

MAG-mv (approved):

PR #698, opened 2025-05-30
[OB-4284] fix: Prevent SpaceEntity::ComponentsLock from doubling locking

MAG-mv:
In the following circumstance, we can cause SpaceEntity::ComponentsLock to be locked twice in the same code path, causing the program to lock up:

- A script exists that modifies a component property when executed
- The script source is then updated
- This causes a patch to be sent, changing the script source
- A patch is received by another client, which locks the ComponentsLock mutex
- ScriptSpaceComponent overrides SetPropertyFromPatch to get them to re-run the script when changed
- This causes the script to run, and the script modifies a property
- Modifying this property calls SpaceEntity::AddDirtyComponent, which also locks ComponentsLock
- The program then freezes

We have currently fixed this by introducing a recursive_mutex, which isn't the best way forward. Ideally, we want to understand our threading and locking model, so we can redesign a better system where we fully understand all of our codepaths, and find a safer way of fixing this without introducing a large amount of locks for each component update.

I unfortunately couldn't create a reliable test for this, as double locking is undefined behaviour. I did manual testing to ensure this bug is fixed.

I also fixed the following bugs introduced by the Serialization refactor:
    - Readded mutex logs for serialzation/deserialization
    - An error log will fire when receiving a patch update stating that the entity could not be found
    - SpaceEntity::UpdateCallback will not fire if only components have changed, and not entity properties

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -516,7 +516,7 @@ private:
 
     CSP_START_IGNORE
     mutable std::mutex EntityMutexLock;
-    mutable std::mutex ComponentsLock;
+    mutable std::recursive_mutex ComponentsLock;
     mutable std::mutex PropertiesLock;
     CSP_END_IGNORE
 
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -436,7 +436,7 @@ ComponentBase* SpaceEntity::AddComponent(ComponentType AddType)
         return nullptr;
     }
 
-    std::scoped_lock<std::mutex> ComponentsLocker(ComponentsLock);
+    std::scoped_lock ComponentsLocker(ComponentsLock);
 
     if (AddType == ComponentType::ScriptData)
     {
@@ -473,7 +473,7 @@ void SpaceEntity::RemoveComponent(uint16_t Key)
         return;
     }
 
-    std::scoped_lock<std::mutex> ComponentsLocker(ComponentsLock);
+    std::scoped_lock ComponentsLocker(ComponentsLock);
 
     if (!TransientDeletionComponentIds.Contains(Key) || Components.HasKey(Key))
     {
@@ -505,7 +505,7 @@ void SpaceEntity::ApplyLocalPatch(bool InvokeUpdateCallback)
     if (!csp::systems::SystemsManager::Get().GetMultiplayerConnection()->GetAllowSelfMessagingFlag())
     {
         std::scoped_lock<std::mutex> PropertiesLocker(PropertiesLock);
-        std::scoped_lock<std::mutex> ComponentsLocker(ComponentsLock);
+        std::scoped_lock ComponentsLocker(ComponentsLock);
 
         auto UpdateFlags = static_cast<SpaceEntityUpdateFlags>(0);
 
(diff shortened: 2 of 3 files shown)


Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -516,7 +516,7 @@ class CSP_API SpaceEntity
 
     CSP_START_IGNORE
     mutable std::mutex EntityMutexLock;
-    mutable std::mutex ComponentsLock;
+    mutable std::recursive_mutex ComponentsLock;
```
MAG-mv, line 519: Read PR description to understand the decision to use this

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1199,20 +1203,25 @@ void SpaceEntity::FromObjectPatch(const mcs::ObjectPatch& Patch)
...
         uint64_t ComponentCount = ComponentUnpacker.GetRuntimeComponentsCount();
 
-        ComponentUpdates = csp::common::Array<ComponentUpdateInfo>(ComponentCount);
-        size_t ComponentIndex = 0;
-
-        for (const auto& ComponentDataPair : *PatchComponents)
+        if (ComponentCount > 0)
         {
-            if (ComponentDataPair.first >= COMPONENT_KEY_END_COMPONENTS)
+            UpdateFlags = static_cast<SpaceEntityUpdateFlags>(UpdateFlags | UPDATE_FLAGS_COMPONENTS);
```
MAG-mv, line 1208: Added if statement to check if there are component updates + flag for UPDATE_FLAGS_COMPONENTS. This is a bug that was introduced due tot he serialization refactor

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1255,7 +1264,7 @@ void SpaceEntity::ComponentFromItemComponentData(uint16_t ComponentId, const mcs
                 Component->Properties[PatchComponentPair.first] = Property;
                 Component->OnCreated();
             }
-
+            std::scoped_lock ComponentsLocker(ComponentsLock);
```
MAG-mv, line 1267: Readded components lock for updates. This is a bug that was introduced due to the serialization refactor

Library/src/Multiplayer/SpaceEntitySystem.cpp
```diff
@@ -1526,6 +1526,7 @@ void SpaceEntitySystem::ApplyIncomingPatch(const signalr::value* EntityMessage)
             if (Entities[i]->GetId() == Patch.GetId())
             {
                 Entities[i]->FromObjectPatch(Patch);
+                EntityFound = true;
```
MAG-mv, line 1529: Readded EntityFound to prevent error log firing. This is a bug that was introduced due to the serialization refactor

MAG-ElliotMorris (approved):

MAG-ChristopherAtkinson (approved):
Looking at the changes they seem reasonable give the execution, I am always hesitant when it comes to using std::recursive_mutex, they generally imply a fundamental issue with the design of an application. 

However given we already use these within the code base, would argue this is out of scope vs the priority of the changes. Still, great work, I can imagine this was problematic to track down.

MAG-mv:
> Looking at the changes they seem reasonable give the execution, I am always hesitant when it comes to using std::recursive_mutex, they generally imply a fundamental issue with the design of an application.
> 
> However given we already use these within the code base, would argue this is out of scope vs the priority of the changes. Still, great work, I can imagine this was problematic to track down.

@MAG-ChristopherAtkinson I am going to create a ticket to investigate and potentially re-design our thread/lock system. While looking into this ticket, it seems flawed, and no one on the team understand the design decisions that went into the current system

PR #699, opened 2025-06-02
[OF-1645] Fix: Conversation tests

MAG-mv:
ConversationComponentSecondClientEventDelayTest -
- Created a second user for the Test runner
- New user has to be set as moderator  to create test object in the space
- Had to disable rate limiting so all patches were sent

ConversationComponentPermissionsTest -
- Had to disable rate limiting so all patches were sent
- Reordered EnterSpace to ensure callback is set before entities are retrieved.

### DIFF
--- a/MultiplayerTestRunner/src/RunnableTests/CreateConversation.cpp
+++ b/MultiplayerTestRunner/src/RunnableTests/CreateConversation.cpp
@@ -41,13 +41,15 @@ void RunTest()
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto& EntitySystem = *SystemsManager.GetSpaceEntitySystem();
 
+    // Ensure patch rate limiting is off, as we're sending patches in quick succession.
+    EntitySystem.SetEntityPatchRateLimitEnabled(false);
+
+    csp::multiplayer::SpaceTransform ObjectTransform { csp::common::Vector3::Zero(), csp::common::Vector4::Zero(), csp::common::Vector3::One() };
+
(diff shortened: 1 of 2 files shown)


MultiplayerTestRunner/src/RunnableTests/CreateConversation.cpp
```diff
@@ -41,13 +41,15 @@ void RunTest()
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto& EntitySystem = *SystemsManager.GetSpaceEntitySystem();
 
-    std::promise<csp::multiplayer::SpaceEntity*> CreateObjectResultPromise;
-    std::future<csp::multiplayer::SpaceEntity*> CreateObjectResultFuture = CreateObjectResultPromise.get_future();
+    // Ensure patch rate limiting is off, as we're sending patches in quick succession.
+    EntitySystem.SetEntityPatchRateLimitEnabled(false);
```
MAG-ChristopherAtkinson, line 45: [Question] What is the current configuration for this on the live services? (Wonder it would be worth considering globally setting this in future:; outside the scope of the changes here)

MAG-AdamThorn (approved):

MultiplayerTestRunner/src/RunnableTests/CreateConversation.cpp
MAG-mv: Yeah, I agree. I will add it as a topic for the weekly sync

PR #700, opened 2025-06-03
[OF-1669] refac: Create IRealtimeEngine interface

MAG-ElliotMorris:
Part of an ongoing modularity effort, create an IRealtimeEngine interface, containing the proposed interface for what is currently the multiplayer module.

This is mostly a port of `SpaceEntitySystem`, removing all the interfaces that are not relevent for a generic multi/single-user abstraction, as well as all the interfaces that are only used internally to what will become the RealtimeEngine module. See the documentation at the bottom of the file for a list of all the functions that have not made the interface, as well as the ones planned for the Multi user concretion.

> [!NOTE]  
> The biggest swing here is the removal of `AddEntity`, which is used [here ](https://github.com/magnopus/ichabod/blob/9bcfe3d072fd4498f5bbd7dcfd2f6e301fe46e9c/ichabod/space_manager.cpp#L1554)in ichabod, although it seems redundant as `CreateObject` is called first. We will need to confirm uses of these methods before we 100% commit to this interface later, as some are simply wrapped and passed on by clients, so they may have downstream dependencies, despite not having internal use. It would be good to remove this though, it makes the interface symmetric. (Despite the create with args pattern not being my favourite here 🤷 )

For the scripting part, it seems all the interfaces with scripting are to do with leader election, which is a purely multiplayer concept. This is a good thing imo, as it means the base interface can just be about entity management, as it should be.

Changed a couple of enums to have declared types so they could be forward declared.

Also went over the documentation with @MAG-mv , fixing up some links, but also rephrasing somewhat to get us off on a good foot here. The vague themes we addressed.

- Removed the `InVariable` pattern
- Added callouts for ownership of pointers
- Defined Entity/Object/Avatar triad. Avatars are entities that contain AvatarSpaceComponents, Objects are entities that do not contain AvatarSpaceComponents.
- Moved away from the `Object` concept as much as possible in the actual inline docs, although it does still remain as we can't fully remove it yet as that would mean changing method names and in-production use cases, (not to mention all the backend types) We'd rather not have this concept in the long run, defining something as an entity that _doesn't_ have a component is a smell in our opinion, as evidenced by the fact that we didn't really know what the terms meant until we researched. 

> [!NOTE]  
> Keep in mind the nature of what this type is, especially when thinking about docs. Anything we write there constitutes an interface guarantee that must apply to _all_ implementations. This sort of thing can shoot you in the foot if you're not careful, as you can't even be sure what data structures any given implementation is using. Concrete implementations will document their behaviours more precisely, as they'll have knowledge of their implementation.
This mainly came up in our `Find` methods, which were more or less fine, but we wanted to make sure not to define exactly how entities must be ordered and searched, as many storage approaches are valid.
>
> Even saying this, the fact that we have `GetByIndex` method does constrain the implementation somewhat. But I think that's acceptable for the moment. This may change if and when we move to key based non-owning access of entities though.

### DIFF
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -19,15 +19,13 @@
 #include "CSP/Common/List.h"
 #include "CSP/Common/String.h"
 
-CSP_START_IGNORE
 namespace csp::multiplayer
 {
-enum class AvatarPlayMode : int;
-enum class AvatarState : int;
+enum class AvatarPlayMode;
+enum class AvatarState;
 class SpaceTransform;
 class SpaceEntity;
 }
-CSP_END_IGNORE
 
 namespace csp::multiplayer
 {
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -95,7 +95,7 @@ public:
     /// which will provide a non-owning pointer to the new SpaceEntity so that it can be used on the local client.
     /// @pre Entity creation callback is non null. This can be set via SetEntityCreatedCallback.
     CSP_ASYNC_RESULT virtual void CreateAvatar(const csp::common::String& Name, const csp::multiplayer::SpaceTransform& SpaceTransform,
-        csp::multiplayer::AvatarState State, const csp::common::String& AvatarId, csp::multiplayer::AvatarPlayMode AvatarPlayMode,
+        const csp::multiplayer::AvatarState& State, const csp::common::String& AvatarId, const csp::multiplayer::AvatarPlayMode& AvatarPlayMode,
         csp::multiplayer::EntityCreatedCallback Callback)
         = 0;
 
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -1,0 +1,241 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "CSP/Common/List.h"
+#include "CSP/Common/String.h"
+
+CSP_START_IGNORE
+namespace csp::multiplayer
+{
+enum class AvatarPlayMode : int;
+enum class AvatarState : int;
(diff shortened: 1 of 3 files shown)


Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -0,0 +1,241 @@
...
+#include "CSP/Common/String.h"
+
+CSP_START_IGNORE
+namespace csp::multiplayer
+{
+enum class AvatarPlayMode : int;
+enum class AvatarState : int;
+class SpaceTransform;
+class SpaceEntity;
+class EventBus;
```
MAG-ElliotMorris: [fix] Redundant forward declaration.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/700/commits/fe0f33d429c269d8eb5159f3e8a497f935051955

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -0,0 +1,240 @@
...
+
+    /// @brief Destroy the specified entity.
+    /// @param Entity csp::multiplayer::SpaceEntity : A non-owning pointer to the entity to be destroyed.
+    /// @param Callback csp::multiplayer::CallbackHandler : A callback that executes when the entity destruction is complete.
+    CSP_ASYNC_RESULT virtual void DestroyEntity(csp::multiplayer::SpaceEntity* Entity, csp::multiplayer::CallbackHandler Callback) = 0;
+
+    /// @brief Sets a callback to be executed when an entity is fully created.
+    ///
+    /// Only one EntityCreatedCallback may be registered, calling this function again will override whatever was previously set.
+    /// This must be set prior to creating an Entity.
```
MAG-AdamThorn: [Question] Is that true, that the callback must be set prior to creating an Entity?
If you are creating a single user experience then the EntityCreatedCallback would not be called in response to someone else creating an Entity.

MAG-AdamThorn (approved):

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
MAG-ElliotMorris: No, it's not strictly true. Matt and I talked about this at some length. 

It came from the doc on `SetEntityCreatedCallback`, `/// If this is not set, some patch functions may fail.` and out of an abundance of caution. My theory is it's better to be stricter here in the case of ambiguity, as users won't mind if they forget to set it but it still works, but they may have issues if they arn't told they have to set it and it fails. The `some` and the `may` in that doc makes me uncomfortable, wanted to promote it to a certainty.

However, thinking about it today I'm less sure. This _could_ be documented on the multiplayer concretion itself, still as a hard precondition, but then we could leave the underlying interface without the restriction. Sort of coming around to that opinion thinking it over again. I think that was your preference anyhow @MAG-mv wasn't it?

MAG-mv: Yeah, I'm happy to remove from the base type. I don't think it makes sense here

MAG-ElliotMorris: Removed the precondition that this is set on the base interface, it's up to the concretions now.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/700/commits/5c75bbda38e7aac44fb4a5f6c641715964ca1e7f

PR #701, opened 2025-06-03
[NT-0] chore: update premake from 5.0-beta2 to 5.0-beta6

MAG-OR:
This is in preparation for https://magnopus.atlassian.net/browse/OF-1640.

Three changes of note: 
- the Android fixes are no longer required (and actively fail the build).
- the config keyword no longer exists. now using filter.
- the gmake2 action in the premake command was no longer exists. it has been renamed to gmake.

Tested on the x64 debug build and the WASM debug build. The CI will tell us about the other ones...

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -186,7 +186,8 @@ if not Project then
             }
 
             libdirs {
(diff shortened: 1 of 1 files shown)


MAG-OR:
> @MAG-OR Can this draft be closed?

I suppose, yes

PR #702, opened 2025-06-03
Prevent referencing native nullptrs

magnopus-swifty:
In some cases, the C++ layer will use `nullptr` in callbacks or return values to indicate failure or a missing object. The C# wrapper generator currently wraps these in a managed object instance; calling any function or property on this object will result in a segfault.

The expected behavior in C# is that objects are valid or `null`.

This PR will:

* Ensure that invalid `NativeClassWrapper` instances cannot be created
* Return a C# `null` any time a native `nullptr` is produced

### DIFF
--- a/Tools/WrapperGenerator/Templates/CSharp/Partials/Class/TaskMethod.mustache
+++ b/Tools/WrapperGenerator/Templates/CSharp/Partials/Class/TaskMethod.mustache
@@ -20,7 +20,18 @@ static void {{ delegate.name }}Function(
     var tcs = (TaskCompletionSource<{{# has_multiple_results }}({{/ has_multiple_results }}{{# results }}{{# type }}{{> Type }}{{/ type }}{{> Comma }}{{/ results }}{{# has_multiple_results }}){{/ has_multiple_results }}>)_handle.Target;
     var _this = ({{ class_name }})tcs.Task.AsyncState;
 
-    var task_result = {{# has_multiple_results }}({{/ has_multiple_results }}{{# results }}{{# type.is_pointer_or_reference }}new {{# type }}{{> Type }}{{/ type }}({{/ type.is_pointer_or_reference }}_{{ name }}{{# type.is_pointer_or_reference }}){{/ type.is_pointer_or_reference }}{{> Comma }}{{/ results }}{{# has_multiple_results }}){{/ has_multiple_results }};
+    var task_result = 
+        {{# has_multiple_results }}({{/ has_multiple_results }}
+        {{# results }}
+            {{# type.is_pointer_or_reference }}
+                _{{ name }}.Pointer == IntPtr.Zero ? null : new {{# type }}{{> Type }}{{/ type }}(_{{ name }})
+            {{/ type.is_pointer_or_reference }}
+            {{^ type.is_pointer_or_reference }}
+                _{{ name }}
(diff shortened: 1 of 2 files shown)


Tools/WrapperGenerator/Templates/CSharp/Partials/Class/TaskMethod.mustache
```diff
@@ -20,7 +20,18 @@ static void {{ delegate.name }}Function(
     var tcs = (TaskCompletionSource<{{# has_multiple_results }}({{/ has_multiple_results }}{{# results }}{{# type }}{{> Type }}{{/ type }}{{> Comma }}{{/ results }}{{# has_multiple_results }}){{/ has_multiple_results }}>)_handle.Target;
     var _this = ({{ class_name }})tcs.Task.AsyncState;
 
-    var task_result = {{# has_multiple_results }}({{/ has_multiple_results }}{{# results }}{{# type.is_pointer_or_reference }}new {{# type }}{{> Type }}{{/ type }}({{/ type.is_pointer_or_reference }}_{{ name }}{{# type.is_pointer_or_reference }}){{/ type.is_pointer_or_reference }}{{> Comma }}{{/ results }}{{# has_multiple_results }}){{/ has_multiple_results }};
+    var task_result = 
+        {{# has_multiple_results }}({{/ has_multiple_results }}
+        {{# results }}
+            {{# type.is_pointer_or_reference }}
+                _{{ name }}.Pointer == IntPtr.Zero ? null : new {{# type }}{{> Type }}{{/ type }}(_{{ name }})
```
magnopus-swifty, line 27: This will change the generated code from (for example):

```csharp
var task_result = new Csp.Systems.NullResult(_Result);
```

to

```csharp
var task_result = _Result.Pointer == IntPtr.Zero ? null : new Csp.Systems.NullResult(_Result);
```

MAG-ElliotMorris (commented):
I don't see any reason not to approve this, a bit surprised this wasn't already the behaviour. Two questions before I do however.

# 1. We don't have a huge gamut of tests for our interop layers at the moment, as far as I know, tends to be part of why we keep putting off small improvements. Is it possible to run this through some sort of test suite at the unity level, or at least a quick battery of manual test cases if you haven't already?
# 2. Do you know if this will require code changes by the unity client? I.e, should it be marked as a breaking change in our release notes?

magnopus-swifty:
> 1. We don't have a huge gamut of tests for our interop layers at the moment, as far as I know. Tends to be part of why we keep putting off small improvements. Is it possible to run this through some sort of test suite at the unity level, or at least a quick battery of manual test cases if you haven't already?

This was successfully run through the (fairly comprehensive) set of integration tests in https://github.com/magnopus/oly-foundation-unity/tree/dev/Foundation.Integration.Unity~/Assets/IntegrationTests

> 2. Do you know if this will require code changes by the unity client? I.e, should it be marked as a breaking change in our release notes?

It only marks a single property as `Obsolete` so code will continue to compile just fine. I went this direction because I don't know all the places where this code is accessed - however, the oly-foundation-unity package is intended to be a complete Adapter to the wrapper generator code so making the changes there should cover the vast majority of cases.

I already have a PR to remove all these calls just waiting for this to be published: https://github.com/magnopus/oly-foundation-unity/pull/294

PR #704, opened 2025-06-06
Refactor tokenizer for wrapper generator parser

magnopus-swifty:
Initial exploratory code improvements for parser

- Fix code style problems
- Add docstrings with examples (tested with doctest)
- Add license headers


Tools/WrapperGenerator/word_reader.py
```diff
@@ -37,30 +48,63 @@ def __init__(self, text: str):
...
+
         start_index = self.index
         # Always consume the first character
         end_index = self.index + 1
 
         while True:
             next_char = self.peek_char()
 
-            if next_char == '\n':
+            if next_char == "\n":
```
MAG-ElliotMorris, line 121: [Question] Is this significant? I would have thought the single quoted version to be the same as this (better even, as it communicates this is a single character and not a string consisting of 2 characters)

MAG-ElliotMorris (approved):
Pretty uncontroversial, thanks for doing a commit-by-commit PR.

I must admit seeing changes enter the parser evokes contrasting feelings in me. Super glad to see it change, but anything more fundamental than these sorts of readability improvements would start to make me antsy regarding acceptance testing.

For example, regarding my nervousness with the existing wrapper gen unit tests, they don't seen to have been run for nearly a year on CI, and arn't functional at the moment.

https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_WrapperGeneratorUnitTests#all-projects

MAG-ElliotMorris:
I'll also say that if we're going to start changing style broadly, (I don't disagree, python is opinionated and we should yield to its opinions), I'd rather we applied a PEP8 linter at the build/CI level and conformed to it, rather than ad-hoc renames as we go, in fear of manual style changes not quite achieving consistency.

Not trying to slow down ad-hoc improvements, I just want to have an eye to the future as well. (and also sidestep any debates on specific points of style that may emerge cracking open this box...)

Tools/WrapperGenerator/word_reader.py
magnopus-swifty: I think this was the IDE "helping". I can revert.

magnopus-swifty: Actually, this made it consistent with the other character strings in the file and `''` for characters is a C-ism so I'll leave it.

MAG-ElliotMorris:
> I think that's reasonable. I'll submit a separate PR containing just file & symbol renames rather than doing it ad-hoc.

I don't think you need to undo anything in this PR, this is reasonable, just if you or us is going to start widely reformatting the codebase that's worth thinking about.

PR #708, opened 2025-06-09
[OF-1690] Fix: Reinstate the unit tests for the wrapper generator

magnopus-swifty:
The wrapper generator unit tests in `UnitTesting` were not compiling due to changes made to the main project that weren't reflected in these tests. This fixes make the project build again so they can be re-enabled in TeamCity.

Key fixes:

I was getting a bunch of problems with the `jinja2-workarounds` module - even though I seem to have it installed, jinja2 was failing to find the module. It doesn't seem to have any effect on the output and the tests run so I just removed those lines, in the sprit of "fewer dependencies is better".

Removed the custom Free logic to match the main project (although I'm not sure why this is necessary) and removed references to mimalloc.

Upgraded the version of Unity (although I'm not sure this is strictly necessary).


UnitTesting/Templates/Exports/exports.cpp.jinja2
```diff
@@ -17,7 +17,7 @@ using namespace std::chrono_literals;
 namespace csp {
     // Helper function to free allocated memory from wrappers
     CSP_API void Free(void* Pointer) {
-        csp::memory::DllFree(Pointer);
```
MAG-ElliotMorris, line 20: [comment] I'm guessing this was failing to compile now, because `csp::,memory::DllFree` is no longer a thing. This is why we need all our code built on CI, or stuff like this gets missed and becomes more and more expensive as people get less and less sure why it's the way it is.

This change is fine, although the concept of the `Free` method at all is redundant now, though that would require larger changes.

UnitTesting/Templates/Exports/exports.cpp.jinja2
```diff
@@ -3,10 +3,10 @@
 // Include the generated C interface so it gets compiled
 #include "../../Tools/WrapperGenerator/Output/C/generated_wrapper.h"
 
-
 #include <cassert>
 #include <cmath>
 #include <limits>
+#include <stdexcept>
```
MAG-ElliotMorris, line 9: [fix] Add `include <chrono>` as well, you need it for the `using namespace std::chrono_literals;` on line 14. Otherwise:

image

I really have no idea how this works for you, and on CI tbh. Where would the `chrono` transient include be coming from? 🤔 Maybe some precompiled header nonsense that depends on the project already having been used a little? (I'm using a fresh checkout to test.)

UnitTesting/premake5.lua
```diff
@@ -77,10 +69,6 @@ project "UnitTestingBinary"
             "MultiProcessorCompile"
         }
 
-        links {
-            "mimalloc"
-        }
```
MAG-ElliotMorris, line 82: [comment] Nice one removing all the old memory crud. It's annoying that it's separate so it can even get out of sync.

UnitTesting/tests/csharp/ProjectSettings/MemorySettings.asset
```diff
@@ -0,0 +1,35 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
```
MAG-ElliotMorris: [comment] A tad uncomfortable having these project files in the repo, just because they'll update every time someone messes with them and we've no real idea if the changes matter or not. Guess we'll just have to hope, only a minor thing anyhow. 🤷

MAG-ElliotMorris (changes requested):
Pulled your fork and ran the flow through, confirm that the tests build and that the unity tests run when you have the correct unity install. I'm quite jazzed to see this integration already exists, albeit a tad disheartened about how little we know at the moment regarding the coverage level it gives us.

I did need to make one source change to get the build going however. Frustratingly I have no idea why it's showing variant behaviour across differing machines.

I also [ran the CI](https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_WrapperGeneratorUnitTests/290505?showLog=290505_1465_1463.1530.1537.1556&logFilter=debug&logView=flowAware), which still fails, but gets further now, building all the tests, although failing to actually run either the unity or the typescript tests. (I realize you were only targeting unity, I suspect we'll need to get the build team involved to deal with having the correct install.)

@magnopus-swifty How far are you planning to take this? Are you going to handle getting the CI running again, or do you need us to pick that up?

We need to do it one way or the other or these fixes will just fall out of sync again the next time we change the library significantly and the tests will stop working. (Plus ... works on my machine is not acceptable for anything C++'ey cause the language _stinks_! :P )

UnitTesting/tests/csharp/ProjectSettings/MemorySettings.asset
magnopus-swifty: Since we're just running tests in a stock empty Unity project we might be able to just exclude these files from versioning. I can explore that - Agreed it's less maintenance and makes the tests inherently compatible with "any version of Unity that can run the code".

UnitTesting/Templates/Exports/exports.cpp.jinja2
magnopus-swifty:
```
1>Note: including file:   C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.36.32532\include\thread
1>Note: including file:    C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.36.32532\include\__msvc_chrono.hpp
```

In my version of MSVC, it's included via `<thread>`. Just as a curiosity. Agree we shouldn't rely on this behaviour.

Surprisingly this also compiles in the linux docker container when building the WASM tests.

UnitTesting/Templates/Exports/exports.cpp.jinja2
magnopus-swifty: I suspect this is necessary due to the peculiar build configuration for these tests. [`premake5.lua`](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/UnitTesting/premake5.lua#L36) includes some of the core files from the library directly.

The official `Free` is in [`CSPFoundation.cpp`](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/src/CSPFoundation.cpp#L415) which obviously these tests want to avoid including.

UnitTesting/tests/csharp/ProjectSettings/MemorySettings.asset
magnopus-swifty: I've removed (and ignored) all the `ProjectSettings` from the project - the tests run just fine this way, Unity creates default settings and it ensures that the tests work in a new empty project.

UnitTesting/Templates/Exports/exports.cpp.jinja2
MAG-ElliotMorris: Aye, I just meant, even the offical `Free` is redundant now, so the whole pattern can be binned in the future.

UnitTesting/Templates/Exports/exports.cpp.jinja2
MAG-ElliotMorris: Oh neat. Where does `thread` even come from in those files. Resisting chasing that down ... not important Elliot :P

Went trawling through the STL implementations for 5 mins, `msvc_chrono` has been in `thread` for a long time, so it's probably about getting `thread` itself. Extremely MSVC specific behaviour though.

It will certainly be a changed internal include in the MSVC STL between `14.36.32532` (yours) and `14.43.34808` (mine), looks like we both agree we shouldn't rely on transient includes however, all good. (I wonder if clang-tidy can detect stuff like this, it's bloody impossible to know about otherwise 🤔 )

Note. I'm going to unresolve this conversation until the change goes in. I wouldn't normally ask one-off external contributors, but if you're planning to commit to the CSP repo a lot, it might smoothen things over if you were to think about following our PR process. https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/PR_GUIDELINES.md

UnitTesting/tests/csharp/ProjectSettings/MemorySettings.asset
MAG-ElliotMorris: Ah! That's awesome. Thanks man, wasn't expecting this to change.

UnitTesting/Templates/Exports/exports.cpp.jinja2
MAG-ElliotMorris: Oh wait, fix is already in. Resolving again. https://github.com/magnopus-opensource/connected-spaces-platform/pull/708/commits/e7da15d6da571b15f61784d108c959b6c9be28fe

UnitTesting/tests/csharp/ProjectSettings/MemorySettings.asset
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/708/commits/2d8813b1fc923455018f1cbfb2610fb93180b1a4

MAG-ElliotMorris (approved):

MAG-ElliotMorris:
You got hit by some of our unfortunately frequent CI runners "hickups". Have restarted the jobs that failed on the other runner, but they may need to get rebooted. For reasons we never managed to get good answers for, our agents occasionally lose all ability to launch docker instances 😞

PR #709, opened 2025-06-10
Misc/of 1640 introduce the clang tidy linter static code analyser

MAG-OR:
This work was meant to integrate clang-tidy into our build.

What this work actually uncovered is that our build system, and to a lesser extent the OS on which we build (Windows), are flawed. The reasons for this are explained in detail in the story (OF-1640).

This PR contains the various ways I have attempted to use clang-tidy. (Note I am not using "integrate", because we can't integrate it in our build system in its current state). 
The unique python script is in no way, shape, or form, intended to be used as is. It just shows various methods that we could use (though the story explains why we likely should not use them) in order to force clang-tidy into our workflow. Should a specific method be chosen, the python script (if it makes sense to keep it) will have to be refactored to remove the other ones.

These were the todos for the story as it was intended in the first place (i.e. assuming there was a way to integrate clang-tidy neatly). If a solution is chosen, some may still be relevant.
- review current .clang-tidy file + existing templates out there
- discuss/agree what team wants in new .clang-tidy file, and commit it
- create a CI job for the python script
- parse the output to something useable? -- output is now readable. do we really need to parse it?
- determine which platform and configurations we want to run this on
- find source of "error while processing: <file>" errors (I think this is just how clang-tidy says that it found an error rather than a warning, i.e. I don't think this is clang-tidy complaining that it couldn't scan the file at all. likely nothing to worry about)

### DIFF
--- a/run_clang-tidy.py
+++ b/run_clang-tidy.py
@@ -6,16 +6,34 @@ import sys
 
 # todo: make these command-line arguments?
 VERBOSE = True
-USE_MSBUILD_CLANGTIDY = True
-USE_CLANGPLUSPLUS = not USE_MSBUILD_CLANGTIDY
-
 PLATFORM = "x64"
 CONFIGURATION = "DebugDLL"
-COMPILE_FILE_NAME = "compile_commands.json"
+
+CALLING_METHOD = 0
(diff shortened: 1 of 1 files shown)


MAG-ElliotMorris:
@MAG-OR Can this draft be closed?

PR #710, opened 2025-06-11
[OB-4404] fix: Fix crash when deserializing invalid component types

MAG-mv:
- Added exception catching when deserializing components
- Currently haven't added logging, as we are changing how this is going to be done in the multiplayer refactor
- We will still receive logs from the exceptions, even though they are caught.

### DIFF
--- a/Library/src/Multiplayer/MCS/MCSTypes.cpp
+++ b/Library/src/Multiplayer/MCS/MCSTypes.cpp
@@ -109,6 +109,38 @@ namespace
             throw std::invalid_argument("Trying to deserialize unsupported ItemComponentDataType");
         }
     }
+
+    void DeserializeComponents(SignalRDeserializer& Deserializer, std::optional<std::map<PropertyKeyType, ItemComponentData>>& Components)
+    {
+        if (Deserializer.NextValueIsNull() == false)
+        {
+            Components = std::map<PropertyKeyType, ItemComponentData> {};
+
+            size_t ComponentsSize = 0;
+            Deserializer.StartReadUintMap(ComponentsSize);
+
+            for (size_t i = 0; i < ComponentsSize; ++i)
+            {
+                try
+                {
+                    std::pair<PropertyKeyType, ItemComponentData> ComponentKeyValue;
+                    Deserializer.ReadKeyValue(ComponentKeyValue);
+
+                    (*Components)[ComponentKeyValue.first] = ComponentKeyValue.second;
+                }
+                catch (const std::exception&)
+                {
+                }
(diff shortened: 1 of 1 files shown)


Library/src/Multiplayer/MCS/MCSTypes.cpp
```diff
@@ -285,7 +313,34 @@ void ObjectPatch::Deserialize(SignalRDeserializer& Deserializer)
             Deserializer.EndReadArray();
         }
 
-        Deserializer.ReadValue(Components);
```
MAG-AdamThorn, line 288: [Consider] Moving this logic to a helper function to be called in these two places.

Library/src/Multiplayer/MCS/MCSTypes.cpp
```diff
@@ -207,7 +207,35 @@ void ObjectMessage::Deserialize(SignalRDeserializer& Deserializer)
...
+            for (size_t i = 0; i < ComponentsSize; ++i)
+            {
+                try
+                {
+                    std::pair<PropertyKeyType, ItemComponentData> ComponentKeyValue;
+                    Deserializer.ReadKeyValue(ComponentKeyValue);
+
+                    (*Components)[ComponentKeyValue.first] = ComponentKeyValue.second;
+                }
+                catch (const std::exception&)
```
MAG-ChristopherAtkinson: [Consider] we should likely still log, I know this is currently being rewritten, but this will be going into a published build

MAG-mv: It would require a large refactor to pass a log callback into each instance that is created, as we are moving away from including the SystemsManager / log system. We still get this info on clients.

MAG-mv: I'm not sure how cout/printf will play on wasm builds, either+

MAG-AdamThorn (approved):

Library/src/Multiplayer/MCS/MCSTypes.cpp
MAG-ChristopherAtkinson: Yeah, that seems reasonable. Happy for this to be resolved

Library/src/Multiplayer/MCS/MCSTypes.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/710/commits/11744d161a10ecaef3f9c01c28ae58b6c3a7b416

PR #712, opened 2025-06-12
[OF-1672] feat: Break Multiplayer Dependencies

MAG-mv:
This is the first PR for this ticket, which contains the following refactors:

- Removed system dependency from PortalSpaceComponent. PortalSpaceComponent will be part of the RealtimeEngine module, and systems code will be part of core. This was done by removing GetSpaceThumbnail from PortalSpcaeComponent.

- Removed SpaceSystem dependency from SpaceEntitySystem. SpaceEntitySystem will be part of the RealtimeEngine module, and SpaceSystem will be part of core.

- Removed SequenceSystem dependency from SpaceEntitySystem. SpaceEntitySystem will be part of the RealtimeEngine module, and SequenceSystem will be part of core.

- Moved CallHelpers to Common, as this will be part of the Shared library, as is used in the RealtimeEngine module and core.

- Removed Asset header from MultiplayerConnection. MultiplayerConnection is part of RealtimeEngine, while Asset is part of core.

- Moved the majority of the web folder to Common. Web code will be shared among core and RealtimeEngine for REST and Multiplayer requests.

- Removed static httpAuth data from WebClient. This is not injected when connectiong to multiplayer


Library/include/CSP/Multiplayer/Components/PortalSpaceComponent.h
```diff
@@ -94,12 +93,6 @@ class CSP_API PortalSpaceComponent : public ComponentBase, public IEnableableCom
     /// @copydoc IEnableableComponent::SetIsEnabled()
     void SetIsEnabled(bool InValue) override;
     /// @}
-
-    /// @brief Retrieves the space thumbnail information associated with the space.
-    /// If the space does not have a thumbnail associated with it the result callback will be successful, the
-    /// HTTP res code will be ResponseNotFound and the Uri field inside the UriResult will be empty.
-    /// @param Callback UriResultCallback : callback when asynchronous task finishes
-    CSP_ASYNC_RESULT void GetSpaceThumbnail(csp::systems::UriResultCallback Callback) const;
```
MAG-ElliotMorris, line 102: [Comment] Please make sure upon merge to explicitly list this, and any other removed public API, in the commit message. We're not great at this, but we must document this somewhere, at least in the PR description if no-where else so it's theoretically traceable.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include "CSP/Multiplayer/Components/ConversationSpaceComponent.h"
 
-#include "CallHelpers.h"
+#include "Common/CallHelpers.h"
```
MAG-ElliotMorris: [Consider] Is this really worth it? `INVOKE_IF_NOT_NULL` is the only thing that's in there and it's pretty naff, I'd recommend just breaking this dependency by using regular `if` statements and letting this thing wither on the vine.

... I do understand the desire to do this if there are really quite a lot of places to do the substitution ... although bet you could completely do it in one pass with a quick script. VS has a button to expand in place for macros even...

Library/src/Common/Web/GraphQLApi/GraphQLApi.cpp
```diff
@@ -13,12 +13,12 @@
...
-#include "Web/GraphQLApi/GraphQLApi.h"
+#include "Common/Web/GraphQLApi/GraphQLApi.h"
 
 #include "CSP/CSPFoundation.h"
-#include "Web/HttpAuth.h"
-#include "Web/HttpPayload.h"
-#include "Web/WebClient.h"
+#include "Common/Web/HttpAuth.h"
+#include "Common/Web/HttpPayload.h"
+#include "Common/Web/WebClient.h"
```
MAG-ElliotMorris: Using this block for a general comment as it has a lot of includes...

[fix] Does the graphQL api really have a justification to be in common. I'm suspecting you've moved too much over here. Why would HttpPayload need to exist in `RealtimeEngine`, etc.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
```diff
@@ -16,8 +16,8 @@
 #include "POCOSignalRClient.h"
 
 #include "CSP/CSPFoundation.h"
+#include "Common/Web/HttpAuth.h"
```
MAG-ElliotMorris: [fix] This seems like too much, all it wants is an access token in this file, which it gets statically. I think we should avoid static data access if at all possible. The AuthDto, or even just a string, would serve just as well as an injection.

Not going to check every file, i'd be grateful if you could make a pass and see if you _really_ need to bring over entire types into common, or if you can extract just the foundational elements of communication, like I have with the enums in my other PR. (A good technique is to remove the include and see what sort of red squigglies you get)

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-ElliotMorris: You also need to think about how this will work when this is a separate library. You won't have access to the `src` folder from other modules, so shared stuff must be in `include`.

Library/src/Common/Web/GraphQLApi/GraphQLApi.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/712/commits/f63f52b1960086f2c4e499cf9b46514f5ac2a751 and https://github.com/magnopus-opensource/connected-spaces-platform/pull/712/commits/f48a9238f28d239c9400091d7036da9222ce209c

HttpPayload is needed in shared though, as it's used by the SignalRClient (We use HTTP for our SignalR connection)

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/712/commits/ce110ea9b5d8a51a8c9ae9f6ebeb180dce482dc4

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/712/commits/85642691c8ea0468b1307036dd6f8947ffb36ca1

Library/src/Common/Web/GraphQLApi/GraphQLApi.cpp
MAG-ElliotMorris: A shame about HttpPayload. I don't conceptually understand why the SignalRClient needs to do HTTP stuff if we are skipping negotiation, but the library certainly thinks it does, so 🤷 .

If it's just this tiny little thing it needs to do, in the future we might consider just duplicating some minor HTTP functionality into the module to avoid this annoying dependency muddle.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-ElliotMorris: Feels better in general. 👍

MAG-ElliotMorris (approved):

PR #713, opened 2025-06-12
Revert "Prevent referencing native nullptrs"

magnopus-swifty:
Reverts magnopus-opensource/connected-spaces-platform#702

While we agree this is a good change, it breaks some behaviour in Array<> that relies on being able to create null instances.

See [this conversation on Slack](https://magnopus.slack.com/archives/C02TKEGRJE6/p1749727816407889).


MAG-OR (approved):

MAG-ElliotMorris (approved):
Looks like I was right to be afraid :(

MAG-ElliotMorris:
Going to go ahead and merge this so we can get moving on it.

PR #716, opened 2025-06-16
[OF-1672] refac!: Remove SystemsManager dependency based on getting the LogSystem from multiplayer.

MAG-ElliotMorris:
> [!IMPORTANT]  
> Breaking changes.
>
> Several types now take a `LogSystem` in their construction, which can be fetched from `SystemsManager`.
> - All Components
> - EntityScript
> - SpaceEntity
>
> For the above types, this is an optional argument, you may pass `nullptr` to these to disable their logging functionality.

Remove all dependencies that deliver `LogSystem` to the multiplayer module.

Some `Debug/Logging` includes remain due to the annoying profiling macros being contained there too, we'll have to either disable that functionality for the realtimeengine module or do the work to port it.

What to expect in this PR
- All components gain an optional `LogSystem` member via injection.
- `SpaceEntities` gain an optional `LogSystem` member via injection
- `SystemBase`, and thus all systems, gain a mandatory `LogSystem` member via injection
- Various other elements of the codebase gain a `LogSystem` member via injection, where there is no public constructors in play, this is expressed when possible via a non-nullable reference, otherwise the non-nullability is only weakly captured via code comment
- Due to unravelling this ball of mud, some transient includes are now manifest and required to be directly included, they are placed with a note that the dependency needs breaking.
- Coincidentally fixed us not deleting the conversation system

No new tests are created, as this is a pure refactor. The most risky change is the change to `ScriptSystem::Log`, as it becomes a capturing lambda now so as not to abuse static data. However, this is quite well tested so I'm not worried.

> [!NOTE]  
> You'll see some churn around moving to and from reference semantics in this PR. This is intentionally left in as I wish to tell the story of how this change _should_ have been made. The final commit changes back to unsafe pointer semantics, (in unavoidable places ... which is a lot) due to the wrapper generator.
>
> If you're looking for a root, holistic cause of why CSP feels unstable ... this sort of thing is a good candidate.

> [!TIP]
> You'll note Components and SpaceEntities gain only optional dependencies on the LogSystem here. (Conceptual optionality ... we can't formally express mandatory invariants due to the wrapper generator reasons above...)
> You are probably asking why I made this choice. It's based on vibe, in that these types are public types used by clients to do things, and I felt they could do with being as flexible as possible, even if that means not logging. All log calls are guarded in these types now, such that if constructed without a logger, they will simply not log. This is a more feeling-based decision than usual, so call out if you think it's wrong.
>
> It probably doesn't manifest currently, as all our construction paths will provide a LogSystem, but flexibility is a temporal quality.


Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -151,7 +156,9 @@ class CSP_API ComponentBase
 
 protected:
     ComponentBase();
-    ComponentBase(ComponentType Type, SpaceEntity* Parent);
+
+    // The LogSystem input may be null, components do not _have_ to log.
+    ComponentBase(ComponentType Type, csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
```
MAG-OR, line 161: [question] out of pure curiosity, why choose to put the log system as the first argument rather than the last?

Library/src/Multiplayer/Components/CustomSpaceComponent.cpp
```diff
@@ -15,17 +15,17 @@
...
 #include <algorithm>
 #include <functional>
 
 namespace csp::multiplayer
 {
 
-CustomSpaceComponent::CustomSpaceComponent(SpaceEntity* Parent)
-    : ComponentBase(ComponentType::Custom, Parent) //, NumProperties(2)
+CustomSpaceComponent::CustomSpaceComponent(csp::common::LogSystem* LogSystem, SpaceEntity* Parent)
+    : ComponentBase(ComponentType::Custom, LogSystem, Parent) //, NumProperties(2)
```
MAG-OR, line 28: [comment] I know you didn't write it, but I am rather unimpressed by the dangling commented out bit of code. (//, NumProperties(2)) (no need to fix this right now)

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -15,36 +15,42 @@
...
 #include "CSP/Multiplayer/Components/ConversationSpaceComponent.h"
+#include "CSP/Common/Systems/Log/LogSystem.h"
 
 #include "CallHelpers.h"
-#include "Debug/Logging.h"
 #include "Multiplayer/Script/ComponentBinding/ConversationSpaceComponentScriptInterface.h"
 #include "Systems/Conversation/ConversationSystemInternal.h"
 #include "Systems/ResultHelpers.h"
 
+// Needs broken
```
MAG-OR, line 25: [fix], sorry, eventually I really _have_ to say something. same for the retro board, seriously. it's "needs + verb-ing", not "needs + past participle". the only way "needs ... + past participle" is correct is like this: "needs to be + past participle". needs breaking. needs improving. or needs improvement. or needs to be improved. not needs improved.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
```diff
@@ -65,8 +71,13 @@ void ConversationSpaceComponent::CreateConversation(const csp::common::String& M
 
     if (!ConversationId.IsEmpty())
     {
-        CSP_LOG_ERROR_MSG("This component does not have an associated conversation. "
-                          "Call CreateConversation to create a new conversation for this component");
+        if (LogSystem != nullptr)
+        {
+            // Review Q, this error is weird ... we're already calling CreateConversation...?
+            LogSystem->LogMsg(csp::common::LogLevel::Error,
```
MAG-OR: [comment] the way I understand this, we attempt to create a conversation, but it comes back with a non-empty conversation ID, which means it already exists? should the error message be instead "error, conversation already exists. delete it before trying to re-create it"? @MAG-mv ?

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -712,81 +712,83 @@ ComponentBase* SpaceEntity::InstantiateComponent(uint16_t InstantiateId, Compone
...
         break;
     case ComponentType::Hotspot:
-        Component = new HotspotSpaceComponent(this);
+        Component = new HotspotSpaceComponent(LogSystem, this);
         break;
     default:
     {
-        CSP_LOG_MSG(csp::common::LogLevel::Warning,
-            csp::common::StringFormat("Unknown Component type of value: %d", static_cast<uint32_t>(InstantiateType)));
+        CSP_LOG_MSG(
```
MAG-OR: [question] did you mean to keep this macro here? as far as I can tell, you got rid of all the error ones. sorry if you already justified this somewhere else

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -350,8 +350,8 @@ class CSP_API AssetSystem : public SystemBase
     CSP_NO_EXPORT void OnEvent(const std::vector<signalr::value>& EventValues) override;
 
 private:
-    AssetSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
-    CSP_NO_EXPORT AssetSystem(csp::web::WebClient* InWebClient, csp::multiplayer::EventBus* InEventBus);
+    AssetSystem(csp::common::LogSystem& LogSystem); // This constructor is only provided to appease the wrapper generator and should not be used
```
MAG-OR: [comment] the silly comment about the wrapper gen is about the default (no args) constructor. do you not still need one?

Library/src/Systems/SystemsManager.cpp
```diff
@@ -168,6 +168,7 @@ void SystemsManager::DestroySystems()
     // Systems must be shut down in reverse order to CreateSystems() to ensure that any
     // dependencies continue to exist until each system is successfully shut down.
     delete SpaceEntitySystem;
+    delete ConversationSystem;
```
MAG-OR, line 171: [comment] well spotted

Library/include/CSP/Multiplayer/Script/EntityScript.h
```diff
@@ -148,6 +153,9 @@ class CSP_API EntityScript
     bool HasBinding;
 
     SpaceEntitySystem* SpaceEntitySystemPtr;
+
+    // may be null
```
MAG-OR, line 157: [question] I get the impression that there are places where the log system is allowed to be null, and places where it is not. will the distinction between "places where the log system is allowed to be null" and "places where the log system can never be null" eventually cause confusion and be a source of bugs? (say you're in a place where it's allowed to be null and you forget to add the check for nullness because you were just working in a place that never allows it to be null)
also, what are the use cases for a never-null log system? do we have classes where it causes problems (as opposed to just being inconvenient) not to print logs? 
I don't know if I am making sense. happy to chat.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
```diff
@@ -25,9 +25,13 @@
 #include <Poco/Net/NetException.h>
 #include <Poco/URI.h>
 #include <chrono>
+#include <fmt/format.h>
 #include <stdexcept>
 #include <thread>
 
+// For the profiling stuff ... needs broken, or just removed.
```
MAG-OR, line 32: [question] when is the last time the profiling stuff was used?

ThirdParty/quickjs/include/quickjspp.hpp
```diff
@@ -29,6 +29,11 @@
 	#include <filesystem>
 #endif
 
+//!!!! Absolutely catastrophic, introduces a reverse dependency in a LIBRARY FILE.
```
MAG-OR, line 32: [comment] yikes.

Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -346,7 +346,7 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
     ///@}
 
 private:
-    SpaceSystem(csp::common::LogSystem& LogSystem); // This constructor is only provided to appease the wrapper generator and should not be used
+    SpaceSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
```
MAG-OR, line 349: [comment] ah, yes.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-ElliotMorris: This is British slang, it's intentionally malformed. As in "That needs broken" or "You need your head screwed on right." If you don't, or think others wont, understand what it means we can change it, but if this is just an aesthetic preference then it doesn't matter. 

Keep in mind that if there is requirements for people, including myself, to form ad-hoc comments in proper grammar, they just won't get written.

Library/include/CSP/Multiplayer/Script/EntityScript.h
MAG-ElliotMorris: Do you not feel this is adequately explained in the PR description? I agree with you, it is a massive source of consternation for me that we are unable to communicate this correctly due to the wrapper generator.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-ElliotMorris: No clue. I don't think very much, but we have [documented it](https://connected-spaces-platform.net/manual/debugging/performance_profiling.html) as a full on _thing_.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ElliotMorris: You probably meant [question].

You do, see the PR description on how I've left this journey in the PR for storytelling purposes. Maybe that was a bad idea.

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-ElliotMorris: It was out of pure instinct. I made no conscious decision here.

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-OR: yeah I saw you reverted it in a later commit. yes you did say it in the description.

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-mv: [Comment] I would have put it at the end, as Parent seems more important for the SpaceEntity to function than logging. Not a massive issue, though

Library/src/Multiplayer/SpaceEntity.cpp
MAG-ElliotMorris: This is addressed in the very next commit. I admit the boundary between them is a little fuzzy, the first commit is supposed to be about injecting Logsystem into components, which included changes to SpaceEntity as it constructs components. However, it does not include updating the macros yet, that comes later.

Library/include/CSP/Multiplayer/Script/EntityScript.h
MAG-ElliotMorris: > also, what are the use cases for a never-null log system? do we have classes where it causes problems (as opposed to just being inconvenient) not to print logs?

I didn't answer this. This is not about behaviour as much as it is about type invariants. Knowing something is never going to be null means you can access it without checks, and never need to worry about invalid memory access causing crashes. They're very good things, from a systems design perspective, you want as many invariants on your types as possible so you can make assumptions safely. 

Not to mention it prevents you from having to document, communicate and test the now unrepresentable states of your application.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-mv: Yep. This logs logic is flipped.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1103,7 +1175,7 @@ mcs::ObjectPatch SpaceEntity::CreateObjectPatch()
     }
 
     // 3. Handle any component deletions
-    ComponentBase DeletionComponent(ComponentType::Delete, const_cast<SpaceEntity*>(this));
+    ComponentBase DeletionComponent(ComponentType::Delete, csp::systems::SystemsManager::Get().GetLogSystem(), const_cast<SpaceEntity*>(this));
```
MAG-mv: [fix] This should be the SpaceEntity's log system

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -820,7 +820,7 @@ void ConversationSystemInternal::GetConversationAnnotation(const csp::common::St
         .then(systems::continuations::SendResult(Callback, "Successfully retrieved annotation."))
         .then(common::continuations::InvokeIfExceptionInChain([Callback](const std::exception& /*exception*/)
             { Callback(MakeInvalid<multiplayer::AnnotationResult>()); },
-            csp::systems::SystemsManager::Get().GetLogSystem()));
+            *csp::systems::SystemsManager::Get().GetLogSystem()));
```
MAG-mv: [question] Should these be using the member LogSystem?

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -533,7 +534,7 @@ void EntityScriptBinding::Bind(int64_t ContextId, csp::systems::ScriptSystem* Sc
     qjs::Context* Context = (qjs::Context*)ScriptSystem->GetContext(ContextId);
     qjs::Context::Module* Module = (qjs::Context::Module*)ScriptSystem->GetModule(ContextId, csp::systems::SCRIPT_NAMESPACE);
 
-    Module->function<&EntityScriptLog>("Log");
+    Module->function("Log", [&LogSystem = this->LogSystem](qjs::rest<std::string> Args) { EntityScriptLog(std::move(Args), LogSystem); });
```
MAG-mv, line 540: [Comment] It's worth verifying that logs still work in scripts! I don't imagine we're testing this...

MAG-ElliotMorris: Oh, you're right! We're doing a bunch of `CSP.Log` in our test scripts themselves, but never actually validating that they log. 🥹 

[fix] to add that, it's serious. I'd assumed that because we were calling it, we were validating.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ElliotMorris: This type needs to be split in two fully, ... however, I guess yes tbh there's no reason not to, it's a systembase inheritor. Will [fix]

... I'd actually forgotten about this when we decided for me to carry on with module splitting. Might need to grab you back for a few hours to pair on splitting this.

Library/include/CSP/Multiplayer/ComponentBase.h
MAG-ElliotMorris: Given that neither of you have marked this as worth [consider]ing, I'm resolving this with no action because I don't want to go to the effort to update all the signatures.

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/716/commits/fa2f1835a6768fe1fde0bea8160cc73adb381ab4

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/716/commits/f08b7c6b662e3ad460ac6762010160c77ba16331

Library/src/Multiplayer/SpaceEntity.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/716/commits/638b3d77f1fb0cea3ba1cd9a19438c806bf89c8e

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-ElliotMorris: Made another PR to deal with this. Removed the review comment :https://github.com/magnopus-opensource/connected-spaces-platform/pull/716/commits/b8916062a9392b2664669ff9d544266692fc7663

MAG-ElliotMorris:
Note to self, this is a breaking change, at least due to the additional `LogSystem` params on the components. TODO mark it breaking and list what the breakages are before merge.

Library/src/Multiplayer/Components/ConversationSpaceComponent.cpp
MAG-OR: I am not sure that "intentionally malformed slang" has its place in a professional setting.
Also, more generally, we are open-source and therefore documentation and comments need to be understood by a multinational audience, therefore avoiding local colloquialisms is probably best, especially when they are objective mistakes/malformed. Apparently, this faulty construction comes from Scotland and has spread to some parts of the UK close to Scotland, and to some areas of the US that have lots of people of Scottish origins (says a Yale linguistics study about this expression that I found), but remains widely unaccepted elsewhere, i.e. it is quite localized and will likely raise eyebrows more often than not.

These comments, as I understand, are short-lived. They are  pretty much short-term todos, right? They will disappear eventually, so you don't need to change them.

MAG-mv (approved):

PR #717, opened 2025-06-16
[NT-0] experiment: Do we need the http client

MAG-ElliotMorris:
Testing a bit of lost knowledge. I'm not sure this is neccesary. If there's no obvious problems, can put it on my list of things to investigate and change.


MAG-ElliotMorris:
Looks like we do! The signalR library itself throws if this isn't set. Want to understand why.

PR #718, opened 2025-06-16
[OF-1671] feat!: promote restful service api uri version information to csp foundation

MAG-ChristopherAtkinson:
This is the CSP portion of the [OF-1671] Promote RESTful Service API URI version information to CSPFoundation work.

The PR contains the following changes:

 - Introduce a new MCSServiceDefinition (Public API break)
 - Updated usage and reference to RESTful service URI. 
 -  Updated CSP to reference the new generated function names (version agnostic csp services generator files)

> [!IMPORTANT]
>
> I have updated the csp services submodule to reference the branch with the required changes. These changes should not be merged until both PR's are approved.

> [!WARNING]
>
> These changes do apply a break in the current public API, As a result, we should get some input from client teams to ensure that the transition is smooth ahead of the next release.

### DIFF
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -18,20 +18,58 @@
 #include "CSP/CSPCommon.h"
 #include "CSP/Common/String.h"
 
+#include <string>
+
 namespace csp
 {
+
+class CSP_API MCSServiceDefinition
+{
+public:
+    MCSServiceDefinition()
+        : URI("")
+        , Version(0)
+    {
+    }
+
+    MCSServiceDefinition(const csp::common::String& InURI, const uint32_t InVersion)
+        : URI(InURI)
+        , Version(InVersion)
+    {
+    }
+
+    /// @brief Gets the root URI for the service endpoint.
+    /// @return csp::common::String : The root URI of the service endpoint.
+    const csp::common::String GetURI() const { return URI; }
+    /// @brief Sets the URI for the service endpoint.
+    /// @param InURI csp::common::String : URI for service endpoint.
+    void SetURI(const csp::common::String& InURI) { URI = InURI; }
+
+    /// @brief Gets the current version of the service endpoint.
+    /// @return int32_t : Representing the current version of the service endpoint.
+    const int32_t GetVersion() const { return Version; }
+    /// @brief Sets the current Version for the service endpoint.
+    /// @param InVersion uint32_t : Version for service endpoint.
+    CSP_NO_EXPORT void SetVersion(const uint32_t InVersion) { Version = InVersion; }
+
+private:
+    csp::common::String URI;
+    uint32_t Version;
+};
 
 /// @brief Holds supported endpoint uris used by Foundation.
 class CSP_API EndpointURIs
 {
 public:
-    csp::common::String UserServiceURI;
-    csp::common::String PrototypeServiceURI;
-    csp::common::String SpatialDataServiceURI;
-    csp::common::String MultiplayerServiceURI;
-    csp::common::String AggregationServiceURI;
-    csp::common::String TrackingServiceURI;
-    csp::common::String MaintenanceWindowURI;
+    MCSServiceDefinition UserService;
+    MCSServiceDefinition PrototypeService;
+    MCSServiceDefinition SpatialDataService;
+    MCSServiceDefinition MultiplayerService;
+    MCSServiceDefinition AggregationService;
+    MCSServiceDefinition TrackingService;
+    MCSServiceDefinition MaintenanceWindow;
+
+private:
 };
 
 /// @brief Holds client data used in requests for all Magnopus Serives.
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -282,12 +282,12 @@ EndpointURIs CSPFoundation::CreateEndpointsFromRoot(const csp::common::String& E
     const std::string MultiplayerServiceURI = TranslateEndpointRootURIToMultiplayerServiceUri(RootURI);
 
     EndpointURIs EndpointsURI;
-    EndpointsURI.UserServiceURI = CSP_TEXT(UserServiceURI.c_str());
-    EndpointsURI.PrototypeServiceURI = CSP_TEXT(PrototypeServiceURI.c_str());
-    EndpointsURI.SpatialDataServiceURI = CSP_TEXT(SpatialDataServiceURI.c_str());
-    EndpointsURI.MultiplayerServiceURI = CSP_TEXT(MultiplayerServiceURI.c_str());
-    EndpointsURI.AggregationServiceURI = CSP_TEXT(AggregationServiceURI.c_str());
-    EndpointsURI.TrackingServiceURI = CSP_TEXT(TrackingServiceURI.c_str());
+    EndpointsURI.UserService = MCSServiceDefinition(CSP_TEXT(UserServiceURI.c_str()), 1U);
+    EndpointsURI.PrototypeService = MCSServiceDefinition(CSP_TEXT(PrototypeServiceURI.c_str()), 1U);
+    EndpointsURI.SpatialDataService = MCSServiceDefinition(CSP_TEXT(SpatialDataServiceURI.c_str()), 1U);
+    EndpointsURI.AggregationService = MCSServiceDefinition(CSP_TEXT(AggregationServiceURI.c_str()), 1U);
+    EndpointsURI.TrackingService = MCSServiceDefinition(CSP_TEXT(TrackingServiceURI.c_str()), 1U);
+    EndpointsURI.MultiplayerService = MCSServiceDefinition(CSP_TEXT(MultiplayerServiceURI.c_str()), 1U);
 
     return EndpointsURI;
 }
--- a/Library/src/Services/ApiBase/ApiBase.h
+++ b/Library/src/Services/ApiBase/ApiBase.h
@@ -218,9 +218,9 @@ using ResponseHandlerPtr = ApiResponseHandlerBase*;
 class ApiBase
 {
 public:
-    ApiBase(csp::web::WebClient* InWebClient, const csp::common::String* InRootUri)
+    ApiBase(csp::web::WebClient* InWebClient, const csp::MCSServiceDefinition* InServiceDefinition)
         : WebClient(InWebClient)
-        , RootUri(InRootUri)
+        , ServiceDefinition(InServiceDefinition)
     {
     }
 
@@ -238,7 +238,7 @@ public:
     }
 
     csp::web::WebClient* WebClient;
-    const csp::common::String* RootUri;
+    const MCSServiceDefinition* ServiceDefinition;
 };
 
 //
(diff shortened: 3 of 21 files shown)


modules/csp-services
MAG-OR, line 1: [comment] unsure if this is the wrong version to update to, but the CI seems to be upset with this

MAG-ChristopherAtkinson: Yeah, the changes on that submodule have not been regenerated. I think the correct course of action is to commit the newly generated files, of course when both PR are complete we will follow the tag process for a release of services

MAG-AdamThorn: @MAG-ChristopherAtkinson can you be more specific about the course of action, are you expecting to get the build checks for this PR all green as part of this review?

MAG-ChristopherAtkinson: Based on the discussion in stand-up, I am committing the generated code to the related csp service branch and updating the reference here; as a result this branch will adopt the new generated api

Library/include/CSP/CSPFoundation.h
```diff
@@ -18,20 +18,58 @@
...
+
+    MCSServiceDefinition(const csp::common::String& InURI, const uint32_t InVersion)
+        : URI(InURI)
+        , Version(InVersion)
+    {
+    }
+
+    /// @brief Gets the root URI for the service endpoint.
+    /// @return csp::common::String : The root URI of the service endpoint.
+    const csp::common::String GetURI() const { return URI; }
```
MAG-ElliotMorris: [fix] Do not return value types as `const`. Just return `csp::common::String`. Same below.

Library/include/CSP/CSPFoundation.h
```diff
@@ -18,20 +18,58 @@
...
+    /// @brief Sets the URI for the service endpoint.
+    /// @param InURI csp::common::String : URI for service endpoint.
```
MAG-ElliotMorris, line 46: [Question] Why does `Get` mention root URI in its comment, but `Set` does not. Is this a different concept?

Library/include/CSP/CSPFoundation.h
```diff
@@ -18,20 +18,58 @@
...
+    void SetURI(const csp::common::String& InURI) { URI = InURI; }
+
+    /// @brief Gets the current version of the service endpoint.
+    /// @return int32_t : Representing the current version of the service endpoint.
+    const int32_t GetVersion() const { return Version; }
+    /// @brief Sets the current Version for the service endpoint.
+    /// @param InVersion uint32_t : Version for service endpoint.
+    CSP_NO_EXPORT void SetVersion(const uint32_t InVersion) { Version = InVersion; }
```
MAG-ElliotMorris, line 54: [Consider] I see why you've wanted to not export this setter, but consider whether it actually achieves anything. Users can still construct these types with whatever version they want. Consider if we should even have a setter at all.

Library/src/CSPFoundation.cpp
```diff
@@ -282,12 +282,12 @@ EndpointURIs CSPFoundation::CreateEndpointsFromRoot(const csp::common::String& E
...
-    EndpointsURI.SpatialDataServiceURI = CSP_TEXT(SpatialDataServiceURI.c_str());
-    EndpointsURI.MultiplayerServiceURI = CSP_TEXT(MultiplayerServiceURI.c_str());
-    EndpointsURI.AggregationServiceURI = CSP_TEXT(AggregationServiceURI.c_str());
-    EndpointsURI.TrackingServiceURI = CSP_TEXT(TrackingServiceURI.c_str());
+    EndpointsURI.UserService = MCSServiceDefinition(CSP_TEXT(UserServiceURI.c_str()), 1U);
+    EndpointsURI.PrototypeService = MCSServiceDefinition(CSP_TEXT(PrototypeServiceURI.c_str()), 1U);
+    EndpointsURI.SpatialDataService = MCSServiceDefinition(CSP_TEXT(SpatialDataServiceURI.c_str()), 1U);
+    EndpointsURI.AggregationService = MCSServiceDefinition(CSP_TEXT(AggregationServiceURI.c_str()), 1U);
+    EndpointsURI.TrackingService = MCSServiceDefinition(CSP_TEXT(TrackingServiceURI.c_str()), 1U);
+    EndpointsURI.MultiplayerService = MCSServiceDefinition(CSP_TEXT(MultiplayerServiceURI.c_str()), 1U);
```
MAG-ElliotMorris: [Question] Are you planning to evolve this in the future. I agree this is a good step for this review, but I want to make sure we're planning to extract/promote this mapping of expected service versions just a tad, even just to a `std::unordered_map` listing at the top of the file.

Library/src/Services/ApiBase/ApiBase.h
```diff
@@ -238,7 +238,7 @@ class ApiBase
     }
 
     csp::web::WebClient* WebClient;
-    const csp::common::String* RootUri;
+    const MCSServiceDefinition* ServiceDefinition;
```
MAG-ElliotMorris: [Consider] I know this is just pattern following, but consider if this really needs to be a potentially nullable pointer. Could you get away with a reference here?

MAG-ElliotMorris (changes requested):
A nice isolated change.

[Question] This new type, it looks as though it's been designed for clients to use, can you quickly tell me what the intent is with that? I'm struggling to see what requirement in the PRD implies the need for public control of this. Maybe it's so clients can query what versions of the services are in play? ... that makes sense.

[Fix] Can you please list exactly what API breaks in the PR description, I may be blind but I can't spot anything that is a breaking change in this PR. Clients will need to know when they update as well.

Library/include/CSP/CSPFoundation.h
```diff
@@ -18,20 +18,58 @@
 #include "CSP/CSPCommon.h"
 #include "CSP/Common/String.h"
 
+#include <string>
+
 namespace csp
 {
 
+class CSP_API MCSServiceDefinition
```
MAG-ElliotMorris: [fix] Needs docstrings if it's going to be a public type (and anyway tbh)

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Amended the doc in [5b74637](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/5b74637b81a80d46ab4a60f3a1b76c373366f37b), this was a result of error and shouldn't have been included

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Added a brief that should have been included initially [5b74637](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/5b74637b81a80d46ab4a60f3a1b76c373366f37b)

Library/src/Services/ApiBase/ApiBase.h
MAG-ChristopherAtkinson: Updated this to reference, here: [b1e8962](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/b1e8962331932dd9679d9399d9416f59938e594e)

the change initially was minimal but good opportunity for a quick improvement

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Yeah, couldn't agree more, just didn't really think about it when I was writing the change initially: [2ec7b8f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/2ec7b8f649993f300d48688d634a9cb002c62ba9)

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: This was initially added for completeness of a full API, the current set is a minimal requirement of the larger idea; moving toward an atomic definition of services. 

I would prefer to leave this as is for the time being, however, I agree this does not offer any additional value

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: Yes, the plan is that when we adopt additional changes for providing status information to clients, we can expand this further; however, I wanted to keep the changes minimal to prevent additional scope.

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: [fix] Typo: Serives > Services

MAG-ChristopherAtkinson: Apologies, fixed here: [f43178d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/f43178d3c9e415eb5b201ae19dbab1f6ef3c4bee)

MAG-ElliotMorris: [Fix] You can't mention magnopus services, it needs to be abstract. ~~You were going to change the name of this type as well weren't you?~~

[Consider] That you haven't really said anything in this docstring You've invented a new term "endpoint description" which is a synonym of the type name "ServiceDefinition." The interesting thing about this type, the fact that it ties a version number to a URL, is not mentioned.

Library/src/CSPFoundation.cpp
MAG-ElliotMorris: This needs to be the plan for P0. It's don't think it's a good idea to leave the source of truth of our version expectations embedded in some function.

Library/include/CSP/CSPFoundation.h
MAG-ElliotMorris: I think that's weak reasoning, extra functionality for completeness's sake is an antipattern. Nonetheless, your call.

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: The name has been changed here: [f8c2b87](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/f8c2b87984dbaf655ae456830eb98feb88ee3a74) forgot to add the commit hash

MAG-ChristopherAtkinson: As for the brief, I will update

MAG-ElliotMorris: 👍 (Annoyingly, cant actually leave a request-changes review formally without having a new comment, so glad that you've seen this will leave my state as is)

MAG-ChristopherAtkinson: How about "Represents atomic definition for identifying and versioning an external service endpoint."

MAG-ElliotMorris: Yeah, that's better, contains the idea that versioning is involved. Though i'd drop the "atomic", what are you trying to get at with that? Is there a better word that dosen't imply that this is thread safe?

MAG-ChristopherAtkinson: Updated to the above string with atomic removed: [fd0e1d8](https://github.com/magnopus-opensource/connected-spaces-platform/pull/718/commits/fd0e1d89a28ade61566c3ad2a08b29b694735b91)

MAG-ElliotMorris (approved):

MAG-AdamThorn (approved):

PR #720, opened 2025-06-17
[OF-1686] feat!: Avatar Space visibility

MAG-AdamThorn:
Added support for visibility control to the AvatarSpaceComponent.

- Implemented the IVisibleComponent interface in the AvatarSpaceComponent.
- Updated the Avatar component script interface to add script support for the new IsVisible and IsARVisible properties.
- Updated Entity script bindings to include new visibility properties.
- Updated the SpaceEntitySystem Avatar creation methods to take an initial Avatar visibility argument.
- Added tests to cover the new visibility property for the Avatar component.

Only 3 commits but probably easier to review them individually as the last commit is just concerned with minor improvements to the existing tests that required updates as part of this work.

> [!NOTE]
> There were no existing tests for the Avatar Space Component so I have added two, one to test Avatar property updating and the other to test the Avatar Script interface. I had intended to add a third to test to ensure that property updates persist between Spaces, but of course Avatars are transient and so this was removed. 

> [!IMPORTANT]
> While do this work I noticed that one of the Avatar components existing properties `LocomotionModel`  had not been exposed via the script interface, so I added that while I was in the area.

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntitySystem.h
@@ -93,12 +93,13 @@ public:
     /// @brief Creates a SpaceEntity with type Avatar, and relevant components and default states as specified.
     /// @param InName csp::common::String : The name to give the new SpaceEntity.
     /// @param InSpaceTransform SpaceTransform : The initial transform to set the SpaceEntity to.
+    /// @param IsVisible bool : The initial visibility of the Avatar.
     /// @param InState AvatarState : The initial Avatar State to set.
     /// @param InAvatarId csp::common::String : The Initial AvatarID to set.
     /// @param InAvatarPlayMode AvatarPlayMode : The Initial AvatarPlayMode to set.
     /// @param Callback EntityCreatedCallback A callback that executes when the creation is complete,
     /// which contains a pointer to the new SpaceEntity so that it can be used on the local client.
-    CSP_ASYNC_RESULT void CreateAvatar(const csp::common::String& InName, const SpaceTransform& InSpaceTransform, AvatarState InState,
+    CSP_ASYNC_RESULT void CreateAvatar(const csp::common::String& InName, const SpaceTransform& InSpaceTransform, bool IsVisible, AvatarState InState,
         const csp::common::String& InAvatarId, AvatarPlayMode InAvatarPlayMode, EntityCreatedCallback Callback);
 
     /// @brief Creates a SpaceEntity of type Object, and relevant default values.
@@ -416,9 +417,9 @@ private:
     CSP_START_IGNORE
     async::shared_task<uint64_t> RemoteGenerateNewAvatarId();
     std::function<async::task<std::tuple<signalr::value, std::exception_ptr>>(uint64_t)> SendNewAvatarObjectMessage(const csp::common::String& Name,
-        const SpaceTransform& Transform, const csp::common::String& AvatarId, AvatarState AvatarState, AvatarPlayMode AvatarPlayMode);
+        const SpaceTransform& Transform, bool IsVisible, const csp::common::String& AvatarId, AvatarState AvatarState, AvatarPlayMode AvatarPlayMode);
     std::function<void(std::tuple<async::shared_task<uint64_t>, async::task<void>>)> CreateNewLocalAvatar(const csp::common::String& Name,
-        const SpaceTransform& Transform, const csp::common::String& AvatarId, AvatarState AvatarState, AvatarPlayMode AvatarPlayMode,
+        const SpaceTransform& Transform, bool IsVisible, const csp::common::String& AvatarId, AvatarState AvatarState, AvatarPlayMode AvatarPlayMode,
         EntityCreatedCallback Callback);
     CSP_END_IGNORE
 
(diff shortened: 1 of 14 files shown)


Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -93,12 +93,13 @@ class CSP_API SpaceEntitySystem
...
     /// @param InName csp::common::String : The name to give the new SpaceEntity.
     /// @param InSpaceTransform SpaceTransform : The initial transform to set the SpaceEntity to.
+    /// @param IsVisible bool : The initial visibility of the Avatar.
     /// @param InState AvatarState : The initial Avatar State to set.
     /// @param InAvatarId csp::common::String : The Initial AvatarID to set.
     /// @param InAvatarPlayMode AvatarPlayMode : The Initial AvatarPlayMode to set.
     /// @param Callback EntityCreatedCallback A callback that executes when the creation is complete,
     /// which contains a pointer to the new SpaceEntity so that it can be used on the local client.
-    CSP_ASYNC_RESULT void CreateAvatar(const csp::common::String& InName, const SpaceTransform& InSpaceTransform, AvatarState InState,
+    CSP_ASYNC_RESULT void CreateAvatar(const csp::common::String& InName, const SpaceTransform& InSpaceTransform, bool IsVisible, AvatarState InState,
```
MAG-ChristopherAtkinson, line 102: [Question] Should this be marked as a breaking change, as we are modifying the existing Public API?

MAG-AdamThorn: Great catch @MAG-ChristopherAtkinson, yes it 100% should. Let me do that now.

Library/src/Multiplayer/Script/ComponentBinding/AvatarSpaceComponentScriptInterface.cpp
```diff
@@ -42,5 +42,8 @@ DEFINE_SCRIPT_PROPERTY_VEC4(AvatarSpaceComponent, HeadRotation);
 DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, float, float, WalkRunBlendPercentage);
 DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, float, float, TorsoTwistAlpha);
 DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::AvatarPlayMode, int32_t, AvatarPlayMode);
+DEFINE_SCRIPT_PROPERTY_TYPE(AvatarSpaceComponent, csp::multiplayer::LocomotionModel, int32_t, LocomotionModel);
```
MAG-ChristopherAtkinson, line 45: [Comment] Thank you for catching cases like these. It seems we have incomplete scripting interfaces. wonder what we would find if we audited them all.

MAG-AdamThorn: Yeah, not sure if it was added later and someone forgot to expose it to scripts.

MAG-AdamThorn:
I had to revert the improvement changes made to the `CreateManyAvatarTest` test as it required a Space to be created with `unlisted` SpaceAttributes.

Fix here: [1df4d5d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/720/commits/1df4d5df16262c1717f3927cc7a752cb0870d115)

MAG-ChristopherAtkinson (approved):
The changes look good to me, glad to see it was a simple revert to resolve the failing test

Tests/src/PublicAPITests/ComponentTests/AvatarComponentTests.cpp
```diff
@@ -0,0 +1,299 @@
...
+    EXPECT_EQ(AvatarComponent->GetTargetHandIKTargetLocation(), csp::common::Vector3::Zero());
+    EXPECT_EQ(AvatarComponent->GetHandRotation(), csp::common::Vector4::Identity());
+    EXPECT_EQ(AvatarComponent->GetHeadRotation(), csp::common::Vector4::Identity());
+    EXPECT_EQ(AvatarComponent->GetWalkRunBlendPercentage(), 0.0f);
+    EXPECT_EQ(AvatarComponent->GetTorsoTwistAlpha(), 0.0f);
+    EXPECT_EQ(AvatarComponent->GetMovementDirection(), csp::common::Vector3::Zero());
+
+    // Set new AvatarSpaceComponent property values
+    constexpr const char* NewAvatarId = "TestAvatarId";
+    AvatarState NewAvatarState = AvatarState::Flying;
```
MAG-mv: [Nit] These could all be const

Tests/src/InternalTests/SpaceEntityTests.cpp
```diff
@@ -83,21 +83,14 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceEntityTests, UpdateSpaceEntityGlobalPositionTest
     auto* SpaceSystem = SystemsManager.GetSpaceSystem();
     auto* EntitySystem = SystemsManager.GetSpaceEntitySystem();
 
-    const char* TestSpaceName = "OLY-UNITTEST-SPACE-REWIND";
```
MAG-mv, line 85: [Comment] Thanks for updating all of these!

MAG-mv:
Looks good. I gave one [Nit]. However, I'm still approving.

Tests/src/PublicAPITests/ComponentTests/AvatarComponentTests.cpp
MAG-AdamThorn: Agreed.

Fixed here: [ed95bb7](https://github.com/magnopus-opensource/connected-spaces-platform/pull/720/commits/ed95bb750301b439b27c42fffc976ce761fb04dd)

MAG-ChristopherAtkinson (approved):

PR #724, opened 2025-06-20
[OF-1672] refac!: Break all ScriptSystem dependencies in RealtimeEngine module

MAG-ElliotMorris:
Decouple dependencies on the ScriptSystem inside RealtimeEngine code.

This is achieved via the creation of a IScriptRunner interface. I didn't expect to be using an interface for this, but it became the natural pattern. This interface encapsulates the behaviours needed to interact with a scripting system from the RealtimeEngine. It is not the greatest abstraction, because it captures already existing behaviours that are perhaps not the best, but it will work.

What you can expect in this PR
- Creation of a new interface, and the moving of an existing one
- Lots of injection of that interface to replace direct `ScriptSystem` calls.

> [!IMPORTANT]  
> Due to wrapper generator restrictions (actually conflicts between how our generated Typescript conceptualizes interfaces and how our generated C# does... ), the final commit does an ugly thing and makes the `IScriptRunner` a fake interface, that is technically instantiable, but hostile to real use. This is necessary for this "interface", but not others, because we pass this interface as a pointer across a public api boundary, which we don't do elsewhere. 
>
> This PR also suffers from the usual inability to declare non-nullable invariants for exposed types :(

TODO: Ach, forgot to list specific breakages again!


Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -14,11 +14,15 @@
  * limitations under the License.
  */
 
-#pragma once
-
 #include "CSP/CSPCommon.h"
 #include "CSP/Common/String.h"
 
+CSP_START_IGNORE
+#pragma once
```
MAG-ElliotMorris: [fix] Move pragma once to top...

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -0,0 +1,199 @@
...
+ * limitations under the License.
+ */
+
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/String.h"
+
+CSP_START_IGNORE
+#pragma once
+#pragma warning(push)
+#pragma warning(disable : 4702) // Unreachable code warning
```
MAG-OR: [consider] why are you suppressing these? consider just removing unreachable code

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * @brief Perform the script binding on any bindings registered via RegisterScriptBinding.
+     * @param ContextId int64_t : The Id of the CSP script context in which to run the script.
+     *                  If the provided context does not exist, the script bind will fail.
+     * @return Whether the context was successfully bound.
+     */
+    CSP_NO_EXPORT virtual bool BindContext(int64_t ContextId)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
```
MAG-OR: [consider] making these error messages different from each other for easier traceability

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * @brief Get the script context object.
+     * @param ContextId int64_t : The Id of the CSP script context in which to run the script.
+     * @return The script context object. This specific type of this is implementation defined. However, if using CSP's ScriptSystem at time of
+     * writing, it will be a csp::systems::ScriptContext*, which you should cast to. Returns nullptr if the provided context does not exist.
+     */
+    CSP_NO_EXPORT virtual void* GetContext(int64_t ContextId)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+        return nullptr;
```
MAG-OR: [question] why are you returning after throwing? that's likely the source of your unreachable code warnings

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -0,0 +1,199 @@
...
+     */
+    CSP_NO_EXPORT virtual bool DestroyContext(int64_t ContextId)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+        return false;
+    }
+
+    /**
+     * @brief Set the javascript source code of a particular module
+     * @param ModuleUrl String : The URL of the module. This is an arbitrary accessor. It's may be a good idea to use something like the scenegraph
```
MAG-OR: [fix] typo it's may be -> it may be / it's

Library/include/CSP/Multiplayer/Script/EntityScript.h
```diff
@@ -156,6 +157,9 @@ class CSP_API EntityScript
 
     // may be null
     csp::common::LogSystem* LogSystem = nullptr;
+
+    // may not be null
+    csp::common::IJSScriptRunner* ScriptRunner;
```
MAG-ChristopherAtkinson, line 162: [Nit] would love to have seen this as a reference, but after discussion with @MAG-mv is a limitation of the wrapper generation process.

MAG-ChristopherAtkinson (approved):
Changes look good to me, a nice straight forward refactor. approved.

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/724/commits/19f4d200bc706d7f04ee57d990f0399d13157d01

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/724/commits/72392c1608a9d080ae631f7101f6993d798edff6

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
MAG-mv: Unnecessary imo, as if one of the functions is called, we know it's a problem with the entire class

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/724/commits/ab029b7acf6683ec10b04e82e9ec8c9113b03767

Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/724/commits/ab029b7acf6683ec10b04e82e9ec8c9113b03767

PR #725, opened 2025-06-20
[OF-1687] refac: localize the multiplayer hub functions to a single point of configuration

MAG-ChristopherAtkinson:
The overall changes are a refactor of the current usage of fixed strings within the multiplayer system and should not affect existing behaviour. As a result, the changes can be described as follows:

 - Add MultiplayerHubMethod enum and mapping utility
 - Update all references in SignalR, Logs, and Tests

### DIFF
--- a/Library/docs/source/manual/documentation/multiplayer_state_replication.md
+++ b/Library/docs/source/manual/documentation/multiplayer_state_replication.md
@@ -37,7 +37,7 @@ UpdatedEntity->SerialisePatch(Serialiser);
 auto SerialisedEntity = Serialiser.Finalise();
 
 std::vector<signalr::value> InvokeArguments = {SerialisedEntity};
-Connection->Invoke("SendObjectPatch", InvokeArguments, LocalCallback);
+Connection->Invoke(MultiplayerHubMethodMap()[MultiplayerHubMethod::SEND_OBJECT_PATCH], InvokeArguments, LocalCallback);
 ```
 
 The serializer is set to parse over the `SpaceEntity` data in a specific way in order to format the message correctly. it starts with the Dto metadata of `Id`, `OwnerId`, `Destroy`, `ParentId`.
--- a/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
+++ b/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
@@ -21,6 +21,7 @@
 #include "CSP/Common/String.h"
 #include "CSP/Multiplayer/Conversation/Conversation.h"
 #include "CSP/Multiplayer/EventParameters.h"
+#include "CSP/Multiplayer/MultiplayerHubMethods.h"
 
 #include <atomic>
 #include <functional>
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -311,7 +311,7 @@ auto MultiplayerConnection::RequestClientId()
             ClientIdRequestedEvent->set(Result.as_uinteger());
         };
 
-        LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling GetClientId");
+        LogSystem.LogMsg(csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::GET_CLIENT_ID]).c_str());
 
         Connection->Invoke(MultiplayerHubMethodMap()[MultiplayerHubMethod::GET_CLIENT_ID], signalr::value(signalr::value_type::array), LocalCallback);
         return ClientIdRequestedContinuation;
@@ -637,7 +637,8 @@ CSP_ASYNC_RESULT void MultiplayerConnection::SetAllowSelfMessagingFlag(const boo
         INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
     };
 
-    LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling SetAllowSelfMessaging");
+    LogSystem.LogMsg(
+        csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::SET_ALLOW_SELF_MESSAGING]).c_str());
 
     const std::vector InvokeArguments = { signalr::value(InAllowSelfMessaging) };
     Connection->Invoke(MultiplayerHubMethodMap()[MultiplayerHubMethod::SET_ALLOW_SELF_MESSAGING], InvokeArguments, LocalCallback);
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -345,7 +345,7 @@ std::function<async::task<void>()> MultiplayerConnection::StartListening()
             StartListeningEvent->set();
         };
 
-        LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling StartListening");
+        LogSystem.LogMsg(csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::START_LISTENING]).c_str());
         Connection->Invoke(
             MultiplayerHubMethodMap()[MultiplayerHubMethod::START_LISTENING], signalr::value(signalr::value_type::array), LocalCallback);
 
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -247,8 +247,9 @@ auto MultiplayerConnection::DeleteEntities(uint64_t EntityId) const
             if (Except != nullptr)
             {
                 auto [Error, ExceptionErrorMsg] = ParseMultiplayerErrorFromExceptionPtr(Except);
-                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(
-                    Error, "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \"DeleteObjects\" invocation.")));
+                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(Error,
+                    "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \""
+                        + MultiplayerHubMethodMap()[MultiplayerHubMethod::DELETE_OBJECTS] + "\" invocation.")));
                 return;
             }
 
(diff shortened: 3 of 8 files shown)


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -247,8 +247,9 @@ auto MultiplayerConnection::DeleteEntities(uint64_t EntityId) const
             if (Except != nullptr)
             {
                 auto [Error, ExceptionErrorMsg] = ParseMultiplayerErrorFromExceptionPtr(Except);
-                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(
-                    Error, "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \"DeleteObjects\" invocation.")));
+                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(Error,
+                    "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \""
+                        + MultiplayerHubMethodMap()[MultiplayerHubMethod::DELETE_OBJECTS] + "\" invocation.")));
```
MAG-AdamThorn: [Question] Is there any value in looking up the string in the hub method map here? This is an exception in the DeleteEntities method and we are already hardcoding the name of the method in the response. Same with the rest of the changes in this specific commit.

[Consider] reverting the changes in this [commit](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/6fcd098450f482138f303b0bed08db32852e434d).

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -593,7 +593,7 @@ void MultiplayerConnection::StopListening(ErrorCodeCallbackHandler Callback)
         INVOKE_IF_NOT_NULL(Callback, ErrorCode::None);
     };
 
-    LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling StopListening");
+    LogSystem.LogMsg(csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::STOP_LISTENING]).c_str());
```
MAG-AdamThorn: [Question] Do you think this is required? This is the log for the StopListening method, the value printed here will never change. As a developer I feel it is easier to read this line without having the parse the lookup.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -345,7 +345,7 @@ std::function<async::task<void>()> MultiplayerConnection::StartListening()
             StartListeningEvent->set();
         };
 
-        LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling StartListening");
+        LogSystem.LogMsg(csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::START_LISTENING]).c_str());
```
MAG-AdamThorn: [Question] Do you think this is required? This is the log for the StartListening method, the value printed here will never change. As a developer I feel it is easier to read this line without having the parse the lookup.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -311,7 +311,7 @@ auto MultiplayerConnection::RequestClientId()
             ClientIdRequestedEvent->set(Result.as_uinteger());
         };
 
-        LogSystem.LogMsg(csp::common::LogLevel::Verbose, "Calling GetClientId");
+        LogSystem.LogMsg(csp::common::LogLevel::Verbose, ("Calling " + MultiplayerHubMethodMap()[MultiplayerHubMethod::GET_CLIENT_ID]).c_str());
```
MAG-AdamThorn: [Question] Do you think this is required? This is the log for the RequestClientId method, the value printed here will never change. As a developer I feel it is easier to read this line without having the parse the lookup.

MAG-ChristopherAtkinson:
> @MAG-ChristopherAtkinson I reviewed the commits in sequence which is why my comments above are repeating the same thing. But generally I question the value of using the map method name lookup for all of the logs.

This was added in the event the name changes; however, that would likely not be the case, so I am happy to revert those changes due to them being superfluous

Library/docs/source/manual/documentation/multiplayer_state_replication.md
```diff
@@ -37,7 +37,7 @@ UpdatedEntity->SerialisePatch(Serialiser);
 auto SerialisedEntity = Serialiser.Finalise();
 
 std::vector<signalr::value> InvokeArguments = {SerialisedEntity};
-Connection->Invoke("SendObjectPatch", InvokeArguments, LocalCallback);
+Connection->Invoke(MultiplayerHubMethodMap()[MultiplayerHubMethod::SEND_OBJECT_PATCH], InvokeArguments, LocalCallback);
```
MAG-mv: [Consider] I would have changed this slightly to make it easier for Callers:

Connection->Invoke(MultiplayerHubMethodMap.Get(MultiplayerHubMethod::SEND_OBJECT_PATCH), InvokeArguments, LocalCallback);

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -247,8 +247,9 @@ auto MultiplayerConnection::DeleteEntities(uint64_t EntityId) const
             if (Except != nullptr)
             {
                 auto [Error, ExceptionErrorMsg] = ParseMultiplayerErrorFromExceptionPtr(Except);
-                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(
-                    Error, "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \"DeleteObjects\" invocation.")));
+                EntitiesDeletedEvent->set_exception(std::make_exception_ptr(ErrorCodeException(Error,
+                    "MultiplayerConnection::DeleteEntities, Unexpected error response from SignalR \""
```
MAG-mv: [Question] Are these calls constructing a new MultiplayerHubMethodMap instance each time and populating the values? This seems inefficient.

[Fix] Construct once and add as a member of MultiplayerConnection

Library/docs/source/manual/documentation/multiplayer_state_replication.md
MAG-AdamThorn: Yeah I much prefer that ^ approach.

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: Reverted all of these changes as they are superfluous : [1dde83f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/1dde83f0edaeb6514541eedda0c6da2bf011a5fe)

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: Answered in a general comment below, but the answer is no, the changes are superfluous, and I agree that they do impact readability and is immutable: [1dde83f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/1dde83f0edaeb6514541eedda0c6da2bf011a5fe)

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: same as above

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: same as above

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: Migrate to MultiplayerConnection: [9138021](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/91380214ca44d321c7d896ae8bf31ec5376566c6)

Yes, these would be constructed each time but due to compiler optimization would not incur a performance impact, however makes sense to have on the MultiplayerConnection for when we need to query this during initialization.

Library/docs/source/manual/documentation/multiplayer_state_replication.md
MAG-ChristopherAtkinson: Agreed and added as part of this change set: [9138021](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/91380214ca44d321c7d896ae8bf31ec5376566c6)

Library/src/Multiplayer/NetworkEventManagerImpl.cpp
```diff
@@ -172,7 +172,8 @@ void NetworkEventManagerImpl::SendNetworkEvent(const csp::common::String& EventN
     std::vector<signalr::value> InvokeArguments;
     InvokeArguments.push_back(EventMessage);
 
-    ISignalRConnectionPtr->Invoke(MultiplayerHubMethodMap()[MultiplayerHubMethod::SEND_EVENT_MESSAGE], InvokeArguments, LocalCallback);
+    ISignalRConnectionPtr->Invoke(
+        MultiplayerConnectionInst->GetMultiplayerHubMethodMap().Get(MultiplayerHubMethod::SEND_EVENT_MESSAGE), InvokeArguments, LocalCallback);
```
MAG-mv: [Consider] Change the name of the getter to GetMultiplayerHubMethods

MAG-ChristopherAtkinson: Updated to the suggested naming: [ad20daa](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/ad20daaa28f07f772cd66e8e5c438da8d3c3f507)

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -196,6 +197,10 @@ class CSP_API MultiplayerConnection
     /// @param Callback ErrorCodeCallbackHandler : a callback with failure state.
     CSP_NO_EXPORT void Disconnect(ErrorCodeCallbackHandler Callback);
 
+    /// @brief Getter for the MultiplayerHubMethodMap
+    /// @return MultiplayerHubMethodMap : the MultiplayerHubMethodMap instance
+    CSP_NO_EXPORT MultiplayerHubMethodMap GetMultiplayerHubMethods() { return MultiplayerHubMethods; }
```
MAG-mv: [Nit] could be const

MAG-ChristopherAtkinson: Yep, missed that. Resolved here: [6bd2380](https://github.com/magnopus-opensource/connected-spaces-platform/pull/725/commits/6bd23800b8a6e0efc48f8e7615336869b249f496)

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #730, opened 2025-06-30
[OF-1691] Added CSP Scene description

MAG-mv:
- Updated MCSComponentUnpacker to be able to deserialize integers of any type.
- Added MCSSceneDescription, which the original JSON file is deserialized to
- Added json serializers/deserializers for MCSObjectMessage and ItemComponentData and the new MCSSceneDescription 
- Added csp::SceneDescription, which is the client-facing data structure
- Added tests for json serialization/deserialization
- Added tests for deserializing 3 different files: empty, minimal, and large

We are still waiting for CHS on the following:
- A file that contains Sequences
- A file that has ParentId
- A file that has OwnerId


Library/src/Json/JsonSerializer.h
```diff
@@ -240,4 +279,39 @@ template <typename T> inline void JsonDeserializer::DeserializeValue(csp::common
...
 
+template <typename T> inline void JsonDeserializer::DeserializeValue(std::vector<T>& Values) const
+{
+    rapidjson::SizeType Size = ValueStack.top()->Size();
+    Values.resize(Size);
+
+    for (rapidjson::SizeType i = 0; i < Size; ++i)
+    {
+        auto JsonValue = &(*ValueStack.top())[i];
+        ValueStack.push(JsonValue);
```
MAG-ElliotMorris: [Question] Confused about what this is doing. It looks like you're taking a a pointer to the i'th element in the stack and then pushing a copy of it to the top of the stack. Before popping it back off below. Why?

Library/src/Json/JsonSerializer.h
```diff
@@ -240,4 +279,39 @@ template <typename T> inline void JsonDeserializer::DeserializeValue(csp::common
     }
 }
 
+template <typename T> inline void JsonDeserializer::DeserializeValue(std::vector<T>& Values) const
+{
+    rapidjson::SizeType Size = ValueStack.top()->Size();
```
MAG-ElliotMorris, line 301: [Consider] Won't this be a nullptr exception if `ValueStack` is empty? You may need to guard.

Tests/src/InternalTests/JsonTests.cpp
```diff
@@ -30,6 +30,8 @@ struct TestObjectProps
     float FloatMember = 0.f;
     float DoubleMember = 0.0;
     csp::common::String StringMember = "";
+    std::vector<float> ArrayMember = { 1.1f, 2.2f, 3.3f };
+    std::map<std::string, float> MapMember = { { "Key1", 1.1f }, { "Key2", 2.2f } };
```
MAG-ElliotMorris, line 34: [Consider] We should test empty containers too.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -0,0 +1,244 @@
...
+    Serializer.SerializeMember("components", StringMap);
+}
+
+template <class T>
+void DeserializeComponentDataStringInternal(const csp::json::JsonDeserializer& Deserializer, csp::multiplayer::mcs::ItemComponentDataVariant& OutVal)
+{
+    T Val;
+    Deserializer.SafeDeserializeMember("item", Val);
+    OutVal = Val;
+}
```
MAG-ElliotMorris: [Consider] Wrapping in an inline namespace to enforce that this isn't a method that should be called externally.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+void DeserializeComponentDataString(
```
MAG-ElliotMorris: [nit] Naming is a bit confusing. I understand that all this data comes from strings, but when there can also be actual string types being deserialized... 😕 Can you think of anything better?

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    else if (Type == SingleType)
+    {
+        DeserializeComponentDataStringInternal<float>(Deserializer, OutVal);
+    }
+    else if (Type == SingleArrayType)
+    {
+        DeserializeComponentDataStringInternal<std::vector<float>>(Deserializer, OutVal);
+    }
+    else if (Type == DoubleType)
```
MAG-ElliotMorris: [nit] Do we have the power to rename these? `SingleType` -> `SinglePrecisionFloatType`, etc. Confusing otherwise.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -0,0 +1,244 @@
...
+    else if (Type == StringType)
+    {
+        DeserializeComponentDataStringInternal<std::string>(Deserializer, OutVal);
+    }
+    else if (Type == UInt16DictionaryType)
+    {
+        DeserializeComponentDataStringInternal<std::map<std::string, csp::multiplayer::mcs::ItemComponentData>>(Deserializer, OutVal);
+
+        // We can only deserialize string maps, so we need to convert to int afterwards.
+        const auto& StringMap = std::get<std::map<std::string, csp::multiplayer::mcs::ItemComponentData>>(OutVal);
```
MAG-ElliotMorris: [nit] We use this term "StringMap" around a lot, but it's always a `map<string, ItemComponentData>`. I'd expect StringMap to mean `map<string, string>`. There a better term or framing I wonder?

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -0,0 +1,244 @@
...
+    std::visit(
+        [&Serializer](const auto& ValueType)
+        {
+            std::string TypeString = csp::multiplayer::mcs::GetComponentString(ValueType);
+            Serializer.SerializeMember("$type", TypeString.c_str());
+
+            csp::multiplayer::mcs::SerializeComponentData(Serializer, ValueType);
+        },
+        Obj.GetValue());
+}
```
MAG-ElliotMorris, line 206: [Question] I'm confused about this, but it's because i'm not very good with the visitor pattern and `std::visit` in particular.

Why do we need to use a visit dispatch here, in the lambda, we're dispatching polymorphically anyway, wouldn't it just work without the visit calling those functions directly?

Library/src/Multiplayer/MCS/MCSSceneDescription.h
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+namespace csp::multiplayer::mcs
+{
+
+}
```
MAG-ElliotMorris: [fix] Mistake? Empty file

Library/src/Multiplayer/MCS/MCSTypes.h
```diff
@@ -128,6 +145,9 @@ class ItemComponentData : public ISignalRSerializable, public ISignalRDeserializ
     bool operator==(const ItemComponentData& Other) const;
 
 private:
+    friend void ::ToJson(csp::json::JsonSerializer& Deserializer, const csp::multiplayer::mcs::ItemComponentData& Obj);
+    friend void ::FromJson(const csp::json::JsonDeserializer& Deserializer, csp::multiplayer::mcs::ItemComponentData& Obj);
```
MAG-ElliotMorris: [fix] Am I missing why this is necessary? Can we not make getters rather than giving such radical access?

Tests/src/InternalTests/MCSSceneDescriptionTests.cpp
```diff
@@ -0,0 +1,163 @@
...
+    mcs::ItemComponentData DeserializedValue {};
+    csp::json::JsonDeserializer::Deserialize(SerializedValue, DeserializedValue);
+
+    EXPECT_EQ(DeserializedValue, ComponentValue);
+}
+
+CSP_INTERNAL_TEST(CSPEngine, MCSSceneDescriptionTests, ItemComponentDataSerializeFloatVectorTest)
+{
+    const std::vector<float> TestValue = { 1.1f, 2.2f, 3.3f };
+    mcs::ItemComponentData ComponentValue { TestValue };
```
MAG-ElliotMorris: [fix] For this and all other container types (including strings). We should test at least the empty case, and potentially also the single element case.

Library/include/CSP/Systems/Assets/Asset.h
```diff
@@ -80,6 +86,8 @@ EAssetPlatform ConvertStringToAssetPlatform(const csp::common::String& Platform)
 /// @brief Converts the EAssetPlatform enum into a string value.
 csp::common::String ConvertAssetPlatformToString(EAssetPlatform Platform);
 
+void AssetDetailDtoToAsset(const csp::services::generated::prototypeservice::AssetDetailDto& Dto, csp::systems::Asset& Asset);
```
MAG-ElliotMorris, line 89: [Consider] Do you think these may be better as static functions on the types themselves rather than free functions? They  all look to map 1->1 to a specific type. Free function symbols can become rather disorganised if you use them too liberally without a strategy.

Library/src/Json/JsonSerializer.h
```diff
@@ -170,6 +170,11 @@ class JsonDeserializer
     /// @return bool : Returns true if the key exists in this object.
     bool HasProperty(const char* Key) const { return ValueStack.top()->HasMember(Key); }
 
+    std::string GetMemberAsString(const char* Key) const;
+
+    void EnterMember(const char* Key) const;
+    void ExitMember() const;
```
MAG-ElliotMorris: [fix] Docs. Doesn't have to be full docstrings as these are internal, but something explaining what these do and when to use them because it's not obvious. Something more than just "Exits a member".

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -1,3 +1,4 @@
+/*
```
MAG-ElliotMorris, line 1: [fix] Typo?

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -178,6 +179,39 @@ void DeserializeComponents(const csp::json::JsonDeserializer& Deserializer, std:
         OutComponents = UIntComponents;
     }
 }
+
+}
+
+void ToJson(csp::json::JsonSerializer& /* Deserializer*/, const csp::multiplayer::mcs::SceneDescription& /* Obj*/) { }
```
MAG-ElliotMorris: [Consider] Throwing an exception to 1. Document why there is no `ToJson` implementation and 2. Prevent people who might accidentally call this method expecting it to do something not realizing there's no implementation.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -230,7 +264,7 @@ void FromJson(const csp::json::JsonDeserializer& Deserializer, csp::multiplayer:
     Deserializer.SafeDeserializeMember("prefabId", Obj.Type);
     Deserializer.SafeDeserializeMember("isTransferable", Obj.IsTransferable);
     Deserializer.SafeDeserializeMember("isPersistent", Obj.IsPersistent);
-    Deserializer.SafeDeserializeMember("ownerUserId", Obj.OwnerId);
+    // Deserializer.SafeDeserializeMember("ownerUserId", Obj.OwnerId);
```
MAG-ElliotMorris, line 253: [Question] Why is this removed?

Library/src/Multiplayer/MCS/MCSSceneDescription.h
```diff
@@ -14,9 +14,26 @@
...
 {
+class SceneDescription
+{
+public:
+    csp::services::generated::userservice::GroupDto Group;
+    std::vector<ObjectMessage> Objects;
+    std::vector<csp::services::generated::prototypeservice::PrototypeDto> Prototypes;
+    std::vector<csp::services::generated::prototypeservice::AssetDetailDto> AssetDetails;
+    std::vector<csp::services::generated::aggregationservice::SequenceDto> Sequences;
+};
```
MAG-ElliotMorris: [Comment] I'll have to expand on this later. You should have at least left a quite detailed explainer in the PR description describing how the design changed over time, and telling reviewers that they will need to keep future commits in mind when doing the review.

However, what really should have been done is this initial change should have been rebased away via a --fixup workflow so the reviewer doesn't need to pre-empt the future so much. It makes it harder to review.

Library/include/CSP/CSPSceneDescription.h
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CSP/Multiplayer/SpaceEntity.h"
+#include "CSP/Multiplayer/SpaceEntitySystem.h"
+#include "CSP/Systems/Assets/Asset.h"
+#include "CSP/Systems/Assets/AssetCollection.h"
+#include "CSP/Systems/Sequence/SequenceSystem.h"
+#include "CSP/Systems/Spaces/Space.h"
```
MAG-ElliotMorris: [Consider] Seems like a lot of includes, I think at least some of these can be forward declared.

Library/src/CSPSceneDescription.cpp
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] 2025

Library/include/CSP/CSPSceneDescription.h
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] 2025

Library/src/CSPSceneDescription.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "Multiplayer/MCS/MCSSceneDescription.h"
+#include "Multiplayer/MCS/MCSTypes.h"
+#include "Services/AggregationService/Dto.h"
+#include "Services/PrototypeService/Dto.h"
+#include "Services/UserService/Dto.h"
+
+namespace csp
+{
+SceneDescription::SceneDescription(
+    const csp::multiplayer::mcs::SceneDescription& MCSSceneDescription, csp::multiplayer::SpaceEntitySystem& /* EntitySystem*/)
```
MAG-ElliotMorris: [Question] Maybe this will come up in a later commit. Do we need to SpaceEntitySystem arg?

Library/src/Multiplayer/MCSComponentPacker.h
```diff
@@ -121,12 +116,18 @@ template <typename T> inline void MCSComponentUnpacker::ReadValue(uint16_t Key,
...
-    Value = static_cast<T>(RawEnumValue);
+    std::visit(
+        [&Value](const auto& ValueType)
+        {
+            using VariantType = std::decay_t<decltype(ValueType)>;
+
+            if constexpr (std::is_convertible<VariantType, T>::value)
+            {
+                Value = static_cast<T>(ValueType);
+            }
```
MAG-ElliotMorris: [Question] So in the case that it isn't convertible, the visit entry does nothing? This is likely me not understanding how visit works, but what about the case where there is no convertible type found? Is that fundamentally possible? Is there a way to inform the developer of that error case (like the equivilent of a `default` in a switch?)

... I've never groked the visitor pattern, it always seemed more overhead than a big ol' switch case for not much benefit. Maybe I can learn something here.

Library/src/Multiplayer/MCSComponentPacker.h
```diff
@@ -121,12 +116,18 @@ template <typename T> inline void MCSComponentUnpacker::ReadValue(uint16_t Key,
...
-inline std::enable_if_t<std::is_enum_v<T>> MCSComponentUnpacker::ReadValue(const mcs::ItemComponentData& ComponentData, T& Value) const
+template <typename T> inline void MCSComponentUnpacker::ReadValue(const mcs::ItemComponentData& ComponentData, T& Value)
 {
-    uint64_t RawEnumValue = 0;
-    ReadValue(ComponentData, RawEnumValue);
-    Value = static_cast<T>(RawEnumValue);
+    std::visit(
+        [&Value](const auto& ValueType)
+        {
+            using VariantType = std::decay_t<decltype(ValueType)>;
```
MAG-ElliotMorris: [Consider] some sort of comment about why it's done this way now would be mondo-appreciated. 

The commit message: "Update the way we read integer values to allow us to read values that arent the exact type" hints at _something_, although I've no idea why that is the case or what an integer that isn't the exact type means in this context. However even that little hint is going to get blown away at squash.

We need to fight against losing information, write it down or lose it forever.

Library/src/CSPSceneDescription.cpp
```diff
@@ -24,10 +24,23 @@
...
+    size_t ObjectsIndex = 0;
+    for (const auto& Object : MCSSceneDescription.Objects)
+    {
+        auto* Entity = new multiplayer::SpaceEntity { &EntitySystem };
+        Entity->FromObjectMessage(Object);
+
+        EntitySystem.AddEntity(Entity);
+        Entities[ObjectsIndex] = Entity;
+        ObjectsIndex++;
+    }
```
MAG-ElliotMorris: [Question] Is this the "issue" from the "Fixed issues with scene description" comment? The lack of information about what each commit is doing is making it hard to know if these are bugfixes driven by CI/Test failures, or complete new bits of functionality. I _think_ this is just a piece of implementation that happens to be miscategorised into this commit.

Library/src/CSPSceneDescription.cpp
```diff
@@ -37,13 +50,17 @@ SceneDescription::SceneDescription(
         PrototypesIndex++;
     }
 
+    Assets = csp::common::Array<csp::systems::Asset>(MCSSceneDescription.AssetDetails.size());
```
MAG-ElliotMorris: [fix] This looks like it can be initialized in the initializer list.

Library/src/CSPSceneDescription.cpp
```diff
@@ -37,13 +50,17 @@ SceneDescription::SceneDescription(
...
+
     size_t AssetDetailIndex = 0;
     for (const auto& AssetDetail : MCSSceneDescription.AssetDetails)
     {
         csp::systems::AssetDetailDtoToAsset(AssetDetail, Assets[AssetDetailIndex]);
         AssetDetailIndex++;
     }
 
+    Sequences = csp::common::Array<csp::systems::Sequence>(MCSSceneDescription.Sequences.size());
```
MAG-ElliotMorris: [fix] This looks like it can be initialized in the initializer list.

Library/src/CSPSceneDescription.cpp
```diff
@@ -24,10 +24,23 @@
 namespace csp
 {
 SceneDescription::SceneDescription(
-    const csp::multiplayer::mcs::SceneDescription& MCSSceneDescription, csp::multiplayer::SpaceEntitySystem& /* EntitySystem*/)
+    const csp::multiplayer::mcs::SceneDescription& MCSSceneDescription, csp::multiplayer::SpaceEntitySystem& EntitySystem)
 {
     csp::systems::GroupDtoToSpace(MCSSceneDescription.Group, Space);
 
+    Entities = csp::common::Array<csp::multiplayer::SpaceEntity*>(MCSSceneDescription.Objects.size());
```
MAG-ElliotMorris: [fix] This looks like it can be initialized in the initializer list.

Library/src/CSPSceneDescription.cpp
```diff
@@ -33,7 +33,7 @@ SceneDescription::SceneDescription(
     size_t ObjectsIndex = 0;
     for (const auto& Object : MCSSceneDescription.Objects)
     {
-        auto* Entity = new multiplayer::SpaceEntity { &EntitySystem };
+        auto* Entity = new multiplayer::SpaceEntity(&EntitySystem);
```
MAG-ElliotMorris: [Question] Honestly just curious about what compiler thought this was bad and why. I've always heard curly brace initialization is generally the best choice.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-ElliotMorris: fixed in https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/a3d5b874361755a1ed373c3c39ae4c90991a37ae

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -1,6 +1,5 @@
 /*
-/*
- * Copyright 2025 Magnopus LLC
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] You've changed from 2025->2023 in the changeset i'm looking at right now.

Tests/src/InternalTests/SceneDescriptionTests.cpp
```diff
@@ -35,7 +35,7 @@ CSP_INTERNAL_TEST(CSPEngine, SceneDescriptionTests, ObjectMessageSerializeTest)
     const uint64_t TestType = 2;
     const bool TestIsTransferable = true;
     const bool TestIsPersistent = true;
-    const uint64_t TestOwnerId = 3;
+    const uint64_t TestOwnerId = 0; // TODO: Set to 3 when this is added to the test files.
```
MAG-ElliotMorris, line 40: [Question] Why does it matter if this is in the test files or not? Isn't this merely testing the Object serialization, that behaviour should work shouldn't it?

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -25,6 +25,44 @@ uint64_t MCSComponentUnpacker::GetRuntimeComponentsCount() const
...
+            using Type = std::decay_t<decltype(ValueType)>;
+
+            if constexpr (std::is_same_v<Type, uint64_t>)
+            {
+                Value = static_cast<Type>(ValueType);
+            }
+            else if constexpr (std::is_same_v<Type, int64_t>)
+            {
+                Value = static_cast<Type>(ValueType);
+            }
```
MAG-ElliotMorris: [Question] It's surprising to see handling of int64_t's in a function that takes uint64_t's as an argument. I'm sure there's a reason, but I can't see it encoded anywhere.

Library/src/Multiplayer/MCSComponentPacker.h
```diff
@@ -116,18 +118,12 @@ template <typename T> inline void MCSComponentUnpacker::ReadValue(uint16_t Key,
...
-
-            if constexpr (std::is_convertible<VariantType, T>::value)
-            {
-                Value = static_cast<T>(ValueType);
-            }
-        },
-        ComponentData.GetValue());
+    uint64_t RawEnumValue = 0;
+    ReadValue(ComponentData, RawEnumValue);
+    Value = static_cast<T>(RawEnumValue);
```
MAG-ElliotMorris: [Comment] This is pretty nifty, and I like how the int/uint/enum cases are seperate now. 
[Consider] Do you need to do the value assign? Dosen't the uint version of `ReadValue` do that for you? (sidenote: Death to out params!)

Library/include/CSP/Multiplayer/CSPSceneDescription.h
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] 2025

Library/include/CSP/Multiplayer/CSPSceneDescription.h
```diff
@@ -0,0 +1,32 @@
...
+namespace csp::multiplayer
+{
+
+class CSPSceneDescription
+{
+public:
+    CSPSceneDescription(const csp::common::String& SceneDescriptionJson, csp::multiplayer::SpaceEntitySystem& EntitySystem,
+        csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner);
+
+    csp::common::Array<csp::multiplayer::SpaceEntity*> Entities;
```
MAG-ElliotMorris: [fix] This needs significant documentation, mostly conceptual about what the intent and ideas behind it are, and how it works hand in hand with `CSPSceneData`

Library/include/CSP/Systems/CSPSceneData.h
```diff
@@ -14,29 +14,19 @@
...
-namespace csp
-{
-
-class SceneDescription
+class CSPSceneData
 {
 public:
-    SceneDescription(const csp::multiplayer::mcs::SceneDescription& MCSSceneDescription, csp::multiplayer::SpaceEntitySystem& EntitySystem,
-        csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner);
+    CSPSceneData(const csp::common::String& SceneDescriptionJson);
```
MAG-ElliotMorris: [fix] This needs significant documentation, same comment as above with `CSPSceneDescription`

Library/include/CSP/Systems/MCS/MCSSceneData.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace csp::json
+{
+class JsonSerializer;
+class JsonDeserializer;
+} // namespace csp::json
+
+namespace csp::systems::mcs
+{
+class SceneData
```
MAG-ElliotMorris, line 33: [fix] This needs not-so significant documentation, but still something. Try to mention this concept of "domain types" you use when talking to me sometime it's quite good.

Library/src/Multiplayer/CSPSceneDescription.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] 2025.

Library/src/Systems/CSPSceneData.cpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [fix] 2025

MAG-ElliotMorris (changes requested):
First off, I think the functionality this delivers is excellent. Decent test coverage too. Nice idea making the fake edge-case files in lack of real ones. 

The following critique is similar to what I've given before, and is primarily about PR practice, which _is_ important, but I don't want you to think I'm not glad to see this PR, _I really am._

**PR Feedback**

You need to explain why you're doing things more, both in the PR description and in the commit messages ... as well as in the source code itself really. There is this declarative tone where you say what you've done, but that's not that helpful to a reviewer or to any future reader coming back to this.

This isn't helped by the way this PR has more after-the-fact changes than it does initial implementation. The burden ends up being placed on the reviewer to predict the future and know that you're going to change something. This is _okay_ but not ideal when guidance on this is provided in the description + commit messages, but it's not here.

You should consider a --fixup workflow where you wait for the tests to go green before undrafting the review, doing a force push to make the history clean before exposing reviewers. I suspect many of my comments are redundant because I never knew if the file/line was the final state of the change.

I'm not trying to be super prescriptive, I compromise and fail on this often too. However, I know you are skilled enough to produce joyful reviews, so it's a waste of talent to see them in this state.

Library/include/CSP/Multiplayer/CSPSceneDescription.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/4e79d25e110823f58175fa94eb768d29d89bff6c + https://github.com/magnopus-opensource/connected-spaces-platform/commit/1bbca66a46a7408b75cae3c110a8e45e4a8b9a83

Library/include/CSP/Systems/CSPSceneData.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/4e79d25e110823f58175fa94eb768d29d89bff6c + https://github.com/magnopus-opensource/connected-spaces-platform/commit/1bbca66a46a7408b75cae3c110a8e45e4a8b9a83

Library/include/CSP/Systems/MCS/MCSSceneData.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/4e79d25e110823f58175fa94eb768d29d89bff6c + https://github.com/magnopus-opensource/connected-spaces-platform/commit/1bbca66a46a7408b75cae3c110a8e45e4a8b9a83

Library/src/Systems/CSPSceneData.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/eaf390a045f36a347348dacd84eb109113aa6ec6

Library/src/Multiplayer/CSPSceneDescription.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/eaf390a045f36a347348dacd84eb109113aa6ec6

Library/src/CSPSceneDescription.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/eaf390a045f36a347348dacd84eb109113aa6ec6

Library/src/Multiplayer/MCS/MCSSceneDescription.h
MAG-mv: This was added in a later commit: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/30fc94a7327c496f364174705e6ed29ae48d2cb2

Library/src/Multiplayer/MCS/MCSTypes.h
MAG-mv: These functions are needed to work with the Json serializer. They are called internally and needed to set the internal value

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: Obj.GetValue() returns a variant. There is an overload of GetComponentString and SerializeComponentData for each type that can exist in the variant. We need to dispatch to get the type, so we call the correct overloaded functions.

Library/src/Json/JsonSerializer.h
MAG-mv: When deserializing containers, we need to iterate over elements, add an element over the stack before reading, so any other deserialize calls happen within that object, and then pop after reading. I have added comments: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/bdf5b60e7f7e2a6021169b1a92cb9ddc461e1c2b

Library/src/Json/JsonSerializer.h
MAG-mv: An element is always added to the stack at the root JsonDeserializer::Deserialize

Tests/src/InternalTests/JsonTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/b23eb93c0ca89d2b1b7f6902619e93483fdedac7

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/ebce9841ee9b0144cf6f35b8a87d11d1d99106bc

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: Slightly better name: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/94329775c2b29a0ab3d97d0bc64fb2e2336b1bfc

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/94329775c2b29a0ab3d97d0bc64fb2e2336b1bfc

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: We use this naming convention throughout the SignalR deserialization process, as well. I dont want to deal with this now.

Library/src/Multiplayer/MCS/MCSTypes.h
MAG-mv: I could fix this by using the constructor, but this would mean constructing a new object, and overriding instead of using the object that was passed in.

Tests/src/InternalTests/MCSSceneDescriptionTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/809ebbbcfe99e8ad9ae4be50d2eafa32df90cefc and https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/167877322850383159efeb4035889ee5bb38ce59

Library/include/CSP/Systems/Assets/Asset.h
MAG-mv: Probably. I just moved the existing functions and don't want to change the behaviour for now.

Library/src/Json/JsonSerializer.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/6fe71e4da333635664a29f5a85508928b7708f30

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: Option 3: Remove as it's not needed: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/8b9cb8c5acc04225d3ea264633cf8677d429cccb

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: "ownerUserId" is not what I expected. I was expecting this to be the OwnerId. However, this seems to be a different variable. This is one of the things we are waiting on CHS for, as mentioned in the PR description.

Library/include/CSP/CSPSceneDescription.h
MAG-mv: The full type info is needed in this case because they are not pointers or references, so forward declaration won't work.

Library/include/CSP/CSPSceneDescription.h
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/eaf390a045f36a347348dacd84eb109113aa6ec6

Library/src/CSPSceneDescription.cpp
MAG-mv: This is needed to correctly initialize SpaceEntities.

Library/src/Multiplayer/MCSComponentPacker.h
MAG-mv: I had another look at this as, and ended up using boolean checks, as we are only expecting integer values in this case. The reason these checks exist is to ensure that integer types are cast, as there's a quirk with the way MCS deserializes integers. This has been documented in the MCS types deserialization.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/8494c06de25fd7fcee1025581f87e20bd96ee921

Library/src/Multiplayer/MCSComponentPacker.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/8494c06de25fd7fcee1025581f87e20bd96ee921

Library/src/CSPSceneDescription.cpp
MAG-mv: That's correct. I forgot to add the entity conversion logic.

Library/src/CSPSceneDescription.cpp
MAG-mv: The way we handle the initialization of CSPSceneData and CSPSceneDescription has changed, which makes this redundant. We now handle the deserialization of the MCS object in the constructor and then convert, so we don't know the size until we deserialize.

Library/src/CSPSceneDescription.cpp
MAG-mv: See above comment

Library/src/CSPSceneDescription.cpp
MAG-mv: See above comment

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/eaf390a045f36a347348dacd84eb109113aa6ec6

Library/src/CSPSceneDescription.cpp
MAG-mv: I also think brace initialization is superior. However, I kept getting: 'initializing': cannot convert from 'initializer list' to 'csp::multiplayer::SpaceEntitySystem'

I now seem to be getting it in other places, so I will investigate.

Library/include/CSP/Multiplayer/CSPSceneDescription.h
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/eaf390a045f36a347348dacd84eb109113aa6ec6

Tests/src/InternalTests/SceneDescriptionTests.cpp
MAG-mv: OwnerId is currently commented out in the serialization process, so we can be sure this is working yet. It is set to 0 in this test because it defaults to 0

Library/src/Multiplayer/MCSComponentPacker.cpp
MAG-mv: Added docs here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/aa0483ca3822141021de1e9a9c2984f69e5f9396

Library/src/Multiplayer/MCSComponentPacker.h
MAG-mv: Another way to do it would be to cast the enum to a uint64_t, so it directly assigns the Value. I decided against this a _ have had issues with casting enums in the past, although its probably easy to do.

Also, out params are essential for this pattern, as it allows compile-time function overloading. I went with this pattern because we expect these classes to be updated often, and enforcing these types and compile time prevents people from forgetting to update the necessary functions.

Library/src/Multiplayer/MCS/MCSSceneDescription.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/4e79d25e110823f58175fa94eb768d29d89bff6c + https://github.com/magnopus-opensource/connected-spaces-platform/commit/1bbca66a46a7408b75cae3c110a8e45e4a8b9a83

Library/src/Json/JsonSerializer.h
MAG-ElliotMorris: Ah right I see, bit of a shame that this relies on outer state, but I get it now ty.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-ElliotMorris: The linked commit doesn't fix the comment. Re-opening for visibility in case you made a mistake, but it's a nit so you can re-resolve if you desire.

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-ElliotMorris: Oh right, it's because it's a variant and not a polymorphic type, so you can't just pass it into overloads. Is the reason to choose this pattern strictly performance related, in that variants have union-like storage behaviour? I still struggle to get why visitor is so popular. 🤷

Library/src/Multiplayer/MCS/MCSTypes.h
MAG-ElliotMorris: Could you not fix it by exposing a public `void SetValue(ItemComponentDataVariant Value)`, and have the functions use that? 

Even a pretty gross mutable getter would be preferable imo, `ItemComponentDataVariant& GetValue()`, if you need to work with the inner object extremely directly. Although if considering that, I might even just make the data public at that point. All better than a friend declaration from my pov.

Library/src/Json/JsonSerializer.h
MAG-ElliotMorris: 🙏

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-ElliotMorris: Ah right, it's probably worth raising this mismatch in expectations more directly if you haven't already with CHS.

Library/src/Multiplayer/MCSComponentPacker.h
MAG-ElliotMorris: Looks nicer to me.

As a general rule, I'd say it's a decent idea to leave a note comment on code that is manifesting quirks, even if it is documented at the outer level, even just a "see main docstring for explanation as to why this seems weird.

Library/src/CSPSceneDescription.cpp
MAG-ElliotMorris: I'm presuming this was the issue caused by not having a fully refreshed local changeset? Feel free to resolve this thread yourself if that is the case, non-issue.

Tests/src/InternalTests/SceneDescriptionTests.cpp
MAG-ElliotMorris: Ah right, I see the connection now with the OwnerID expectation issue.

Library/src/Multiplayer/MCSComponentPacker.cpp
MAG-ElliotMorris: Very nice! (I think i made a sanctimonious comment on this code earlier about how i'd love to see exactly this sort of description, so take that me from 5 minutes ago!)

Library/include/CSP/Multiplayer/CSPSceneDescription.h
MAG-ElliotMorris: Good enough for now. If my suspicions about this become a mainline thing are true, I suspect this doc will want to grow. Thanks!

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
MAG-mv: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/b5c3a654fbed29c651184c5364e4f4af3c114d0d

Library/src/Multiplayer/MCS/MCSTypes.h
MAG-mv: Fixed by using constructors: https://github.com/magnopus-opensource/connected-spaces-platform/pull/730/commits/d18b669a13c46c0f9e1c47416fbe0aa4f8cb2245

MAG-ElliotMorris: Thanks. Sidenote, if your initial reticence to do this was around not wanting to double construct, that's the perfect time to implement some move operations, which would make it identical-cost to the prior implementation.

MAG-ElliotMorris (approved):
Nice one. 

The failing builds looks to be teamcity config, not actual failures, so pre-emptively approving. Let us know if it is actually a test failure, and that test failure ends up requiring significant code changes. Otherwise merge when ready.

PR #736, opened 2025-07-09
[NT-0] chore: Replace OLY space/asset names with CSP

MAG-ElliotMorris:
Oly is an old name (Olympus), as is Rewind. Migrate to the new name. Should have no behavioural differences.
Stop this pattern propagating via copy paste, and avoid having to think about it ever again.

### DIFF
--- a/Library/docs/source/manual/documentation/script_system.md
+++ b/Library/docs/source/manual/documentation/script_system.md
@@ -69,7 +69,7 @@ So when one entity posts a `HeyThere` message...
 ```js
 PostAMessage()
 {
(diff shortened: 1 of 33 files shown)


MAG-OR (approved):
thanks for doing that.

I see most of the space names and descriptions are actually identical, so we could either do globals and just use them in most places, or modify the CreateSpace wrapper that's already in SpaceSystemTests.cpp to use defaults if not specified (and move the unique naming in there, instead of doing every single time), but there is no need to undo this PR.

MAG-SamBirley:
Nice one.

PR #737, opened 2025-07-10
[NT-0] ci: Add GH Action for PR review guidelines

MAG-AdamThorn:
Added a GitHub Action that will automatically add a comment with PR review guidelines when a PR is opened.

### DIFF
--- a/.github/workflows/add-pr-guidance.yml
+++ b/.github/workflows/add-pr-guidance.yml
@@ -1,0 +1,59 @@
+# Name of the workflow as it appears in the GitHub Actions tab
+name: Add PR Review Guidance Comment
+
+# Trigger the workflow when a pull request is opened / reopened
+on:
+  pull_request:
+    types: [opened, reopened]
+
+# Grant write access to pull requests (this is necessary to add comments)
+permissions:
+  pull-requests: write
+
+# Define a single job named 'add-guidance-comment'
+jobs:
+  add-guidance-comment:
+    # Specify the runner environment for this job
+    runs-on: ubuntu-latest
+
+    # Define the step: Add a comment to the Pull Request with review guidance
+    steps:
+      - name: Add PR Review Guidance Comment
+        # Use the 'peter-evans/create-or-update-comment' action to manage PR comments
+        uses: peter-evans/create-or-update-comment@v3
+        with:
+          # Specify the ID of the pull request to comment on
+          issue-number: ${{ github.event.pull_request.number }}
+          # Define the content of the comment.
+          body: |
+            # 🚀 Pull Request Review Guidelines
+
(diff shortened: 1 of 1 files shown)


.github/workflows/add-pr-guidance.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  add-guidance-comment:
+    # Specify the runner environment for this job
+    runs-on: ubuntu-latest
+
+    # Define the step: Add a comment to the Pull Request with review guidance
+    steps:
+      - name: Add PR Review Guidance Comment
+        # Use the 'peter-evans/create-or-update-comment' action to manage PR comments
+        uses: peter-evans/create-or-update-comment@v3
```
MAG-ElliotMorris, line 23: [Comment] Not sure if we have guidance on this yet, but it's important to have a scan through non-official actions like this to make sure they're not sketchy.

Imo my very rough-and-ready advice is if they have a reasonable number of stars then you're pretty okay trusting them, I don't want to imply we need to do full audits, but it should be an expectation that they're investigated.

(You probably have already done this, more commenting to broadcast the intent of having a norm on this)

.github/workflows/add-pr-guidance.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          # Specify the ID of the pull request to comment on
+          issue-number: ${{ github.event.pull_request.number }}
+          # Define the content of the comment.
+          body: |
+            # 🚀 Pull Request Review Guidelines
```
MAG-ElliotMorris, line 29: [Comment] Tempted to say we should pull this from a data-file, but tbh I appreciate the direct simplicity for something like this.

.github/workflows/add-pr-guidance.yml
MAG-ElliotMorris: 819 stars, all good.

MAG-AdamThorn: Yes I did review the repo and it looks legit. The version tag is mutable so there is nothing to stop them from changing the code underneath, however permissions are set our side so I think we are good. And as you say 819 stars.

.github/workflows/add-pr-guidance.yml
MAG-AdamThorn: @MAG-ElliotMorris sorry not sure I follow.

MAG-ElliotMorris: Just saying that this is technically duplication, we could pull this text from the guidelines.md file, but I sort of like it this way cause it's simpler.

MAG-AdamThorn: We have just discussed. All good to merge.

PR #738, opened 2025-07-10
[BP-1287] ci: Added exception throwing

mag-aidangustard:
Added exception throwing to stop script silently failing if the response is not 200.


teamcity/TestReportUploader.py
```diff
@@ -42,12 +42,11 @@ def upload_test_report(args):
...
-
-    log_path = os.path.dirname(os.path.abspath(args.junit_file)) + "TestReportUploader.log"
-    with open(log_path, "w") as file:
-        file.write(response.text)
-        print(f"Log written to {log_path}")
+        log_path = os.path.dirname(os.path.abspath(args.junit_file)) + "TestReportUploader.log"
+        with open(log_path, "w") as file:
+            file.write(response.text)
+            print(f"Log written to {log_path}")
+        raise Exception(f"Failed to import results: {response.status_code}")
```
MAG-ElliotMorris, line 49: [Question] Was it intentional to change the indentation such that the file is now only written in the failure case. Previously it looks like you got a file output no matter what branch the above if took.

mag-aidangustard: yes.  There is no reason to log nothing which is what gets logged if it's successful

MAG-ElliotMorris (approved):

MAG-ElliotMorris:
I'll ping in the channel about getting a reviewer on this soonish, as Orianne has gone home for the day and you probably want unblocked.

teamcity/TestReportUploader.py
MAG-AdamThorn: [Question] Why remove the log for the failure case?

[Consider] Can you add a comment about what that file upload is doing and why it only happens in the case of failure?

mag-aidangustard: 1.  I moved the log so it only happens when the build fails.  But now that it throws an exception the log would never get written to unless the upload was successful so I moved the log before the exception to be part of the failure condition
2. nothing about the upload has changed.  What I've done is made sure that if the response from the API is anything other than 200 it will throw an exception so that it will not silently fail as it's been doing currently

PR #739, opened 2025-07-10
[OF-1701] feat: Add http request response logging

MAG-AdamThorn:
Add http request and response logging to the POCOWebClient.

- This logging will only occur when the loglevel is set to VeryVerbose.
- While we intended to add similar logging to the EmscriptenWebClient this is currently not possible due to a known issue with cross-thread callbacks on WASM ([OB-3782](https://magnopus.atlassian.net/browse/OB-3782)). Once that bug is addressed logging on WASM can be added.

A ticket ([OF-1719](https://magnopus.atlassian.net/browse/OF-1719)) has been created to track the work required to add Rest http requests / responses logging to the Emscripten Web Client once OB-3782 has been resolved.

[OB-3782]: https://magnopus.atlassian.net/browse/OB-3782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OF-1719]: https://magnopus.atlassian.net/browse/OF-1719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
+++ b/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
@@ -53,10 +53,10 @@ void LogHttpResponseIfLoglevelVeryVerbose(
     // If the LogSystem LogLevel has been set to VeryVerbose, log the response.
     if (LogSystem != nullptr && LogSystem->GetSystemLevel() == csp::common::LogLevel::VeryVerbose)
     {
-        csp::common::String Status = std::to_string(static_cast<int>(PocoResponse.getStatus())).c_str();
-
         LogSystem->LogMsg(csp::common::LogLevel::VeryVerbose,
-            fmt::format("HTTP Response\n{0} {1}\nStatus: {2} - {3}", Verb, Request.GetUri().GetAsString(), Status, PocoResponse.getReason()).c_str());
+            fmt::format("HTTP Response\n{0} {1}\nStatus: {2} - {3}", Verb, Request.GetUri().GetAsString(), static_cast<int>(PocoResponse.getStatus()),
+                PocoResponse.getReason())
+                .c_str());
     }
 }
 


Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -202,7 +202,7 @@ class CSP_API ComponentBase
 
     ComponentScriptInterface* ScriptInterface;
 
-    // May be null, should check before use.
+    // This will be null for the SignalR client as by default we do not want to log Multiplayer traffic.
```
MAG-AdamThorn: [fix] Remove this comment change.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -340,6 +355,9 @@ void POCOWebClient::Head(HttpRequest& Request)
     {
         ProcessResponseAsync(*ClientSession, PocoResponse, ResponseStream, Request);
     }
+
+    // Call will log the response if the LogSystem LogLevel has been set to VeryVerbose.
```
MAG-OR: [comment] is this repeated comment really necessary?

MAG-OR (approved):

Library/src/Common/Web/EmscriptenWebClient/EmscriptenWebClient.h
```diff
@@ -17,6 +17,13 @@
 
 #include "Common/Web/WebClient.h"
 
+namespace csp::common
+{
+
+class LogSystem;
+
+}
```
MAG-ElliotMorris: [nit] Redundant whitespace.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -141,6 +144,9 @@ void POCOWebClient::Get(HttpRequest& Request)
...
+    LogHttpResponseIfLoglevelVeryVerbose(LogSystem, "GET", Request, PocoResponse);
```
MAG-ElliotMorris: [Nit] You've done the thing I like, having a very descriptive method name, and the bonus is that you probably don't need to duplicate the explanation in the comment.

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
```diff
@@ -90,6 +97,10 @@ class POCOWebClient : public WebClient
 
     std::vector<Poco::Net::HTTPCookie>* Cookies;
     std::mutex CookiesMutex;
+
+private:
+    static void LogHttpResponseIfLoglevelVeryVerbose(
+        csp::common::LogSystem* LogSystem, const char* Verb, const HttpRequest& Request, const Poco::Net::HTTPResponse& PocoResponse);
```
MAG-ElliotMorris: [Consider] private static makes little sense as a valid combination to my mind. You're saying this is something that may only be used in the private implementation (the .cpp), and yet has external static linkage, which is redundant. I'd just use an anonymous namespace method in the cpp, functionally identical but without the wasted work for the linker.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -537,4 +555,18 @@ void POCOWebClient::SetFileUploadContent(HttpPayload* Payload, Poco::Net::PartSo
...
 }
 
+void POCOWebClient::LogHttpResponseIfLoglevelVeryVerbose(
+    csp::common::LogSystem* LogSystem, const char* Verb, const HttpRequest& Request, const Poco::Net::HTTPResponse& PocoResponse)
+{
+    // If the LogSystem LogLevel has been set to VeryVerbose, log the response.
+    if (LogSystem != nullptr && LogSystem->GetSystemLevel() == csp::common::LogLevel::VeryVerbose)
+    {
+        csp::common::String Status = std::to_string(static_cast<int>(PocoResponse.getStatus())).c_str();
+        csp::common::String StatusReason = PocoResponse.getReason().c_str();
```
MAG-ElliotMorris: [consider] These are redundant conversions, fmt::format can take all the basic types, so you do't need to convert `Status` to a string (can just be an int), and you don't need to convert `PocoResponse.getReason` from `std::string` to `csp::string` at all. (You'll need to include `fmt_formatters` to get the conversion behaviour there though)

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
```diff
@@ -22,6 +22,13 @@
 #include <Poco/Net/PartSource.h>
 #include <Poco/Net/PrivateKeyPassphraseHandler.h>
 
+namespace csp::common
+{
+
+class LogSystem;
+
+}
```
MAG-ElliotMorris: [nit] Redundant whitespace.

Library/src/Common/Web/WebClient.cpp
```diff
@@ -464,4 +472,70 @@ void WebClient::PrintClientErrorResponseMessages(const HttpResponse& Response)
...
+    case ERequestVerb::Delete:
+        Verb = "DELETE";
+        break;
+    case ERequestVerb::Head:
+        Verb = "HEAD";
+        break;
+    default:
+        Verb = "UNKNOWN";
+        break;
+    }
```
MAG-ElliotMorris: [Consider] Writing a [fmt::formatter](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/include/CSP/Common/fmt_Formatters.h) for this type. It may be a good way to encode conversions like this, and it means you can pass the enum directly into `fmt::format`. Also means one-off implementations like this can be reused elsewhere.

I think all you need to do is copy the existing one, change `csp::common::String` to `ERequestVerb`. Put this exact implementation in the `format` method, and then ` return formatter<std::string_view>::format(std::string_view(Verb.c_str(), Verb.Length()), ctx);` at the bottom

Be interesting to establish a pattern on this, depending on what your take is. We log a lot of our enums as just ints for convenience, would love to start logging them more as strings as you have here.

Alternatively, you could also [consider] writing a formatter for the `HttpRequest` type _as a whole_. Effectively replacing this entire function. That means any time we wanted to log a HttpRequest, it would be completely trivial, as well as always formatted identically, with all the formatting code nicely tucked away!

(I love string manipulation it's the best thing :D )

Library/src/Common/Web/WebClient.cpp
```diff
@@ -464,4 +472,70 @@ void WebClient::PrintClientErrorResponseMessages(const HttpResponse& Response)
...
+
+    csp::common::String Url = Request->GetUri().GetAsString();
+
+    const auto& RequestHeaders = Request->GetPayload().GetHeaders();
+    csp::common::String Headers;
+
+    for (const auto& Header : RequestHeaders)
+    {
+        Headers.Append(fmt::format("\t{}: {}\n", Header.first, Header.second).c_str());
+    }
```
MAG-ElliotMorris: [Comment] I know you can do fancy things with fmt::format like auto-list formatting and whatnot. I don't know how to do it though, so not an actual recommend, just a callout.

Library/src/Common/Web/WebClient.cpp
```diff
@@ -464,4 +472,70 @@ void WebClient::PrintClientErrorResponseMessages(const HttpResponse& Response)
...
+
+    const csp::common::String& RequestPayload = Request->GetPayload().ToJson();
+
+    csp::common::String RequestBody;
```
MAG-ElliotMorris: [nit] Default initialize with `()`. `{}` or ` = ""` to prevent garbage in release builds in the case of an empty payload.

Library/src/Common/Web/WebClient.cpp
MAG-ElliotMorris: ... ah, the one thing that strikes me is that I've put `fmt_formatters.h` in `common`, and I guess you need a formatter file per module to avoid dependency bleeds. A tad annoying, maybe there's a better arrangement.

Library/src/Common/Web/WebClient.cpp
```diff
@@ -464,4 +472,70 @@ void WebClient::PrintClientErrorResponseMessages(const HttpResponse& Response)
...
+    {
+        rapidjson::Document RequestJson;
+        RequestJson.Parse(RequestPayload);
+
+        if (RequestJson.IsObject())
+        {
+            for (rapidjson::Value::ConstMemberIterator itr = RequestJson.MemberBegin(); itr != RequestJson.MemberEnd(); ++itr)
+            {
+                // Obfuscate the users password when logging the Http request body
+                if (std::strcmp(itr->name.GetString(), "password") == 0)
```
MAG-ElliotMorris: [Question] Any reason to use `strcmp` here rather than the more natural equality operator?

Library/src/Multiplayer/SignalR/SignalRClient.cpp
```diff
@@ -136,10 +136,11 @@ class SignalRResponseReceiver : public SignalRResponseWaiter, public csp::web::I
 
 CSPHttpClient::CSPHttpClient()
 {
+// Passing null for the LogSystem to the POCO/Emscripten web client ctor to avoid logging high frequency multiplayer API exchange.
```
MAG-ElliotMorris, line 139: [Comment] 👍 leaving explanations.

MAG-ElliotMorris (approved):
Approving as no hard blockers, only nits, comments and considers.

MAG-AdamThorn:
> Approving as no hard blockers, only nits, comments and considers.

Great feedback thank you, addressing now.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: Fixed: [77ed149](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/77ed149c21e1c41e0a66bc01bc482b10ad65bc37)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: Fixed: [77ed149](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/77ed149c21e1c41e0a66bc01bc482b10ad65bc37)

Library/src/Common/Web/EmscriptenWebClient/EmscriptenWebClient.h
MAG-AdamThorn: Fixed: [3a9b114](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/3a9b1148a075ffb72c7d8d55be03f20052c5ca68)

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
MAG-AdamThorn: Agreed. Fixed: [db5af10](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/db5af103ca2f53a0c9d2e673be2be202d96162c8)

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
MAG-AdamThorn: Fixed: [3a9b114](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/3a9b1148a075ffb72c7d8d55be03f20052c5ca68)

Library/src/Common/Web/WebClient.cpp
MAG-AdamThorn: Great feedback. I elected to go for option 2 and have implemented a formatter for HttpRequest.

Change: [4828a32](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/4828a325d72474a354e08fe9be5e68c9cd71a8f8)

Library/src/Common/Web/WebClient.cpp
MAG-AdamThorn: I did investigate options here. Including `fmt_range` which allowed me to replace this loop with a one liner, but that printed all headers on a single line. To log each header on it's own line I would have had to use fmt::join, but that would have required another custom formatter so I elected to leave it as it was.

Library/src/Common/Web/WebClient.cpp
MAG-AdamThorn: Fixed: [4828a32](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/4828a325d72474a354e08fe9be5e68c9cd71a8f8)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: The first conversion is there as I want to log the string representation of the enum value, so for example `GET`. I agree though that the second conversation is redundant and have removed.

Fix: [cc36118](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/cc3611845531d28d8fd137616cd0bc7873d709ae)

Library/src/Common/Web/WebClient.cpp
MAG-AdamThorn: GetString() returns a const char* so would compare the string literal with the memory address. I have updated it to use a std::string_view comparison instead.

Change: [4828a32](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/4828a325d72474a354e08fe9be5e68c9cd71a8f8)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-ElliotMorris: `std::to_string` will not get you that string unless you've manually provided overloads (sort of what formatter overloads are), you'll just get "0" as a string. (or whatever the number happens to be) Am I missing something?

Library/src/Common/Web/WebClient.cpp
MAG-ElliotMorris: 🤩 So good.

Library/src/Common/Web/WebClient.cpp
MAG-ElliotMorris: Aye, I would have made the same choice probably.

Library/src/Common/Web/WebClient.cpp
MAG-ElliotMorris: Ah I see. I probably wouldn't have caught that first time round and made the mistake. Calling a `const char*` return `GetString` in a c++ codebase feels like they're intentionally trying to trick you!

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: Ha I am an idiot. You are of course right. In my mind I was thinking of the verb enum and needing to convert that to a string.

Fix: [7189067](https://github.com/magnopus-opensource/connected-spaces-platform/pull/739/commits/7189067af409215e3a1d03cc29e2ced6a035e117)

PR #740, opened 2025-07-11
[OF-1688] feat: provide service and multiplayer hub status information to clients

MAG-ChristopherAtkinson:
The changes here are prerequisite components to handle resolving services and multiplayer hub methods ahead of integration. 
The changes include the following:
 * Added methods to resolve services 
 * Added test to cover potential success and failure paths for services
 * Added methods to resolve the multiplayer hub function  
 * Added test to cover potential success and failure paths for the multiplayer hub function  
 * Added template result information ahead of integration 

_This PR does not include changes to the current initialisation process; this requirement has been moved to the integration ticket, as we are unsure when it will be applicable. It was best not to require the client to adopt a breaking change until necessary._ 

Originally, I included a system for Status; however, this was removed as we currently do not have a use for it, but kept the project structure so we can support it in the future.


Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+
+/// @brief Stores the version and deprecation information for a services
+class CSP_API ServiceVersionInfo
+{
+public:
+    csp::common::String Version;
+    csp::common::String DeprecationDatetime;
+};
+
+/// @brief Stores information about a service include available versions
```
MAG-ElliotMorris: [nit] ...including available versions

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+CSP_START_IGNORE
+template <typename T, typename U, typename V, typename W> class ApiResponseHandler;
+CSP_END_IGNORE
+
+} // namespace csp::services
+
+namespace csp::systems
+{
+
+/// @brief Stores the version and deprecation information for a services
```
MAG-ElliotMorris: [nit] "a services" should be either "a service" or just "services"

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+    csp::common::String DeprecationDatetime;
+};
+
+/// @brief Stores information about a service include available versions
+class CSP_API ServiceInfo
+{
+public:
+    csp::common::String ReverseProxy;
+    csp::common::String Name;
+    csp::common::Array<ServiceVersionInfo> ApiVersions;
```
MAG-ElliotMorris: [Comment] Good type composition. Annoys me that we can't make these things `structs` because of the wrapper gen, oh well, not the biggest affront.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+public:
+    csp::common::String Version;
+    csp::common::String DeprecationDatetime;
+};
+
+/// @brief Stores information about a service include available versions
+class CSP_API ServiceInfo
+{
+public:
+    csp::common::String ReverseProxy;
```
MAG-ElliotMorris: [consider] Can you doc what this is? I don't think it's self-evident. (Ie, I don't know what a reverse proxy is and if I need to care)

If this is like, an address? Then `ReverseProxyAddress` as a name would serve, because then I know this is an address to a "thing" and that's fine.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+} // namespace csp::services
+
+namespace csp::systems
+{
+
+/// @brief Stores the version and deprecation information for a services
+class CSP_API ServiceVersionInfo
+{
+public:
+    csp::common::String Version;
```
MAG-ElliotMorris: [consider] A comment showing how the version is formatted, helpful when all you can see is string types.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+{
+public:
+    csp::common::String ReverseProxy;
+    csp::common::String Name;
+    csp::common::Array<ServiceVersionInfo> ApiVersions;
+    csp::common::String CurrentApiVersion;
+};
+
+/// @brief Store information about the current service deployment
+class CSP_API StatusInfo
```
MAG-ElliotMorris: [Consider] Naming this something else. It's just that `StatusInfo` and `ServiceInfo` are so similar, and "The information about a status" doesn't make much sense to me. Is this more like `ContainerInfo`? 

... It's the collection of all the services right, what's the concept for that called... maybe just api, `ApiInfo`? Encoding the fact that our backends happen to be run in a container may not be the best thing.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+{
+public:
+    csp::common::String ContainerVersion;
```
MAG-ElliotMorris: [consider] Following on from the above point, consider `ApiVersion` or something else instead of `ContainerVersion` (although I grant, using different words to CHS has bit us in the past...)

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+    csp::common::Array<ServiceInfo> Services;
+};
+
+/// @ingroup CSPFoundation
+/// @brief Data class used to contain information when a Response is received from Status Server
+class CSP_API StatusInfoResult : public csp::systems::ResultBase
```
MAG-ElliotMorris: [Consider] Think about what this type name means.

"Status" - Information about something
"Info" - Information about something
"Result" - Information about something (to be fair, it's a tad more specific, Information about an _action_)

This type could be alternately spelled `InfoInfoInfo` 😆 

(This is half a joke, I know the word `Result` is a pattern we apply to mean "Object that contains the prefixing type", so I'm not suggesting you address that.)

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+{
+    /** @cond DO_NOT_DOCUMENT */
+    CSP_START_IGNORE
+    friend class CSPFoundation;
+    template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+    CSP_END_IGNORE
+    /** @endcond */
+
+public:
+    /// @brief Will return info for the latest available status information from the server
```
MAG-ElliotMorris: [Nit] Reads awkwardly. "Info for status information", what does that mean? (All these comments are on the same theme, status information is the wrong concept.)

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+    /// @return StatusInfo : the closest maintenance window information
+    [[nodiscard]] const StatusInfo& GetLatestStatusInfo() const;
```
MAG-ElliotMorris: [Consider] This is new, "maintenance window". What does that mean? This is a surprising thing to read in a docstring when the term hasn't been mentioned anywhere else. You should explain it if it's necessary information.

Library/src/Systems/Status/Status.cpp
```diff
@@ -0,0 +1,25 @@
+/*
```
MAG-ElliotMorris: [fix] I've made a lot of comments on this point, using this one as the main blocker, commenting on the filename.

`Status` is too generic a concept, and is a modifier word. Find a concept that addresses the actual thing you're talking about. I recommend `RestApi`, but feel free to choose anything that conveys the correct concept. Apply that concept to the rest of the types and documentation. 

Now that i'm moving to the next commit, i'll stop bringing this point up again and again.

Library/include/CSP/CSPFoundation.h
```diff
@@ -167,6 +172,11 @@ class CSP_API CSPFoundation
     /// @return csp::common::String&
     static const csp::common::String& GetTenant();
 
+    /// @brief Resolves Service Definition aginst the provided StatusInfo
```
MAG-ElliotMorris: [nit] This is simply repeating the method signature. You should say something that provides new information. What do you mean by "Resolve"?

Library/include/CSP/CSPFoundation.h
```diff
@@ -167,6 +172,11 @@ class CSP_API CSPFoundation
...
+    /// Will log information about the state of each of the services, such as depreciation, and newer versions
+    /// @return bool : true if all services are available, false otherwise
+    CSP_NO_EXPORT static bool ResolveServiceDefinition(const ServiceDefinition& ServiceDefinition, const csp::systems::StatusInfo& StatusInfo);
```
MAG-ElliotMorris: [Consider] It seems strange to me that these are separate objects. `StatusInfo` is the status of an api, `ServiceDefinition` is the endpoints and version of an Api. Why do we not extend `ServiceDefinition` with this status fetching functionality, and avoiding having to pass two objects that are intrinsically linked to each other.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+bool CSPFoundation::ResolveServiceDefinition(const ServiceDefinition& ServiceDefinition, const csp::systems::StatusInfo& StatusInfo)
+{
+    // Extract the reverse proxy from the URI to be used as key in StatusInfo
+    auto const ReverseProxy = [](const csp::common::String URI) -> csp::common::String
+    {
+        if (auto const Split = URI.Split('/'); Split.Size() != 0)
+            return Split[Split.Size() - 1];
+
+        return "";
+    }(ServiceDefinition.GetURI());
```
MAG-ElliotMorris: [fix] Using an immediately invoked lambda is redundant. Use an anonymous namespace if you want function like-semantics without external linkage. Although I don't see why you wouldn't just perform this calculation inline as the use is never repeated.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
 
 const csp::common::String& CSPFoundation::GetTenant() { return *Tenant; }
 
+bool CSPFoundation::ResolveServiceDefinition(const ServiceDefinition& ServiceDefinition, const csp::systems::StatusInfo& StatusInfo)
+{
+    // Extract the reverse proxy from the URI to be used as key in StatusInfo
+    auto const ReverseProxy = [](const csp::common::String URI) -> csp::common::String
```
MAG-ElliotMorris: [fix] The CSP codebase uses west const almost everywhere.

Library/include/CSP/Systems/Status/Status.h
MAG-mv: [Fix] If this is client-facing API, we should doc the public vars

Library/include/CSP/Systems/Status/Status.h
MAG-mv: [Fix] Same as above

Library/include/CSP/Systems/Status/Status.h
MAG-mv: [Fix] Same as above

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
 
 const csp::common::String& CSPFoundation::GetTenant() { return *Tenant; }
 
+bool CSPFoundation::ResolveServiceDefinition(const ServiceDefinition& ServiceDefinition, const csp::systems::StatusInfo& StatusInfo)
+{
+    // Extract the reverse proxy from the URI to be used as key in StatusInfo
```
MAG-ElliotMorris: [fix] Please give an example as a comment of what these URIs look like. Following split logic is impossible without one.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+
+/// @brief Store information about the current service deployment
+class CSP_API StatusInfo
+{
+public:
+    csp::common::String ContainerVersion;
+    csp::common::Array<ServiceInfo> Services;
+};
+
+/// @ingroup CSPFoundation
```
MAG-mv: [Fix] I dont think this should be ingroup CSPFoundation. You should probaly add groups to the rest of your types, as well.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+    auto const ReverseProxy = [](const csp::common::String URI) -> csp::common::String
+    {
+        if (auto const Split = URI.Split('/'); Split.Size() != 0)
+            return Split[Split.Size() - 1];
+
+        return "";
+    }(ServiceDefinition.GetURI());
+
+    // Find the ServiceInfo that correlate with the reverse proxy
```
MAG-ElliotMorris: [nit] "...correlate**s** with..."

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+    /** @cond DO_NOT_DOCUMENT */
+    CSP_START_IGNORE
+    friend class CSPFoundation;
+    template <typename T, typename U, typename V, typename W> friend class csp::services::ApiResponseHandler;
+    CSP_END_IGNORE
+    /** @endcond */
+
+public:
+    /// @brief Will return info for the latest available status information from the server
+    /// @return StatusInfo : the closest maintenance window information
```
MAG-mv: [Nit] Capital should go after colon

Library/src/Systems/Status/Status.cpp
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+#include "CSP/Systems/Status/Status.h"
+
+#include "Debug/Logging.h"
+#include "Services/ApiBase/ApiBase.h"
+#include "Systems/ResultHelpers.h"
+
+const csp::systems::StatusInfo& csp::systems::StatusInfoResult::GetLatestStatusInfo() const { return StatusInfoResponse; }
```
MAG-mv: [Question] Any reason this isnt in the systems namespace?

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+    auto const ServiceInfo = std::find_if(StatusInfo.Services.begin(), StatusInfo.Services.end(),
+        [ReverseProxy](csp::systems::ServiceInfo ServiceInfo) { return ServiceInfo.ReverseProxy == ReverseProxy; });
+
+    if (!ServiceInfo)
+    {
+        auto const message = fmt::format("Unable to resolve {0} Reverse Proxy in Status Info", ReverseProxy.c_str());
+
+        CSP_LOG_MSG(csp::common::LogLevel::Error, message.c_str());
+        return false;
+    }
```
MAG-ElliotMorris: [comment] Good logging and early return.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+        return "";
+    }(ServiceDefinition.GetURI());
+
+    // Find the ServiceInfo that correlate with the reverse proxy
+    auto const ServiceInfo = std::find_if(StatusInfo.Services.begin(), StatusInfo.Services.end(),
+        [ReverseProxy](csp::systems::ServiceInfo ServiceInfo) { return ServiceInfo.ReverseProxy == ReverseProxy; });
+
+    if (!ServiceInfo)
+    {
+        auto const message = fmt::format("Unable to resolve {0} Reverse Proxy in Status Info", ReverseProxy.c_str());
```
MAG-ElliotMorris: [fix] #include fmt_formatters.h and you won't have to convert to c_str() here.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+
+        CSP_LOG_MSG(csp::common::LogLevel::Error, message.c_str());
+        return false;
+    }
+
+    // Find the ServiceVersionInfothat correlate with the expected version of the service
```
MAG-ElliotMorris: [nit] " Find the ServiceVersionInfo that correlate**s** with the expected version of the service

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
 
 const csp::common::String& CSPFoundation::GetTenant() { return *Tenant; }
 
+bool CSPFoundation::ResolveServiceDefinition(const ServiceDefinition& ServiceDefinition, const csp::systems::StatusInfo& StatusInfo)
+{
+    // Extract the reverse proxy from the URI to be used as key in StatusInfo
+    auto const ReverseProxy = [](const csp::common::String URI) -> csp::common::String
+    {
+        if (auto const Split = URI.Split('/'); Split.Size() != 0)
```
MAG-mv: [Consider] Small comment explaining what the if does would be useful to readers

Library/src/CSPFoundation.cpp
MAG-mv: [Nit] Random lowercase var

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+        auto const message = fmt::format("Unable to resolve {0} Reverse Proxy in Status Info", ReverseProxy.c_str());
+
+        CSP_LOG_MSG(csp::common::LogLevel::Error, message.c_str());
+        return false;
+    }
+
+    // Find the ServiceVersionInfothat correlate with the expected version of the service
+    auto const ServiceVersionInfo = std::find_if(ServiceInfo->ApiVersions.begin(), ServiceInfo->ApiVersions.end(),
+        [ServiceDefinition](csp::systems::ServiceVersionInfo ServiceVersionInfo)
+        { return ServiceVersionInfo.Version.c_str() == fmt::format("v{0}", ServiceDefinition.GetVersion()); });
```
MAG-ElliotMorris: [comment] It's fine like this, but just to let you know that you don't need the number in the `{0}` thing, by default it will substitute your args in order of placement if you just use `{}`. I've used them before sometimes with multi-args, but tbh I think even that was a mistake.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+
+    auto const Documentation = "https://connected-spaces-platform.net/index.html";
+
+    // Validate that the current service has being retired by the live services, and promote information to the user
```
MAG-ElliotMorris: [nit] Phrasing here makes it sounds like we're expecting the service to have been retired. "Validate" has a strong implication that you expect things to be a certain way, and would be surprised if they are not.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+        CSP_LOG_MSG(csp::common::LogLevel::Error, message.c_str());
+        return false;
+    }
+
+    // Find the ServiceVersionInfothat correlate with the expected version of the service
+    auto const ServiceVersionInfo = std::find_if(ServiceInfo->ApiVersions.begin(), ServiceInfo->ApiVersions.end(),
+        [ServiceDefinition](csp::systems::ServiceVersionInfo ServiceVersionInfo)
+        { return ServiceVersionInfo.Version.c_str() == fmt::format("v{0}", ServiceDefinition.GetVersion()); });
+
+    auto const Documentation = "https://connected-spaces-platform.net/index.html";
```
MAG-mv: [Nit] Could be static constexpr

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+
+    // Validate that the current service has being retired by the live services, and promote information to the user
+    if (ServiceInfo != StatusInfo.Services.end()
+        && ServiceVersionInfo == ServiceInfo->ApiVersions.end()) // The current version in use has been retired.
+    {
+        auto const message = fmt::format("{0} v{1} has been retired, the latest version is {2}. For more information please visit: {3}",
```
MAG-ElliotMorris: [Comment] I like the message here, clear and to the point.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+    if (ServiceInfo != StatusInfo.Services.end()
+        && !ServiceVersionInfo->DeprecationDatetime.IsEmpty()) // The current version in use has been marked as deprecated.
+    {
+        auto const message = fmt::format("{0} v{1} will be deprecated as of {2}, the latest version is {3}. For more information please visit: {4}",
+            ServiceInfo->Name.c_str(), ServiceDefinition.GetVersion(), ServiceVersionInfo->DeprecationDatetime.c_str(),
+            ServiceInfo->CurrentApiVersion.c_str(), Documentation);
+
+        CSP_LOG_MSG(csp::common::LogLevel::Warning, message.c_str());
+        return true;
+    }
```
MAG-ElliotMorris: [Consider] Writing some anon-namespace function with names to encapsulate this logic. Then your main function could be readable.

```
ResolveServiceDefinition(Urls)
{
  ReverseProxyAddress = ExtractReverseProxy(Urls);
  CheckForServiceRetirement(ReverseProxyAddress, Urls);
  CheckForServiceDeprecations(ReverseProxyAddress, Urls);
  //etc
}
```

Then you'd have room to leave comments on what these functions mean (ie, what does Retirement and Deprecation mean), in this more direct method.

It would also make the return exit's clearer, as when phrased like this, it's clear that you've done it in an exclusive priority order, ie log about retirement most importantly, deprecations next, and newly-available last, which is clever, but unclear.

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+
+    // Validate that the current service has a newer version available on the live services, and promote information to the user
+    if (ServiceInfo != StatusInfo.Services.end()
+        && ServiceVersionInfo->Version != ServiceInfo->CurrentApiVersion) // The current version in use is not the latest available.
+    {
+        auto const message = fmt::format("{0} v{1} is not the latest available, the latest version is {2}. For more information please visit: {3}",
```
MAG-ElliotMorris: [Question] I'm a bit surprised we're informing the user about this. Do they need/want to know if the service they're on is out of date. ...Tbh, it might be a good idea, just wondering if this is directly from the requirements doc or something we've added?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -189,6 +189,23 @@ MultiplayerConnection::~MultiplayerConnection()
     }
 }
 
+bool MultiplayerConnection::ResolveMultiplayerHubMethods(const csp::common::Array<csp::common::String>& Methods)
+{
+    for (const auto& Method : MultiplayerHubMethodMap())
```
MAG-mv: [Question] Why is the hub map being recreated here? Can we use the instance?

Library/src/CSPFoundation.cpp
```diff
@@ -396,6 +398,73 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+    {
+        if (auto const Split = URI.Split('/'); Split.Size() != 0)
+            return Split[Split.Size() - 1];
+
+        return "";
+    }(ServiceDefinition.GetURI());
+
+    // Find the ServiceInfo that correlate with the reverse proxy
+    auto const ServiceInfo = std::find_if(StatusInfo.Services.begin(), StatusInfo.Services.end(),
+        [ReverseProxy](csp::systems::ServiceInfo ServiceInfo) { return ServiceInfo.ReverseProxy == ReverseProxy; });
```
MAG-ElliotMorris: [Question] Can you explain why the way we identify these is via the reverse proxy? This is really hard to follow, partly because I don't know what a reverse proxy is, and partly because all the `Info` objects are very similar and don't convey their domain of responsibility that well.

Tests/src/PublicAPITests/MCSExternalDependencyTests.cpp
```diff
@@ -0,0 +1,214 @@
...
+using namespace csp::common;
+
+namespace
+{
+/* We need to unset the mock logger before CSP shuts down,
+ * because you get interdependent memory errors in the "Foundation shutdown"
+ * log if you don't. (Another reason we don't want to be starting/stopping
+ * ALL of CSP in these tests really.)
+ */
+struct RAIIMockLogger
```
MAG-mv: [Fix] I believe in a recent commit, we put te RAII logger in a specific location to prevent duplication. @MAG-ElliotMorris ?

Library/include/CSP/CSPFoundation.h
```diff
@@ -167,6 +172,11 @@ class CSP_API CSPFoundation
     /// @return csp::common::String&
     static const csp::common::String& GetTenant();
 
+    /// @brief Resolves Service Definition aginst the provided StatusInfo
+    /// Will log information about the state of each of the services, such as depreciation, and newer versions
```
MAG-ElliotMorris: [Question] Is this true? Does it log information on the state of a singular service, or lots of services? Each of the services implies there's more than one service in play, which I don't think there is in this method.

I was expecting to see a loop of some kind and am confused.

Library/include/CSP/Systems/Status/Status.h
```diff
@@ -0,0 +1,89 @@
...
+/// @brief Stores the version and deprecation information for a services
+class CSP_API ServiceVersionInfo
+{
+public:
+    csp::common::String Version;
+    csp::common::String DeprecationDatetime;
+};
+
+/// @brief Stores information about a service include available versions
+class CSP_API ServiceInfo
```
MAG-mv: [Question] Service feels like a very generic name for these types. Is there a name that is more expressive?

Library/src/CSPFoundation.cpp
MAG-mv: [Nit] Missing space + correlate -> correlates

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -193,6 +193,10 @@ class CSP_API MultiplayerConnection
     /// @return MultiplayerHubMethodMap : the MultiplayerHubMethodMap instance
     CSP_NO_EXPORT MultiplayerHubMethodMap GetMultiplayerHubMethods() const { return MultiplayerHubMethods; }
 
+    /// @brief Resolves Multiplayer Hub Methods aginst the provided array of available methods
```
MAG-ElliotMorris: [fix] Stop saying the word "Resolve" without defining your meaning, it's not used for the thing you're actually doing in common technical parlance.

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -193,6 +193,10 @@ class CSP_API MultiplayerConnection
...
+    /// @return bool : true if all available methods match, false otherwise
```
MAG-ElliotMorris: [Consider] Match what? Explain.

MAG-mv (changes requested):

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -193,6 +193,10 @@ class CSP_API MultiplayerConnection
...
+    CSP_NO_EXPORT static bool ResolveMultiplayerHubMethods(const csp::common::Array<csp::common::String>& Methods);
```
MAG-ElliotMorris: [nit] Calling the arg `MethodNames` or `MethodSignatures` or something that more clearly documents what the intended content of these strings should be.

Tests/src/PublicAPITests/MCSExternalDependencyTests.cpp
MAG-ElliotMorris: Yep, you can just include it now.

Tests/src/PublicAPITests/MCSExternalDependencyTests.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    RAIIMockLogger() { csp::systems::SystemsManager::Get().GetLogSystem()->SetLogCallback(MockLogCallback.AsStdFunction()); }
+    ~RAIIMockLogger() { csp::systems::SystemsManager::Get().GetLogSystem()->SetLogCallback(nullptr); }
+    ::testing::MockFunction<void(const String&)> MockLogCallback;
+};
+
+csp::systems::ServiceVersionInfo CreateServiceVersionInfoFromParams(const String& Version, const String& DeprecationDatetime)
```
MAG-ElliotMorris: [nit] `FromParams` is redundant and implicit in the fact that this is a method.

MAG-ElliotMorris (changes requested):
I think we should have gone further than this and bedded the checks into the initialization code behind a feature flag. I know we don't have our "feature flag" system, but #ifdefs or just if statements work perfectly well. 

Then we could have had integration tests and simply removed the feature flag when the time came, rather than having to split and track additional work.

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: That is an error on my part, should be referencing status info, will address in a bulk change with other comment sections

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: Yeah, to confirm this is due to propagating the potential states back to the user, for newer versions from a client perspective would not be too useful, but still has value for third parties.

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Ah yeah apologies, this was an error when I changed wrapping all services to a single service, When the integration aspect is done, the initialisation will loop over all available services, call this for each

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: because the method is static, so the instance is not guaranteed to exist; this raises the question should static be removed from this instance.

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Hey @MAG-mv , this is not a pattern that we currently adopt. I do agree that we should, but wanted to check that the naming of the variable and class are not aiding in the confusion around the data being stored?

MAG-mv: We do in some cases. Take a look at our NetworkEventData classes

MAG-ChristopherAtkinson:
There seems to be some confusion around the new Info classes. I wanted to just post here my thoughts on the suggested changes: 

StatusInfo -> ServicesDeploymentStatus or DeployedContainerStatus: It reflects the current deployment state, listing container version and deployed services. This should help to resolve the dependency with MCS without an explicit callout. 

ServiceInfo -> ServiceStatus: It holds descriptive and structural metadata about a service: Name, Reverse Proxy Address, etc... The new name clearly expresses that it's not the service itself, but information about it.

ServiceVersionInfo -> VersionMetadata: This class stores metadata about each API version, including version string and deprecation date.

@MAG-mv @MAG-ElliotMorris would appreciate your thoughts ahead of the change

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: I addressed this with other inline documentation in a bulk change here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: I addressed this with other inline documentation in a bulk change here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: I addressed this with other inline documentation in a bulk change here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: I added the group 'Status System', which will be expanded in the future with the integration work: [179c7a3](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/179c7a3ade32b3d309b98e0e773dcf89b0bea990)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: fixed in [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: fixed when resolving const correctness in that area: [63bcc79](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/63bcc79f570d018cee0bcd4d13fdbaf7a279e0e5)

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: I ended up moving the implementation of this function to the 'MultiplayerHubMethodMap', please see here: [c6f5cb1](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/c6f5cb1bbcd1c83c1ca6fc9fcc15e83102b3aa70)

Tests/src/PublicAPITests/MCSExternalDependencyTests.cpp
MAG-ChristopherAtkinson: really appreciate this changes, updated here: [e3537a9](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/e3537a9a9990a76892b4659982096d26ecdbb257)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: fixed during bulk inline documentation changes: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: addressed here as part of bulk changes: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Update here as part of bulk inline documentation changes: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Added needed information here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

I kept this as ReverseProxy for now, as I have an open question and can rework this name at the same time

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Added an example to the expected format of the version string: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Reframed the brief here, as I agree this was not very readable: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: Updated to reference the correct structure [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/Systems/Status/Status.cpp
MAG-ChristopherAtkinson: Yeah I have added an open question to the PR around some suggested changes

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: This has been expanded to be more specific: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: I have moved these to their respective classes now and generalised the name to remove confusion around status: [25b1361](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/25b136117e832cca5a0d9e82e3925b823a75375b)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: The self-invoked lambda is just a means to manage scope, the same could have been achieved using braces, this was just a choice due to the utilisation of the nameless namespace

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: fixed here: [63bcc79](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/63bcc79f570d018cee0bcd4d13fdbaf7a279e0e5)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: example provided in bulk changes here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: resolved here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: Amazing: [0d83dc6](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/0d83dc6aafe933a7b112e8fe705c2d4cbc53c400)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: This has been updated: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: Reframes these to remove the expectation, but to be clear around the validation: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: The information that MCS will provide will only contain a name or a reverse proxy to identify the service. Given that we currently store the reverse proxy in the service definition, it made sense to use that over the name, which is used when displaying information as a human-readable string 

Please see the [TDD](https://github.com/magnopus/oko-tech-docs/blob/main/features/dependency-pinning/dependency-pinning-tdd.md) for example of the json structure

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
MAG-ChristopherAtkinson: The use of resolve has been removed: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
MAG-ChristopherAtkinson: The inline documentation has been updated here as it was also missing params: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
MAG-ChristopherAtkinson: Updated here for clarify: [c10e5ce](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/c10e5ce35c45379b636e4e7a42fa29008812476e)

Tests/src/PublicAPITests/MCSExternalDependencyTests.cpp
MAG-ChristopherAtkinson: Agreed, remove here: [fa9f16c](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/fa9f16c4e286c5881453281417ddf5a5ade0affc)

Library/src/Systems/Status/Status.cpp
MAG-ChristopherAtkinson: Move into the namespace here: [d152744](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/d152744d5ac4234db2108c69e0376cc2e0834d76)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: A comment was added with the bulk changes here: [16edb1a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/16edb1ab9f88e970ddd84d29dc298f1782acf727)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: I was just checking, and this introduces a bug; the code assumes that the remainder after the last '/' is the reverse proxy, which is true for all services except multiplayer. I will see about applying a fix, but in this case may be better to use the name of the service as the key.

Library/include/CSP/CSPFoundation.h
MAG-ElliotMorris: Yeah, `CheckPrerequisites` is better.

Library/src/CSPFoundation.cpp
MAG-ElliotMorris: > The self-invoked lambda is just a means to manage scope, the same could have been achieved using braces, this was just a choice due to the utilisation of the nameless namespace

I suppose that's technically fair enough, I can understand reaching for immediately invoked lambdas for complicated variable initialization, although I think it's poor style when an anon-namespace (or as you say, if you only care about scope, inline braces), provides all the same benefits. You create the possibility of an allocation for no reason doing this.

Nonetheless, i'll drop this to [consider] if you wish to reject it.

MAG-ChristopherAtkinson: Sorry, I didn't make that clear. I do agree with the suggestion; however, due to the issue with using the reverse proxy not being reliable, I would like to refactor this out completely. I plan to revisit this later today. 

Unfortunately, I triggered the re-review before I found the issue, but the rest of the changes are suitable for review

MAG-ElliotMorris: Ok, if this code is being deleted feel free to resolve this comment thread as redundant.

Library/src/CSPFoundation.cpp
MAG-ElliotMorris: > Yeah, to confirm this is due to propagating the potential states back to the user, for newer versions from a client perspective would not be too useful, but still has value for third parties.

Trying to interpret this ... you're saying that you added this because you thought it was a good idea to let the user know if they're not using the latest available service version?

I agree, although my question was about whether this behaviour is something that you decided to add yourself, or something from the PRD.

Just going to resolve this thread ... not worth going back and forth on this over a matter of procedure.

Library/src/CSPFoundation.cpp
MAG-ElliotMorris: Makes sense. [Consider] leaving a note comment that we're using the reverse proxy as an identifier just because it's convenient, and not because of any special need to care about the reverse proxy.

MAG-ElliotMorris:
Your name changes fulfil my two requirements 👍 

#1 : That they are distinct from eachother enough to tell the difference
#2 : That `StatusInfo` is changed to say what status or info it is referring to.

Use `ServicesDeploymentStatus ` over `DeployedContainerStatus` imo. The fact that MCS happens to deploy in a container is an implementation detail, and not something we should encode.

Library/src/CSPFoundation.cpp
```diff
@@ -262,6 +265,102 @@ namespace
 
         return MultiplayerServiceURI;
     }
+
+    /// @brief Extracts and returns the last segment of a URI path by splitting it on '/'.
+    /// e.g. 'https://www.magnopus.com/connected-spaces-platform' -> 'connected-spaces-platform'
+    /// @return csp::common::String : empty string if not successful, otherwise the remaining stting.
+    csp::common::String ExtractReverseProxyFromUri(const csp::common::String& URI)
```
MAG-ChristopherAtkinson: [Fix] This function assumes that the current provided URI will have the key required to resolve the service at the end, which is not the case for mag-multiplayer. 

`http://localhost:8081/mag-multiplayer/hubs/v1/multiplayer`, which will resolve as `multiplayer`; not the expected `mag-multiplayer`

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: The name for this class has now been updated based on the proposal: [554088f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/554088f1b57dcf37391e419952eac03fdbff6395)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: These have been updated based on our previous discussion,  please see: [554088f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/554088f1b57dcf37391e419952eac03fdbff6395)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: ContainerVersion has now being updated to just Version as the class provides the additional context: [ec60b9b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/ec60b9bb64349a5ddfe4dd826af29443fd4f0082)

Library/include/CSP/Systems/Status/Status.h
MAG-ChristopherAtkinson: These have been updated based on our previous discussion, please see: [554088f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/554088f1b57dcf37391e419952eac03fdbff6395)

Now it is no longer `InfoInfoInfo` 😆

Library/src/Systems/Status/Status.cpp
MAG-ChristopherAtkinson: Sorry to repeat this again, but these have been updated based on our previous discussion, please see: [554088f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/554088f1b57dcf37391e419952eac03fdbff6395)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: I have refactored this whole section based on this feedback, also added likely way to many comment to really drive the point of these functions: [af46df1](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/af46df10eeb8c60dd61b246c2ac0c5b524cb6312)

Additionally to this, I have resolved the issue with the reverse proxy assumption, but validating based off the available services to identify the reverse proxy in the URI: [4bb202d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/4bb202dfa7e134e81ec4b6440857e2a323f96e05)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: With the refactor, I am hoping that this is much clearer now: [af46df1](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/af46df10eeb8c60dd61b246c2ac0c5b524cb6312)

Library/src/CSPFoundation.cpp
MAG-ChristopherAtkinson: This has been resolved by validating based on the available services to identify the reverse proxy in the URI: :[4bb202d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/4bb202dfa7e134e81ec4b6440857e2a323f96e05)

Hence the assumption is no longer present.

MAG-ChristopherAtkinson:
> Your name changes fulfil my two requirements 👍
> 
> #1 : That they are distinct from eachother enough to tell the difference #2 : That `StatusInfo` is changed to say what status or info it is referring to.
> 
> Use `ServicesDeploymentStatus ` over `DeployedContainerStatus` imo. The fact that MCS happens to deploy in a container is an implementation detail, and not something we should encode.

For transparency, the changes for the naming based on the above discussion can be found here: [554088f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/554088f1b57dcf37391e419952eac03fdbff6395)

MAG-mv (approved):

Library/src/Systems/Status/Status.cpp
MAG-ElliotMorris: > commenting on the filename.

[fix] The file name (Status.h) is still very generic. `ServiceStatus` would do. I'd also move it out of a `Status` folder, I know you mentioned you kept it in the folder for a reason.

> Originally, I included a system for Status; however, this was removed as we currently do not have a use for it, but kept the project structure so we can support it in the future.

But I don't know what this means, we shouldn't be making projects structured around speculative systems we don't have any plans to introduce.

MAG-ChristopherAtkinson: Thank you completely went over my head, will update shortly

Library/src/CSPFoundation.cpp
MAG-ElliotMorris: Better. You're right, too many comments, but I'd always prefer more explanation over less. The beauty of using methods rather than inline code is that the method names say what is happening.

MAG-ElliotMorris (changes requested):
One remaining conversation about changing a filename/location.

Library/src/Systems/Status/Status.cpp
MAG-ChristopherAtkinson: Renamed and moved here: [ce6c9ac](https://github.com/magnopus-opensource/connected-spaces-platform/pull/740/commits/ce6c9ac5223585753795783f988928a66309926b)

MAG-ElliotMorris (approved):

PR #742, opened 2025-07-14
[Ob-4479] Fix: Updating conversation was causing annotation to be lost

MAG-ThomasGreenhalgh:
When updating the conversation message using `ConversationSystemInternal::UpdateConversation` the conversation asset collection metadata was being overwritten with version containing only the new message string. This was resulting in any previously set metadata values pertaining to the annotation to be lost.
The fix is to merge the updated conversation message metadata values with the existing values for the conversation metadata.

There was a similar problem with updating messages (replies) when the message had an annotation. The same fix has been applied in this case.

### DIFF
--- a/Library/src/Systems/Conversation/ConversationSystemHelpers.cpp
+++ b/Library/src/Systems/Conversation/ConversationSystemHelpers.cpp
@@ -268,4 +268,15 @@ multiplayer::MessageInfo GetConversationInfoFromConversationAssetCollection(cons
     return Info;
 }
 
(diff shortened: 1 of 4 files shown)


MAG-ElliotMorris:
Assigning @MAG-mv  as a reviewer as he's far and away the most knowledgeable in this area.

MAG-mv (approved):
Nice fix! And thanks for adding tests.

MAG-ElliotMorris:
Retriggered the timed-out tests.

PR #743, opened 2025-07-14
[OF-1708] feat: Break WebClient Dependencies

MAG-mv:
Please write a short description of what this change does and what problem it solves.
For more information, see our [CONTRIBUTING](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/.github/CONTRIBUTING.md) page.


Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -171,9 +191,8 @@ void UserSystem::Login(const csp::common::String& UserName, const csp::common::S
                 };
 
                 auto* MultiplayerConnection = SystemsManager::Get().GetMultiplayerConnection();
-                MultiplayerConnection->Connect(ErrorCallback, csp::multiplayer::MultiplayerConnection::MakeSignalRConnection(),
```
MAG-mv, line 174: Below changes shouldn't exist in commit [4af12df](https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/4af12dfbdd213b265cd24dfdffa7ba407d913ce9). Our commit hooks did this.

MAG-ElliotMorris: Infuriating isn't it. We should just remove them they seem more trouble than they're worth, and I think half the team is defaulting to skipping the hooks at this point.

Library/src/Common/Web/EmscriptenWebClient/EmscriptenWebClient.h
```diff
@@ -49,9 +32,8 @@ class EmscriptenWebClient : public WebClient
     void SetFileUploadContentFromBuffer(HttpPayload* Payload, const char* Buffer, size_t BufferLength, const csp::common::String& FileName,
         const char* Version, const csp::common::String& MediaType) override;
 
-protected:
```
MAG-mv, line 53: Constructor needs to be public as we have removed friendship

Library/src/Common/Web/WebClient.cpp
```diff
@@ -110,13 +108,7 @@ void WebClient::RefreshIfExpired()
         return;
     }
 
-    if (LoginState == nullptr)
```
MAG-mv, line 116: This isn't needed anymore as we arent using the UserSystem, and these is no point storing the LoginState

Library/src/Common/Web/WebClient.cpp
```diff
@@ -128,35 +120,34 @@ void WebClient::RefreshIfExpired()
         RefreshNeeded = true;
 #endif
 
-        csp::systems::NullResultCallback NullResCallback = [this](const csp::systems::NullResult& NullRes)
-        {
```
MAG-mv, line 135: This chunk looks complicated, but I have just replaced the UserSystem::RefreshSession call with the new IAuthContext::RefreshToken call

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -37,6 +38,28 @@ class WebClient;
...
 namespace csp::systems
 {
+class UserSystem;
+
+// This class exists purely to appease the wrapper generator.
+// IAuthContext was previously implemented by the UserSystem.
+// However, due to limitations with function pointers as parameters, we needed to hide the interface implementation.
+// Now, AuthContext, uses the functionality from the UserSystem to act as the AuthContext.
+CSP_START_IGNORE
+class AuthContext : public csp::common::IAuthContext
```
MAG-ElliotMorris, line 48: [Comment] The correct solution in our current context, absolutely infuriating none the less.

Library/include/CSP/Common/Interfaces/IAuthContext.h
```diff
@@ -0,0 +1,48 @@
...
+#pragma once
+
+#include "CSP/Common/LoginState.h"
+
+/**
+ * @class IAuthContext
+ * @brief Represents an object which holds information about the authentication session state,
+ * and refreshing the refresh token to keep the session in a valid state.
+ *
+ * This was created to seperate systems logic from multiplayer
```
MAG-AdamThorn: [Consider] This comment is a bit of a mouthful which makes it harder to reason about. How about something like:

"This interface decouples systems logic from multiplayer by offering a public interface for WebClient functionality, that doesn't rely on any systems types. This is necessary because the WebClient is utilized in both systems and multiplayer contexts."

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -48,12 +71,12 @@ class CSP_API UserSystem : public SystemBase
     friend class SystemsManager;
     friend class LoginStateResult;
     friend class csp::web::WebClient;
+    friend class AuthContext;
```
MAG-ElliotMorris: [fix] Don't do this, use public functionality. We should prefer making `RefreshSession` public over doing this.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -245,6 +268,8 @@ class CSP_API UserSystem : public SystemBase
     /// @param EventValues std::vector<signalr::value> : event values to deserialise
     CSP_NO_EXPORT void OnAccessControlChangedEvent(const csp::common::NetworkEventData& NetworkEventData);
 
+    CSP_NO_EXPORT AuthContext& GetAuthContext();
```
MAG-ElliotMorris: [fix] I don't think you should use a ref-return here, it's fragile. `AuthContext` is a trivially copyable type, holding just a reference to the UserSystem, you should design it as a value type and use value semantics, rather than making people think about where the memory lives.

Library/src/Systems/SystemsManager.cpp
```diff
@@ -134,8 +134,11 @@ void SystemsManager::CreateSystems()
 #else
     WebClient = new csp::web::POCOWebClient(80, csp::web::ETransferProtocol::HTTPS, LogSystem);
 #endif
-    ScriptSystem = new csp::systems::ScriptSystem();
 
+    UserSystem = new csp::systems::UserSystem(WebClient, NetworkEventBus, *LogSystem);
+    WebClient->SetAuthContext(UserSystem->GetAuthContext());
```
MAG-ElliotMorris: [Comment] A bit annoying the construction ordering here preventing you creating a true type invariant in WebClient. Oh well, I have no suggestions for fixing that.

MAG-ElliotMorris: Other than following our plan, which (probably) would make this construction happen outside at the client level post initialization, thus allowing constructors to be used.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -90,6 +90,13 @@ POCOWebClient::POCOWebClient(const Port InPort, const ETransferProtocol Tp, csp:
     Cookies = new std::remove_pointer_t<decltype(Cookies)>();
 }
 
+POCOWebClient::POCOWebClient(
+    const Port InPort, const ETransferProtocol Tp, csp::common::IAuthContext& AuthContext, csp::common::LogSystem* LogSystem, bool AutoRefresh)
+    : POCOWebClient(InPort, Tp, LogSystem, AutoRefresh)
```
MAG-AdamThorn, line 95: [Question] Why doesn't the POCOWebClient call the WebClient ctor overload that takes the AuthContext as the EmscriptenWebClient does rather than call the SetAuthContext() method inside the ctor?

Library/src/Common/Web/WebClient.h
```diff
@@ -107,6 +103,8 @@ class WebClient
         const char* Version, const csp::common::String& MediaType)
         = 0;
 
+    void SetAuthContext(csp::common::IAuthContext& AuthContext);
```
MAG-AdamThorn, line 110: [Consider] Adding a comment to make it clear when the different WebClient ctors are called and this function.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: I would argue that this class is an extension of UserSystem, so it is appropriate here. However, due to the fact that we can set CSP_NO_EXPORT, I will make it public

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: I have thoughts on this. It would be best to talk in a meet

Tests/src/InternalTests/WebClientTests.cpp
```diff
@@ -78,8 +79,8 @@ class ResponseReceiver : public ResponseWaiter, public csp::web::IHttpResponseHa
 class TestWebClient : public EmscriptenWebClient
 {
 public:
-    TestWebClient(const Port InPort, const ETransferProtocol Tp, csp::common::LogSystem* LogSystem)
-        : EmscriptenWebClient(InPort, Tp, LogSystem, false)
+    TestWebClient(const Port InPort, const ETransferProtocol Tp, csp::common::IAuthContext& AuthContext, csp::common::LogSystem* LogSystem)
```
MAG-AdamThorn: [Consider] Both flavours of the TestWebClient use the ctor that takes the AuthContext as an argument. Neither tests construction of a web client where we instead set the auth context via the WebClient::SetAuthContext() method. I think we should add tests to also validate that flow.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: I know we discussed other methods, but they required a large refactor and added complexity due to the way LoginResult sets variables on the LoginState. I fixed it by making AuthContext private:

https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/f7f8de8ba023c55cc554e092086468cb7bf7b2b9

Library/include/CSP/Common/Interfaces/IAuthContext.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/8b195b866444da437590bd352a7ec960887b8f59

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ElliotMorris: It's strange to me that this works, I can't say I've ever declared a private class like this, but one would assume it's definition would be inaccessible to other types, which confuses me as we pass it to the WebClient so it must exist in the interface.

[Question] What's that all about? Is this friendship abuse by the backdoor? Do we have `friend UserSystem` on these webclients or something?

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-mv: I created the additional call on POCO as it has a lot of setup , so I didnt want to duplicate all of the code.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Removed friendship after our conversation: https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/bd7c046106cfb32c49672065c360d129689adf18

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/16fe5d959830e0ade398a24b8517144f44f39bb4

Library/src/Common/Web/WebClient.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/bd7c046106cfb32c49672065c360d129689adf18

Tests/src/InternalTests/WebClientTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/743/commits/a25a66404fb5a2011275daa26d0830e28f42fc19

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ElliotMorris: Just to capture, I'd not fully followed why this was necessary, the dependency breaking concerns overlap with the difficulty in copying virtual types, i'd sort of been forgetting virtual constructors don't exist.

MAG-ElliotMorris (approved):

PR #746, opened 2025-07-16
[OF-1710] feat: expose public api for user service application settings

MAG-ChristopherAtkinson:
In support of feature flagging, CSP will need to expose the current application setting RESTful service being provided by MCS. The PR exposes the following endpoint through the implementation of a new Application Settings System:

- https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindByContext
- https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindByContextAnonymous
- https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindAllContexts

### DIFF
--- a/Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
+++ b/Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
@@ -96,6 +96,7 @@ private:
     ~ApplicationSettingsSystem();
 
     /// @brief Asynchronously creates or updates application settings for a specific context.
+    /// This function is made private because it requires elevated permissions and is intended to be used only by internal tests.
     /// Elevated Permissions Required: This function requires elevated permissions to execute successfully. If the user does not possess any of the
     /// necessary roles, the operation will fail with a 403 Forbidden error.
     /// @param ApplicationSettings ApplicationSettings& : The settings object containing application name, context, and key-value pairs to be stored.


Library/include/CSP/Systems/Settings/ApplicationSettings.h
```diff
@@ -0,0 +1,136 @@
...
+    /// @brief The name of the application.
+    /// Used to identify the application this settings instance is associated with.
+    csp::common::String ApplicationName;
+
+    /// @brief A context identifier for the settings.
+    /// This can be used to scope or separate settings across different environments or modules.
+    csp::common::String Context;
+
+    /// @brief Flag indicating whether anonymous access is allowed.
+    /// If true, the application can operate without user authentication.
```
MAG-AdamThorn: This doesn't impact operation of the application. If true an anonymous user could access these settings.

MAG-ChristopherAtkinson: Update the inline documentation here: [df12983](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/df1298321ed60b72be23d770a51d5d0ac4336453)

MAG-ChristopherAtkinson:
> In the process of writing tests, I have run into the following issues:
> 
> * [FindByContextAnonymous](https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindByContextAnonymous) was failing to resolve the tenant on the local CHS:
>   
>   * This seems to have been resolved since yesterday, when I had brought up a dialog with @jeromagnopus
> * [FindByContextAnonymous](https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindByContextAnonymous) is returning a '404 Not Found' when intentionally failing the request to validate the failure path:
>   
>   * This is not documented in the Swagger documentation, which suggests that the correct response should be '204 No Content'
> * [FindByContext](https://docs.magnopus-mgmt.cloud/#/ApplicationSettings/ApplicationSettings_FindByContext) is returning a '200 Response OK', with a partial data set when intentionally failing the request to validate the failure path:
>   
>   * The Swagger documentation specifies '204 No Content', which is what I would expect in this case
> 
> Given this information, we need further dialogue with CHS to ensure the expectations of these endpoints.

These have now been resolved and return a standardized and expected output from the services, if you are using local CHS you may need to pull the latest changes. 

Please see changes here: [0e37ecb](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/0e37ecb8a1e6d34ebddf09e5e7acf1723c56d289)

Library/include/CSP/Systems/Settings/ApplicationSettings.h
```diff
@@ -0,0 +1,104 @@
...
+
+    /// @brief A key-value store of arbitrary application settings.
+    /// The keys and values are both strings and represent additional configurable options.
+    csp::common::Map<csp::common::String, csp::common::String> Settings;
+};
+
+/// @brief Holds a list of application settings contexts.
+/// This class is used to manage and represent multiple configuration contexts
+/// that can be associated with an application.
+class CSP_API ApplicationSettingsContexts
```
MAG-AdamThorn: [Question] Is this used anywhere? It does not appear to be. Likely left over from when we had that additional endpoint. Please remove if not required.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -0,0 +1,94 @@
...
+    /** @cond DO_NOT_DOCUMENT */
+    friend class SystemsManager;
+    /** @endcond */
+    CSP_END_IGNORE
+
+public:
+    /// @brief Asynchronously creates or updates application settings for a specific context.
+    /// @param ApplicationSettings ApplicationSettings& : The settings object containing application name, context, and key-value pairs to be stored.
+    /// @param Callback ApplicationSettingsResultCallback : Callback when asynchronous task finishes.
+    CSP_NO_EXPORT CSP_ASYNC_RESULT void CreateSettingsByContext(
```
MAG-AdamThorn: [Question] Why are we exposing this endpoint? The TDD does not have a create endpoint and the PRD specifies Read-Only.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -0,0 +1,94 @@
...
+        const ApplicationSettings& ApplicationSettings, ApplicationSettingsResultCallback Callback);
+
+    /// @brief Asynchronously retrieves application settings for a specific context.
+    /// @param ApplicationName csp::common::String : The name of the application for which settings are requested.
+    /// @param Context csp::common::String : The specific context whose settings should be retrieved.
+    /// @param Keys csp::common::Optional<csp::common::String> : List of setting keys to retrieve. If not provided, all settings for the context are
```
MAG-AdamThorn: [fix] It is an optional array rather than a list.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -0,0 +1,342 @@
...
+    auto ApplicationSettings = ::ApplicationSettings();
+    ApplicationSettings.ApplicationName = "MAG_APPLICATION_SETTINGS_TESTS";
+    ApplicationSettings.Context = "MAG_APPLICATION_SETTINGS_CONTEXT_TESTS";
+    ApplicationSettings.AllowAnonymous = false;
+    ApplicationSettings.Settings = Settings;
+
+    return ApplicationSettings;
+}
+
+ApplicationSettings GetApplicationSettingsAnonymousTestData()
```
MAG-AdamThorn: [Question] is there a reason we aren't just passing an allowAnonymous argument to the function and then we wouldn't need a second utility function?

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -105,6 +105,18 @@ void LogInAsNewTestUser(csp::systems::UserSystem* UserSystem, csp::common::Strin
     LogIn(UserSystem, OutUserId, NewTestUser.Email, GeneratedTestAccountPassword, AgeVerified, ExpectedResultCode, ExpectedResultFailureCode);
 }
 
+void LogInAsLocalAdminUser(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, bool AgeVerified)
+{
+    // Attempt to log in with an account with elevated permissions.
```
MAG-AdamThorn: [Question] I am concerned that someone evaluating our library using Application Security Testing tools, especially those that detect hard coded secrets are going to flag this. Is there another option rather than hardcoding it here?

Also can the comment be clearer? And by clearer I mean make it 'clear' that there is no value in someone naughty grabbing these secrets.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-AdamThorn: I see now (and am remembering earlier conversations) that this is required for testing local chs. Do we have any concerns about clients using this as an opaque datastore which may impact interop? This is not a leading question, I am not sure.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+void ApplicationSettingsSystem::GetSettingsByContext(
+    const String& ApplicationName, const String& Context, const Optional<Array<String>>& Keys, ApplicationSettingsResultCallback Callback)
+{
+    GetSettingsByContext(ApplicationName, Context, Keys)
+        .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<ApplicationSettingsResult>(Callback,
+            "ApplicationSettingsSystem::GetSettingsByContext successfully retrieved application settings", "Failed to get application settings", {},
+            {}, {}))
+        .then(systems::continuations::SendResult(Callback, "Successfully retrieved application settings."))
```
MAG-AdamThorn: [Consider] This is going to log the same information twice on successful retrieval. I would consider removing the second log.

Same applies to the second getter below and the create endpoint above.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -0,0 +1,342 @@
...
+        RAIIMockLogger MockLogger {};
+        csp::systems::SystemsManager::Get().GetLogSystem()->SetSystemLevel(LogLevel::Log);
+
+        // Set an expectation that the mock logger will receive message for a successful result retrieval of settings by context
+        const String GetSettingsByContextMsg = "ApplicationSettingsSystem::GetSettingsByContext successfully retrieved application settings";
+        EXPECT_CALL(MockLogger.MockLogCallback, Call(GetSettingsByContextMsg)).Times(1);
+
+        // Set an expectation that the mock logger will receive message for a successful result retrieval
+        const String SendResultMsg = "Successfully retrieved application settings.";
+        EXPECT_CALL(MockLogger.MockLogCallback, Call(SendResultMsg)).Times(1);
```
MAG-AdamThorn: [nit] I added a comment on the ApplicationSettingsSystem. Two almost identical logs here feels spammy to me.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+    }
+
+    // Log out
+    LogOut(UserSystem);
+}
+
+}
+
```
MAG-AdamThorn: [fix] We do not have any tests for the new CreateSettingsByContext method. This endpoint is only available to those with elevated permissions, is that correct? We should confirm the correct behaviour for those attempting to use it without elevated permissions. We should also test edge cases for those with elevated permissions such as passing an invalid context object.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-AdamThorn: This endpoint is only available to those with elevated permissions right? We should call that out in the doxygen comments along with the error they will receive if the user does not have the appropriate permissions.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -0,0 +1,342 @@
...
+        EXPECT_EQ(ApplicationSettings.AllowAnonymous, ApplicationSettingsTestData.AllowAnonymous);
+        EXPECT_EQ(ApplicationSettings.ApplicationName, ApplicationSettingsTestData.ApplicationName);
+        EXPECT_EQ(ApplicationSettings.Context, ApplicationSettingsTestData.Context);
+        EXPECT_EQ(ApplicationSettings.Settings.Size(), Keys.Size());
+        EXPECT_EQ(ApplicationSettings.Settings[Keys[0]], "TestData_3");
+    }
+}
+
+CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidSettingsByContextAnonymousTest)
```
MAG-AdamThorn, line 375: [fix] We should add another test here in which the ApplicationName and Context are valid but the Tenant is not.

Library/include/CSP/Systems/Settings/ApplicationSettings.h
MAG-ChristopherAtkinson: This was left over when I removed the previous changes. Apologies, this should be promoted to fix

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: Yeah, that is a good point. I could make this private and expose it to the tests through friendship?

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: The structure only allows for a context to be anonymous or private, so we need to separate context, but there is no reason why we can't pass the name as well as allowAnonymous

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: We could provide these through the use of an environment variable, for example, but would require all team members and the CI to adopt the change. Happy to change if anyone has further suggestions

Library/include/CSP/Systems/Settings/ApplicationSettings.h
```diff
@@ -0,0 +1,136 @@
...
+{
+
+/// @ingroup Application Settings System
+/// @brief Represents configuration settings for an application context.
+class CSP_API ApplicationSettings
+{
+public:
+    /// @brief The name of the application.
+    /// Used to identify the application this settings instance is associated with.
+    csp::common::String ApplicationName;
```
MAG-mv, line 46: [Question] What would an example of an application name be? Is this being set by clients? I'll probably find out later in the PR

Library/src/Systems/Settings/ApplicationSettings.cpp
```diff
@@ -0,0 +1,94 @@
...
+    if (Dto.HasContext())
+        ApplicationSettings.Context = Dto.GetContext();
+
+    if (Dto.HasAllowAnonymous())
+        ApplicationSettings.AllowAnonymous = Dto.GetAllowAnonymous();
+
+    if (Dto.HasSettings())
+    {
+        for (const auto& Pair : Dto.GetSettings())
+        {
```
MAG-mv: [Consider] We have convert functions in Convert.h to convert between csp and mcs types

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-mv: The type Array is also missing from the left-hand side of the :

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-mv: Imo this should be a [fix]

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-mv: Agreed. We cant have this in the codebase

Library/include/CSP/Systems/Settings/ApplicationSettings.h
MAG-ChristopherAtkinson: The application name is an arbitrary string used for routing, same as context. As a result, this is upto the consumer to decide what would be most suitable from their end. 

An example of such a route could be `/mag-user/api/v1/applications/magnopus/settings/public`

Library/include/CSP/Systems/Settings/ApplicationSettings.h
MAG-ChristopherAtkinson: removed in commit: [0ae7d55](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/0ae7d556801e2b361b9dae143fb0d52b04cb5a39)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: for now, I have just removed the async, for some reason I thought this was required. Looking back the duplicate definition, made no sense: [a4d5710](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/a4d571039cb88079f62966408f84cf8876a26684)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: great catch, i missed then when loop back over my comments: [92e5987](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/92e598770a3695f034d10643239c43ce26d0c46e)

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: Yeah the message where included because of the signature for sendResult, just directly handle the callback now: [15f5806](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/15f58065f87591a23b5f3b0f3617fa2996fa94b8)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: same as above: [15f5806](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/15f58065f87591a23b5f3b0f3617fa2996fa94b8)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: I have added some additional test coverage for the create function here: [96ef1ed](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/96ef1ed78e67deaed824bac855ed0c312eba93d3)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: Added a test for invalid tenant: [f0c6c2a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/f0c6c2a5a7d9897b28250ab2e1e9b2706eb0545c)

Library/src/Systems/Settings/ApplicationSettings.cpp
MAG-ChristopherAtkinson: Yeah I have used this previous, just forgot it existed updated here: [74f6401](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/74f64015b6296c3ac4ad3934735cc9b758ec6163)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: I have combined them into a single function allowing for Context and AllowAnonymous to be configured: [5d1f253](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/5d1f25302edd97742e6177263a5bd1f748e69ae0)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-AdamThorn: I am happy with that change. @MAG-mv please resolve if you are happy too.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -0,0 +1,95 @@
...
+    /** @endcond */
+    CSP_END_IGNORE
+
+public:
+    /// @brief Asynchronously creates or updates application settings for a specific context.
+    /// Elevated Permissions Required: This function requires elevated permissions to execute successfully. If the user does not possess any of the
+    /// necessary roles, the operation will failwith a 403 Forbidden error.
+    /// @param ApplicationSettings ApplicationSettings& : The settings object containing application name, context, and key-value pairs to be stored.
+    /// @param Callback ApplicationSettingsResultCallback : Callback when asynchronous task finishes.
+    CSP_NO_EXPORT void CreateSettingsByContext(const ApplicationSettings& ApplicationSettings, ApplicationSettingsResultCallback Callback);
```
MAG-mv: [fix] If this isn't meant to be called by regular clients, we should make this private and use friend access for our tests

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-AdamThorn: You could just construct the context name depending on whether you passed true of false in for the AllowAnonymous argument.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: the creds are only seeded as part of local CHS, we discussed this as part of the weekly sync and agreed that we could just populate these directly; however, I was also hesitant against this. If you have any suggestions for a better method, I would appreciate it.

MAG-AdamThorn: What are we using to define endpoint url/s for local CHS, environment variables right?

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-AdamThorn: @MAG-ChristopherAtkinson I think your idea of making it private and exposing it to tests via friendship is the right way to go.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: Yeah, we have a single environment variable that configures the URI for local CHS, if we took this approach we would need to introduce two: MAGNOPUS_SERVICES_EMAIL, and MAGNOPUS_SERVICES_PASSWORD

This would then need to be populated on each developer machine and on the CI instances, which to be clear I am not against, just wanted to highlight that it invoke additional changes

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -102,6 +102,142 @@ void AttemptToSeedLocalCHSApplicationSettings()
...
+    {
+        RAIIMockLogger MockLogger {};
+        csp::systems::SystemsManager::Get().GetLogSystem()->SetSystemLevel(LogLevel::Log);
+
+        // Set an expectation that the mock logger will receive message for a successful creation of settings by context
+        const String GetSettingsByContextMsg = "ApplicationSettingsSystem::CreateSettingsByContext successfully created application settings";
+        EXPECT_CALL(MockLogger.MockLogCallback, Call(GetSettingsByContextMsg)).Times(1);
+
+        // Set an expectation that the mock logger will receive message for a successful creation
+        const String SendResultMsg = "Successfully created application settings.";
```
MAG-AdamThorn: [fix] This second log needs to be removed as previously discussed.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: actually likely want them to be MAGNOPUS_SERVICES_ADMIN_EMAIL, and MAGNOPUS_SERVICES_ADMIN_PASSWORD

MAG-ChristopherAtkinson: and if this is the case we can remove the branching logic in the tests as we can provide a support account for the team city instance

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: perfect, if this is the case, guess the current nameless namespace function will no longer function

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -102,6 +102,142 @@ void AttemptToSeedLocalCHSApplicationSettings()
...
+        EXPECT_EQ(ApplicationSettings.ApplicationName, ApplicationSettingsTestData.ApplicationName);
+        EXPECT_EQ(ApplicationSettings.Context, ApplicationSettingsTestData.Context);
+        EXPECT_EQ(ApplicationSettings.Settings, ApplicationSettingsTestData.Settings);
+    }
+
+    // Log out
+    LogOut(UserSystem);
+}
+
+CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, CreateSettingsByContextAnonymousTest)
```
MAG-AdamThorn: [fix] How is this test anonymous? You are logging in as the local admin user with elevated permissions and calling the same endpoint as above. The concept of anonymity doesn't exist in the context of application settings creation.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-AdamThorn: Same for the logging below.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: Which branching logic are you referring to?

Can you think of any alternate approaches worth considering?

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -476,3 +476,31 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidSettingsByC
...
+        csp::systems::SystemsManager::Get().GetLogSystem()->SetSystemLevel(LogLevel::Log);
+
+        // Set an expectation that the mock logger will receive message for a failed result 404 no payload/error message.
+        const String GetRequestErrorMsg = "has returned a failed response (404) but with no payload/error message.";
+        EXPECT_CALL(MockLogger.MockLogCallback, Call(testing::HasSubstr(GetRequestErrorMsg))).Times(1);
+
+        const String ErrorMsg = "Failed to get application settings";
+        EXPECT_CALL(MockLogger.MockLogCallback, Call(ErrorMsg)).Times(1);
+
+        auto [Result] = AWAIT(ApplicationSettingsSystem, GetSettingsByContextAnonymous, "", ApplicationSettingsTestData.ApplicationName,
```
MAG-AdamThorn: [fix] This test raises an interesting point. Testing `Tenant == ""` is different to testing `Tenant == "garbage"`. I think we should be guarding against the former and then forwarding the error response from CHS for the latter - what do you think? Allowing clients to pass an empty string makes no sense and is a wasted call to my mind.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,137 @@
...
+            {}, {}, {}))
+        .then([Callback](const ApplicationSettingsResult& Result) { Callback(Result); })
+        .then(common::continuations::InvokeIfExceptionInChain(
+            [Callback](const std::exception& /*exception*/) { Callback(MakeInvalid<ApplicationSettingsResult>()); }, *LogSystem));
+}
+
+void ApplicationSettingsSystem::GetSettingsByContext(
+    const String& ApplicationName, const String& Context, const Optional<Array<String>>& Keys, ApplicationSettingsResultCallback Callback)
+{
+    GetSettingsByContext(ApplicationName, Context, Keys)
```
MAG-AdamThorn, line 51: [Question] I added another comment to one of your new tests that relates to this. Should we be checking that users are not passing empty strings for both ApplicationName and Context? Would be pointless to make the request to CHS if either were empty.

MAG-ChristopherAtkinson:
I am currently expecting the CI to fail as the credentials are now configured through environment variables [b3452e2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/b3452e235e752505fb31a3a3007b139e6c12bbee)

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: I have promoted the credentials to environment variables; this will currently make the CI fail. because we are now providing a means to configure, we no longer need to branch the logic for the test. [b3452e2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/b3452e235e752505fb31a3a3007b139e6c12bbee)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: I have demoted the access modifier for this function and added friendship for test support [5cd97d9](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/5cd97d9982f2ecd416a80744cee8ca7bce90d197)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -81,6 +95,13 @@ class CSP_API ApplicationSettingsSystem : public SystemBase
     CSP_NO_EXPORT ApplicationSettingsSystem(csp::web::WebClient* InWebClient, csp::common::LogSystem& LogSystem);
     ~ApplicationSettingsSystem();
 
+    /// @brief Asynchronously creates or updates application settings for a specific context.
+    /// Elevated Permissions Required: This function requires elevated permissions to execute successfully. If the user does not possess any of the
+    /// necessary roles, the operation will failwith a 403 Forbidden error.
```
MAG-AdamThorn: [fix] Missing space 'failwith' > 'fail with'

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -315,11 +312,25 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidSettingsByC
...
     auto& SystemsManager = csp::systems::SystemsManager::Get();
+    auto* UserSystem = SystemsManager.GetUserSystem();
     auto* ApplicationSettingsSystem = SystemsManager.GetApplicationSettingsSystem();
 
+    // Seed application settings test data
+    {
+        csp::common::String UserId;
+        LogInAsAdminUser(UserSystem, UserId);
+
+        auto ApplicationSettings = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] spelling mistake: ANONTMOUS > ANONYMOUS

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -315,11 +312,25 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidSettingsByC
...
+        auto [Result] = AWAIT(ApplicationSettingsSystem, CreateSettingsByContext, ApplicationSettings);
+
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+        // Log out
+        LogOut(UserSystem);
+    }
+
     auto const ApplicationSettingsTestData = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -348,11 +359,25 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetSettingsByContextA
...
     auto& SystemsManager = csp::systems::SystemsManager::Get();
+    auto* UserSystem = SystemsManager.GetUserSystem();
     auto* ApplicationSettingsSystem = SystemsManager.GetApplicationSettingsSystem();
 
+    // Seed application settings test data
+    {
+        csp::common::String UserId;
+        LogInAsAdminUser(UserSystem, UserId);
+
+        auto ApplicationSettings = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -348,11 +359,25 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetSettingsByContextA
...
+        auto [Result] = AWAIT(ApplicationSettingsSystem, CreateSettingsByContext, ApplicationSettings);
+
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+        // Log out
+        LogOut(UserSystem);
+    }
+
     auto const ApplicationSettingsTestData = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -460,7 +443,7 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidTentantSett
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto* ApplicationSettingsSystem = SystemsManager.GetApplicationSettingsSystem();
 
-    auto ApplicationSettingsTestData = GetApplicationSettingsAnonymousTestData();
+    auto ApplicationSettingsTestData = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -163,7 +146,7 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, CreateSettingsByConte
     if (UserId.IsEmpty())
         return;
 
-    auto ApplicationSettingsTestData = GetApplicationSettingsAnonymousTestData();
+    auto ApplicationSettingsTestData = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -82,15 +65,15 @@ void AttemptToSeedLocalCHSApplicationSettings()
...
+        auto ApplicationSettings = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_CONTEXT_TESTS", false);
         auto [Result] = AWAIT(ApplicationSettingsSystem, CreateSettingsByContext, ApplicationSettings);
 
         EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
     }
 
     // Seed anonymous application settings test data
     {
-        auto ApplicationSettings = GetApplicationSettingsAnonymousTestData();
+        auto ApplicationSettings = GetApplicationSettingsTestData("MAG_APPLICATION_SETTINGS_ANONTMOUS_CONTEXT_TESTS", true);
```
MAG-AdamThorn: [fix] Same spelling mistake as above

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -49,19 +49,14 @@ ApplicationSettings GetApplicationSettingsTestData(const csp::common::String& Co
...
+void SeedApplicationSettings()
 {
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto* UserSystem = SystemsManager.GetUserSystem();
     auto* ApplicationSettingsSystem = SystemsManager.GetApplicationSettingsSystem();
 
     csp::common::String UserId;
-    LogInAsLocalAdminUser(UserSystem, UserId);
-
-    // In the event that login fails, we do not want to attempt to seed the service with application settings.
```
MAG-AdamThorn: [Question] Why have you removed this comment? Login as admin user will still fail when not attempting to login to local CHS will it not?

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: This has been renamed to 'CreateAnonymousSettingsByContextTest'

Tests/src/PublicAPITests/UserSystemTestHelpers.h
```diff
@@ -32,7 +32,9 @@ void LogInAsNewTestUser(csp::systems::UserSystem* UserSystem, csp::common::Strin
     csp::systems::EResultCode ExpectedResultCode = csp::systems::EResultCode::Success,
     csp::systems::ERequestFailureReason ExpectedResultFailureCode = csp::systems::ERequestFailureReason::None);
 
-void LogInAsLocalAdminUser(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, bool AgeVerified = true);
+void LogInAsAdminUser(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, bool AgeVerified = true,
```
MAG-AdamThorn, line 35: [Question] Why have you removed 'Local' from the name? I wonder if we should be making it super clear that you are logging in with local chs only admin credentials.

MAG-AdamThorn (changes requested):

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: These have already been remove, and look like these comments are from stale changes

Tests/src/PublicAPITests/UserSystemTestHelpers.h
MAG-ChristopherAtkinson: because this is now configurable through environment vars, it has no concept of local

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: It will fail for a valid reason now, such as a 403 Forbidden error. The concept of local has been removed as we can configure these creds to target any environment external to the source

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: Apologies, this seems to be a copy-paste error

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: The branching logic was concerned with seeding the initial test data, as we can configure a support account for any service on the CI. and/or local development, we no longer need to validate whether we are using local CHS. This means that the login as admin function can provide additional validation

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: Yes, agreed, when going through the changes for this, I realised that we cannot test for garbage data, as all strings are valid as they are simply used for routing

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: fixed all of these in a bulk change found here: [8afb48f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/8afb48f986e58ef1a7a4cdb712f5a49b32656561)

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: Updated here: [a9e4e94](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/a9e4e94514d8d97920a5abe8cd5166b8446a4857)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: I have included guard for the methods for empty string values found here: [db6fea8](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/db6fea8013ef552088a51a2c810e80d17ea4ddac)

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: I have included guard for the methods for empty string values found here: [db6fea8](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/db6fea8013ef552088a51a2c810e80d17ea4ddac)

^ repeat for transparency, we can treat the other conversation as the main thread

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-mv: What was the outcome of this?

MAG-ChristopherAtkinson: For transparency, the change where made here: [5cd97d9](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/5cd97d9982f2ecd416a80744cee8ca7bce90d197)

With the caveat that we cannot friend the nameless namespace function, hence the slight duplication.

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: see above ^

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: see above ^

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
MAG-ChristopherAtkinson: see above ^

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -0,0 +1,116 @@
...
+    ApplicationSettingsSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
+    CSP_NO_EXPORT ApplicationSettingsSystem(csp::web::WebClient* InWebClient, csp::common::LogSystem& LogSystem);
+    ~ApplicationSettingsSystem();
+
+    /// @brief Asynchronously creates or updates application settings for a specific context.
+    /// Elevated Permissions Required: This function requires elevated permissions to execute successfully. If the user does not possess any of the
+    /// necessary roles, the operation will fail with a 403 Forbidden error.
+    /// @param ApplicationSettings ApplicationSettings& : The settings object containing application name, context, and key-value pairs to be stored.
+    /// @param Callback ApplicationSettingsResultCallback : Callback when asynchronous task finishes.
+    CSP_NO_EXPORT void CreateSettingsByContext(const ApplicationSettings& ApplicationSettings, ApplicationSettingsResultCallback Callback);
```
MAG-AdamThorn, line 104: [Consider] Adding a comment to say why this method is private rather than public.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {
+        CSP_LOG_ERROR_MSG(
+            "The retrieval of an application setting was issued without a application name. You have to provide an asset application name.");
+        OnCompleteEvent->set_exception(std::make_exception_ptr(std::exception()));
+        return OnCompleteTask;
+    }
+
+    if (Context.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG("The retrieval of an application setting was issued without a context. You have to provide an asset context.");
```
MAG-AdamThorn: [fix] Same as above/ I would also consider rewording this log, perhaps:

"Your request for application settings was made without a valid context. A valid application context must be provided"

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+async::task<ApplicationSettingsResult> ApplicationSettingsSystem::GetSettingsByContext(const csp::common::String& ApplicationName,
+    const csp::common::String& Context, const csp::common::Optional<csp::common::Array<csp::common::String>>& Keys)
+{
+    auto OnCompleteEvent = std::make_shared<async::event_task<ApplicationSettingsResult>>();
+    async::task<ApplicationSettingsResult> OnCompleteTask = OnCompleteEvent->get_task();
+
+    if (ApplicationName.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG(
+            "The retrieval of an application setting was issued without a application name. You have to provide an asset application name.");
```
MAG-AdamThorn: [fix] Why does your comment mention an 'asset'? I also think the message could be clearer, consider:

"Your request for application settings was made without a valid application name. A valid application name must be provided"

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {
+        CSP_LOG_ERROR_MSG(
+            "The creation of an application setting was issued without a application name. You have to provide an asset application name.");
+        OnCompleteEvent->set_exception(std::make_exception_ptr(std::exception()));
+        return OnCompleteTask;
+    }
+
+    if (ApplicationSettings.Context.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG("The creation of an application setting was issued without a context. You have to provide an asset context.");
```
MAG-AdamThorn: [fix] Same as comments below re use of the term 'asset' and clarity of the message.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+async::task<ApplicationSettingsResult> ApplicationSettingsSystem::CreateSettingsByContext(const ApplicationSettings& ApplicationSettings)
+{
+    auto OnCompleteEvent = std::make_shared<async::event_task<ApplicationSettingsResult>>();
+    async::task<ApplicationSettingsResult> OnCompleteTask = OnCompleteEvent->get_task();
+
+    if (ApplicationSettings.ApplicationName.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG(
+            "The creation of an application setting was issued without a application name. You have to provide an asset application name.");
```
MAG-AdamThorn: [fix] Same as comments below re use of the term 'asset' and clarity of the message.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+    const csp::common::String& ApplicationName, const csp::common::String& Context,
+    const csp::common::Optional<csp::common::Array<csp::common::String>>& Keys)
+{
+    auto OnCompleteEvent = std::make_shared<async::event_task<ApplicationSettingsResult>>();
+    async::task<ApplicationSettingsResult> OnCompleteTask = OnCompleteEvent->get_task();
+
+    if (ApplicationName.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG(
+            "The retrieval of an application setting was issued without a application name. You have to provide an asset application name.");
```
MAG-AdamThorn: [fix] Same comment as above re use of term 'asset' and clarity of the log message.

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
```diff
@@ -0,0 +1,182 @@
...
+        OnCompleteEvent->set_exception(std::make_exception_ptr(std::exception()));
+        return OnCompleteTask;
+    }
+
+    if (Context.IsEmpty())
+    {
+        CSP_LOG_ERROR_MSG("The retrieval of an application setting was issued without a context. You have to provide an asset context.");
```
MAG-AdamThorn: [fix] Same comment as above re use of term 'asset' and clarity of the log message.

Tests/src/PublicAPITests/UserSystemTestHelpers.h
MAG-AdamThorn: So I could provide admin credentials that work on the OKO tenant for example and then use this endpoint there?

MAG-AdamThorn (changes requested):

Tests/src/PublicAPITests/UserSystemTestHelpers.h
MAG-ChristopherAtkinson: yes, you can provide creds for any tenants for testing, as long as they match with MAGNOPUS_SERVICES_ENDPOINT URI environment variable

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: 'This function is made private because it requires elevated permissions and is intended to be used only by internal tests. ' <- something in the vein?

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: I have updated the message here: [8e8f106](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/8e8f106e3c19969f721ad98b8574993ef96672b7)

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Settings/ApplicationSettingsSystem.cpp
MAG-ChristopherAtkinson: The comment includes assets because I messed up, an honest mistake and has been corrected

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: updated here, happy to change if you feel it is not suitable: [712b636](https://github.com/magnopus-opensource/connected-spaces-platform/pull/746/commits/712b636311fcf559dff061a8406d362819891f98)

PR #747, opened 2025-07-17
[NT-0] ci: Update the PR Guidance workflow to support forks

MAG-AdamThorn:
When a pull request is created from a fork, GitHub Actions workflows run in a more restricted context for security reasons. The `pull_request_target` event is used to give the workflow the appropriate write permissions and runs the workflow in the context of the base repository (the one the PR is targeting).

The `pull_request_target` event runs with elevated permissions so we must be careful with its use in workflows. We can mitigate this risk by not allowing workflows with this event to check out and run untrusted code from the pull request or to run untrusted scripts. 

This change means that when a PR is raised from a fork, the PR Guidance comment will be added.

### DIFF
--- a/.github/workflows/add-pr-guidance.yml
+++ b/.github/workflows/add-pr-guidance.yml
@@ -1,14 +1,16 @@
 # Name of the workflow as it appears in the GitHub Actions tab
 name: Add PR Review Guidance Comment
 
-# Trigger the workflow when a pull request is opened / reopened
+# Trigger the workflow when a pull request is opened / reopened - this will run in the context of the base repository
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, reopened]
 
 # Grant write access to pull requests (this is necessary to add comments)
 permissions:
   pull-requests: write
+  # Comments are considered issues in GitHub
+  issues: write
 
 # Define a single job named 'add-guidance-comment'
 jobs:


MAG-AdamThorn:
> can you provide an example run of the workflow on a fork (for new behaviour) and on a branch (for non-regression)?

That is a good suggestion and it made me realise the updated workflow had not run as part of this PR as I would have expected (there is no commit with the PR Guidance) so I did some investigation.

Investigating this it is expected behaviour. When you create a pull request with a new or edited workflow that uses the 'pull_request_target' the workflow will not trigger on that pull request because the new or changed YAML file is not yet in the base branch. This behavior is a security measure. The `pull_request_target` trigger uses a GITHUB_TOKEN with write permissions to the repository. Someone could therefore raise a PR with a tampered workflow file and have it run immediately with those permissions - stealing secrets or making unauthorized changes.

So unfortunately in this case we will have to merge this change and then I can confirm that it works correctly in all cases. To recap, the changes are as follows:

`pull_request_target`
A workflow trigger that runs in the context of the base repository's ref (usually the default branch) rather than the pull request's branch.

`issues: write`
The issues: write permission is used to grant the workflow the ability to create, update, or delete comments on a pull request.

MAG-ChristopherAtkinson (approved):

PR #748, opened 2025-07-21
[OB-4396] fix: create space fails if an empty array of user invites is passed

MAG-ChristopherAtkinson:
During https://github.com/magnopus-opensource/connected-spaces-platform/pull/675, an edge case was missed where the CSP would guarantee that the provided invite list was not empty. This PR provides the following changes:

 - Reintroduce the validation for the edge case in  'SpaceSystem::BulkInviteUsersToSpaceIfNeccesary'
 - Add explicit test for empty bulk invites in create space

### DIFF
--- a/Tests/src/PublicAPITests/SpaceSystemTests.cpp
+++ b/Tests/src/PublicAPITests/SpaceSystemTests.cpp
@@ -503,6 +503,43 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, CreateSpaceWithBulkInviteTest)
     LogOut(UserSystem);
 }
 
+CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, CreateSpaceWithEmptyBulkInviteTest)
+{
+    SetRandSeed();
+
+    auto& SystemsManager = ::SystemsManager::Get();
+    auto* UserSystem = SystemsManager.GetUserSystem();
+    auto* SpaceSystem = SystemsManager.GetSpaceSystem();
+
+    const char* TestSpaceName = "CSP-UNITTEST-SPACE-MAG";
+    const char* TestSpaceDescription = "CSP-UNITTEST-SPACEDESC-MAG";
+
+    char UniqueSpaceName[256];
+    SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+    String UserId;
+
+    // Log in
+    LogInAsNewTestUser(UserSystem, UserId);
+
+    // Create space
+    ::Space Space;
+    InviteUserRoleInfoCollection InviteUsers;
+    CreateSpace(SpaceSystem, UniqueSpaceName, TestSpaceDescription, SpaceAttributes::Private, nullptr, InviteUsers, nullptr, nullptr, Space);
+
+    auto [GetInvitesResult] = AWAIT_PRE(SpaceSystem, GetPendingUserInvites, RequestPredicate, Space.Id);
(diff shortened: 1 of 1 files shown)


MAG-ElliotMorris:
[Question] In reference to your comment on Jira. 

If MCS is returning an error 400 for this, and not the error 204 as we would expect, would the reasonable thing to do not be either:
- Update [our code](https://github.com/magnopus-opensource/connected-spaces-platform/blob/0ddd6163e1e8dfb932e79744e7170a0dfb483bab/Library/src/Systems/Spaces/SpaceSystem.cpp#L805) that specifies which error code we're expecting
- Log a bug with MCS because we think they're sending us the wrong error

MAG-ChristopherAtkinson:
> [Question] In reference to your comment on Jira.
> 
> If MCS is returning an error 400 for this, and not the error 204 as we would expect, would the reasonable thing to do not be either:
> 
> * Update [our code](https://github.com/magnopus-opensource/connected-spaces-platform/blob/0ddd6163e1e8dfb932e79744e7170a0dfb483bab/Library/src/Systems/Spaces/SpaceSystem.cpp#L805) that specifies which error code we're expecting
> * Log a bug with MCS because we think they're sending us the wrong error

Yeah, I think this is something we can still guard against to remove the need for clients to persist their changes. While informing  MCS to resolve this at the service level, in case this API is consumed externally.

As for updating the expected response code, a 400 should not be applicable; the correct result should be a 204.

MAG-ChristopherAtkinson:
> > Guess there is also the argument for if we know it will fail, why send the request at all?
> 
> That's a good point, convinces me that this change should go in. Conceptually, doing a space invite on an empty/null list makes no sense.
> 
> ~I just do not want to lose the thread of why this bug manifested as a crash., Even if we slip up we want MCS to respond appropriately and for us to forward errors, not crash. I think it's something like "Unexpected error codes from MCS cause crashes", this is a good case for improving resiliance.~
> 
> It didn't crash ... okay this is all redundant. Thinking it through, error 400 sort of makes sense if we're sending a null list, it's what i'd expect.

Yeah, I was under the impression this was also causing a crash, but it seems that may have been the case at the client level. Likely due to an assumption that the space was created? 

But yeah, a 400 does make sense; however, so would a 204 would also make sense in this context, it would be successful in creating an empty list.

MAG-ElliotMorris (approved):

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -503,6 +503,43 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, CreateSpaceWithBulkInviteTest)
...
+
+    String UserId;
+
+    // Log in
+    LogInAsNewTestUser(UserSystem, UserId);
+
+    // Create space
+    ::Space Space;
+    InviteUserRoleInfoCollection InviteUsers;
+    CreateSpace(SpaceSystem, UniqueSpaceName, TestSpaceDescription, SpaceAttributes::Private, nullptr, InviteUsers, nullptr, nullptr, Space);
```
MAG-AdamThorn, line 528: [Question] is it worth us asserting that we receive a 200 here?

MAG-AdamThorn (approved):

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-ChristopherAtkinson: I don't think this would add any value, as the test is only concerned that it being successful, not by what means, as it is an early return behaviour, and 200 is just the accepted response to continue the chain.

PR #752, opened 2025-07-22
[OF-1708] feat: Break system dependencies on WebSocketClient

MAG-mv:
This ticket addressed the CSPFoundation dependency in POCOSignalrRClient. The Multiplayer URI is now passed through. This may seem confusing as it isn't done for the Emscripten version, but Emscripten uses a different URI, which is passed in when Start is called. This URI is modified by SignalR, which is needed for Emscripten, but not for POCO. I don't claim to understand the reasons for this modification.

I also removed a couple of redundant includes in the PR.

### DIFF
--- a/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
+++ b/Library/include/CSP/Multiplayer/MultiPlayerConnection.h
@@ -140,7 +140,8 @@ public:
     /// @param SpaceEntitySystem SpaceEntitySystem& : System provided such that it can create bindings at the appropriate point in the connection
     /// flow, prior to entity fetches.
     CSP_NO_EXPORT void Connect(ErrorCodeCallbackHandler Callback, ISignalRConnection* SignalRConnection,
-        csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, const csp::common::String& AccessToken, const csp::common::String& DeviceId);
+        csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, [[maybe_unused]] const csp::common::String& MultiplayerUri,
+        const csp::common::String& AccessToken, const csp::common::String& DeviceId);
 
     /// @brief Indicates whether the multiplayer connection is established
     /// @return bool : true if connected, false otherwise
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -352,7 +352,8 @@ std::function<async::task<void>()> MultiplayerConnection::StartListening()
 }
 
 void MultiplayerConnection::Connect(ErrorCodeCallbackHandler Callback, ISignalRConnection* SignalRConnection,
-    csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, const csp::common::String& AccessToken, const csp::common::String& DeviceId)
+    csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, [[maybe_unused]] const csp::common::String& MultiplayerUri,
+    const csp::common::String& AccessToken, const csp::common::String& DeviceId)
 {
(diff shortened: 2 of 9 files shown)


MAG-ChristopherAtkinson (commented):
The changes look good to me, not a little confused to why the URI is modified would be good to know the reason.

Happy to approve once the conflicts have been resolved

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -40,27 +40,6 @@ namespace csp::systems
 {
 class UserSystem;
 
-// This class exists purely to appease the wrapper generator.
-// IAuthContext was previously implemented by the UserSystem.
-// However, due to limitations with function pointers as parameters, we needed to hide the interface implementation.
-// Now, AuthContext, uses the functionality from the UserSystem to act as the AuthContext.
-CSP_START_IGNORE
```
MAG-mv: This change should be in the parent PR :(. I seemed to have failing with my rebase here

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -81,14 +81,14 @@ csp::common::String FormatScopesForURL(csp::common::Array<csp::common::String> S
...
 }
 
-AuthContext::AuthContext(csp::systems::UserSystem& UserSystem)
+UserSystem::AuthContext::AuthContext(csp::systems::UserSystem& UserSystem)
     : UserSystem { UserSystem }
 {
 }
 
-const csp::common::LoginState& AuthContext::GetLoginState() const { return UserSystem.GetLoginState(); }
+const csp::common::LoginState& UserSystem::AuthContext::GetLoginState() const { return UserSystem.GetLoginState(); }
```
MAG-mv: Same with these changes. They were done in the parent PR

MAG-ElliotMorris (commented):
Putting myself to comment status because underlying branch is changing, so this PR can't really be reviewed yet. You should consider moving PR's that arn't reviewable back to draft.

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -140,7 +140,8 @@ class CSP_API MultiplayerConnection
     /// @param SpaceEntitySystem SpaceEntitySystem& : System provided such that it can create bindings at the appropriate point in the connection
     /// flow, prior to entity fetches.
     CSP_NO_EXPORT void Connect(ErrorCodeCallbackHandler Callback, ISignalRConnection* SignalRConnection,
-        csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, const csp::common::String& AccessToken, const csp::common::String& DeviceId);
+        csp::multiplayer::SpaceEntitySystem& SpaceEntitySystem, [[maybe_unused]] const csp::common::String& MultiplayerUri,
```
MAG-ElliotMorris, line 143: [nit] Tiny issue, not sure about `maybe_unused` in declarations. Arguably it serves as documentation, but it's also arguably leaking implementation details to the interface.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -366,10 +367,13 @@ void MultiplayerConnection::Connect(ErrorCodeCallbackHandler Callback, ISignalRC
         delete Connection;
     }
 
+// You will notice that the Emscripten webclient doesn't take the Uri into its constructor.
+// This is because it uses the url passed into its Start function, which gets modified by SignalR.
+// This modified version isn't compatible with our POCO implementation, so we need to directly use the MultiplayerUri.
```
MAG-ElliotMorris, line 372: [comment] Down with this sort of thing! Will need to make these interfaces, and hopefully even implementations, more similar eventually ... :(

MAG-ElliotMorris (approved):

PR #755, opened 2025-07-23
[OF-1670] refac!: Online Realtime Engine

MAG-ElliotMorris:
## Online Realtime Engine
This PR converts `SpaceEntitySystem` to `OnlineRealtimeEngine`, providing an interface for clients to pass arbitrary `IRealtimeEngine` instances to `EnterSpace`, such that CSP can branch on different engine behaviour. The most immediate desire for this being the support of an `OfflineRealtimeEngine` which is coming soon, but also opens up lots of alternate behaviour, such as a more lean `ServerSideScriptRunnerRealtimeEngine` or even a `MockRealtimeEngine`, etc. I'm quite excited about how this interface point makes experimental implementations of radically new communication/entity management techniques possible.

This has many implications for our memory management during login and space entry. In general, you will see a removal of manual state management in objects, and an addition of constructor arguments being used to establish invariants. This all takes into account the future direction of the EnterSpace flow, and is a solid step towards the Summer Party Vision, as well as being bettter in general.

An important factor is the seperation of SignalR binding and dispatch, you'll see this in early commits. `MultiplayerConnection` is now explicitly bound during `Connect`, and maintains those bindings for its entire lifetime. Those bindings are only actually dispatched to a `RealtimeEngine` if one has been set, which is something that is invariant to the temporal scope of entering a space (as defined by the scope between the instruction when `EnterSpace` is first called, and the receiving instruction of the `ExitSpace` callback)
So long as this invariant holds, it is not possible to dispatch an off-thread network event to null memory. Remember, whilst `RealtimeEngine` has been extracted the way we want it, `MultiplayerConnection` as yet **has not**, so a shim is neccesary.

There are a **ton** of not-great decisions made here due to the wrapper generator, it's really been pushed to its limit during this work. (See [this](https://docs.google.com/document/d/13oJrWh2x07QdTPUdtwTR7tf4AOfTiNhEAlk0oTqYvtE/edit?tab=t.0#heading=h.w7400dxtrnuy) for a listing of some of the issues that occurred during this work.)

The most major of these compromises is the need for clients to manually ensure `RealtimeEngine` lifetime for the scope of being in a space. Ideally this would be expressed in the interface via a `shared_ptr/weak_ptr` link, or a different memory management strategy, but we're not there yet.

> [!NOTE]  
> `IRealtimeEngine`, whilst being an improved interface, is still 95% designed around its original legacy as `SpaceEntitySystem`. This has unfortunate consequences, but we knew we'd have a quirky interface going in. The biggest deal here is that even implementing a `SingleplayerRealtimeEngine` will require the usage of a patch processing model, as those interfaces (`ProcessPending`, etc) remain on `IRealtimeEngine` such that clients can have an easier time transitioning.
>
> However, I do hope to change this in the future. Having to account for late-update semantics, even in a fake way, is an uncomfortable burden. It's irrelevent in an offline engine, and there's many other threading models we could use to manage better even in the online model ... unreasonably restrictive.

## Reviewer Notes

- This PR shows the transitional steps to make this refactor, you will see in the initial commits the use of `weak/shared_ptr` that will disappear, as well as a bunch of static/global shims to maintain lifetime. This was in order to do red/green testing during the refactor. These shims and interfaces be removed over the next few commits. These are called out in code comments so you shouldn't get lost. Test passes were maintained between each major commit
- Tests are almost all shown at the end, despite me keeping them up to date as I went, with all the mass procedural changes to tests being isolated to [one commit,](https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/4b3f472e339895efe4ffa8a1df5f32b1bc5831b1) where the more interesting ones are in other commits.
- Near the end, there is a change to our init flow to allow injecting custom SignalRConnections, mostly driven by a need to inject `SignalRConnectionMock` more fundamentally into our actual systems. This wouldn't be necessary upon the completion of our full transformation plan, as "injection" would simply be a system property, not an explicit carveout. The temporaryness of this is called out in comments
- Much of the changes that may seem strange make more sense if you consider space re-entry or re-login. Because our library has been written in an assumed non-instantiable, static way, much of the handling around this has felt coincidental up until this point. These changes are more explicit about it, and I suspect will fix latent bugs around this sort of thing. The emergency re-enter space regression test has been removed, since the state it tested for is no longer possible.
- **Commit mistakes** : Quite a few, it's been a long PR. Sorry about these, hooks killed me.
  - Made a change to convert NullResult -> SpaceResult in EnterSpace, but split this over commits incorrectly. 
  - Also made a mistake splitting the moving of `SetSelectionState`. 
  - Interface changes to `IRealtimeEngine` about adding a parentID are rolled into the commit that does the fake interface throw, wouldn't be like that ideally.

## Breaking Changes
- `SpaceEntitySystem` removed
- New type `OnlineRealtimeEngine : public IRealtimeEngine` added, which serves as the replacement to `SpaceEntitySystem`
- `EnterSpace` now takes an `IRealtimeEngine` pointer. The client is responsible for instantiating this object, and maintaining its lifetime for the scope of being in a space. Deletion can happen once the `ExitSpace` callback has been received.
- Convenience method added : `SystemsManager::makeOnlineRealtimeEngine` for transitional convenience. However, this is a temporary method and is already deprecated. We still recommend its use during this transition however. Alternately, you may directly construct an `OnlineRealtimeEngine` by providing its constructor arguments (which are all systems that can be retrieved from `SystemsManager`)
- `EnterSpace` callback now returns a `SpaceResult` instead of a `NullResult`, for convenience
- `OnlineRealtimeEngine` conforms to the `IRealtimeEngine` interface, which has some changes from the previous `SpaceEntitySystem` interface (it is still mostly the same however)
  - `CreateAvatar` now takes an optional ParentID, this should be set to an empty optional if you wish to create a root entity (which was the previous behaviour).
  - `CreateObject` now correctly named `CreateEntity` (See IRealtimeEngine doc for new documentation on what these words actually mean)
  - `SetInitialEntitiesRetrievedCallback(CallbackHandler)` renamed `SetEntityFetchCompleteCallback(EntityFetchCompleteCallback)`.
  - `EntityFetchCompleteCallback` must be set before calling `EnterSpace`. `EnterSpace` will emit an error if this is not so.
  - `EntityFetchCompleteCallback` now provides a uint_32, that being the number of entities fetched in the initial entity retrieval
  - `SetSelectionStateOfEntity(SelectedState, SpaceEntity)` removed from `SpaceEntitySystem`, now call directly on SpaceEntity, `SpaceEntity()->Select` \ `SpaceEntity->Deselect() `
  - `SetScriptSystemReadyCallback(CallbackHandler)` renamed to `SetScriptLeaderReadyCallback(ScriptLeaderReadyCallback)` to better reflect what this actually does.
- `RegisterEntityScriptAsModule` and `BindEntityScript` removed, as we could find no evidence of their use, and original implementors left a note that they were only made public accidentally. (If you have any, let us know so we can reinstate them.)

## Example
Javascript example of new way to enter a space. From our tests.
```
//Create multiplayer realtime engine
const realtimeEngine = Systems.SystemsManager.get().makeOnlineRealtimeEngine();

//Set mandatory callback, you probably want to react to this in someway. Don't mutate entities before this is called.
realtimeEngine.setEntityFetchCompleteCallback((entitiesFetchedNum) => {})

//Enter the space
const spaceSystem = Systems.SystemsManager.get().getSpaceSystem();
const enterSpaceResult = await spaceSystem.enterSpace(spaceId, realtimeEngine);

if (enterSpaceResult.getResultCode() == Systems.EResultCode.Success){
  console.log("Successfully entered space");
}
else {
  console.error("Failed to enter space");
}

enterSpaceResult.delete(); 

//Exit the space
const exitSpaceResult = await spaceSystem.exitSpace(spaceId);

//Delete the realtime engine now that we've exited the space. Do not delete the engine before this.
realtimeEngine.delete();
```

Note deleting the realtime engine. The only condition is that the realtime engine must stay in-memory for the period between invoking `EnterSpace` and receiving the final callback from `ExitSpace`. You are free to defer these deletes, or even never do them at all.

## Wrapper Gen
This change required a change to the wrapper gen in order to add virtual/override functionality to events and async tasks in the C# generator, in order to do dynamic dispatch from an interface. I've assigned you @magnopus-swifty specifically to review this [one commit](https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/102bc0b24ee5d166b9006005479732ecead28471), as we don't have a good change strategy around C# changes. At least with JS we can observe one real working test.

This is so we can get the `virtual` (on the base) and `override` (on the derived) keywords here in event declarations:
```
bool OnEntityCreatedInitialised = false;
event EventHandler<Csp.Multiplayer.SpaceEntity>? RealOnEntityCreated;
public virtual event EventHandler<Csp.Multiplayer.SpaceEntity> OnEntityCreated
{...}
```

We don't normally change the wrapper gen _at all_, we treat it like this magic black box, so this is a fear. This should be an explicit callout for C# integration as I don't know the implications of doing this sort of thing.

### DIFF
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -25,6 +25,11 @@ namespace csp::systems
 class ServicesDeploymentStatus;
 } // namespace csp::systems
 
+namespace csp::multiplayer
+{
+class ISignalRConnection;
+}
+
 namespace csp
 {
 
@@ -119,6 +124,12 @@ public:
     /// interact with each other.
     /// @return bool : True for successful initialisation.
     static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant);
+
+    // Hidden function for testing. Lets us pass in state that would otherwise be injected in a set way in the SystemsManager.
+    // In a different, perhaps better api, this wouldn't be necessary as constructors would inject this at client level and the configurability would
+    // be there by default
+    CSP_NO_EXPORT static bool InitialiseWithInject(
+        const csp::common::String& EndpointRootURI, const csp::common::String& Tenant, csp::multiplayer::ISignalRConnection* SignalRInject);
 
     /// @brief This should be used at the end of the application lifecycle.
     /// Clears event queues and destroys foundation systems.
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -45,6 +45,16 @@ typedef std::function<void(csp::multiplayer::SpaceEntity*)> EntityCreatedCallbac
 namespace csp::common
 {
 
+// This callback should be fired during IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers when the system is ready for
+// csp::systems::SpaceSystem::EnterSpace to yield control back to calling clients. This may be done prior to actually completing fetching all
+// entities, as that could be a long operation and the specific RealtimeEngine implementation may not wish to block clients entering a space.
+typedef std::function<void()> EntityFetchStartedCallback;
+// This callback should be fired once all the entities have been fetched upon initial space setup, invoked via
+// IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers. This serves as a public notification to clients that the RealtimeEngine is in a valid state
+// and entity inspection and mutation may begin.
+// Provides the number of entities fetched as an argument
+typedef std::function<void(std::uint32_t)> EntityFetchCompleteCallback;
+
 // This, frustratingly, cannot be an in-class type due to the code generator.
 /// @brief Enum of concrete types of RealtimeEngines.
 enum class RealtimeEngineType : int
@@ -118,6 +128,24 @@ public:
     ///
     /// @param Callback csp::multiplayer::EntityCreatedCallback : the callback to execute.
     CSP_EVENT virtual void SetEntityCreatedCallback(csp::multiplayer::EntityCreatedCallback Callback) = 0;
+    /**
+     * @brief Fetch space entities from the RealtimeEngine data source and perform initial setup to populate internal buffers
+     *
+     * This method is called during csp::systems::SpaceSystem::Enterspace.
+     *
+     * @param SpaceId const csp::common::String& : The identifier of the space to fetch entities from. Format depends on how your specific backend
+     * store indexes spaces.
+     * @param FetchStartedCallback EntityFetchStartedCallback : Callback notifying for when csp::systems::SpaceSystem::EnterSpace can yield control
+     * back to clients
+     * @params FetchCompleteCallback EntityFetchCompleteCallback : Callback notifying when all entities are fetched and stored appropriately, and
+     * clients may start entity mutation.
+     *
+     * @post @param FetchStartedCallback will have been called, and @param FetchCompleteCallback will either have been called, or an async job started
+     * with an intent to call it once fetch is complete.
+     */
+    CSP_NO_EXPORT virtual void FetchAllEntitiesAndPopulateBuffers(const csp::common::String& SpaceId, EntityFetchStartedCallback FetchStartedCallback,
+        csp::common::EntityFetchCompleteCallback FetchCompleteCallback)
+        = 0;
 
     /***** ENTITY ACCESS *****************************************************/
 
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -164,10 +164,19 @@ public:
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
     }
 
+    /// @brief Adds an entity to the set of selected entities
+    /// @param Entity csp::multiplayer::SpaceEntity* Entity to set as selected
+    /// @return True if the entity was succesfully added, false otherwise. Refer to your specific IRealtimeEngine instantiation for specific failure
+    /// cases.
     virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
     {
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
     }
+
+    /// @brief Removes an entity to the set of selected entities
+    /// @param Entity csp::multiplayer::SpaceEntity* Entity to set as selected
+    /// @return True if the entity was succesfully removed, false otherwise. Refer to your specific IRealtimeEngine instantiation for specific failure
+    /// cases.
     virtual bool RemoveEntityFromSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
     {
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -168,7 +168,7 @@ public:
     /// @param Entity csp::multiplayer::SpaceEntity* Entity to set as selected
     /// @return True if the entity was succesfully added, false otherwise. Refer to your specific IRealtimeEngine instantiation for specific failure
     /// cases.
-    virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    CSP_NO_EXPORT virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
     {
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
     }
@@ -177,7 +177,7 @@ public:
     /// @param Entity csp::multiplayer::SpaceEntity* Entity to set as selected
     /// @return True if the entity was succesfully removed, false otherwise. Refer to your specific IRealtimeEngine instantiation for specific failure
     /// cases.
-    virtual bool RemoveEntityFromSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    CSP_NO_EXPORT virtual bool RemoveEntityFromSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
     {
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
     }
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -16,7 +16,10 @@
 
 #pragma once
 
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/Interfaces/InvalidInterfaceUserError.h"
 #include "CSP/Common/List.h"
+#include "CSP/Common/Optional.h"
 #include "CSP/Common/String.h"
 
 namespace csp::multiplayer
@@ -40,20 +43,30 @@ typedef std::function<void(bool)> CallbackHandler;
 
 // Callback that provides a non-owning pointer to a SpaceEntity object.
 typedef std::function<void(csp::multiplayer::SpaceEntity*)> EntityCreatedCallback;
+
 }
 
 namespace csp::common
 {
+// Note: It's not great that we require std (function) dependencies in interop types, seems to defeat the whole point.
 
-// This callback should be fired during IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers when the system is ready for
-// csp::systems::SpaceSystem::EnterSpace to yield control back to calling clients. This may be done prior to actually completing fetching all
-// entities, as that could be a long operation and the specific RealtimeEngine implementation may not wish to block clients entering a space.
+/**
+ * This callback should be fired during IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers when the system is ready for
+ * csp::systems::SpaceSystem::EnterSpace to yield control back to calling clients. This may be done prior to actually completing fetching all
+ * entities, as that could be a long operation and the specific RealtimeEngine implementation may not wish to block clients entering a space.
+ */
 typedef std::function<void()> EntityFetchStartedCallback;
-// This callback should be fired once all the entities have been fetched upon initial space setup, invoked via
-// IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers. This serves as a public notification to clients that the RealtimeEngine is in a valid state
-// and entity inspection and mutation may begin.
-// Provides the number of entities fetched as an argument
-typedef std::function<void(std::uint32_t)> EntityFetchCompleteCallback;
+
+/**
+ * This callback should be fired once all the entities have been fetched upon initial space setup, invoked via
+ * IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers. This serves as a public notification to clients that the RealtimeEngine is in a valid state
+ * and entity inspection and mutation may begin.
+ *
+ * Provides the number of entities fetched as an argument
+ */
+typedef std::function<void(uint32_t)> EntityFetchCompleteCallback;
+
+class LoginState;
 
 // This, frustratingly, cannot be an in-class type due to the code generator.
 /// @brief Enum of concrete types of RealtimeEngines.
@@ -83,6 +96,9 @@ enum class RealtimeEngineType : int
  * - @b Entity: All items in a space are entities.
  * - @b Avatar: A specialization of Entity representing an avatar. Defined by whether the entity contains an AvatarSpaceComponent.
  * - @b Object: An entity that is not an avatar. Defined by that entity not containing an AvatarSpaceComponent.
+ *
+ * @important This type is not a true interface, instead throwing default implementations.
+ *            This is due to wrapper generator constraints, hopefully temporary ones.
  */
 class CSP_API IRealtimeEngine
 {
@@ -91,43 +107,72 @@ public:
     virtual ~IRealtimeEngine() = default;
 
     /// @brief Returns the concrete type of the instantiation of the abstract IRealtimeEngine.
-    virtual RealtimeEngineType GetRealtimeEngineType() const = 0;
+    virtual RealtimeEngineType GetRealtimeEngineType() const { throw InvalidInterfaceUseError("Illegal use of \"abstract\" type."); }
 
     /***** ENTITY MANAGEMENT *************************************************/
 
     /// @brief Create and add a SpaceEntity with type Avatar, and relevant components and default states as specified.
     /// @param Name csp::common::String : The entity name of the newly created avatar entity.
+    /// @param LoginState const csp::common::Optional<csp::common::LoginState>& : Login state tokens for authenticating on the remote server. Only
+    /// required if backend data store requires authentication.
     /// @param SpaceTransform csp::multiplayer::SpaceTransform : The initial transform to set the SpaceEntity to.
-    /// @param State csp::multiplayer::AvatarState : The initial Avatar State to set.
+    /// @param IsVisible bool : Whether the avatar defaults to being visible.
+    /// @param AvatarState csp::multiplayer::AvatarState : The initial Avatar State to set.
     /// @param AvatarId csp::common::String : The ID to be set on the AvatarSpaceComponent
     /// @param AvatarPlayMode csp::multiplayer::AvatarPlayMode : The Initial AvatarPlayMode to set.
     /// @param Callback csp::multiplayer::EntityCreatedCallback A callback that executes when the creation is complete,
     /// which will provide a non-owning pointer to the new SpaceEntity so that it can be used on the local client.
-    CSP_ASYNC_RESULT virtual void CreateAvatar(const csp::common::String& Name, const csp::multiplayer::SpaceTransform& SpaceTransform,
-        const csp::multiplayer::AvatarState& State, const csp::common::String& AvatarId, const csp::multiplayer::AvatarPlayMode& AvatarPlayMode,
-        csp::multiplayer::EntityCreatedCallback Callback)
-        = 0;
+    CSP_ASYNC_RESULT virtual void CreateAvatar(const csp::common::String& Name, const csp::common::Optional<csp::common::LoginState> LoginState,
+        const csp::multiplayer::SpaceTransform& SpaceTransform, bool IsVisible, const csp::multiplayer::AvatarState& AvatarState,
+        const csp::common::String& AvatarId, const csp::multiplayer::AvatarPlayMode& AvatarPlayMode, csp::multiplayer::EntityCreatedCallback Callback)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
 
     /// @brief Create and add a SpaceEntity, with relevant default values.
     /// @param Name csp::common::String : The name of the newly created SpaceEntity.
     /// @param SpaceTransform csp::multiplayer::SpaceTransform : The initial transform to set the SpaceEntity to.
+    /// @param ParentID csp::common::Optional<uint64_t> : ID of another entity in the space that this entity should be created as a child to. If
+    /// empty, entity is created as a root entity.
     /// @param Callback csp::multiplayer::EntityCreatedCallback : A callback that executes when the creation is complete,
     /// which will provide a non-owning pointer to the new SpaceEntity so that it can be used on the local client.
-    CSP_ASYNC_RESULT virtual void CreateEntity(
-        const csp::common::String& Name, const csp::multiplayer::SpaceTransform& SpaceTransform, csp::multiplayer::EntityCreatedCallback Callback)
-        = 0;
+    CSP_ASYNC_RESULT virtual void CreateEntity(const csp::common::String& Name, const csp::multiplayer::SpaceTransform& SpaceTransform,
+        const csp::common::Optional<uint64_t>& ParentID, csp::multiplayer::EntityCreatedCallback Callback)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
+
+    virtual void AddEntity(csp::multiplayer::SpaceEntity* EntityToAdd) { throw InvalidInterfaceUseError("Illegal use of \"abstract\" type."); }
 
     /// @brief Destroy the specified entity.
     /// @param Entity csp::multiplayer::SpaceEntity : A non-owning pointer to the entity to be destroyed.
     /// @param Callback csp::multiplayer::CallbackHandler : A callback that executes when the entity destruction is complete.
-    CSP_ASYNC_RESULT virtual void DestroyEntity(csp::multiplayer::SpaceEntity* Entity, csp::multiplayer::CallbackHandler Callback) = 0;
+    CSP_ASYNC_RESULT virtual void DestroyEntity(csp::multiplayer::SpaceEntity* Entity, csp::multiplayer::CallbackHandler Callback)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
 
     /// @brief Sets a callback to be executed when an entity is fully created.
     ///
     /// Only one EntityCreatedCallback may be registered, calling this function again will override whatever was previously set.
+    /// The better way to set this and avoid initialisation race conditions is via passing this in the constructor, only use this if you wish to
+    /// override or remove this callback.
     ///
     /// @param Callback csp::multiplayer::EntityCreatedCallback : the callback to execute.
-    CSP_EVENT virtual void SetEntityCreatedCallback(csp::multiplayer::EntityCreatedCallback Callback) = 0;
+    CSP_EVENT virtual void SetEntityCreatedCallback(csp::multiplayer::EntityCreatedCallback Callback)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
+
+    virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
+    virtual bool RemoveEntityFromSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
+
     /**
      * @brief Fetch space entities from the RealtimeEngine data source and perform initial setup to populate internal buffers
      *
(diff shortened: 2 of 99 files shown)


MultiplayerTestRunner/src/SpaceRAII.cpp
```diff
@@ -49,11 +49,11 @@ SpaceRAII::SpaceRAII(std::optional<std::string> ExistingSpaceId)
     }
 
     // Enter space
-    std::promise<csp::systems::NullResult> ResultPromise;
-    std::future<csp::systems::NullResult> ResultFuture = ResultPromise.get_future();
+    std::promise<csp::systems::SpaceResult> ResultPromise;
```
MAG-ElliotMorris: [consider] You got burned by hooks again, let people know, or do the rebase ... (probably wont bother to do the rebase)

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -40,17 +43,37 @@ typedef std::function<void(bool)> CallbackHandler;
...
+typedef std::function<void()> EntityFetchStartedCallback;
+
+/**
+ * This callback should be fired once all the entities have been fetched upon initial space setup, invoked via
+ * IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers. This serves as a public notification to clients that the RealtimeEngine is in a valid state
+ * and entity inspection and mutation may begin.
+ *
+ * Provides the number of entities fetched as an argument
+ */
+typedef std::function<void(uint32_t)> EntityFetchCompleteCallback;
```
MAG-ElliotMorris, line 67: [consider] Maybe `Initial` is a good word to keep here, it had that before. Like `InitialEntityFetchCompleteCallback`

MAG-ElliotMorris: You know actually, I don't think the intial nature of things is _really_ important. You could fetch entities whenever you want in theory. Even if _right now_ that flow might break, it shouldn't in the perfectest system. Would rather keep the naming like this to gesture at this.

Library/include/CSP/Systems/SystemsManager.h
```diff
@@ -151,7 +152,9 @@ class CSP_API SystemsManager
     /// @return HotspotSequenceSystem : pointer to the HotspotSequenceSystem system class
     HotspotSequenceSystem* GetHotspotSequenceSystem();
 
-    csp::multiplayer::SpaceEntitySystem* GetSpaceEntitySystem();
+    CSP_NO_EXPORT std::weak_ptr<csp::common::IRealtimeEngine> GetRealtimeEngine();
+    CSP_NO_EXPORT std::weak_ptr<csp::common::IRealtimeEngine> InstantiateMultiplayerRealtimeEngine();
+    CSP_NO_EXPORT void DestroyRealtimeEngine();
```
MAG-ElliotMorris: [Comment] Just for clarity, also temp hacks, the idea of instantiating/destroying a realtime engine ourselves dissapears when the ownership transfers

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -373,8 +380,12 @@ void MultiplayerConnection::Connect(ErrorCodeCallbackHandler Callback, ISignalRC
...
 
     Connection = SignalRConnection;
-    NetworkEventManager->SetConnection(Connection);
-    SpaceEntitySystem.SetConnection(Connection);
+    NetworkEventManager->SetConnection(*SignalRConnection);
+
+    BindOnObjectMessage();
+    BindOnObjectPatch();
+    BindOnRequestToSendObject();
+    BindOnRequestToDisconnect();
```
MAG-ElliotMorris, line 386: [Comment] These are the bindings that constitute the data sources for a  connected-spaces-online-reciever, or whatever better name we come up with. Would be nice to make an interface out of them at some point, do a clean observer system. The beginnings of one are already here.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -642,4 +653,67 @@ CSP_ASYNC_RESULT void MultiplayerConnection::SetAllowSelfMessagingFlag(const boo
...
+        [this](const signalr::value& Params)
+        {
+            if (auto EntitySystem = SpaceEntitySystemWeak.lock())
+            {
+                EntitySystem->OnObjectMessage(Params);
+            }
+            else
+            {
+                LogSystem.LogMsg(
+                    common::LogLevel::Log, "Recieved OnObjectMessage without an alive EntitySystem. This is expected if leaving a space.");
```
MAG-ElliotMorris: [consider] I've been burned before with being too assumptive in these error messages. There may be other valid states where this log occurs. Should I remove the `This is expected if leaving a space?`? Unsure, I don't want to spook folks when they get this log ... maybe it should be verbose at least.?

In fact, yeah, [fix], make this verbose.

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -91,43 +107,72 @@ class CSP_API IRealtimeEngine
...
+    }
+
+    virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
+    virtual bool RemoveEntityFromSelectedEntities(csp::multiplayer::SpaceEntity* Entity)
+    {
+        throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
+    }
```
MAG-ElliotMorris: [fix] You need to doc these

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -350,6 +346,9 @@ class CSP_API SpaceEntity
     /// @return True if deselection occurred. False if not.
     bool Deselect();
 
+    // ClientID is the clientID that comes from the multiplayer connection, ie the clientID of the client that is executing the code at the time.
+    bool SetSelectionState(bool SelectionState, uint64_t ClientId);
```
MAG-ElliotMorris: [fix] Should doc this better, with actual doxy formatting

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -194,7 +194,10 @@ class CSP_API MultiplayerConnection
     CSP_NO_EXPORT MultiplayerHubMethodMap GetMultiplayerHubMethods() const { return MultiplayerHubMethods; }
 
     /// @brief Sets the internal reference to the SpaceEntitySystem. Should be called when entering a space.
-    CSP_NO_EXPORT void SetSpaceEntitySystem(std::shared_ptr<SpaceEntitySystem>& SpaceEntitySystem);
+    /// @param SpaceEntitySystem SpaceEntitySystem* : Non-owning pointer. Ideally this would be an owned structure, but we can't due to the wrapper
+    /// generator. Remember to null this when exiting a space as object dispatch depends on that, until clients are capable of managing this
+    /// themselves. Unfortunate manual state management.
```
MAG-ElliotMorris: [nit] You didn't really mean "clients", you meant "client interop" in reference to owned pointer structures.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -428,6 +436,12 @@ void SpaceSystem::ExitSpace(NullResultCallback Callback)
                             return;
                         }
 
+                        // Null the realtime engine pointer in the multiplayer connection such that it stops dispatching signalR updates.
+                        // (Error paths are messy, what does failing to leave a space mean memory wise? This is why owned types with RAII work so much
+                        // better).
+                        MultiplayerConnection->SetSpaceEntitySystem(nullptr);
```
MAG-ElliotMorris: [Question] For everyone, what does failing to leave a space mean, in reference to the if case above?

Are we still in the space, and we shouldn't null the SpaceEntitySystem(RealtimeEngine) here? Or are we not. I've made the decision that we're still in the space, because the error is like "Failed to leave space", but i'm unsure. I don't think we accounted for this case before, but it'd be nice to be clearer.

MultiplayerTestRunner/src/InternalUnitTests/Test_SpaceRAII.cpp
```diff
@@ -121,7 +121,7 @@ TEST_F(SpaceRAIITest, TestUseInvalidExistingSpace)
     try
     {
         ::testing::internal::CaptureStdout();
-        SpaceRAII Space({ Invalid_SpaceID });
+        SpaceRAII Space(std::make_optional(Invalid_SpaceID));
```
MAG-ElliotMorris, line 124: [Comment] This is a consequence of weird arcane stuff around overload resolution of deleted private constructors + initializer list preference.

Library/src/Multiplayer/SignalR/SignalRConnection.cpp
```diff
@@ -103,11 +106,27 @@ void SignalRConnection::SetDisconnected(const std::function<void(std::exception_
...
+    try
+    {
+        Connection.on(EventName, Handler);
+        return true;
+    }
+    catch (const signalr::signalr_exception& exception)
+    {
+        // If you register an event twice, you'll throw. Very hard to debug if this happens off thread, which it does with re-logging in.
+        // Not really an error though, expected behaviour for our flow, just log to help in debugability
+        LogSystem.LogMsg(csp::common::LogLevel::Log,
```
MAG-ElliotMorris: [fix] Make this verbose

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -2814,7 +2810,7 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, WhenSignalRStartErrorsThenDisconnec
 {
     csp::common::LogSystem LogSystem;
     SignalRConnectionMock* SignalRMock = new SignalRConnectionMock();
-    csp::multiplayer::MultiplayerConnection Connection { LogSystem };
+    csp::multiplayer::MultiplayerConnection Connection { LogSystem, *SignalRMock };
```
MAG-ElliotMorris: [Comment] I probably could have refactored these to use the new test macro, but there was no need. Not all tests need the full system to be instantiated. This is the difference between unit tests and integration tests, but we don't make a real split (and good tbh, who can be arsed with the hassle of that level of categorization)

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -276,7 +276,7 @@ class CSP_API IRealtimeEngine
     /// This callback must be set prior to entering a space.
     /// @param FetchCompleteCallback csp::common::EntityFetchCompleteCallback : Callback that notifies when the RealtimeEngine has completed initial
     /// entity fetch
-    void SetEntityFetchCompleteCallback(EntityFetchCompleteCallback Callback) { EntityFetchCompleteCallback = Callback; }
+    CSP_EVENT void SetEntityFetchCompleteCallback(EntityFetchCompleteCallback Callback) { EntityFetchCompleteCallback = Callback; }
```
MAG-ElliotMorris, line 350: [comment] Glad this error happened, I realised we had no other callbacks sets that wern't marked CSP_EVENT. One generates a delegate, one an event. I don't know the implicative differences in how our C# api uses them enough to justify diverging, and I guess there's no need.

MAG-ElliotMorris (commented):
Let me request changes on my own PR you cowards @github!

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -654,7 +621,9 @@ bool MultiplayerConnection::GetAllowSelfMessagingFlag() const { return AllowSelf
 
 void MultiplayerConnection::BindOnObjectMessage()
 {
-    GetSignalRConnection()->On("OnObjectMessage",
+
+    GetSignalRConnection()->On(
+        "OnObjectMessage",
```
magnopus-swifty: Should we/can we do something about these magic strings?

Tests/src/PublicTestBase.cpp
```diff
@@ -54,6 +55,42 @@ void PublicTestBase::TearDown()
...
+    ::testing::Test::SetUp();
+
+    // Yield SignalRMock ownership
+    InitialiseFoundationWithUserAgentInfo(EndpointBaseURI(), SignalRMock);
+
+    csp::systems::SystemsManager::Get().GetLogSystem()->SetSystemLevel(csp::common::LogLevel::VeryVerbose);
+
+    csp::systems::SystemsManager::Get().GetLogSystem()->SetLogCallback([](csp::common::String Message) { fprintf(stderr, "%s\n", Message.c_str()); });
+
+    csp::systems::SystemsManager::Get().GetLogSystem()->LogMsg(csp::common::LogLevel::Verbose, "Foundation initialised!");
```
magnopus-swifty: [nit] Maybe cache the result of `GetLogSystem()` so this is less verbose?

Tests/src/PublicAPITests/ComponentTests/CollisionComponentTests.cpp
```diff
@@ -64,16 +63,19 @@ CSP_PUBLIC_TEST(CSPEngine, CollisionTests, CollisionComponentTest)
     CreateSpace(
         SpaceSystem, UniqueSpaceName, TestSpaceDescription, csp::systems::SpaceAttributes::Private, nullptr, nullptr, nullptr, nullptr, Space);
 
-    auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id);
+    std::unique_ptr<csp::multiplayer::SpaceEntitySystem> RealtimeEngine { SystemsManager.MakeOnlineRealtimeEngine() };
+    RealtimeEngine->SetEntityFetchCompleteCallback([](uint32_t) {});
+
+    auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id, RealtimeEngine.get());
 
     EXPECT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
```
magnopus-swifty, line 71: Looks like a lot of `EXPECT_EQ`s need changing for `ASSERT_EQ`s, perhaps in a future PR.

Library/src/Systems/Spaces/SpaceSystem.cpp
magnopus-swifty: As a user I'd expect leaving a space to be a non-negotiable action, more of a "FYI I'm leaving now" than a "please may I leave?"  As such I'd expect the library to clean everything up and send any pending messages, with the expectation that I'm now able to enter a different space.

I'm not sure I'd do with a failure in this context. I might want to inform the user that some data has been lost?

Tools/WrapperGenerator/Templates/CSharp/Partials/Class/Event.mustache
```diff
@@ -54,7 +54,8 @@ event EventHandler<
 {{# has_multiple_parameters }}
     ){{/ has_multiple_parameters }}
 >? Real{{ name }};
-public event EventHandler<
+public {{# is_virtual }}{{^ is_override }}virtual{{/ is_override }}{{/ is_virtual }}
+       {{# is_override }}override{{/ is_override }} event EventHandler<
```
magnopus-swifty, line 58: [Consider] Took me a minute to get my head around this. Is this actually necessary? We could potentially avoid outputting anything for overriding classes because the overrides generate identical code.

Line 75: `{{ unique_method_name }}(_ptr, {{ name }}DelegateInstance, (IntPtr)handle);`

So long as _this_ method does the overrides correctly, subclasses shouldn't need to override the base implementation. Am I understanding this right?

Library/include/CSP/CSPFoundation.h
```diff
@@ -120,6 +125,12 @@ class CSP_API CSPFoundation
     /// @return bool : True for successful initialisation.
     static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant);
 
+    // Hidden function for testing. Lets us pass in state that would otherwise be injected in a set way in the SystemsManager.
+    // In a different, perhaps better api, this wouldn't be necessary as constructors would inject this at client level and the configurability would
+    // be there by default
+    CSP_NO_EXPORT static bool InitialiseWithInject(
```
magnopus-swifty, line 131: [Nit] nitpick but we should probably be consistently using US English spelling, including `Initialise` above

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -0,0 +1,442 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ChristopherAtkinson, line 2: [Question] Should this be updated to 2025 or persisted as it is a refactor of the previous system

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: [Fix] These can be replaced now with the introduction of Library/src/Multiplayer/MultiplayerHubMethods.cpp

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -155,8 +155,8 @@ ISignalRConnection* MultiplayerConnection::MakeSignalRConnection()
         std::make_shared<csp::multiplayer::CSPWebsocketClient>());
 }
 
-MultiplayerConnection::MultiplayerConnection(csp::common::LogSystem& LogSystem)
-    : Connection(nullptr)
+MultiplayerConnection::MultiplayerConnection(csp::common::LogSystem& LogSystem, csp::multiplayer::ISignalRConnection& Connection)
+    : Connection(&Connection)
```
MAG-ChristopherAtkinson, line 160: [Comment] love to see this, guaranteeing objects as part of RAII is great pattern to adopt, and has the benefits of removing all the nullptr checks in dependant system

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -642,4 +619,76 @@ CSP_ASYNC_RESULT void MultiplayerConnection::SetAllowSelfMessagingFlag(const boo
...
+                    common::LogLevel::Log, "Received OnObjectMessage without an alive EntitySystem. This is expected if leaving a space.");
+            }
+        },
+        LogSystem);
+}
+
+void MultiplayerConnection::BindOnObjectPatch()
+{
+    GetSignalRConnection()->On(
+        "OnObjectPatch",
```
MAG-ChristopherAtkinson: [Fix] There are a few of these strings in this file that can be replaced now with the introduction of Library/src/Multiplayer/MultiplayerHubMethods.cpp

MAG-ChristopherAtkinson:
> > Unfortunately I only devoted 1/2 day on this based on your suggestion due to complexity, but happy to revisit again and devote some additional time.
> 
> When I said that I meant more as : "Don't expect this to only take an hour", not as a cap. I think this is going to be pretty fundamental so If you think you'll have more insights with more time, from my perspective, take it.
> 
> However, IANYM (I am not your manager :P )

Haha, yeah, I should clarify, I mean today. I have some other work that requires my attention, so will be waiting until Monday to have my attention again

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -212,6 +215,19 @@ class CSP_API MultiplayerConnection
 
     void Stop(ExceptionCallbackHandler Callback) const;
 
+    /*
+     * Bind the SignalR messages that are recieved from MCS to facilitate realtime communication.
+     * These are bound for the entire lifetime of the MultiplayerConnection (conceptually Login/Logout scoped).
+     * These messages are dispatched to the SpaceEntitySystem if it exists.
+     * For the future, consider how pleasant it might be if we bound everything all at once here, and had a single
```
MAG-mv, line 232: [Comment] Sounds exactly like what we did with he event bus, but only for specific events

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -365,6 +367,30 @@ class CSP_API SpaceEntitySystem
     /// @return: MultiplayerConnection*
     CSP_NO_EXPORT MultiplayerConnection* GetMultiplayerConnectionInstance();
 
+    /**
+     * @brief Fetch all entities in the space from MCS
+     *
+     * Uses SignalR to fetch all the entities from MCS and populate the entities buffer.
```
MAG-mv: [Question] I see a lot of comments around SignalR. I imagine in the future these should be replaced, as we want Realtime to be agnostic to the  underlying connection implementation

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -91,43 +107,72 @@ class CSP_API IRealtimeEngine
...
+    /// @param AvatarState csp::multiplayer::AvatarState : The initial Avatar State to set.
     /// @param AvatarId csp::common::String : The ID to be set on the AvatarSpaceComponent
     /// @param AvatarPlayMode csp::multiplayer::AvatarPlayMode : The Initial AvatarPlayMode to set.
     /// @param Callback csp::multiplayer::EntityCreatedCallback A callback that executes when the creation is complete,
     /// which will provide a non-owning pointer to the new SpaceEntity so that it can be used on the local client.
-    CSP_ASYNC_RESULT virtual void CreateAvatar(const csp::common::String& Name, const csp::multiplayer::SpaceTransform& SpaceTransform,
-        const csp::multiplayer::AvatarState& State, const csp::common::String& AvatarId, const csp::multiplayer::AvatarPlayMode& AvatarPlayMode,
-        csp::multiplayer::EntityCreatedCallback Callback)
-        = 0;
+    CSP_ASYNC_RESULT virtual void CreateAvatar(const csp::common::String& Name, const csp::common::Optional<csp::common::LoginState> LoginState,
```
MAG-mv: [Question] I believe we had this conversation before - Why do we need to pass LoginState through? Can it not be passed to the RealtimeEngine when we connect?

MAG-mv: Also, under what circumstances would we not need authentication? Is this just future proof?

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -367,6 +175,144 @@ class CSP_API SpaceEntitySystem
...
+    void UnlockEntityUpdate() const;
+
+    /// @brief Queues a specific entity to update. Used in SpaceEntity to queue updates via passing the this pointer
+    void QueueEntityUpdate(SpaceEntity* EntityToUpdate);
+
+    /// @brief "Resolves" the entity heirarchy, such that the entity is parented appropriately, and internal buffers are populated appropriately.
+    /// (Vague, need more understanding about what this does)
+    void ResolveEntityHierarchy(SpaceEntity* Entity);
+
+    /********************** REALTIME ENGINE INTERFACE ************************/
```
MAG-mv: [Consider] putting RealtimeEngine interface at the top for better visibility

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -953,6 +951,31 @@ void SpaceEntity::Unlock()
 
 bool SpaceEntity::IsLocked() const { return EntityLock != LockType::None; }
 
+bool SpaceEntity::SetSelectionState(bool SelectionState, uint64_t ClientId)
```
MAG-mv: [Comment] These are the kind of functions that wil make more sense to be on SpaceEntitySystem/Realtime engine in the future.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
MAG-ElliotMorris: I _think_ we wanna keep it at the earliest date the file was made, as this represents when copyright protection started from.

But I'll admit I didn't think about it, just did a bit of research just now. I don't think a rename constitutes a new file, but 🤷

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/786938a41a2e9f84be11ed26d7f51fbadda0f7df

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/786938a41a2e9f84be11ed26d7f51fbadda0f7df

Tests/src/PublicAPITests/ComponentTests/CollisionComponentTests.cpp
MAG-ElliotMorris: Aye, I changed a few in this PR that caused issues from wanting to be ASSERT, but not going to go through here.

Tools/WrapperGenerator/Templates/CSharp/Partials/Class/Event.mustache
MAG-ElliotMorris: I think you need more than that, as the C# build will complain about function hiding for the event itself. [error CS0114](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0114)

> 1>C:\dev\connected-spaces-platform\Tools\WrapperGenerator\Output\CSharp\Multiplayer\OnlineRealtimeEngine.cs(525,50,525,62): error CS0114: 'OnlineRealtimeEngine.CreateAvatar(string, LoginState?, SpaceTransform, bool, AvatarState, string, AvatarPlayMode)' hides inherited member 'IRealtimeEngine.CreateAvatar(string, LoginState?, SpaceTransform, bool, AvatarState, string, AvatarPlayMode)'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword.

This is a warning (as error), but a very reasonable one imo. I think you need both virtual and override to satisfy it.

Library/include/CSP/CSPFoundation.h
MAG-ElliotMorris: Now ... putting cultural imperialism aside, I agree with you on this. _However_, `Initialise` is called by every client, and changing the spelling would be a break. I'm going to remain file consistent for this one, because I don't want to muddy up an already quite breaky PR with other unrelated breakages.

Nonetheless, we should do a big replace of all the ones that have crept in at some point.

Tests/src/PublicTestBase.cpp
MAG-ElliotMorris: Sure, can't see a downside. https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/5459662e10e1afe9c211f5d389aa2c01d45ea368

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
```diff
@@ -161,7 +161,7 @@ class CSP_API SpaceEntitySystem : public csp::common::IRealtimeEngine
     /// @param NetworkEventBus csp::multiplayer::NetworkEventbus& : Reference the the network event bus, used for leadership election messaging.
     /// @param RemoteScriptRunner csp::common::IJSScriptRunner& : Object capable of running a script. Called to execute scripts when the leader
     /// election system
-    CSP_NO_EXPORT SpaceEntitySystem(MultiplayerConnection* InMultiplayerConnection, csp::common::LogSystem& LogSystem,
```
MAG-mv, line 354: [Comment] TODO for me: Check if javascript tests have been updated to prove these constuctors work for clients. This would be a nice thing to check in c# too, but we dont have tests anymore

Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -72,13 +72,17 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
      *   @{ */
 
     /// @brief Enter a space if you have permission to, based on the Space settings.
-    /// This includes setting scopes (and toggling event listening in order to set the scope).
-    /// It also retrieves all entities in the space. Ensure Connect is called prior to this.
+    /// Registers the user as in the space on the backend service, and calls csp::common::IRealtimeEngine::FetchAllEntitiesAndPopulateBuffers.
```
MAG-mv, line 75: [Consider] Make comment more generic as public docs are harder to keep up-to-date if we're mentioning internal function calls.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: The only reason I can think that would make this fail is a connection error as you are leaving the space, or this function is being called while not in a space. I imagine connection error scenario would be cleaned up by MCS, anyhow.

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -118,6 +128,24 @@ class CSP_API IRealtimeEngine
...
+     * This method is called during csp::systems::SpaceSystem::Enterspace.
+     *
+     * @param SpaceId const csp::common::String& : The identifier of the space to fetch entities from. Format depends on how your specific backend
+     * store indexes spaces.
+     * @param FetchStartedCallback EntityFetchStartedCallback : Callback notifying for when csp::systems::SpaceSystem::EnterSpace can yield control
+     * back to clients
+     * @params FetchCompleteCallback EntityFetchCompleteCallback : Callback notifying when all entities are fetched and stored appropriately, and
+     * clients may start entity mutation.
+     *
+     * @post @param FetchStartedCallback will have been called, and @param FetchCompleteCallback will either have been called, or an async job started
```
MAG-mv: [Comment] I'm pretty sure linking like this doesn't work: and @param FetchCompleteCallback

WasmTesting/html_tests/EnterSpace.html
```diff
@@ -29,9 +29,13 @@
           console.error("Failed to log in");
         }
 
+        //Create multiplayer realtime engine
+        const realtimeEngine = Systems.SystemsManager.get().makeOnlineRealtimeEngine();
```
MAG-mv, line 33: [Question] Is it possible to add a test for the realtimeengine constructor at this point?

MultiplayerTestRunner/src/SpaceRAII.cpp
```diff
@@ -49,11 +49,11 @@ SpaceRAII::SpaceRAII(std::optional<std::string> ExistingSpaceId)
...
+    std::future<csp::systems::SpaceResult> ResultFuture = ResultPromise.get_future();
 
     SpaceSystem.EnterSpace(SpaceId.c_str(), RealtimeEngine.get(),
```
MAG-mv, line 60: [Comment] I think the result should represent what has been modified/retrieved/deleted by the call. The space technically hasn't been modified here. In the future, I imagine we would want to include more in-space related details (if we have any) here instead.

I'm not fussed about this change, though

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
MAG-ChristopherAtkinson: Yeah this is always something I have to check myself, but that sounds reasonable, will resolve the conversation

MAG-ChristopherAtkinson (commented):
That for addressing the minor comments, there are some comments you left yourself that I would like to be included related to logging, and inline documentation 

Happy to promote these to 'request changes' since GH won't allow you to do this on your own PR?

MAG-ElliotMorris:
Aye, that's cool. I'm working on those changes, am just going easiest -> hardest.

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/2cca3e1ebc6f95d79c9fe0bddcd26bde40cf8df7

Library/src/Multiplayer/SignalR/SignalRConnection.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/2cca3e1ebc6f95d79c9fe0bddcd26bde40cf8df7

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/45225db156d7bd6f84d59f35d6459c42c65222b8

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-ElliotMorris: Actually just removed this, after talking to @MAG-mv 

This exists purely because of the weird coupling between `SpaceEntity` and `SpaceEntitySystem`, as some of that has been removed in this PR, this method is redundant : https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/be8dc7f83db015b3d4ce4da645fc94cf4150f817

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: Yeah, I think I agree.

I've decided to call this path a fatal error, which should imply that the user should shut down their application. I don't think we're at any risk of data loss however, just invalid state.

Have nulled the realtime engine out in both cases, that way, the chance of the engine trying to send state updates and confusing the server is removed. 
https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/b8332bf726f1a357e2d148eb90b39fc7d285dd91

MAG-mv (changes requested):

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
MAG-ElliotMorris: Remember this is the OnlineRealtimeEngine, I think this sort of specific comment is actually exactly what you want in the concretion.

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
MAG-ElliotMorris: We had an offline chat about this, Matt has a good insight here I think, the loginstate is redundant for CreateAvatar in an offline flow. I didn't follow that he was talking about behaviour divergence, not just interface preference.

Will [fix] this, he's right.

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
MAG-ElliotMorris: There is a test that enters a space, confirming the concept works.

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-ElliotMorris: Disagree on this, I think `FetchAllEntitiesAndPopulateBuffers` is a pretty important function to specifically call out, it's not incidental, I added it quite intentionally to be a point of interface stability, the thing that controls when the concrete realtime engine is happy to yield backs to clients.

You (as a realtime engine author) _must_ be prepared for it to be called here in this flow, and you _must_ respond appropriately with the callbacks as documented in the interface, as failing to do so will cause us not to notify completion at the correct times. You get to make the choice as to whether you call both callbacks syncronously, or defer one for an async fetch after having entered the space (as the online engine does ... but I expect the offline engine probably will not)

WasmTesting/html_tests/EnterSpace.html
MAG-ElliotMorris: I'm not sure what you mean? Like assert that it's constructed correctly in this test? It must be functioning here, or the EnterSpace flow would fail.

MultiplayerTestRunner/src/SpaceRAII.cpp
MAG-ElliotMorris: I just thought that since we had a SpaceResult, it seemed silly not to provide it to clients. Though I did expect pushback. 
Unless you think this should be reverted, will leave it in.

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
MAG-ElliotMorris: Yeah. I've just removed that, you're right, dosen't work.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/e1c20e4103c48632da5c47dda3c6d8096f348977

Library/include/CSP/Multiplayer/SpaceEntitySystem.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/c68133f6fd419b0bc31ce8885b0c93f65d04f57e

Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
MAG-ElliotMorris: So, can't be passed to RealtimeEngine when we connect because then we'd need to design a mechanism to keep it up to date as the token is refreshed. Doing this via a pointer is a bad idea, because whilst it probably won't be, we don't really have a good guarantee that the source of the pointer won't be reseated, it's just fragile.

Instead, have sharpened the interface to what is actually neccesary, and am now passing through just a UserID. Which I grant is a bit tenuous as to whether you need that in a single player interface, but I can imagine it having a use. Not the best, one of those not-great interfaces we spoke about when we made the initial plans.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/2bc70b14e173a05af3b3cdc87e27f28e6802b542

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-mv: Makes sense

MultiplayerTestRunner/src/SpaceRAII.cpp
MAG-mv: All good

WasmTesting/html_tests/EnterSpace.html
MAG-mv: We're only testing the makeOnlineRealtimeEngine, when clients can call "new OnlineRealtimeEngine(...)". We should test that the client ownership is working

MAG-ChristopherAtkinson (commented):
Hey sorry for the delay getting back on this, I am happy with the changes from my point of view, pending conflict resolution and open comments. Happy to approve.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -0,0 +1,452 @@
...
+    ///
+    /// Only one EntityCreatedCallback may be registered, calling this function again will override whatever was previously set.
+    ///
+    /// @param Callback csp::multiplayer::EntityCreatedCallback : the callback to execute.
+    CSP_EVENT virtual void SetEntityCreatedCallback(csp::multiplayer::EntityCreatedCallback Callback) override;
+
+    /// @brief Adds an entity to the set of selected entities
+    /// @param Entity csp::multiplayer::SpaceEntity* Entity to set as selected
+    /// @return True if the entity was succesfully added, false if the entity already existed in the selection and thus could not be added.
+    virtual bool AddEntityToSelectedEntities(csp::multiplayer::SpaceEntity* Entity) override;
```
MAG-mv: [fix] I discovered this issue while working offline. I think this should be CSP_NO_EXPORT, as we don't want this to be called by clients. 

Thinking about this more... I don't know who this plays with inheritance in our wrapper gen

WasmTesting/html_tests/EnterSpace.html
MAG-ElliotMorris: We discussed this offline and traced through the interop code, the concern was over potential double copies and we're satisfied it's not going to do that.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755/commits/ba2f0108aada8d7b2e30b2499264c9482d83d7a8

I'm going to go ahead and rebase and resolve conflicts now, as this is the last change.
Sorry, normally i'd wait for confirmation but am under time crunch.

MAG-ChristopherAtkinson (approved):

MAG-ElliotMorris:
That should be all changes addressed. As far as I'm concerned, if the checks pass, this can go in. (Don't merge it though to anyone reading this, we're managing a release)

PR #757, opened 2025-08-01
[OF-1732] fix!: Shorthand signature wrapper gen bug workaround

MAG-ElliotMorris:
Workaround the wrapper generator bug with shorthand signature generation.

There is an issue where if the return type of the accepting function is different from the return type of the registering callback (at least in cases like this where we're registering for later behaviour, not making a task/promise sort of construct), then the shorthand type sig used for binding callbacks will generate wrong, using the return type of the enclosing function.

Coincidently, we'd never done this before.

Wrapper generator bug still exists, this just avoids it.

Add a regression test, which I saw fails before the change, and passes after.

**Breaking changes**
Return types of following methods changed from bool -> void
- NetworkEventBus::ListenNetworkEvent(NetworkEventRegistration, NetworkEventCallback)
- NetworkEventBus::StopListenNetworkEvent(NetworkEventRegistration Registration)
- NetworkEventBus::StopListenAllNetworkEvents(const csp::common::String& EventReceiverId)

### DIFF
--- a/WasmTesting/csp-wasm.test.ts
+++ b/WasmTesting/csp-wasm.test.ts
@@ -59,6 +59,28 @@ test('Cross Thread Callbacks From Log Callback, OB-3782', async () => {
   
 });
 
+test ('SendReceiveNetworkEvent', async() => {
+  // This test was added as a regression test against `RuntimeError: null function or function signature mismatch`
+  // Caused by a wrapper gen bug when you make a return type of an enclosing function different for the return type of the callback
+  // We didn't actually fix it at time of writing, change `ListenNetworkEvent` to return a bool and you'll see what I mean.
+  const user = await CreateTestUser();
+  await LoginAsUser(user);
+  const spaceId = await CreatePublicTestSpace();;
+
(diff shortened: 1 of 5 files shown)


WasmTesting/csp-wasm.test.ts
```diff
@@ -59,6 +59,28 @@ test('Cross Thread Callbacks From Log Callback, OB-3782', async () => {
   
 });
 
+test ('SendReceiveNetworkEvent', async() => {
+  // This test was added as a regression test against `RuntimeError: null function or function signature mismatch`
+  // Caused by a wrapper gen bug when you make a return type of an enclosing function different for the return type of the callback
+  // We didn't actually fix it at time of writing, change `ListenNetworkEvent` to return a bool and you'll see what I mean.
+  const user = await CreateTestUser();
+  await LoginAsUser(user);
+  const spaceId = await CreatePublicTestSpace();;
```
MAG-ChristopherAtkinson: [Consider] remove double ';' here, it doesn't prevent functionality hence not promoted to fix

MAG-ChristopherAtkinson (approved):
Changes look good, minor consider but doesn't affect functionality

WasmTesting/csp-wasm.test.ts
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/757/commits/6ee45052022e0376e810c1f37b237155f8240f1a

MAG-AdamThorn (approved):

PR #759, opened 2025-08-05
OF-1714 build: Ensure internal headers compile with warnings

MAG-AdamThorn:
We have enabled Warnings as Errors, however we currently classify all includes as external includes which mean they are getting compiled without warnings.

To ensure that includes are correctly classified I have updated the premake file to remove src and include directories from `externalincludedirs` and to add them to `includedirs` instead.

This resulted in a number of wErrors which needed to be resolved. I have endeavoured to group these changes by error type as much as possible. This is reflected in the commits.

### DIFF
--- a/Library/include/CSP/Systems/Settings/ApplicationSettings.h
+++ b/Library/include/CSP/Systems/Settings/ApplicationSettings.h
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/Array.h"
+#include "CSP/Common/Map.h"
+#include "CSP/Common/String.h"
+#include "CSP/Systems/WebService.h"
+
+namespace csp::services
+{
+
+class ApiResponseBase;
+
+CSP_START_IGNORE
+template <typename T, typename U, typename V, typename W> class ApiResponseHandler;
+CSP_END_IGNORE
+
+} // namespace csp::services
+
+namespace csp::systems
+{
+
+/// @ingroup Application Settings System
+/// @brief Represents configuration settings for an application context.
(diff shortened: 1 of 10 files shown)


Library/include/CSP/Common/Interfaces/IJSScriptRunner.h
```diff
@@ -65,13 +65,23 @@ class CSP_API IJSScriptRunner
...
      */
-    virtual bool RunScript(int64_t ContextId, const String& ScriptText) { throw InvalidInterfaceUseError("Illegal use of \"abstract\" type."); }
+    virtual bool RunScript(int64_t ContextId, const String& ScriptText)
+    {
+        // Marking parameters as unused by casting them to void to suppress warnings.
+        // All other methods are marked as CSP_NO_EXPORT which means that parameter names can be commented out to suppress warnings,
+        // however this method is exported and the wrapper generator does not support that approach.
+        // C++ 17 adds support for the [[maybe_unused]] attribute, but again the wrapper generator does not support it.
+        (void)ContextId;
+        (void)ScriptText;
```
MAG-ElliotMorris, line 75: [comment] Gross, but good solve. Damn you wrapper generator!

Library/src/Common/Algorithm.h
```diff
@@ -21,15 +21,16 @@
...
 namespace csp::common
 {
 
 /// @brief Finds a given value in an array.
 /// @param Array const csp::common::Array<T>& : Array to search
 /// @param Value const T& : Value to search for
-/// @return int : Index of first found element or -1 if not found
-template <typename T> int Find(const Array<T>& Array, const T& Value)
+/// @return std::optional<size_t> : Index of first found element or nullopt if not found
+template <typename T> std::optional<size_t> Find(const Array<T>& Array, const T& Value)
```
MAG-ElliotMorris, line 33: [comment] Oh hey this is a sort of big deal! Nice.

MAG-ElliotMorris: ... although why on earth do we have a custom algorithm header anyway... Adding iterators to `Array` would let us do this. 🤷

Library/include/CSP/Multiplayer/Conversation/Conversation.h
```diff
@@ -72,6 +72,7 @@ class CSP_API MessageInfo
         const csp::common::String& ConversationId, bool IsConversation, const csp::common::String& Message, const csp::common::String& MessageId);
 
     MessageInfo(const MessageInfo& MessageData);
+    MessageInfo& operator=(const MessageInfo& MessageData);
```
MAG-ElliotMorris, line 75: [comment] Another real thing!

MAG-ElliotMorris (approved):
🙏 Nice. Thanks for doing commit isolation.

Library/src/Common/Algorithm.h
MAG-AdamThorn: That's a good call. Only one of the methods is in use as well so another reason to do it. I will capture that work and fit it in.

PR #760, opened 2025-08-05
[OF-1696] docs: added cppcheck section to external tools section

MAG-ChristopherAtkinson:
This PR introduces a new section to the project's documentation covering the use of Cppcheck, a static analysis tool for C++ code.

The primary goal of this documentation is to inform external developers about the tool, its purpose, and its specific configuration within the project's build process. This provides clarity on how we leverage static analysis to maintain code quality, even though its use is currently optional.


MAG-ChristopherAtkinson:
> Did we ever fix the errors that were found with Cpp check? Without those fixed, I assume it wont actually be supported for CSP?

These changes don't enforce cppcheck in any capacity. When this is integrated with CI, we will disable everything. From there, we can systematically document and prune as we enable subsets 

If you are interested in the current state, there is xml file attached to the ticket that can be opened in cppcheck GUI or text editor

MAG-AdamThorn (commented):
[Question] Where is it being enforced? It is not being enforced as part of this PR.

MAG-ChristopherAtkinson:
> > These changes don't enforce cppcheck in any capacity. When this is integrated with CI, we will disable everything. From there, we can systematically document and prune as we enable subsets
> 
> I am slightly confused. The PR is a documentation only change. The documentation states:
> 

From the discussion in stand-up the other day, we decided that we would document its use and ticket separately for the integration with CI pipeline. 

>
> "While Cppcheck is not currently an enforced part of the solution's build process, it is configured for optional use to assist developers in maintaining code quality."
> 

This should be updated to reflect that we provide configuration:

"While Cppcheck is not currently an enforced part of the solution's build process, we provide instructions to configure Cppcheck for optional use to assist developers in maintaining code quality."

>
> Where is it configured and how? We should be adding it as part of this PR otherwise we have documentation about something that doesn't exist.
>

The documentation provides instructions for use, but not any direct configuration

MAG-mv:
So, based on the xml, it is working? We're also encountering some unexpected errors eround includes, like so?
image

PR #761, opened 2025-08-08
[OF-1730] feat: Added new Screen Sharing component

MAG-AdamThorn:
Added a new Screen Sharing component which is designed to facilitate the sharing of a user's screen within a connected space.
Added tests to ensure the component can be constructed and updated correctly, as well as to ensure component property script bindings work correctly.

**Note for the reviewer:** While I was working on the new component tests, I noticed a couple of minor issues with our existing component tests, namely inconsistent test space creation and some erroneous comments. While I was in the area I elected to resolve these issues which is why there are so many file changes. This PR has two commits: Commit 1 has all of the changes required to add the new Screen Sharing component and it's tests. Commit 2 has all of the existing component updates. It is worth reviewing this PR by commit.

### DIFF
--- a/Library/src/Multiplayer/Components/ScreenSharingSpaceComponent.cpp
+++ b/Library/src/Multiplayer/Components/ScreenSharingSpaceComponent.cpp
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "CSP/Multiplayer/Components/ScreenSharingSpaceComponent.h"
+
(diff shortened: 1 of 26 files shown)


Library/src/Multiplayer/Components/ScreenSharingSpaceComponent.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+ScreenSharingSpaceComponent::ScreenSharingSpaceComponent(csp::common::LogSystem* LogSystem, SpaceEntity* Parent)
+    : ComponentBase(ComponentType::ScreenSharing, LogSystem, Parent)
+{
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::Position)] = csp::common::Vector3::Zero();
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::Rotation)] = csp::common::Vector4::Identity();
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::Scale)] = csp::common::Vector3::One();
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::IsVisible)] = true;
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::IsARVisible)] = true;
+    Properties[static_cast<uint32_t>(ScreenSharingPropertyKeys::IsShadowCaster)] = true;
```
MAG-ChristopherAtkinson: [Fix] the default for this property should be 'false'

MAG-AdamThorn: Ah yes, well spotted.

Fix: [1da3c47](https://github.com/magnopus-opensource/connected-spaces-platform/pull/761/commits/1da3c4730edca9dfb56452555afcfc19cd0e65aa)

PR #762, opened 2025-08-08
[OF-1724] fix: race condition in conversation flows

MAG-ChristopherAtkinson:
## Race condition in conversation flows

Our async code takes callbacks, which is what blocks external users from continuing. These async chains trigger the callback in some of the intermediary functions, rather than deferring it until the entire chain is complete. Depending on the execution speed of the subsequent chains/the main thread, this may or may not manifest as a crash.

Following this, the PR addresses these issues with the following changes:

 - removed the concept of callback being managed by the intermediary functions
 - introduce ambiguous callable to expected vs unexpected params
 - updated usage of continuations throughout to reflect these changes 
 - ensured that any chain handles the callback for both success and failure cases where appropriate 
 - update the test to reflect the expectation for callbacks in  intermediary functions
 - made some general improvements around continuations  (comments, formatting, default scheduler)
 - added hierarchical exception handling  
 
As part of testing this I have run the CSPEngine.ConversationTests using the --gtest_repeat=<Quantity> flag, and I am now seeing reliable execution for the test. 

## Exception handling 

Exception Hierarchy: All "expected" errors (like a failed HTTP request or a multiplayer connection issue) are derived from a common, non-templated base class, ExpectedExceptionBase. All other errors remain as standard std::exception types. This creates a clear distinction between anticipated errors and unexpected, unrecoverable ones.

Specialized Handler: A function, InvokeIfExceptionInChain, is used at the end of the continuation chain. This function has two distinct catch blocks. The first catch (const ExpectedExceptionBase&) specifically handles all anticipated errors. The second, more general catch (const std::exception&), acts as a fallback for all other, unexpected exceptions.

This design ensures that expected errors are handled gracefully with specific logic and rich data, while unexpected errors are caught and logged appropriately, preventing application crashes.

### Advantages

Safety and Robustness: The biggest advantage is compile-time safety. The compiler enforces that you handle different exception types separately, preventing you from accidentally forgetting to handle an important error. This design is robust because it relies on the C++ type system, not on a fragile dynamic_cast at runtime. ( We have limitations around the use of dynamic_cast due to WASM  as I understand it)

Clarity and Readability: The function signature InvokeIfExceptionInChainP1(ExpectedCallable, UnexpectedCallable, ...) clearly communicates its purpose. It's designed to handle two distinct categories of errors. The code is self-documenting, making it easy for other developers to understand at a glance.

Centralized Logic: All the logic for distinguishing between expected and unexpected errors is centralized in one place: the InvokeIfExceptionInChainP1 function. If you need to add a new ExpectedExceptionBase type, you only need to modify this function, not every single place it's called.

Scalability: This approach is highly scalable. You can add new expected error types (e.g., DatabaseException) to your exception hierarchy, and the system will automatically handle them through the generic catch (const ExpectedExceptionBase&) block.

### Disadvantages 

Slightly More Verbose at the Call Site: The user has to provide two separate callbacks, even if one is just a default "do nothing" lambda. This can feel slightly more cumbersome compared to a single-callback API.

Can Be Harder for New Developers to Grasp: The concept of a type hierarchy and providing two separate callbacks might have a slightly steeper learning curve for developers not familiar with these design patterns. However, the resulting code is much safer and more maintainable.

## Reviewer Notes

As part of these changes, I have updated the test to reflect the expectation that the callback will not be invoked by the intermediary functions. I would pay close attention here, as I believe I have made the correct choice in explicitly checking this behaviour.

### DIFF
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -481,7 +481,6 @@ async::task<NullResult> AssetSystem::DeleteAssetCollection(const AssetCollection
 
     if (PrototypeId.IsEmpty())
     {
-        CSP_LOG_ERROR_MSG("A delete of an asset collection was issued without an ID. You have to provide an asset collection ID.");
         OnCompleteEvent.set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
             "A delete of an asset collection was issued without an ID. You have to provide an asset collection ID.", MakeInvalid<NullResult>())));
 
@@ -1110,7 +1109,6 @@ async::task<AssetsResult> AssetSystem::GetAssetsByCriteria(const csp::common::Ar
 
     if (AssetCollectionIds.IsEmpty())
     {
-        CSP_LOG_ERROR_MSG("You have to provide at least one AssetCollectionId");
         OnCompleteEvent.set_exception(std::make_exception_ptr(
             csp::common::continuations::ResultException("You have to provide at least one AssetCollectionId", MakeInvalid<AssetsResult>())));
 
@@ -1269,7 +1267,6 @@ async::task<UriResult> AssetSystem::UploadAssetDataEx(const AssetCollection& Ass
 
     if (Asset.Name.IsEmpty())
     {
-        CSP_LOG_ERROR_MSG("Asset name cannot be empty");
         OnCompleteEvent.set_exception(
             std::make_exception_ptr(csp::common::continuations::ResultException("Asset name cannot be empty", MakeInvalid<UriResult>())));
 


Library/include/CSP/Systems/ContinuationUtils.h
```diff
@@ -33,16 +33,11 @@ using namespace csp::systems;
...
+inline void LogHTTPErrorAndCancelContinuation(std::string ErrorMsg, EResultCode ResultCode, csp::web::EResponseCodes HttpResultCode,
+    ERequestFailureReason FailureReason, csp::common::LogLevel LogLevel = csp::common::LogLevel::Log)
 {
     CSP_LOG_MSG(LogLevel, ErrorMsg.c_str());
-    ErrorResultT FailureResult(ResultCode, HttpResultCode, FailureReason);
-    if (Callback)
-    {
-        Callback(FailureResult);
-    }
+    ResultBase FailureResult(ResultCode, static_cast<std::underlying_type<csp::web::EResponseCodes>::type>(HttpResultCode), FailureReason);
```
MAG-ElliotMorris: [Question] What's the point of this? It seems to be a completely unused type. You'd probably get an error if this was compiled against the warnings-as-errors change Adam just made.

Library/include/CSP/Systems/WebService.h
```diff
@@ -38,6 +38,8 @@ class CSP_API ResultBase
     /// @brief Constructs an empty result.
     ResultBase();
 
+    ResultBase(csp::systems::EResultCode ResCode, uint16_t HttpResCode, csp::systems::ERequestFailureReason Reason);
```
MAG-ElliotMorris: [Question] I am presuming this was necessary due to the need to construct one in the prior commit? See prior comment on that.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -803,23 +803,23 @@ void ConversationSystemInternal::GetConversationAnnotation(const csp::common::St
...
         .then(GetAnnotationAsset(AssetSystem, ConversationAssetCollection))
-        .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetsResult>(Callback,
+        .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetsResult>(
             "ConversationSystemInternal::GetConversationAnnotation, successfully retrieved annotation asset", "Failed to get annotation asset.", {},
             {}, {}))
         .then(SetAnnotationAssetFromAssets(AnnotationAsset))
         // 4. Get annotation thumbnail asset
         .then(GetAnnotationThumbnailAsset(AssetSystem, ConversationAssetCollection))
-        .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetsResult>(Callback,
+        .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetsResult>(
```
MAG-ElliotMorris, line 818: [fix] Here and everywhere else.

By making such a direct change, you have removed the possibility of providing useful values to the arguments passed in the error callbacks. I note we have almost never (perhaps actually never) used that facility, but the interface remains to provide the values which are now not-capable of being used.

I don't dislike the idea of always handling errors in an exception catch, but you need to think it right through to the user. You probably want to make a `ResultException` and throw that in the cancellations, such that when you catch it you can use that result in the callback.

That ... or perhaps the arguments should be fully expunged? Would there be any consequences to that?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -186,19 +186,17 @@ std::function<async::task<UriResult>()> SpaceSystem::CreateAndUploadSpaceThumbna
...
                 "SpaceSystem::CreateAndUploadSpaceThumbnailToSpace, successfully created space thumbnail asset collection.",
                 "Failed to create space thumbnail asset collection.", {}, {}, {}))
             .then(systems::continuations::GetResultFromContinuation<AssetCollectionResult>(ThumbnailAssetCollection))
             .then(CreateSpaceThumbnailAsset(Space, ThumbnailAssetCollection))
-            .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetResult>(Callback,
+            .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<AssetResult>(
                 "SpaceSystem::CreateAndUploadSpaceThumbnailToSpace, successfully created space thumbnail asset.",
                 "Failed to create space thumbnail asset.", {}, {}, {}))
             .then(UploadSpaceThumbnailAsset(ThumbnailAssetCollection, *Data))
-            .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<UriResult>(Callback,
```
MAG-ElliotMorris, line 202: [Question] Is there a test to confirm that this behaves as you expect? We're now relying on a exception being thrown via a `std::function` into another chain, presumably to be caught at the end of it. Be easy to accidentally not catch.

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
```diff
@@ -64,14 +64,13 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceEntitySystemTests, TestSuccessInRemoteGenerateNe
             });
 
     SpaceEntitySystem->RemoteGenerateNewAvatarId()
-        .then(async::inline_scheduler(),
```
MAG-ElliotMorris: [Consider] Please double check that these are actually unnecessary in all projects. (I'm not sure how, maybe you can debug, or print the thread-ids.)

I know Matt made a change on this, but as far as I know there is no testing for it, and I'm dubious if it actually propagates everywhere for 2 reasons. 1 - As far as I know it was defined in a header, what if it's not included. and 2, How do we know it propagates to the test project?

Keep in mind that to us, this will look almost identical whether it's using inline_scheduler or threadpool, but it is important to be using inline even in tests.

Forgive me if there actually are tests for this. Although I _am_ worried about the reliance on remembering to have done an include.

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
```diff
@@ -74,14 +74,13 @@ CSP_PUBLIC_TEST(CSPEngine, GeneralContinuationsTests, TestLogErrorAndCancel)
     NullResult ExpectedResult(ResultCode, HttpResultCode, FailureReason);
 
     ::testing::MockFunction<void(const NullResult& Result)> MockResultCallback;
-    // Expect that the callback is called with the result constructed as expected.
-    EXPECT_CALL(MockResultCallback, Call(ExpectedResult)).Times(1);
+    // Expect that the callback is not called from the intermediate function.
+    EXPECT_CALL(MockResultCallback, Call(ExpectedResult)).Times(0);
```
MAG-ElliotMorris: [fix] This is irrelevant. There is no need to test a callback that is never passed to anything. Here and elsewhere.

Library/include/CSP/Systems/ContinuationUtils.h
MAG-ChristopherAtkinson: Yeah, this was missed when refactoring out the callback. Previously, it was returned, but should be promoted to fix

Library/include/CSP/Systems/WebService.h
MAG-ChristopherAtkinson: This can be reverted once the previous issue has been resolved

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ChristopherAtkinson: Yes, I see your point. Currently, our user consideration is done through logging, and return logic for each continuation. Handling a result exception would be my preference here.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ChristopherAtkinson: The changes do not provide a guarantee that the callback will be invoked; as a result, this becomes the responsibility of CSP. Each implementation will need to have suitable coverage for both success and failure paths that would catch if the assumption is not met

Of course, if you have a good method to test this, I would be interested to hear.

MAG-ChristopherAtkinson: For this case, as the continuation does not hold the responsibility for invoking the callback, it is deferred to the owning chain, in this case, we have several failure paths that test the behaviour, for example, 'CreateSpaceWithInvalidThumbnailTest'

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
MAG-ChristopherAtkinson: That is a great point, I will see what I can find to prove this, otherwise I would recommend we promote this to a fix if we cannot ensure behaviour

MAG-ChristopherAtkinson: I took some time to look into this. Currently, we depend on including <async++.h> through CSPAsyncScheduler.h, in the event this was missed, it is likely that we would default back to the threadpool scheduler, which is not ideal.

Based on the documentation from async++ I this does look like it is being populated throughout the projects assuming the assumption is meet for includes as we explicitly define the pre-processor and default scheduler in CSPAsyncScheduler.h

https://github.com/Amanieu/asyncplusplus/wiki/API-Reference

However, this is not tested, and those concerns are valid. I would argue we promote this as: [fix] revert reliance on the default scheduler

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
MAG-ChristopherAtkinson: These have now been removed please see: [9eb5c10](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/9eb5c10f1cf6d52853cb7bf3e43fb6c5545f5333)

Library/include/CSP/Systems/ContinuationUtils.h
MAG-ChristopherAtkinson: Due to the rebase, these changes are no longer populated.

Library/include/CSP/Systems/WebService.h
MAG-ChristopherAtkinson: Due to the rebase, these changes are no longer populated.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ChristopherAtkinson: I ended up going with the suggestion to introduce the `ResultException` in its current form, which is a partial implementation that can be expanded in the future. 

I am considering whether we should populate this in our continuation chain currently to distinguish between result and standard exceptions for clarity.  @MAG-ElliotMorris what are your thoughts?

Tests/src/PublicAPITests/SpaceEntitySystemTests.cpp
MAG-ChristopherAtkinson: I did consider whether we should enforce defining inline schedulers. For the time being, I have opted not to enforce this, but maintain the current set. I will produce a ticket so we can dive a bit deeper into this as the potential to revert to the `default_threadpool_scheduler` is concerning

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ElliotMorris: > I am considering whether we should populate this in our continuation chain currently to distinguish between result and standard exceptions for clarity

This is vague and could mean a lot of things, try to be specific.

I'm going presume you're asking about how to adapt `InvokeIfExceptionInChain` to this new approach. 

Currently, it demands you pass a callable that can be provided a `std::exception`, which is workable, albeit annoying because it means in all error handling at the end of chains each implementor will need to inspect the exception type to see if it's a result exception vs another (unexpected) exception, and get the result out of it if appropriate, or do more basic fallback behaviour. 

We do need to be capable of handling both these cases, the category of exceptions that will fall under `ResultException` will be our expected errors, but non-expected errors do happen, and will fall under `std::exception`

My mind goes to 2 solutions:
- Make the function take 2 callbacks, `InvokeIfExpectedErrorCallable` and `InvokeIfUnexpectedErrorCallable`. This is the least prone to forgetting to catch errors, but also the most burdensome, and may cause some annoying duplication.
- Establish a pattern where the callables provided to `InvokeIfExceptionInChain` always check the exception type to see if there's a result, to decide if they're providing a rich error callback, or an emergency `MakeInvalid` one with an "unexpected error" sort of message. It's possible this makes `InvokeIfExceptionInChain` sort of redundant, as the case by case handling will have to be done by every user now, but I can't be sure without doing the working.

Of these 2 (and there may be more options), I'm not sure which is better. If it were me I'd prototype both in a few instances and see how the usage patterns felt.

MAG-ElliotMorris: Also just as a side note, do keep in mind the multiplayer error side of this. `ResultException` makes sense for `Results`, but not everything has results.

MAG-ChristopherAtkinson: > Also just as a side note, do keep in mind the multiplayer error side of this. `ResultException` makes sense for `Results`, but not everything has results.

Yeah in this case, I was cautious that we are using the current continuation, so this should only consider the chains for https results. Regarding multiplayer, we are currently logging and still throwing a standard exception.

MAG-ChristopherAtkinson: > 
> This is vague and could mean a lot of things, try to be specific.
> 

Yeah, I mean, do we want to adopt this as a manual process or build it into our utility functions? 

I will give this some thought and take a look at both options

MAG-ChristopherAtkinson: The prototype has now been applied to the solution [df960ed](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/df960edff204b018fabb593dfe7f2adf7e06110b)

I am going to update the description for the PR with additional information regarding the changes

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -23,10 +23,78 @@ CSP_START_IGNORE
...
+        : std::runtime_error(message)
+    {
+    }
+
+    virtual const ExceptionType GetExceptionType() const = 0;
+};
+
+/**
+ * @brief An exception class for API request results.
+ * @details This template class captures a the result object alongside the
```
MAG-AdamThorn: [fix] Remove the unnecessary 'a'.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -23,10 +23,78 @@ CSP_START_IGNORE
...
+
+    const csp::systems::ResultBase& GetResult() const { return Result; }
+
+private:
+    const csp::systems::ResultBase& Result;
+};
+
+/**
+ * @brief An exception class for Multiplayer Error code.
+ * @details This template class captures a the Multiplayer Error code object alongside the
```
MAG-AdamThorn: [fix] Same as above with an unnecessary 'a'.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -23,10 +23,78 @@ CSP_START_IGNORE
...
+ * exception. It's ideal for handling errors from Multiplayer where specific
+ * error codes need to be preserved.
+ */
+class ErrorCodeException : public csp::common::continuations::ExpectedExceptionBase
```
MAG-AdamThorn: [fix] this exception type has the full base class namespace whereas the one above does not.

MAG-AdamThorn: Same with ExceptionType in the getter.

Library/include/CSP/Systems/Assets/Asset.h
```diff
@@ -213,6 +213,9 @@ class CSP_API AssetResult : public csp::systems::ResultBase
     CSP_NO_EXPORT AssetResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode)
```
MAG-AdamThorn, line 213: [Question] Is there any reason not to update HttpResCode to be a web::EResponseCodes as below?

MAG-AdamThorn: Same with below.

Library/include/CSP/Systems/Assets/AssetCollection.h
```diff
@@ -142,6 +142,10 @@ class CSP_API AssetCollectionResult : public csp::systems::ResultBase
     CSP_NO_EXPORT AssetCollectionResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode)
```
MAG-AdamThorn, line 142: [Question] Is there any reason not to update HttpResCode to be a web::EResponseCodes as below?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -548,18 +555,18 @@ void SpaceSystem::CreateSpaceWithBuffer(const String& Name, const String& Descri
...
             "SpaceSystem::CreateSpaceWithBuffer, successfully created space metadata asset collection.",
             "Failed to create space metadata asset collection.", {}, {}, {}))
         .then(CreateAndUploadSpaceThumbnailWithBufferToSpace(CurrentSpaceResult, Thumbnail))
         .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<UriResult>(
-            Callback, "SpaceSystem::CreateSpaceWithBuffer, successfully created thumbnail.", "Failed to create thumbnail.", {}, {}, {}))
+            "SpaceSystem::CreateSpaceWithBuffer, successfully created thumbnail.", "Failed to create thumbnail.", {}, {}, {}))
         .then(BulkInviteUsersToSpaceIfNeccesary(this, CurrentSpaceResult, InviteUsers))
         .then(systems::continuations::AssertRequestSuccessOrErrorFromResult<NullResult>(
-            Callback, "SpaceSystem::CreateSpaceWithBuffer, successfully invited users to space.", "Failed to invited users to space.", {}, {}, {}))
+            "SpaceSystem::CreateSpaceWithBuffer, successfully invited users to space.", "Failed to invited users to space.", {}, {}, {}))
```
MAG-AdamThorn: [Fix] "Failed to invited users to space." > "Failed to invite users to space."

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -23,10 +23,78 @@ CSP_START_IGNORE
...
 namespace csp::common::continuations
 {
 
+enum class ExceptionType
+{
+    Result,
+    Multiplayer
+};
+
+class ExpectedExceptionBase : public std::runtime_error
```
MAG-ElliotMorris, line 44: [Consider] A brief line (doesn't need to be formal docs as this isn't public), describing that this is the base type for errors we expect to be thrown during normal CSP execution.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -43,33 +111,53 @@ inline void LogErrorAndCancelContinuation(
...
         {
             ExceptionTask.get();
         }
-        catch (const std::exception& exception)
+        catch (const ExpectedExceptionBase& ExpectedException)
         {
-            LogSystem.LogMsg(
-                csp::common::LogLevel::Verbose, "Caught exception during async++ chain. Invoking callable from InvokeIfExceptionInChain");
-            InvokeIfExceptionCallable(exception);
+            LogSystem.LogMsg(LogLevel::Verbose, "Caught Exception during async++ chain. Invoking callable from InvokeIfExceptionInChain");
```
MAG-ElliotMorris: [nit] decapitalise `exception`, really just so it mirrors the casing of the non-expected output.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -43,33 +111,53 @@ inline void LogErrorAndCancelContinuation(
...
         catch (const csp::common::InvalidInterfaceUseError& exception)
         {
-            LogSystem.LogMsg(
-                csp::common::LogLevel::Verbose, "Error, CSP expects derived IRealtimeEngine type, but has received a base instantiation. ");
-            InvokeIfExceptionCallable(std::runtime_error(exception.msg));
+            // Keep specific custom catches here for other types.
+            LogSystem.LogMsg(LogLevel::Verbose, "Error, CSP expects derived IRealtimeEngine type, but has received a base instantiation.");
+            InvokeIfUnexpectedErrorCallable(std::runtime_error(exception.msg));
+        }
+        catch (const std::exception& exception)
```
MAG-ElliotMorris, line 149: [nit] Maybe put the expected/unexpected catch blocks next to eachother. Having the bad-interface error in between them makes it a tad harder to parse.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -43,33 +111,53 @@ inline void LogErrorAndCancelContinuation(
...
+            InvokeIfUnexpectedErrorCallable(exception);
         }
     };
 }
 
+template <typename ExpectedCallable>
+inline auto InvokeIfExceptionInChain(csp::common::LogSystem& LogSystem, ExpectedCallable&& InvokeIfExpectedErrorCallable)
+{
+    return InvokeIfExceptionInChain(
+        LogSystem, std::forward<ExpectedCallable>(InvokeIfExpectedErrorCallable), []([[maybe_unused]] const std::exception& exception) {});
```
MAG-ElliotMorris, line 162: [fix] We should log something here, so it's not just an invisible error when folk "forget" to add an unexpected case handler.

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
```diff
@@ -173,35 +153,28 @@ CSP_PUBLIC_TEST(CSPEngine, GeneralContinuationsTests, TestAssertRequestSuccessOr
...
+        ASSERT_ANY_THROW(csp::common::continuations::AssertRequestSuccessOrErrorFromMultiplayerErrorCode(SuccessMsg.c_str(),
+            MakeInvalid<NullResult>(), *csp::systems::SystemsManager::Get().GetLogSystem(), csp::common::LogLevel::Log)(ErrorCode));
 
-        ::testing::Mock::VerifyAndClearExpectations(&MockResultCallback);
         ::testing::Mock::VerifyAndClearExpectations(&MockLogger.MockLogCallback);
     }
     {
-        // When we pass empty optionals, we expect default values to be used to construct the result passed in the callback //
+        // When we pass empty optionals, we expect default values to be used to construct the result passed in the callback
         NullResult DefaultFailureResult = MakeInvalid<NullResult>();
```
MAG-AdamThorn: [fix] We don't appear to be using this any more. The 2nd part of this test seems to be doing the same thing as the first.

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -43,33 +111,53 @@ inline void LogErrorAndCancelContinuation(
...
+        {
+            // Catches all other standard exceptions (the generic fallback).
+            LogSystem.LogMsg(LogLevel::Fatal, "Caught unexpected exception during async++ chain. Invoking callable from InvokeIfExceptionInChain");
+            InvokeIfUnexpectedErrorCallable(exception);
         }
     };
 }
 
+template <typename ExpectedCallable>
+inline auto InvokeIfExceptionInChain(csp::common::LogSystem& LogSystem, ExpectedCallable&& InvokeIfExpectedErrorCallable)
```
MAG-ElliotMorris, line 159: [Comment] Glad to see this. Was expecting a default argument, but this works just as well.

Library/include/CSP/Systems/ContinuationUtils.h
```diff
@@ -31,19 +31,12 @@ using namespace csp::systems;
...
+inline void LogHTTPErrorAndCancelContinuation(std::string ErrorMsg, ResultT Result, csp::common::LogLevel LogLevel = csp::common::LogLevel::Log)
 {
     CSP_LOG_MSG(LogLevel, ErrorMsg.c_str());
-    ErrorResultT FailureResult(ResultCode, HttpResultCode, FailureReason);
-    if (Callback)
-    {
-        Callback(FailureResult);
-    }
-    throw std::runtime_error("Continuation cancelled"); // Cancels the continuation chain.
+    throw csp::common::continuations::ResultException("Continuation cancelled", Result); // Cancels the continuation chain.
```
MAG-ElliotMorris: [fix] Perfectly forward `Result` here. 

... I know this probably feels redundant because we don't bother with movable types ... but still nice to write generic code correctly.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -481,7 +482,8 @@ async::task<NullResult> AssetSystem::DeleteAssetCollection(const AssetCollection
     if (PrototypeId.IsEmpty())
     {
         CSP_LOG_ERROR_MSG("A delete of an asset collection was issued without an ID. You have to provide an asset collection ID.");
-        OnCompleteEvent.set_exception(std::make_exception_ptr(std::exception()));
+        OnCompleteEvent.set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
+            "A delete of an asset collection was issued without an ID. You have to provide an asset collection ID.", MakeInvalid<NullResult>())));
```
MAG-ElliotMorris: [Nit] Here and elsewhere, would be nice if these strings were single source...

```
static constexpr std::string_view Error= "A delete of an asset collection was issued without an ID. You have to provide an asset collection ID.";
CSP_LOG_ERROR_MSG(Error);
OnCompleteEvent.set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(Error, MakeInvalid<NullResult>())));
```

or something like that. 

(Read the next comment before deciding on this, there's another approach)

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -78,7 +78,8 @@ namespace
             if (Result.GetAssetCollection().ParentId != ConversationId)
             {
                 CSP_LOG_ERROR_MSG("Given message doesn't exist on the conversation.");
-                throw std::exception();
+                throw csp::common::continuations::ResultException(
```
MAG-ElliotMorris, line 80: [Consider] Now that you're catching these exceptions which can carry messages, one could theoretically remove the logs here and only print to the logger in the exception catch. Might be nicer than these duplicate strings.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
```diff
@@ -803,31 +806,34 @@ void ConversationSystemInternal::GetConversationAnnotation(const csp::common::St
...
         .then(SetAnnotationAssetFromAssets(AnnotationThumbnailAsset))
         // 5. Process result
         .then(CreateAnnotationResult(ConversationAssetCollection, AnnotationAsset, AnnotationThumbnailAsset))
         .then(systems::continuations::SendResult(Callback, "Successfully retrieved annotation."))
         .then(common::continuations::InvokeIfExceptionInChain(
-            [Callback](const std::exception& /*exception*/) { Callback(MakeInvalid<multiplayer::AnnotationResult>()); }, *LogSystem));
+            *LogSystem,
+            [Callback]([[maybe_unused]] const csp::common::continuations::ExpectedExceptionBase& exception)
+            { Callback(MakeInvalid<multiplayer::AnnotationResult>()); },
+            []([[maybe_unused]] const std::exception& exception) {}));
```
MAG-ElliotMorris: [fix] You should use the overload with the default if you're going to be providing an empty unexpected error handler. That way if we put logging and such into the default, you'll get it for free. (Here and elsewhere)

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -187,6 +187,9 @@ std::function<async::task<UriResult>()> SpaceSystem::CreateAndUploadSpaceThumbna
         const auto SpaceId = Space->GetSpace().Id;
         auto ThumbnailAssetCollection = std::make_shared<AssetCollectionResult>();
 
+        // This task is designed to be a link in another continuation chain
+        // It is not responsible for handling its own callbacks (success or failure). Instead,
+        // it returns the task that allows the next task in the chain to be executed.
```
MAG-ElliotMorris, line 192: [Comment] Good.

If we had infinite time, there would be a higher evolution of this that lets us specify partial chains at the type level I'm sure, and guarentee "full" chains had exception handling on them, but the docs are at least something.

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
```diff
@@ -214,36 +187,51 @@ CSP_PUBLIC_TEST(CSPEngine, GeneralContinuationsTests, TestWhenNoExceptionThrownI
...
+    EXPECT_CALL(MockExpectedHandlerCallable, Call(::testing::_)).Times(1);
+
+    ::testing::MockFunction<void(const std::exception&)> MockUnexpectedHandlerCallable;
+    EXPECT_CALL(MockUnexpectedHandlerCallable, Call(::testing::_)).Times(0);
+
+    csp::systems::continuations::detail::testing::SpawnChainThatThrowsGeneralExceptionWithHandlerAtEnd(MockExpectedHandlerCallable.AsStdFunction(),
+        MockUnexpectedHandlerCallable.AsStdFunction(), csp::common::continuations::ResultException("", MakeInvalid<NullResult>()));
+}
+
+CSP_PUBLIC_TEST(CSPEngine, GeneralContinuationsTests, TestWhenUnexpectedExceptionThrownInContinuationChainThenHandlerCalled)
```
MAG-ElliotMorris, line 191: [comment] 👍

Library/include/CSP/Systems/ContinuationUtils.h
```diff
@@ -55,7 +55,11 @@ inline auto AssertRequestSuccessOrErrorFromResult(std::string SuccessMsg, std::s
     {
         if (Result.GetResultCode() != EResultCode::Success)
         {
-            LogHTTPErrorAndCancelContinuation<ResultT>(std::move(ErrorMsg), Result, LogLevel);
+            auto ResultCodeToUse = ResultCode.value_or(Result.GetResultCode());
+            auto HTTPResultCodeToUse = HttpResultCode.value_or(static_cast<csp::web::EResponseCodes>(Result.GetHttpResultCode()));
+            auto FailureReasonToUse = FailureReason.value_or(Result.GetFailureReason());
+            ResultT InternalResult(ResultCodeToUse, HTTPResultCodeToUse, FailureReasonToUse);
+            LogHTTPErrorAndCancelContinuation<ResultT>(std::move(ErrorMsg), InternalResult, LogLevel);
```
MAG-ElliotMorris: [Question] Does `LogHttpErrorAndCancelContinuation` take `InternalResult` by value or const ref?
if value : [consider]`std::move` it
if const ref: Close this conversation with no change.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ElliotMorris: I see in a later commit you've done this here, although I believe there's still cases in other files where there's duplication.

MAG-AdamThorn (changes requested):

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: Yeah, that is a good suggestion, I can move the bad-interface before so they are coupled together

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: Yeah I did try to capture a default argument; however, it seems that it is problematic. It was failing to resolve the interface, even with the correct lambda signature.

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ChristopherAtkinson: Thank you, I will double-check this

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: hmm, are you suggesting some in the vein of  `(LogLevel::Log, "unexpected exception handling has been defaulted.")`

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ChristopherAtkinson: Yeah there is quite a lot we can do in this regard, going forward may be a good candidate for weekly sync

Library/include/CSP/Systems/ContinuationUtils.h
MAG-ChristopherAtkinson: yeah this is by value so we can update to use std::move

Library/include/CSP/Systems/Assets/AssetCollection.h
MAG-ChristopherAtkinson: There is no reason why we couldn't, from my understanding. However, this would mean you need to cast down to the base type during construction `static_cast<std::underlying_type<csp::web::EResponseCodes>::type>(HttpResCode)` to initialise the `csp::systems::ResultBase`

Also, this is an established pattern, which would require all classes that inherit from `csp::systems::ResultBase` to be updated for consistency (not against the idea, but do not think it offers much value)

Library/include/CSP/Systems/Assets/Asset.h
MAG-ChristopherAtkinson: same as below.

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: I have added some inline documentation to class: [4691407](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/4691407bc908a607c7170ca8ef1cb41e8a4a0e3e)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: When going through and making some other changes realised this has already been updated

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: group them together here: [fbcacb2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/fbcacb2f4f4e259f5e0c49287592373f0699c98c)

Library/include/CSP/Systems/ContinuationUtils.h
MAG-ChristopherAtkinson: Yeah this should have been done from the star: [4f906c4](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/4f906c467085111562fc9ead52e754f8ff2dc421)

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ChristopherAtkinson: I have updated the continuation where needed to use the overload: [923028b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/923028b35f047172e3d37f8bea3fe652c5b711c9)

Library/include/CSP/Systems/ContinuationUtils.h
MAG-ChristopherAtkinson: Fixed here: [7c61d88](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/7c61d88fc126defa71ce02da8aaaaa1a6f5e9967)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ChristopherAtkinson: I have updated the error message here: [f29c23a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/f29c23acc59c4e03d8addf8493bc0ee7d3663809)

Tests/src/PublicAPITests/GeneralContinuationsTests.cpp
MAG-ChristopherAtkinson: Yeah, that was missed on my part when removing the validation around the callback: [781593a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/781593ab0c7f433647de655e3bdc947ee53df375)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: Apologies for accidentally batching these with namespace changes: [51b5c5b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/51b5c5b32df48921f808de837e564d1d7648d6df)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: Apologies for accidentally batching these with namespace changes: [51b5c5b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/51b5c5b32df48921f808de837e564d1d7648d6df)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: same as above

Library/src/Systems/Conversation/ConversationSystemInternal.cpp
MAG-ChristopherAtkinson: Yeah I did miss a few due to namespaces; fixed here: [392f40b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/commits/392f40b3447c9be1b48b4e67c7881315be4e33c5)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ElliotMorris: Yes, although something more direct, "been defaulted" won't mean anything to someone who dosen't know the underlying implementation. I'd suggest "Caught unexpected exception, aborting".

MAG-ElliotMorris (changes requested):
Just the logging comment remains from me.

Library/include/CSP/Common/ContinuationUtils.h
MAG-ElliotMorris: Should probably be at `Warning` level as well.

MAG-ChristopherAtkinson: @MAG-ElliotMorris Was giving this some thought, as this will call the primary `InvokeIfExceptionInChain`, which will call `LogSystem.LogMsg(LogLevel::Fatal, "Caught unexpected exception during async++ chain. Invoking callable from InvokeIfExceptionInChain");`, this would introduce duplicated logs for std::Exception in that path

https://github.com/magnopus-opensource/connected-spaces-platform/pull/762/files/df960edff204b018fabb593dfe7f2adf7e06110b#:~:text=LogSystem.LogMsg(LogLevel%3A%3AFatal%2C%20%22Caught%20unexpected%20exception%20during%20async%2B%2B%20chain.%20Invoking%20callable%20from%20InvokeIfExceptionInChain%22)%3B

Library/include/CSP/Systems/Assets/AssetCollection.h
MAG-AdamThorn: So you have two versions of the ctor:

1. CSP_NO_EXPORT AssetCollectionResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode)
2. CSP_NO_EXPORT AssetCollectionResult(csp::systems::EResultCode ResCode, csp::web::EResponseCodes HttpResCode, csp::systems::ERequestFailureReason Reason)

The second ctor takes an additional Reason argument. The second ctor, which you have added, takes the HttpResCode as a 'csp::web::EResponseCodes' and performs the underderlying type cast you have mentioned. The first ctor takes it as a uint16_t.

So are you saying that it is a pattern we follow elsewhere, to specify one as a uint16_t and the other as a EResponseCodes?

[Update] I just took a look at the codebase. I see we follow that ^ pattern in most places like Conversation.h, ApplicationSettings.h and Space.h, but not in Sequence.h. SystemsResult.h supports both for NullResult.

MAG-AdamThorn (approved):

Library/include/CSP/Common/ContinuationUtils.h
MAG-ElliotMorris: Ah right. Good-o. 👍

MAG-ElliotMorris (approved):

PR #763, opened 2025-08-08
[OF-1689] feat: OfflineRealtimeEngine

MAG-mv:
This PR contains the work for implementing the OfflineRealtimeEngine. The implementation is a lot simpler than the Online version, as we don't need to deal with replication, and there is a lot of shared functionality with the Online implementation. Because of this, most of the operations are now synchronous, meaning when Entities are created or destroyed, this happens instantaneously, and isn't deferred until ProcessPendingEntityOperations is called. This function is still in use for Entity and component updates, as this functionality is built into the SpaceEntity itself.

To get around code duplication, a RealtimeEngineUtils has been created. In the future, we likely want to find a better way of handling this, as Online is essentially an extension of the Offline version. To keep this PR simple, and to get around time restraints, this file is a compromise.

This is not in a finished state, as the Online version has not been merged, and this work contains some critical changes for this to work properly. Things that will likely need to change are:

- FetchAllEntitiesAndPopulateBuffers hasn't been implemented yet, as this is called from EnterSpace. We may want to defer the Entity creation until this point, as its currently done in the constructor.
- RealtimeEngineUtils also needs to be used in the Online version to remove duplicated code.
- ParentTest hasn't been successfully run yet, as the current SpaceEntity still relies on a SpaceEntitySystem reference, which has been removed in the online refactor.
- There is a hack in CSPSceneDescription::CreateEntities as I am reinterpret_casting the realtimeengine pointer to a spaceentity system. This is the reason we are getting crashes in the ParentTest, as the SpaceEntity is trying to use this pointer. We can remove this cast after the merge

### DIFF
--- a/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
+++ b/Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
@@ -357,7 +357,7 @@ public:
     /// @brief Adds an entity to a list of entities to be updated when ProcessPendingEntityOperations is called.
     /// From a client perspective, ProcessPendingEntityOperations is normally called via the CSPFoundation::Tick method.
     /// @param Entity SpaceEntity : A non-owning pointer to the entity to be marked.
-    virtual void MarkEntityForUpdate(csp::multiplayer::SpaceEntity* Entity)
+    virtual void QueueEntityUpdate(csp::multiplayer::SpaceEntity* Entity)
     {
         throw InvalidInterfaceUseError("Illegal use of \"abstract\" type.");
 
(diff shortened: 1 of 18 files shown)


Library/include/CSP/Common/Interfaces/IRealtimeEngine.h
```diff
@@ -357,7 +357,7 @@ class CSP_API IRealtimeEngine
     /// @brief Adds an entity to a list of entities to be updated when ProcessPendingEntityOperations is called.
     /// From a client perspective, ProcessPendingEntityOperations is normally called via the CSPFoundation::Tick method.
     /// @param Entity SpaceEntity : A non-owning pointer to the entity to be marked.
-    virtual void MarkEntityForUpdate(csp::multiplayer::SpaceEntity* Entity)
```
MAG-mv, line 360: MarkEntityForUpdate is not necessary, as it has the same functionality as QueueEntityUpdate

PR #765, opened 2025-08-14
[OF-1696] fix: cppcheck static analysis

MAG-ChristopherAtkinson:
After QA, we found that the previous changes were misconfigured and did not provide suitable coverage over the connected spaces platform source files. This PR introduces some needed fixes to address the configuration issues:

 - Migrate the files to a more suitable location 'Tools'
 - Resolved the incorrect batch command-line arguments (incorrect quoting of the path)
 - Include the additional src directory to command-line
 - Added suppression for 'syntaxError'

> [!IMPORTANT]
>  Note that the suppression is only needed due to cpp failing to parse the file and flagging an abort, which prevent the rest of the file from being parsed.


Library/docs/source/manual/documentation/external_tools.md
```diff
@@ -45,13 +45,14 @@ The primary goal of using Cppcheck is to:
...
 ```
+cd Tools\CppCheck
 run_cppcheck_static_analysis.bat
 ```
 
 Alternatively, Cppcheck can be configured to run with the following command-line arguments to define the analysis scope and output format:
 
 ```
-"C:\Program Files\Cppcheck\cppcheck.exe" --xml --xml-version=2 --output-file=cppcheck_output.xml --enable=all --suppressions-list=cppcheck_suppressions.txt -I"Library/include" Library
+"C:\Program Files\Cppcheck\cppcheck.exe" --xml --xml-version=2 --output-file=cppcheck_output.xml --enable=all --suppressions-list=Tools\CppCheck\cppcheck_suppressions.txt -I"Library/include" -I"Library/src" Library
```
MAG-ElliotMorris: [Consider] Mentioning that you need to run this from the root of the repo. 

Alternatively, delete this whole section as it's redundant, the command is contained within the batch file.

Tools/CppCheck/cppcheck_suppressions.txt
```diff
@@ -1,3 +1,4 @@
 unknownMacro
 missingInclude
 missingIncludeSystem
+syntaxError
```
MAG-ElliotMorris, line 5: [consider] Good spotting this, although it seems arbitrary that you're suppressing this but not dealing with the `toomanyconfigs` errors throughout the output. Maybe suppress/deal with that as well. Presuming the goal is to have only "real" issues in the output.

MAG-ElliotMorris (approved):

MAG-ChristopherAtkinson:
> Whilst I'm absolutely not complaining, it's also somehow covering generated headers from the wrapper generator, which surprised me as the path to wrapper-generated headers isn't specified in the args we provide via the batch script. How do they get pulled in, out of interest?

The issue is likely due to Cppcheck finding and analysing header files from directories that are not explicitly part of the intended analysis scope. The Tools directory could be included in the analysis because one of the files being analysed references a file within it. Hence, Cppcheck resolved the include.

https://github.com/magnopus-opensource/connected-spaces-platform/blob/8b79a4472ceea680159b187bdc65f87acff51761/Library/src/CSPFoundation.cpp#L18

However, that is my best informed guess, and would kind of explain why csharp and typescript are omitted as Cppcheck will not resolve their file extensions.

MAG-ChristopherAtkinson:
> 
> A very plausible explanation. Thank you for looking into it! I guess an easy way to confirm the theory would be to remove the line, run CppCheck and see if the wrapper generator header is still covered in the output.
> 

Yeah, that was 100% the case

Tools/CppCheck/cppcheck_suppressions.txt
MAG-ChristopherAtkinson: Yeah `toomanyconfigs` does not cause an abort of the current file, which was the primary reason; however, I do see the point and should include resolve for this as well

Library/docs/source/manual/documentation/external_tools.md
MAG-ChristopherAtkinson: Yeah it was mainly left as an artifact from when this did not have a support script. I think it would be good to remove this to prevent confusion

MAG-ChristopherAtkinson: I opted to remove the section as the preferred methods would be to use the script: [28c4dbd](https://github.com/magnopus-opensource/connected-spaces-platform/pull/765/commits/28c4dbd831b46dff25016a4eaf505f1276dd179d)

Tools/CppCheck/cppcheck_suppressions.txt
MAG-ChristopherAtkinson: I also included this in the suppression list: [7c95970](https://github.com/magnopus-opensource/connected-spaces-platform/pull/765/commits/7c95970e4803b8633edebff4992cb7a436eaabfb)

PR #766, opened 2025-08-15
[OF-1727] fix: remove redundant getspace call from enterspace flow

MAG-ChristopherAtkinson:
This PR removes the redundant MCS call from EnterSpace flow. 

> [!NOTE]
> The public API has not been modified, hence the change has not been marked as breaking.

### DIFF
--- a/Library/include/CSP/Systems/Spaces/SpaceSystem.h
+++ b/Library/include/CSP/Systems/Spaces/SpaceSystem.h
@@ -232,7 +232,12 @@ public:
     /// @param UserId csp::common::String : unique ID of user
     /// @param Callback SpaceResultCallback : callback when asynchronous task finishes
     CSP_ASYNC_RESULT void AddUserToSpace(const csp::common::String& SpaceId, const csp::common::String& UserId, SpaceResultCallback Callback);
-    CSP_NO_EXPORT async::task<SpaceResult> AddUserToSpace(const SpaceResult& Space, const csp::common::String& UserId);
+
+    /// @brief Adds a user to a space.
+    /// @param Result SpaceResult : result object of the space to which the user will be added
+    /// @param UserId csp::common::String : unique ID of user
+    /// @return async::task<SpaceResult> : a task containing the result of the operation
+    CSP_NO_EXPORT async::task<SpaceResult> AddUserToSpace(const SpaceResult& Result, const csp::common::String& UserId);
 
     /// @brief Creates new Site information and associates it with the Space.
     /// @param SpaceId csp::common::String : Space ID to associate the Site information with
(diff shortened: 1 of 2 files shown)


Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -232,7 +232,7 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
     /// @param UserId csp::common::String : unique ID of user
     /// @param Callback SpaceResultCallback : callback when asynchronous task finishes
     CSP_ASYNC_RESULT void AddUserToSpace(const csp::common::String& SpaceId, const csp::common::String& UserId, SpaceResultCallback Callback);
-    CSP_NO_EXPORT async::task<SpaceResult> AddUserToSpace(const csp::common::String& SpaceId, const csp::common::String& UserId);
+    CSP_NO_EXPORT async::task<SpaceResult> AddUserToSpace(const SpaceResult& Space, const csp::common::String& UserId);
```
MAG-ElliotMorris: [consider] This should probably be `Space` instead of `SpaceResult` as the type, more flexible.

MAG-SamBirley: [Fix] Since this is a separate function, feels like it warrants its own inline documentation.
It's necessary in order for intellisense to correctly render tooltip text, if nothing else.

MAG-ChristopherAtkinson: Yeah, I had some confusion around this. Space does not have a  property for 'GroupCode', and did not want to make the assumption that this was going to be the same as 'ID'

@MAG-ElliotMorris, do you know if this is the case

MAG-SamBirley (changes requested):

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-ChristopherAtkinson: Just gave it a quick test and seems that they are not equivalent, tbh expected that considering the inline documentation

MAG-ChristopherAtkinson: I have added additional inline documentation as well as renamed the parameter 'Space' to reflect that this is now a result: [b9e7056](https://github.com/magnopus-opensource/connected-spaces-platform/pull/766/commits/b9e705615849da50ac6c9920bcea904cc391eb32)

MAG-SamBirley: Yeah they're not equivalent. GroupCode is something I've never fully understood. It's a bit of an oddity from an MCS perspective as far as I can see.

MAG-SamBirley (approved):
Thanks for addressing my change request!

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-ChristopherAtkinson: Thank @MAG-SamBirley, for the clarification, in this case, I will opt to resolve this thread and leave the implementation as is

PR #767, opened 2025-08-19
[NT-0] fix: Fix OnlineRealtimeEngine::CreateAvatar hanging

MAG-mv:
- Hanging was due to passing enums by reference, breaking the wrapper generator; these are now passed by value.
- fprintf wasn't logging to the console, so we weren't seeing the error logs coming back from the wrapper generator. This has also been changed to printf
- A new wasm test has been created for CreateAvatar

### DIFF
--- a/WasmTesting/html_tests/CreateAvatar.html
+++ b/WasmTesting/html_tests/CreateAvatar.html
@@ -4,7 +4,7 @@
     <meta http-equiv="Cross-Origin-Opener-Policy" content="same-origin">
     <meta http-equiv="Cross-Origin-Embedder-Policy" content="require-corp">
     <meta charset="utf-8">
-    <title>CSP WASM Browser Test, EnterSpace</title>
+    <title>CSP WASM Browser Test, CreateAvatar</title>
     <script type="module">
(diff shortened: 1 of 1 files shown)


WasmTesting/html_tests/CreateAvatar.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Cross-Origin-Opener-Policy" content="same-origin">
+    <meta http-equiv="Cross-Origin-Embedder-Policy" content="require-corp">
+    <meta charset="utf-8">
+    <title>CSP WASM Browser Test, EnterSpace</title>
```
MAG-ChristopherAtkinson: [Fix] looks like an artifact from EnterSpace flow, which I assumed this was based on

MAG-ChristopherAtkinson (changes requested):

MAG-AdamThorn (approved):

WasmTesting/html_tests/CreateAvatar.html
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/767/commits/8cc95a6f403b53f491e1a2bce24c937804071fc6

PR #768, opened 2025-08-19
[NT-0] refac: Remove use of custom Algorithm header

MAG-AdamThorn:
We had a csp::common::Algorithm header which contained custom implementations of find, find_if and sort. Arrays now have iterators which mean the std algorithm library can be used instead.

Usage of the custom algorithm methods have been replaced by calls to std::find_if and std::sort.


MAG-ChristopherAtkinson (approved):
Great to see these changes, looks good.

PR #770, opened 2025-08-20
[OF-1738] refac: Remove legacy Analytics System implementation

MAG-AdamThorn:
OF-1738 is concerned with the implementation of a new Analytics System. Before starting work on that I am stripping out the legacy Analytics System which has never been used.

I am doing this in a separate PR to help avoid confusion later.


MAG-ElliotMorris (approved):
💣 Be gone with ya!

PR #772, opened 2025-08-22
[OF-1741] feat: client applications can express components as holdouts

MAG-ChristopherAtkinson:
## Client applications can express components as holdouts

The following PR introduces the following new properties to the existing `IVisibleComponent`:

- `IsVRVisible` - A complement to the current `IsARVisible` flag, allowing applications to express components as visible in AR but not in virtual context.

The following PR introduces the following new properties to the new `IRenderBehaviourComponent`:

- `ShowAsHoldout` - Determines whether the component should render normally (when flag is false) or as a holdout (when true) in default virtual environments.
- `ShowAsHoldoutInAR` - Determines whether the component should render normally (when flag is false) or as a holdout (when true) in AR.
- `ShowAsHoldoutInVR`- Determines whether the component should render normally (when flag is false) or as a holdout (when true) in virtual contexts.

> [!NOTE]
> The original interface has changed due to discussion on the external thread, I have updated the ticket to capture this discussion and linked the original thread. 

The new properties are also included as part of the basic component test and scripting interface test. During development, I noticed a few inconsistencies in the scripting interface, which are included in the changes:

- Not all components that inherit from `IVisibleComponent` expose all methods through the scripting interface 
- Not all components that implement the scripting interface had suitable or missing tests (Added new test and updated existing)
- Scripting interface test properties out of order (changes ordering for readability)

### DIFF
--- a/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
@@ -60,9 +60,9 @@ enum class AnimatedModelPropertyKeys
     ThirdPartyComponentRef,
     IsShadowCaster,
     MaterialOverrides,
-    IsVRVisible,
+    IsVirtualVisible,
     ShowAsHoldoutInAR,
-    ShowAsHoldoutInVR,
+    ShowAsHoldoutInVirtual,
     Num
 };
 
@@ -172,10 +172,10 @@ public:
     bool GetIsARVisible() const override;
     /// @copydoc IVisibleComponent::SetIsARVisible()
     void SetIsARVisible(bool InValue) override;
-    /// @copydoc IVisibleComponent::GetIsVRVisible()
-    bool GetIsVRVisible() const override;
-    /// @copydoc IVisibleComponent::SetIsVRVisible()
-    void SetIsVRVisible(bool InValue) override;
+    /// @copydoc IVisibleComponent::GetIsVirtualVisible()
+    bool GetIsVirtualVisible() const override;
+    /// @copydoc IVisibleComponent::SetIsVirtualVisible()
+    void SetIsVirtualVisible(bool InValue) override;
     /// @}
 
     /// \addtogroup IThirdPartyComponentRef
@@ -200,10 +200,10 @@ public:
     bool GetShowAsHoldoutInAR() const override;
     /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInAR()
     void SetShowAsHoldoutInAR(bool InValue) override;
-    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInVR()
-    bool GetShowAsHoldoutInVR() const override;
-    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInVR()
-    void SetShowAsHoldoutInVR(bool InValue) override;
+    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInVirtual()
+    bool GetShowAsHoldoutInVirtual() const override;
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInVirtual()
+    void SetShowAsHoldoutInVirtual(bool InValue) override;
     /// @}
 };
 
--- a/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
@@ -60,9 +60,9 @@ enum class AnimatedModelPropertyKeys
     ThirdPartyComponentRef,
     IsShadowCaster,
     MaterialOverrides,
-    IsVirtualVisible,
+    IsVRVisible,
     ShowAsHoldoutInAR,
-    ShowAsHoldoutInVirtual,
+    ShowAsHoldoutInVR,
     Num
 };
 
@@ -172,10 +172,10 @@ public:
     bool GetIsARVisible() const override;
     /// @copydoc IVisibleComponent::SetIsARVisible()
     void SetIsARVisible(bool InValue) override;
-    /// @copydoc IVisibleComponent::GetIsVirtualVisible()
-    bool GetIsVirtualVisible() const override;
-    /// @copydoc IVisibleComponent::SetIsVirtualVisible()
-    void SetIsVirtualVisible(bool InValue) override;
+    /// @copydoc IVisibleComponent::GetIsVRVisible()
+    bool GetIsVRVisible() const override;
+    /// @copydoc IVisibleComponent::SetIsVRVisible()
+    void SetIsVRVisible(bool InValue) override;
     /// @}
 
     /// \addtogroup IThirdPartyComponentRef
@@ -200,10 +200,10 @@ public:
     bool GetShowAsHoldoutInAR() const override;
     /// @copydoc IVisibleComponent::SetShowAsHoldoutInAR()
     void SetShowAsHoldoutInAR(bool InValue) override;
-    /// @copydoc IVisibleComponent::GetShowAsHoldoutInVirtual()
-    bool GetShowAsHoldoutInVirtual() const override;
-    /// @copydoc IVisibleComponent::SetShowAsHoldoutInVirtual()
-    void SetShowAsHoldoutInVirtual(bool InValue) override;
+    /// @copydoc IVisibleComponent::GetShowAsHoldoutInVR()
+    bool GetShowAsHoldoutInVR() const override;
+    /// @copydoc IVisibleComponent::SetShowAsHoldoutInVR()
+    void SetShowAsHoldoutInVR(bool InValue) override;
     /// @}
 };
 
--- a/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
@@ -23,10 +23,10 @@
 #include "CSP/Common/String.h"
 #include "CSP/Multiplayer/ComponentBase.h"
 #include "CSP/Multiplayer/Components/Interfaces/IExternalResourceComponent.h"
+#include "CSP/Multiplayer/Components/Interfaces/IRenderBehaviourComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IShadowCasterComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IThirdPartyComponentRef.h"
 #include "CSP/Multiplayer/Components/Interfaces/ITransformComponent.h"
-#include "CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h"
 #include "CSP/Multiplayer/SpaceTransform.h"
 
@@ -61,6 +61,7 @@ enum class AnimatedModelPropertyKeys
     IsShadowCaster,
     MaterialOverrides,
     IsVRVisible,
+    ShowAsHoldout,
     ShowAsHoldoutInAR,
     ShowAsHoldoutInVR,
     Num
@@ -77,7 +78,7 @@ class CSP_API AnimatedModelSpaceComponent : public ComponentBase,
                                             public IExternalResourceComponent,
                                             public IThirdPartyComponentRef,
                                             public IShadowCasterComponent,
-                                            public IVisibilityBehaviourComponent
+                                            public IRenderBehaviourComponent
 {
 public:
     /// @brief Constructs the animated model space component, and associates it with the specified Parent space entity.
@@ -194,15 +195,19 @@ public:
     void SetIsShadowCaster(bool Value) override;
     /// @}
 
-    /// \addtogroup IVisibilityBehaviourComponent
+    /// \addtogroup IRenderBehaviourComponent
     /// @{
-    /// @copydoc IVisibleComponent::GetShowAsHoldoutInAR()
+    /// /// @copydoc IRenderBehaviourComponent::GetShowAsHoldout()
+    bool GetShowAsHoldout() const override;
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldout()
+    void SetShowAsHoldout(bool InValue) override;
+    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInAR()
     bool GetShowAsHoldoutInAR() const override;
-    /// @copydoc IVisibleComponent::SetShowAsHoldoutInAR()
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInAR()
     void SetShowAsHoldoutInAR(bool InValue) override;
-    /// @copydoc IVisibleComponent::GetShowAsHoldoutInVR()
+    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInVR()
     bool GetShowAsHoldoutInVR() const override;
-    /// @copydoc IVisibleComponent::SetShowAsHoldoutInVR()
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInVR()
     void SetShowAsHoldoutInVR(bool InValue) override;
     /// @}
 };
--- a/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
@@ -26,6 +26,7 @@
 #include "CSP/Multiplayer/Components/Interfaces/IShadowCasterComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IThirdPartyComponentRef.h"
 #include "CSP/Multiplayer/Components/Interfaces/ITransformComponent.h"
+#include "CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h"
 #include "CSP/Multiplayer/SpaceTransform.h"
 
@@ -60,6 +61,8 @@ enum class AnimatedModelPropertyKeys
     IsShadowCaster,
     MaterialOverrides,
     IsVirtualVisible,
+    ShowAsHoldoutInAR,
+    ShowAsHoldoutInVirtual,
     Num
 };
 
@@ -73,7 +76,8 @@ class CSP_API AnimatedModelSpaceComponent : public ComponentBase,
                                             public IVisibleComponent,
                                             public IExternalResourceComponent,
                                             public IThirdPartyComponentRef,
-                                            public IShadowCasterComponent
+                                            public IShadowCasterComponent,
+                                            public IVisibilityBehaviourComponent
 {
 public:
     /// @brief Constructs the animated model space component, and associates it with the specified Parent space entity.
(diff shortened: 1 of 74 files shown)


Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
```diff
@@ -43,6 +43,14 @@ CSP_INTERFACE class CSP_API IVisibleComponent
     /// @param InValue True if the component is visible in AR mode, false otherwise.
     virtual void SetIsARVisible(bool InValue) = 0;
 
+    /// @brief Checks if the component is visible when in virtual modalities.
+    /// @return True if the component is visible when in virtual modalities, false otherwise.
```
MAG-ElliotMorris: [consider] "Virtual modalities" isn't a term people are going to know without context. Can you say this more directly? 

The ticket says "allowing applications to express components as visible in AR but not in virtual context.", is that not more akin to what you want to say? (Although ... we went through this with Sam and he asserted none of these flags were exclusive ... I think we at least need to explain what a virtual context is.)

Note: I'm not saying "virtual modalities" is incorrect, it makes sense when you have all the other context, it's just that I don't think most people will follow that phrasing without at least a little further explanation.

Tests/src/PublicAPITests/ComponentTests/ComponentTests.cpp
```diff
@@ -164,6 +164,35 @@ CSP_PUBLIC_TEST(CSPEngine, ComponentTests, ARVisibleTest)
...
+    csp::common::LogSystem* LogSystem = csp::systems::SystemsManager::Get().GetLogSystem();
+
+    std::vector<ComponentBase*> Components { new AnimatedModelSpaceComponent(LogSystem, MySpaceEntity),
+        new ButtonSpaceComponent(LogSystem, MySpaceEntity), new ImageSpaceComponent(LogSystem, MySpaceEntity),
+        new LightSpaceComponent(LogSystem, MySpaceEntity), new StaticModelSpaceComponent(LogSystem, MySpaceEntity),
+        new VideoPlayerSpaceComponent(LogSystem, MySpaceEntity) };
+
+    for (auto Component : Components)
+    {
+        auto* VisibleComponent = dynamic_cast<IVisibleComponent*>(Component);
```
MAG-ElliotMorris, line 180: [nit] No need to `dynamic_cast` if you know the type is valid and you're not checking for null, can just `static_cast`

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
```diff
@@ -0,0 +1,50 @@
...
+/// @brief Holdout visibility control for components.
+
+#pragma once
+
+#include "CSP/CSPCommon.h"
+
+namespace csp::multiplayer
+{
+
+/// @brief Controls the holdout visibility behaviour of the component when in default mode, AR mode or Virtual mode.
```
MAG-ElliotMorris: [Consider] I don't think we explain what the term holdout means anywhere? We shouldn't be using concepts that we don't explain. It's not a common enough term for that imo.

If we're going to insist on using the term in the API, I think a sentence or two here in the main description of the interface describing the concept would behoove us.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
```diff
@@ -0,0 +1,50 @@
...
+CSP_INTERFACE class CSP_API IVisibilityBehaviourComponent
```
MAG-ElliotMorris: [Question] I don't understand why we have the bools on `IVisibleComponent` but also these interfaces. Naively, it seems like these are the same concept so I don't get why there's these separate interfaces. Can someone explain it to me? I'm confused.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
```diff
@@ -23,7 +23,7 @@
 namespace csp::multiplayer
 {
 
-/// @brief Controls the visibility of the component when in default mode or in AR mode.
+/// @brief Controls the visibility of the component when in default mode, AR mode or Virtual mode.
```
MAG-ElliotMorris, line 26: [Question] Are these exclusive modes? Is default and virtual not the same thing more or less? Another thing we could do with explaining fully rather than just alluding to.

MAG-ElliotMorris (commented):
Not an approval just because I don't understand the interface expression of the behaviour type, I don't think any of the implementation is wrong, just need it explained to me.

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
```diff
@@ -59,6 +60,9 @@ enum class AnimatedModelPropertyKeys
     ThirdPartyComponentRef,
     IsShadowCaster,
     MaterialOverrides,
+    IsVirtualVisible,
```
MAG-mv, line 63: [Consider] It wouldbe nice if these names following a similar convention. either-
- ShowAsHoldoutInVirtual+ ShowAsHoldoutInAugmented
- ShowAsHoldoutInVR+ ShowAsHoldoutInAR

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-mv: I was also going to ask why these aren't added to IVisible

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
```diff
@@ -0,0 +1,50 @@
...
+{
+public:
+    /// @brief Checks if the component is shown as holdout when in AR mode.
+    /// @return True if the component is shown as holdout when in AR mode, false otherwise.
+    virtual bool GetShowAsHoldoutInAR() const = 0;
```
MAG-mv: [Fix] I had to look at the ticket to understand what holdout means. Can we add this in the docs somewhere?

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-mv: I expect default would be using on a desktop or phone, without mixed reality, while VR is in a headset, and AR is an augmented compatible device

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-mv: Thinking about this more, I think a seperate compoennt is ok. Most clients will only want to deal with basic visibility. This is more about configuring visibility for diferent platforms, so I unerstand keeping it seperated.

MAG-mv (changes requested):

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: I can expand and modify the documentation to improve clarity, as I agree there is definitely missing context and could be made clearer.

Tests/src/PublicAPITests/ComponentTests/ComponentTests.cpp
MAG-ChristopherAtkinson: Yeah this was done just due to cohesion, as we specify different components to test we know the type. If this were all components then we would need to `dynamic_cast`

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: Yeah that would help with the context issues mentioned previously, can add some additional documentation

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: Yeah there was prior discussion ahead of the ticket, from my understanding, `IVisibleComponent` is concerned with visibility in the scene, where as `IVisibilityBehaviourComponent` is concerned with how these are rendered. hence not everything that is `IVisibleComponent` is also `IVisibilityBehaviourComponent`

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: Yeah i think the wording here can cause some confusion as this is equivalent to Desktop, AR, and VR if I am understanding correctly

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
MAG-ChristopherAtkinson: Yeah I did give this some thought, my preference would be:

- ShowAsHoldoutInAR -> IsARHoldout
- ShowAsHoldoutInVirtual -> IsVirtualHoldout

This would give it consistency with the naming convention from `IVisibleComponent`

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: Yeah this was raise earlier and will add some additional documentation to resolve this

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: Actually double-checked with docs, and it seems Virtual is any form of virtual space not explicitly VR so this would be the following:

- Everywhere 
- AR
- Virtual (VR, and Desktop)

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
MAG-mv: I prefer those names, but my point still stands. Wouldn't it be clearer to be either:

- IsARHoldout + IsVRHoldout or
- IsAugmentedHoldout +  IsVirtualHoldout

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-mv: So:

Everywhere = AR + Virtual?

If so, we need something specifically for VR?

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
MAG-ChristopherAtkinson: Virtual is not just VR, it any virtual representation, where as AR is Augmented Reality only

MAG-mv: What do you mean by virtual representation?

MAG-ChristopherAtkinson: 3D scene

magnopus-swifty: Agree we should have consistent language and not mix abbreviations and words. I would be OK with using VR in this context - we're still viewing a _virtual environment_ even if we're not literally using a VR headset.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: I added some additional documentation to the class docs to provide some context on holdouts: [1c429f2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/1c429f2a0e3f601eb931b6f1ff48745d2fabf40f)

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: I added some additional documentation to the class docs to provide some context on holdouts: [1c429f2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/1c429f2a0e3f601eb931b6f1ff48745d2fabf40f)

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
MAG-ChristopherAtkinson: I have updated the naming convention to use VR in place of Virtual based on this discussion: [5204a8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/5204a8e0e3027f91f1aeae4d09eddf17609d175a)

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: Yeah, this has been updated to reflect Virtual as VR in this context, based on discussion below. Here are the relevant changes: [1c429f2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/1c429f2a0e3f601eb931b6f1ff48745d2fabf40f)

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ElliotMorris: I wonder about `IRenderingBehaviour` instead of conflating this visibility concept. If visibility is "Do we render anything at all", then this component is "Assuming we're rendering, how do we render?". Seems evident to me that it's a different concept, and should have a different word. The current way it's setup doesn't feel like a joyful interface to use, feels a bit design by committee tbh.

The comment says 
> /// @brief Controls the holdout visibility behaviour

Which means this component is only for holdouts, so why isn't it called `IHoldoutSomething` ?

I understand this was designed by a feature group, so I won't block the PR on this business, but i'm a tad unhappy with it.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: I have updated the naming convention to use VR in place of Virtual based on discussion below: [5204a8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/5204a8e0e3027f91f1aeae4d09eddf17609d175a)

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-ChristopherAtkinson: Yeah I had this same discussion with @MAG-mv externally as well, since this really is related to the depth buffer / render behaviour, it should be something more akin to `IDepthBufferRenderingBehaviour`, but the current naming is drawn from the spec provided

MAG-ChristopherAtkinson: From external discussion: we ended going with `IRenderBehaviour` as this would capture the meaning better from CSP perspective.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibleComponent.h
MAG-ChristopherAtkinson: From external discussion: due to points around the use of Virtual, we have gone for a more by modality approach [default, AR, VR]. This can be expanded in the future to adopt more, such as XR

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-SamBirley: Happy with this naming choice, thanks! Closing this one out.

Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
MAG-SamBirley: Looks good. Closing this one out.

Library/include/CSP/Multiplayer/Components/Interfaces/IVisibilityBehaviourComponent.h
MAG-SamBirley: Changes look good.

Library/include/CSP/Multiplayer/Components/Interfaces/IRenderBehaviourComponent.h
```diff
@@ -0,0 +1,74 @@
...
+public:
+    /// @defgroup HoldoutConcept Holdouts and Data Access
+    /// A holdout is an object that is rendered as a mask, cutting it out from the final image.
+    /// It still participates in depth testing, allowing objects to move behind it, but does not contribute pixels to the final image.
+
+    /// @brief Checks if the component is shown as holdout when in default mode.
+    /// @return True if the component is shown as holdout when in default mode, false otherwise.
+    ///
+    /// @see HoldoutConcept
+    virtual bool GetShowAsHoldout() const = 0;
```
MAG-SamBirley: Why does this 'default' holdout state do? What purpose does it serve? I get the desire to be symmetric with visibility flags, but it wasn't asked for in the ticket and it's not obvious to me what it affords since AR and VR are already covered by the flags below.

MAG-mv: I believe this relates to: https://github.com/magnopus-opensource/connected-spaces-platform/pull/772#discussion_r2293185367

MAG-ChristopherAtkinson: This is distinguished between VR and virtual environment, which is confusing, see @MAG-mv link above

MAG-ChristopherAtkinson: I should clarify that VR in this term means virtual reality, as in HMD, so desktop is not a covered case. Resulting in a default being for standard virtual environments like desktops

MAG-SamBirley: Ok, I'd like for us to remove the default state, and consider HMD VR and desktop VR to both be covered by the 'VR' state instead.

I appreciate that the term 'VR' is overloaded for this purpose, however the decision has been made on Slack over [here](https://magnopus.slack.com/archives/C098Y6QDW8N/p1756314511382979?thread_ts=1755692928.165809&cid=C098Y6QDW8N).

MAG-ChristopherAtkinson: Great will update that now

MAG-ChristopherAtkinson: The default has now been removed here: [82b934d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/772/commits/82b934d3bc379ef04ec8d0b24934b258f34f0463)

I updated the comments slightly to include Desktop for VR, just to add some clarity however minor that may be

MAG-SamBirley: Nice one. Looks good, thanks!

MAG-AdrianMeredith:
Who's the feature owner for this?  may I ask that any and all appropriate changes to the specs are updated here https://github.com/magnopus/oko-tech-docs

MAG-ChristopherAtkinson:
> Who's the feature owner for this? may I ask that any and all appropriate changes to the specs are updated here https://github.com/magnopus/oko-tech-docs

Yeah, you are quite right. Now that the suggested changes have been approved, I will submit an NT-0 change to capture this

Library/include/CSP/Multiplayer/Components/Interfaces/IRenderBehaviourComponent.h
MAG-AdamThorn: @MAG-SamBirley sorry I am late to the party. But to confirm this means that we are combining the option for `Holdouts in VR` with `Holdouts in default` (non-AR/VR/XR))? In the link above David posted his preference but stated that he was up for suggestions which were made below. I would like to finish that conversation before we bake it into CSP.

MAG-ChristopherAtkinson: Unresolved conversation while there is an open question

PR #774, opened 2025-08-26
Start a suite of unit tests for the wrapper generator

magnopus-swifty:
This change introduces a test suite for the Wrapper Generator.

A small change is required to the Parser: a log file is passed to the constructor rather than being created inside the top-level `parse` function. We could switch to the standard `logger` package instead but I took this approach to keep the change minimal.

To support this change I've extracted a couple of methods from the top-level `parse` function. This wasn't strictly necessary for this PR but it makes the code slightly easier to follow.

The actual tests just exercise some of the parsing specifically for enums to demonstrate the idea.

You can run the tests with `py -m unittest` and test discovery will find them correctly. I've tested with `coverage` and measured a whopping 8% coverage on Parser.py with these tests!

### DIFF
--- a/Library/include/CSP/Common/ContinuationUtils.h
+++ b/Library/include/CSP/Common/ContinuationUtils.h
@@ -23,9 +23,80 @@ CSP_START_IGNORE
 #include "CSP/Common/CSPAsyncScheduler.h"
 #include "CSP/Common/Interfaces/InvalidInterfaceUserError.h"
 #include "CSP/Common/Systems/Log/LogSystem.h"
+#include "CSP/Systems/SystemsResult.h"
+#include "CSP/Systems/WebService.h"
 
 namespace csp::common::continuations
 {
+
+enum class ExceptionType
+{
+    Result,
+    Multiplayer
+};
+
+/**
+ * @brief Serves as the base class for all expected, or business-logic, exceptions.
+ * @details This class provides a common interface for exceptions that are
+ * part of the normal program flow, distinguishing them from unexpected system-level
+ * errors like out-of-memory or file not found.
+ */
+class ExpectedExceptionBase : public std::runtime_error
+{
+public:
+    ExpectedExceptionBase(const std::string& message)
+        : std::runtime_error(message)
+    {
+    }
+
+    virtual const ExceptionType GetExceptionType() const = 0;
+};
+
+/**
+ * @brief An exception class for API request results.
+ * @details This template class captures the result object alongside the
+ * exception. It's ideal for handling errors from HTTP requests where specific
+ * result codes, HTTP status, and failure reasons need to be preserved.
+ */
+class ResultException : public ExpectedExceptionBase
+{
+public:
+    ResultException(const std::string& Message, const csp::systems::ResultBase& Result)
+        : ExpectedExceptionBase(Message)
+        , Result(Result)
+    {
+    }
+
+    const ExceptionType GetExceptionType() const override { return ExceptionType::Result; }
+
+    const csp::systems::ResultBase& GetResult() const { return Result; }
+
+private:
+    const csp::systems::ResultBase& Result;
+};
+
+/**
+ * @brief An exception class for Multiplayer Error code.
+ * @details This template class captures the Multiplayer Error code object alongside the
+ * exception. It's ideal for handling errors from Multiplayer where specific
+ * error codes need to be preserved.
+ */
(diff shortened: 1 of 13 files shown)


magnopus-swifty:
Demonstrating the value of automated testing, in writing the tests I identified that we were just throwing away docstrings for enum values, so I changed the code to preserve those values (#83eb1d1). However running the integration tests highlighted that the Typescript templates don't account correctly for this.

Good reminder that the best way to be confident of changes to the wrapper generator is still to diff the output prior to your changes with the output afterwards and visually inspect the changes.

MAG-ChristopherAtkinson (approved):
The PR looks good! It's awesome that we're finally getting some tests in there. It's a huge step forward for the project, and the initial changes to the parser to make it more testable are a great move.

However, I've got a couple of questions and some ideas for what we should do next.

First off, the logging. I wonder if we should consider switching to Python's actual logging package down the line. It would be a bigger change, for sure, since we'd have to overhaul our current setup. But it would give us so much more control, like being able to configure severity levels. We could set it to only show warnings or errors in production, but get a flood of debug info when we're working on something. It's a bit of work, but it would make debugging way more flexible.

And second, that 8% test coverage. It's great to have a starting point, but let's be real—8% doesn't give me much confidence. I'm assuming the plan is to keep building on this and expand our test suite in the future, right? We need to get that number way up to feel good about the parser's reliability.

Overall, it's a solid start. Let's keep the conversation going on how we can improve these areas. Not going to block this as the changes are still a worth while addition.

magnopus-swifty:
> I wonder if we should consider switching to Python's actual logging package down the line.

I thought about doing this but I wanted to focus this PR on getting a test suite started. Once we have a working test suite (running as part of the CI process) I can feel more confident about making more bold changes to the codebase.

> I'm assuming the plan is to keep building on this and expand our test suite in the future, right?

Yes. Following the philosophy of "the first test is the hardest one to write", it should be much easier for anyone to follow my lead and write additional tests around edge cases or regressions they find. I didn't want to submit a PR with 1000+ test cases before there's an agreement that my approach is OK - choice of test library, code style etc.

MAG-ChristopherAtkinson:
> > I wonder if we should consider switching to Python's actual logging package down the line.
> 
> I thought about doing this but I wanted to focus this PR on getting a test suite started. Once we have a working test suite (running as part of the CI process) I can feel more confident about making more bold changes to the codebase.
> 
> > I'm assuming the plan is to keep building on this and expand our test suite in the future, right?
> 
> Yes. Following the philosophy of "the first test is the hardest one to write", it should be much easier for anyone to follow my lead and write additional tests around edge cases or regressions they find. I didn't want to submit a PR with 1000+ test cases before there's an agreement that my approach is OK - choice of test library, code style etc.

Great, that was the impression I got, just wanted to make sure I wasn't making assumptions

MAG-ChristopherAtkinson:
>  an agreement that my approach is OK - choice of test library, code style etc.

And personally, I tend to always go with Pytest but both are powerful. 

I would also like to add a linting (PEP8 compliance) process to the CI in the future but yeah, completely out of scope here

Tools/WrapperGenerator/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+* [C](./CWrapperGenerator.py)
+* [Typescript/Javascript](./TypeScriptWrapperGenerator.py) (with [WebAssembly] interop)
+* [C#](./CSharpWrapperGenerator.py)
+
+The Typescript and C# wrappers rely on the C wrapper, which is always generated by default.
+
+Each wrapper generator is responsible for taking the generic metadata types output from the Parser, translating them into the appropriate concepts and naming conventions for the target language and invoking code generation.
+
+Code generation is implemented using the [Mustache] templating language, through the [Chevron] implementation in Python.
```
MAG-AdamThorn, line 36: [Comment] Typescript also uses Jinja2, though not sure if worth calling out as only used for global_functions and helpers.

MAG-AdamThorn:
Thank you for your continued Wrapper Gen efforts @magnopus-swifty !

Tools/WrapperGenerator/README.md
magnopus-swifty: I'll leave this comment as-is for the moment. I started writing more exhaustive docs on how each wrapper generator works but I felt like it didn't really belong in this PR. I'll make a follow-up PR that expands on this README.

PR #775, opened 2025-08-27
test: initial tests for WordReader

magnopus-swifty:
Initial set of tests for `WordReader`.

With these changes we're at 95%+ test coverage for the `WordReader` class, although I think that's misleading because that last 5% is where the obscure edge cases live.

### DIFF
--- a/Tools/WrapperGenerator/tests/test_word_reader.py
+++ b/Tools/WrapperGenerator/tests/test_word_reader.py
@@ -1,0 +1,101 @@
+import unittest
+from word_reader import WordReader
+
+class TestWordReader(unittest.TestCase):
+    def test_next_word_basic(self):
+        w = WordReader("Hello, world!")
+        self.assertEqual(w.next_word(), "Hello")
+        self.assertEqual(w.next_word(), ",")
+        self.assertEqual(w.next_word(), "world!")
+        self.assertIsNone(w.next_word())
+
+    def test_skip_whitespace_and_peek_char(self):
+        w = WordReader("   \t\n  Meow")
+        w.skip_whitespace()
+        self.assertEqual(w.peek_char(), "M")
+        self.assertEqual(w.next_word(), "Meow")
+        self.assertIsNone(w.next_word())
+
+    def test_next_word_with_custom_delimiters(self):
+        w = WordReader("a|b|c")
+        self.assertEqual(w.next_word(delimiters={"|"}), "a")
+        self.assertEqual(w.next_word(delimiters={"|"}), "|")
+        self.assertEqual(w.next_word(delimiters={"|"}), "b")
+        self.assertEqual(w.next_word(delimiters={"|"}), "|")
+        self.assertEqual(w.next_word(delimiters={"|"}), "c")
+        self.assertIsNone(w.next_word(delimiters={"|"}))
+
+    def test_next_word_return_empty(self):
+        w = WordReader("a,,b")
+        self.assertEqual(w.next_word(delimiters={","}, return_empty=True), "a")
+        self.assertEqual(w.next_word(delimiters={","}, return_empty=True), ",")
+        self.assertEqual(w.next_word(delimiters={","}, return_empty=True), ",")
+        self.assertEqual(w.next_word(delimiters={","}, return_empty=True), "b")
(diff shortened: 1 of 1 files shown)


Tools/WrapperGenerator/tests/test_word_reader.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_peek_char_and_skip_char(self):
+        w = WordReader("abc")
+        self.assertEqual(w.peek_char(), "a")
+        w.skip_char()
+        self.assertEqual(w.peek_char(), "b")
+        w.skip_char(2)
+        self.assertIsNone(w.peek_char())
+
+    def test_skip(self):
+        w = WordReader("a\nb\nc")
```
MAG-ChristopherAtkinson, line 45: [Question] same as above ^

Tools/WrapperGenerator/tests/test_word_reader.py
MAG-ChristopherAtkinson, line 1: [Question] I wonder if there is any value in testing against a utf-16 characters, not sure off the top of my head how Python or our implementation handles string encoding

Tools/WrapperGenerator/tests/test_word_reader.py
```diff
@@ -0,0 +1,101 @@
...
+        w.skip(2)  # skips 'a' and '\n'
+        self.assertEqual(w.current_line, 2)
+        self.assertEqual(w.peek_char(), "b")
+        w.skip(2)  # skips 'b' and '\n'
+        self.assertEqual(w.current_line, 3)
+        self.assertEqual(w.peek_char(), "c")
+
+    def test_skip_line(self):
+        w = WordReader("foo\nbar\nbaz")
```
MAG-ChristopherAtkinson, line 54: [Question] same as above ^

Tools/WrapperGenerator/tests/test_word_reader.py
```diff
@@ -0,0 +1,101 @@
...
+        w.skip(4)
+        self.assertEqual(w.find_next_of({"e", "f"}), "e")
+
+    def test_find_next_of_raises_index_error(self):
+        w = WordReader("abc")
+        with self.assertRaises(IndexError):
+            w.find_next_of({"z"})
+
+    def test_next_word_handles_only_whitespace(self):
+        w = WordReader("   \t\n  ")
```
MAG-ChristopherAtkinson, line 77: [Question] same as above ^

Tools/WrapperGenerator/tests/test_word_reader.py
```diff
@@ -0,0 +1,101 @@
...
+class TestWordReader(unittest.TestCase):
+    def test_next_word_basic(self):
+        w = WordReader("Hello, world!")
+        self.assertEqual(w.next_word(), "Hello")
+        self.assertEqual(w.next_word(), ",")
+        self.assertEqual(w.next_word(), "world!")
+        self.assertIsNone(w.next_word())
+
+    def test_skip_whitespace_and_peek_char(self):
+        w = WordReader("   \t\n  Meow")
```
MAG-ChristopherAtkinson, line 13: [Question] Should we be including a carriage return here as well? \r\n

MAG-ElliotMorris (approved):
Looks good, although not too fussed about WordReader. It's always seemed solid, and I doubt it survives/needs changing into whatever the next implementation of the parser is. (tbh, I'm hopeful the python language won't survive, dynamically typed languages are bad at this sort of thing imo)

Tools/WrapperGenerator/tests/test_word_reader.py
magnopus-swifty: I think we'd handle that downstream. Assuming you open the file with the correct encoding then we just pass the tokens on and it's up to the parser to decide if it cares (right now there's not much validation).

Tools/WrapperGenerator/tests/test_word_reader.py
magnopus-swifty: Assuming you open the file as text (`open("...", "t")`) then it'll convert line endings so I think we can just hand off responsibility that way.

MAG-ChristopherAtkinson (approved):

PR #776, opened 2025-08-27
Tests for `Parser.__parse_type()`

magnopus-swifty:
Added fairly thorough tests for the `__parse_type()` function. These tests cover _almost_ every line of the function.

These tests build on each other, starting from checking simple types (e.g. `test_parse_primitive_type`), then using the parser to construct tests for templated types (`test_parse_template_type`) to finally checking templated `std::function<>` types with templated arguments.

`test_parse_pointer_to_pointer_to_pointer_type_fails` demonstrates that there are certain edge case types (`int***`) that the parser handles incorrectly.

### DIFF
--- a/Tools/WrapperGenerator/tests/test_parser.py
+++ b/Tools/WrapperGenerator/tests/test_parser.py
@@ -262,14 +262,10 @@ class TestParserParseType(unittest.TestCase):
 
     def test_parse_optional_type(self):
         """ Test the special parsing of optional types (using csp::common::Optional). """
-        result = self.__parse_type("csp::common::Optional<int>")
+        result = self.__parse_type("csp::common::Optional<csp::common::String>")
 
-        expected = MetadataTypes.TypeMetadata("", "int")
-        expected.is_const = False
-        expected.is_primitive = True
+        expected = self.__parse_type("csp::common::String")
         expected.is_optional = True
-        expected.template_name = "int"
-        expected.template_arguments = []
 
         self.assertDictEqual(result.__dict__, expected.__dict__)
 
--- a/Tools/WrapperGenerator/tests/test_parser.py
+++ b/Tools/WrapperGenerator/tests/test_parser.py
@@ -253,3 +253,17 @@ class TestParserParseType(unittest.TestCase):
         expected.template_arguments = []
 
         self.assertDictEqual(result.__dict__, expected.__dict__)
+
+    def test_parse_function_type(self):
+        """ Test parsing a function type. """
+        result = self.__parse_type("std::function<void()>")
+
+        function_metadata = MetadataTypes.FunctionMetadata(None, 0, 0, None, None, None, False, False, [])
+
+        expected = MetadataTypes.TypeMetadata("std", "function")
+        expected.is_function_signature = True
+        expected.function_signature = function_metadata
+        expected.template_arguments = []
+        expected.template_name = "function"
+
+        self.assertDictEqual(result.__dict__, expected.__dict__)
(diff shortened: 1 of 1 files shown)


Tools/WrapperGenerator/tests/test_parser.py
```diff
@@ -78,3 +78,243 @@ def test_parse_enum_field_values(self):
...
+        expected = MetadataTypes.TypeMetadata("", "int")
+        expected.is_reference = True
+        expected.is_pointer_or_reference = True
+        expected.is_primitive = True
+        expected.template_name = "int"
+        expected.template_arguments = []
+
+        self.assertDictEqual(result.__dict__, expected.__dict__)
+
+    def test_parse_const_pointer_type(self):
```
MAG-mv, line 174: [Consider] What about const int&

Tools/WrapperGenerator/tests/test_parser.py
```diff
@@ -78,3 +78,243 @@ def test_parse_enum_field_values(self):
...
+        """ Test parsing a const pointer type. """
+        result = self.__parse_type("const int*")
```
MAG-mv, line 176: [Question] I wonder if we support int const *?

Tools/WrapperGenerator/tests/test_parser.py
```diff
@@ -78,3 +78,243 @@ def test_parse_enum_field_values(self):
...
+            MetadataTypes.TemplateArgumentMetadata(
+                type=int_type,
+                is_last=True
+            )
+        ]
+
+        self.assertDictEqual(result.__dict__, expected.__dict__)
+
+    def test_parse_multi_template_type(self):
+        """ Test parsing a template type with two arguments to the template. """
```
MAG-mv, line 226: [Question] Is multiple template args actually fully supported? Or do we have a special case for maps?

MAG-mv: Asking because we have a lot of special cases for certain types, although this may be in the language generation step

Tools/WrapperGenerator/tests/test_parser.py
```diff
@@ -78,3 +78,243 @@ def test_parse_enum_field_values(self):
...
+        expected = MetadataTypes.TypeMetadata("std", "function")
+        expected.is_function_signature = True
+        expected.function_signature = function_metadata
+        expected.template_arguments = []
+        expected.template_name = "function"
+
+        self.assertDictEqual(result.__dict__, expected.__dict__)
+
+    def test_parse_function_with_return_type(self):
+        """ Test parsing a function type with a return type. """
```
MAG-mv, line 327: [Consider] What about CSP macros? CSP_ASYNC_RESULT, CSP_EVENT, CSP_START_IGNORE etc

MAG-mv:
[Comment] It looks like our parser seems well supported. A lot of our issues and limitations seem to come from the language generation step. Which makes sense, I guess.

Tools/WrapperGenerator/tests/test_parser.py
magnopus-swifty: I think the wrapper generator does (in theory) support it, at least for C#. The other tests (#779) were partly motivated by needing a place to hand-write tests that verify this.

Tools/WrapperGenerator/tests/test_parser.py
magnopus-swifty: See 1b4138c

Tools/WrapperGenerator/tests/test_parser.py
magnopus-swifty: Not right now. See 1b4138c.

Tools/WrapperGenerator/tests/test_parser.py
magnopus-swifty: I think I'd rather do a follow-up PR for testing `__parse_function` where I test function types specifically. I'll leave these tests here just to demonstrate how function metadata ends up in a type declaration but expect a future PR containing a `TestParserParseFunction` fixture.

MAG-AdamThorn (approved):

PR #779, opened 2025-08-28
[NT-0] test: Hand-written integration test for the Wrapper Generator

magnopus-swifty:
Implements a trivial hand-rolled integration test (in `UnitTesting`, ironically) for the Wrapper Generator.

* Move all generated files into `generated/`
* Hand-wrote a header/source file in C++
* Wrote corresponding tests for it in C#
* Fixed ignore files

This tweaks the file layout to make writing tests slightly less inscrutable.

### DIFF
--- a/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/AnimatedModelSpaceComponent.h
@@ -23,6 +23,7 @@
 #include "CSP/Common/String.h"
 #include "CSP/Multiplayer/ComponentBase.h"
 #include "CSP/Multiplayer/Components/Interfaces/IExternalResourceComponent.h"
+#include "CSP/Multiplayer/Components/Interfaces/IRenderBehaviourComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IShadowCasterComponent.h"
 #include "CSP/Multiplayer/Components/Interfaces/IThirdPartyComponentRef.h"
 #include "CSP/Multiplayer/Components/Interfaces/ITransformComponent.h"
@@ -59,6 +60,9 @@ enum class AnimatedModelPropertyKeys
     ThirdPartyComponentRef,
     IsShadowCaster,
     MaterialOverrides,
+    IsVirtualVisible,
+    ShowAsHoldoutInAR,
+    ShowAsHoldoutInVirtual,
     Num
 };
 
@@ -72,7 +76,8 @@ class CSP_API AnimatedModelSpaceComponent : public ComponentBase,
                                             public IVisibleComponent,
                                             public IExternalResourceComponent,
                                             public IThirdPartyComponentRef,
-                                            public IShadowCasterComponent
+                                            public IShadowCasterComponent,
+                                            public IRenderBehaviourComponent
 {
 public:
     /// @brief Constructs the animated model space component, and associates it with the specified Parent space entity.
@@ -167,6 +172,10 @@ public:
     bool GetIsARVisible() const override;
     /// @copydoc IVisibleComponent::SetIsARVisible()
     void SetIsARVisible(bool InValue) override;
+    /// @copydoc IVisibleComponent::GetIsVirtualVisible()
+    bool GetIsVirtualVisible() const override;
+    /// @copydoc IVisibleComponent::SetIsVirtualVisible()
+    void SetIsVirtualVisible(bool InValue) override;
     /// @}
 
     /// \addtogroup IThirdPartyComponentRef
@@ -183,6 +192,18 @@ public:
     bool GetIsShadowCaster() const override;
     /// @copydoc IShadowCasterComponent::SetIsShadowCaster()
     void SetIsShadowCaster(bool Value) override;
+    /// @}
+
+    /// \addtogroup IRenderBehaviourComponent
+    /// @{
+    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInAR()
+    bool GetShowAsHoldoutInAR() const override;
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInAR()
+    void SetShowAsHoldoutInAR(bool InValue) override;
+    /// @copydoc IRenderBehaviourComponent::GetShowAsHoldoutInVirtual()
+    bool GetShowAsHoldoutInVirtual() const override;
+    /// @copydoc IRenderBehaviourComponent::SetShowAsHoldoutInVirtual()
+    void SetShowAsHoldoutInVirtual(bool InValue) override;
     /// @}
 };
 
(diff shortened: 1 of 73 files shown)


UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
```
MAG-ElliotMorris: [Fix] This is a unity specific metadata file? Seems irrelevant to a c# test. Unless i'm mistaken, remove, and consider adding *.meta to gitignore for this directory.

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Csp.Tests
+{
+    public class ClassTests
+    {
+        [Test]
+        public void CreateAndDestroySimpleClass()
+        {
+            var simpleClass = new SimpleClass();
+
+            simpleClass.Dispose();
```
MAG-ElliotMorris: [Consider] I'd prefer if this test could perform an assertion, like checking if the underlying pointer has actually been deleted after `Dispose`. (and possibly a manual `GC` invocation?) Worried about false confidence.

UnitTesting/include/classes.h
```diff
@@ -0,0 +1,28 @@
+/*
```
MAG-ElliotMorris, line 1: [Consider] Do you think these source files/entire project would be better in a `UnitTesting/include/wrappergen` or something like that. (Or to be honest, I wouldn't mind seeing a `WrapperGenTests` folder at the top level, similar to how we have `WasmTesting`)

Just a bit concerned about getting confounded, there's old stuff and new stuff and it's hard to know what's what's for what in this directory.

MAG-ElliotMorris (changes requested):

UnitTesting/include/classes.h
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "CSP/CSPCommon.h"
+
+namespace csp::Tests
+{
+class CSP_API SimpleClass
```
MAG-ChristopherAtkinson, line 22: [Consider] whether we want to include base classes that utilizes more than our CSP_API define, granted we do have a specific define for concepts such as interfaces 'CSP_INTERFACE' and visibility around functions 'CSP_NO_EXPORT' just to name a few

We can then expand the current test cases to have better coverage over our expected behaviour from the defines

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs.meta
MAG-ChristopherAtkinson, line 1: [Fix] looks like an accidental metadata artifact, should be removed if not needed

MAG-ChristopherAtkinson (changes requested):

MAG-ChristopherAtkinson:
Also, as a side note, can we update the name of the PR and adopt conventional commits: [NT-0] test: Hand-written integration test for the Wrapper Generator

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
magnopus-swifty: I definitely plan on writing more tests that exercise the interop code more so this test doesn't stand on its own. I have mixed feelings about the user having access to the underlying pointer _to be able to_ do assertions on it -- but for now since it's there, I will.

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs.meta
magnopus-swifty: This whole directory (`Assets/...`) is a Unity project so conventionally you check in .meta files.

If we're treating all the Unity artefacts as side-effects then we need clearer documentation on expectations for how to maintain the files in this project. I can make an educated guess based on this feedback.

UnitTesting/include/classes.h
magnopus-swifty: Yes, as with my other recent PRs this is just to demonstrate "the first test" to get feedback on the approach before I write the actual tests that this motivated.

UnitTesting/include/classes.h
magnopus-swifty: Everything in `UnitTesting` concerns integration tests for the wrapper generator so the whole thing is poorly named. (See https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/UnitTesting/README.md)

I think it would be helpful to rename UnitTesting to WrapperGenTests or maybe InteropTests since this is where we can write tests that run in the target environment.

It'd also be nice to incorporate this stuff into the main build script (the generation and native code building at least). That feels like a task for if/when CSP adopts CMake or something that can manage the complicated build process.

UnitTesting/include/classes.h
MAG-ChristopherAtkinson: Great, happy to resolve

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs.meta
MAG-ChristopherAtkinson: my first thought is that the generator is not concerned with Unity directly, but rather csharp. In this case, the Unity associated files should not be included?

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs.meta
magnopus-swifty: @MAG-ChristopherAtkinson made the same point elsewhere. While these tests are de facto Unity tests (note `Assets` in the path) I think I agree that we want to treat the actual Unity project files as unversioned artefacts - this way we support the widest possible range of Unity versions.

 I'll update the .gitignore files & docs to reflect this expectation.

magnopus-swifty: Removed all artefacts at ab42eaf and added a README at ab42eaf.

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
magnopus-swifty: Added some juicy tests at d1787ba. I could expand on this (a lot) but I'd rather do that in follow-up PRs.

UnitTesting/tests/csharp/README.md
```diff
@@ -0,0 +1,32 @@
+# Unity Tests
+
+This directory is the root of a [Unity] project used for running a mixture of generated and hand-written integration tests for the Wrapper Generator.
```
MAG-ChristopherAtkinson, line 3: [Nit] minor comment about the use of Unity here, given the changes made for metadata files, this should likely reference C# in place of Unity

magnopus-swifty: Given the infrastructure (both CI and scripts) are specifically for testing end-to-end in Unity I'd prefer to document what's currently here. You're welcome to rewrite this if and when a generic C# testing framework is implemented.

PR #780, opened 2025-09-01
[NT-0] fix: updated mcs endpoints to use new provided types

MAG-ChristopherAtkinson:
PR resolved changes needed to support the new service generation 
- Updated mcs endpoints to use new provided types

> [!NOTE]
> The generated services need to be committed before this, and then the module will need to be updated to reflect the changes. 
>
> https://github.com/magnopus-opensource/csp-services/pull/16


Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -571,31 +571,20 @@ void AssetSystem::CopyAssetCollectionsToSpace(csp::common::Array<AssetCollection
         = PrototypeAPI->CreateHandler<AssetCollectionsResultCallback, AssetCollectionsResult, void, csp::services::DtoArray<chs::PrototypeDto>>(
             Callback, nullptr);
 
+    auto PrototypeFilters = std::shared_ptr<chs::PrototypeFilters>();
+    PrototypeFilters->SetHasGroup(true);
+
+    auto DuplicateGroupPrototypesOptions = std::shared_ptr<chs::DuplicateGroupPrototypesOptions>();
+    DuplicateGroupPrototypesOptions->SetOriginalGroupId(SourceSpaceId);
+    DuplicateGroupPrototypesOptions->SetNewGroupId(DestSpaceId);
+    DuplicateGroupPrototypesOptions->SetAdditionalFilters(PrototypeFilters);
```
MAG-AdamThorn, line 580: [Question] What about the other filters CHS expose?

MAG-ChristopherAtkinson: They are left to default, which in this case is `std::nullopt`, matching the previous values

MAG-AdamThorn: Ok and we can expose them later as required.

MAG-AdamThorn (approved):

PR #781, opened 2025-09-02
[OF-1738] refac: Pass ClientUserAgentInfo to CSP Initialise method

MAG-AdamThorn:
Currently, for the CSP Initialisation flow, Clients call CSPFoundation::Initialise() and then they call CSPFoundation::SetClientUserAgentInfo(). The Initialise method would default initialise the ClientUserAgentInfo only to then have clients write over it with the correct data.

I require this ClientUserAgentInfo data for the AnalyticsSystem. By having it passed in to Initialisation (which is a better flow), it will be available when the SystemsManager creates the systems and can be passed in to the AnalyticsSystem.


Library/src/Common/Web/HttpPayload.cpp
```diff
@@ -34,12 +34,6 @@ HttpPayload::HttpPayload()
 {
     std::string ResponseContent = csp::CSPFoundation::GetClientUserAgentString().c_str();
```
MAG-AdamThorn, line 35: I have confirmed locally by setting log level to very verbose, that the correct user agent info is still being sent in the header.

Tests/src/TestHelpers.h
```diff
@@ -201,7 +199,7 @@ inline void InitialiseFoundationWithUserAgentInfo(const csp::common::String& End
     ClientHeaderInfo.ClientEnvironment = "ODev";
     ClientHeaderInfo.CHSEnvironment = "oDev";
 
-    csp::CSPFoundation::SetClientUserAgentInfo(ClientHeaderInfo);
+    csp::CSPFoundation::InitialiseWithInject(EndpointRootURI, "OKO_TESTS", ClientHeaderInfo, SignalRMock);
```
MAG-ChristopherAtkinson, line 202: ~~[Fix] this be using CSPFoundation::GetTenant() in place of hard coded string~~

MAG-ChristopherAtkinson (changes requested):

Tests/src/TestHelpers.h
MAG-AdamThorn: GetTenant() returns the tenant value which is set inside the InitialiseWithInject() method. So it will not be possible to use that in this case.

MAG-ChristopherAtkinson: Yeah my mistake, thought that we were forcing the tenant name in test, but it seems that is not the case

Library/include/CSP/CSPFoundation.h
```diff
@@ -122,14 +122,16 @@ class CSP_API CSPFoundation
     /// @param EndpointRootURI csp::common::String : Root URI for service endpoints
     /// @param Tenant csp::common::String : Tenant for Magnopus Services. Data is not shared between tenants so clients using separate tenants cannot
     /// interact with each other.
+    /// @param ClientUserAgentHeader csp::ClientUserAgent : The Client Info data
```
MAG-mv: fix: We should add const& do the doc type

MAG-mv: May as well do the others while you're there

MAG-AdamThorn: Agreed have done them all in the file.

Change: [8b73e85](https://github.com/magnopus-opensource/connected-spaces-platform/pull/781/commits/8b73e85c5e8b3d4953c94805b544d4b18b59a24f)

MAG-mv (approved):

PR #783, opened 2025-09-04
[OF-1744] Update CSPSceneDescription / CSPSceneData to work with new checkpoint files

MAG-mv:
This PR doesn't fully address the ticket. However, this is necessary to complete first, as the checkpoint files have undergone some changes, and we now have a mechanism to generate the checkpoint files manually.

Changes in this PR are:

- Fix key names of checkpoint elements by renaming to lowercase. This was needed to work correctly with our current json DTO parsers
- Regenerate all tests using the new endpoints
- Add anchors to CSPSceneData (Although, these aren't currently added to the  checkpoint file)
- Update multiplayer type strings to use the new format. This significantly reduces file sizes for component properties.
- Added a new function: GetMaterialFromUri, which allows us to get materials from a SceneDescription file, as the ld function would get the asset collections and assets remotely

Limitations:

- We found when doing this PR, that certain features are not supported due to assumptions from the AssetSystem. In an offline context, we already have asset collections and assets, and we don't want to attempt to get them remotely. Due to this, we don't currently support:
    - Conversations
    - LODs

A separate ticket will be created to address these issues.

- Anchors are also not being generated in checkpoint files, so this cannot be tested


Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1682,52 +1682,60 @@ void AssetSystem::GetMaterial(const csp::common::String& AssetCollectionId, cons
 
             // 3. Download material
```
MAG-mv, line 1684: This change looks confusing. All I did was create the GetMaterialFromUri function, and move the DownloadAssetData call to it from the GetMaterial function. GetMaterialFromUri is then called internally by GetMaterials

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -321,12 +321,23 @@ class CSP_API AssetSystem : public SystemBase
...
         const csp::common::String& AssetCollectionId, const csp::common::String& AssetId, MaterialResultCallback Callback);
 
+    /// @brief Gets a material using it's Uri.
+    /// @details This function does not retrieve the AssetCollection or Asset remotely. It downloads the material directly from the uri.
+    /// This should be used in an offline context where The AssetCollection and Asset has been parsed from a SceneData file.
+    /// @param AssetCollection const csp::system::AssetCollection& : The asset collection this material is associated with.
+    /// @param AssetId const csp::common::String& : The asset id this material is associated with.
+    /// @param Uri const csp::common::String& : The uri this material was uploaded to.
+    /// @param Callback MaterialResultCallback : Callback when asynchronous task finishes.
+    CSP_ASYNC_RESULT void GetMaterialFromUri(const csp::systems::AssetCollection& AssetCollection, const csp::common::String& AssetId,
```
MAG-AdamThorn, line 338: [Question] Is there no way around them having to pass the AssetCollection in? It strikes me that if they havn't kept hold of this they will have to first get the AssetCollection first.

MAG-AdamThorn: Perhaps they could rather pass the Ids and ShaderType

Library/src/Multiplayer/MCS/MCSSceneDescription.cpp
```diff
@@ -24,28 +24,15 @@
...
-                                          "System.Private.CoreLib]], Magnopus.Service.Multiplayer.Contracts";
-static constexpr const char* UInt16DictionaryType
-    = "Magnopus.Service.Multiplayer.Messages.Components.ItemComponentData`1[[System.Collections.Generic.IDictionary`2[[System.UInt16, "
-      "System.Private.CoreLib],[Magnopus.Service.Multiplayer.Messages.Components.IComponentData, Magnopus.Service.Multiplayer.Contracts]], "
-      "System.Private.CoreLib]], Magnopus.Service.Multiplayer.Contracts";
-static constexpr const char* StringDictionaryType
-    = "Magnopus.Service.Multiplayer.Messages.Components.ItemComponentData`1[[System.Collections.Generic.IDictionary`2[[System.String, "
-      "System.Private.CoreLib],[Magnopus.Service.Multiplayer.Messages.Components.IComponentData, Magnopus.Service.Multiplayer.Contracts]], "
-      "System.Private.CoreLib]], Magnopus.Service.Multiplayer.Contracts";
+static constexpr const char* BooleanType = "ComponentData[[Boolean]]";
```
MAG-AdamThorn, line 27: Wow this is a bit easier to read!

Tests/assets/checkpoint-empty.json
```diff
@@ -0,0 +1,35 @@
...
+    "data": {
+        "group": {
+            "id": "68addce4985d7612f76b9461",
+            "createdBy": "68addce0985d7612f76b945e",
+            "createdAt": "2025-08-26T16:12:20.701+00:00",
+            "groupOwnerId": "68addce0985d7612f76b945e",
+            "groupType": "space",
+            "name": "checkpoint-empty",
+            "users": [
+                "68addce0985d7612f76b945e"
```
MAG-AdamThorn, line 15: [Question] Is it correct that these are private and undiscoverable spaces with only a single user? I presume this is the temporary test account? Will this work for over test accounts?

MAG-AdamThorn: If an issue you could set `"requiresInvite": true,` to false.

Tests/src/InternalTests/SceneDescriptionTests.cpp
```diff
@@ -307,98 +337,88 @@ CSP_INTERNAL_TEST(CSPEngine, SceneDescriptionTests, SceneDescriptionDeserializeT
...
     csp::multiplayer::NetworkEventBus NetworkEventBus { &Connection, LogSystem };
     csp::multiplayer::OnlineRealtimeEngine EntitySystem(Connection, LogSystem, NetworkEventBus, ScriptRunner);
 
     CSPSceneDescription SceneDescription { Json.c_str(), EntitySystem, LogSystem, ScriptRunner };
     CSPSceneData SceneData { Json.c_str() };
 
-    // Test csp scene description values are correct
-
-    // Space
-    EXPECT_EQ(SceneData.Space.Id, "66c65e8d9821e1cc2b51dc0d");
```
MAG-AdamThorn, line 318: [Question] Is there a reason we are nolonger checking stuff like the OwnderId value?

Tests/src/PublicAPITests/OfflineRealtimeEngineTests.cpp
```diff
@@ -730,14 +738,20 @@ CSP_PUBLIC_TEST(CSPEngine, OfflineRealtimeEngineTests, MarkEntityForUpdate)
 }
 
 /*
-    The CSPSceneDecription has already been thoroughly tested in our SceneDescription tests.
-    This just ensures that the correct amount of objects are populated.
+    This is a basic integration test, showing that an empty CSPSceneDescription is correctly processed in the full online flow.
```
MAG-AdamThorn: [fix] This should say "...  processed in the full offline flow." Not "online flow."

Tests/src/PublicAPITests/OfflineRealtimeEngineTests.cpp
```diff
@@ -751,8 +765,101 @@ CSP_PUBLIC_TEST(CSPEngine, OfflineRealtimeEngineTests, ConstructorTest)
...
+    EXPECT_EQ(RealtimeEngine->GetAllEntities()->Size(), 0);
+
+    // Cleanup
+    AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
+
+    LogOut(UserSystem);
+}
+
+/*
+    This is a basic integration test, showing that a basic CSPSceneDescription with one entity is correctly processed in the full online flow.
```
MAG-AdamThorn: [fix] As above this should be 'offline' flow.

MAG-AdrianMeredith:
I've discussed with matt around the need for a proper api for GetMaterialFromUri.  This isn't something we want to be doing (using different apis just for offline engine). instead if assets could be cached and referenced from the existing functions that would be ideal (so materials don't try and fetch asset collections etc).  Happy to get it in as is for now but we will consider any of these apis as essentially "deprecated" and add TODOs to remove them once theres a suitable alternative.

In short the client shouldn't writing any "if (offline) {" code and all apis should work transparently (within reason)

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-mv: This will be used for offline, so they can get the AssetCollection form the file.

> Perhaps they could rather pass the Ids and ShaderType

They won't know the shadertype until calling this function, unless they parse the metadata themselves.

We could pass in the metadata, along with the id, but this feels worse imo.

MAG-mv: I don't intend this change to be permanent, as we want to rethink the asset system

Tests/assets/checkpoint-empty.json
MAG-mv: This data is irrelevant  for offline mode, at least for now

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -577,34 +577,28 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, CreateUserTest)
...
+    const char* TestUserName = "Matt";
+    const char* TestDisplayName = "Matt";
 
     csp::common::String CreatedUserId;
 
     // Create new user
     {
-        auto [Result] = AWAIT_PRE(UserSystem, CreateUser, RequestPredicate, UniqueUserName, TestDisplayName, UniqueEmail,
-            GeneratedTestAccountPassword, true, true, nullptr, nullptr);
+        auto [Result] = AWAIT_PRE(UserSystem, CreateUser, RequestPredicate, TestUserName, TestDisplayName, "matthew.voisey+2@magnopus.com",
```
MAG-ChristopherAtkinson: [Question] not sure what has happened here, but likely this was not an intentional commit

Tests/src/PublicAPITests/OfflineRealtimeEngineTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/479d1d5cb2952f79d74a0fa1717fc9b1f1576bd1

Tests/src/PublicAPITests/OfflineRealtimeEngineTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/479d1d5cb2952f79d74a0fa1717fc9b1f1576bd1

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -17,9 +17,9 @@
 
 #include "CSP/CSPFoundation.h"
 #include "CSP/Common//Systems/Log/LogSystem.h"
```
MAG-ChristopherAtkinson: [Comment] just noticed there is a double slash include here

Tests/assets/checkpoint-material.json
MAG-ChristopherAtkinson, line 1: [Question] Do we intend to expand this future? I know these current checkpoint files only cover a subset for parity

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -321,12 +321,23 @@ class CSP_API AssetSystem : public SystemBase
     CSP_ASYNC_RESULT void GetMaterials(const csp::common::String& SpaceId, MaterialsResultCallback Callback);
 
     /// @brief Gets a material using its AssetCollection and Asset Id.
```
MAG-ChristopherAtkinson, line 323: [Question] do we know what factor this is?

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1682,52 +1682,60 @@ void AssetSystem::GetMaterial(const csp::common::String& AssetCollectionId, cons
...
-                // Create material of the specific derived type.
-                Material* FoundMaterial = InstantiateMaterialOfType(ShaderType.value(), "", FoundAssetCollection.Id, FoundAsset.Id);
+        const char* MaterialData = static_cast<const char*>(DownloadResult.GetData());
 
-                // Deserialse material data.
-                auto DeserializationResult = DeserializeIntoMaterialOfType(MaterialData, ShaderType.value(), FoundMaterial);
+        std::optional<csp::systems::EShaderType> ShaderType = GetShaderTypeFromMaterialCollection(AssetCollection);
+        if (!ShaderType.has_value())
+        {
+            INVOKE_IF_NOT_NULL(Callback, MakeInvalid<MaterialResult>());
```
MAG-ChristopherAtkinson, line 1712: [Consider] Logging suitable error here, I would assume that if we are unable to determine the shader type, like the data would be malformed

MAG-ChristopherAtkinson (changes requested):

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-mv: It makes 2 extra api calls. I'm not sure how this could be expressed

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-mv: Yeah, it was a clang-tidy change. I'm pretty sure it's valid though

Tests/src/InternalTests/SceneDescriptionTests.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/85a971f5d27f633140ec204a9fbd260c77ab8924

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
MAG-mv: fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/367831491a8d5a545106a7b996c9eb2726eb67ca

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: What reason is there to use a specific email "matthew.voisey+2@magnopus.com" tied to a user, and considering that email is a alt

Tests/assets/checkpoint-material.json
MAG-mv: I dont think the checkpoint files themselves will add more behaviour, unless chs add more dtos, but we need to modify the AssetSystem to be able to find Assets/Collections from a file instead of chs

Library/include/CSP/Systems/Assets/AssetSystem.h
MAG-ChristopherAtkinson: Yeah in this scope that is substantial, resolving thread

Library/src/Systems/Assets/AssetSystem.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/616294e7b852a3b56dcf7732a6bfe5cad40e5069

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-mv: Oh, damn, I missed the email change

MAG-mv: Fixed in: [5b457663056f2fedf66eb0e1b586c0824f1841e2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/783/commits/5b457663056f2fedf66eb0e1b586c0824f1841e2)

MAG-ChristopherAtkinson (approved):

MAG-AdamThorn (approved):

PR #785, opened 2025-09-05
[OF-1533] fix!: token expiration

MAG-ChristopherAtkinson:
Currently, token expiration is disabled on the CI because the test has to sleep for 40 minutes. This PR exposes the new property from MCS for configuring the token expiration length, and re-enables the test. 

During these changes, we identified a bug in our refresh token flow, which would cause the refresh token flow to trigger `LoginStateResCallback` early with an invalid response due to not honouring the expected states.  

> [!NOTE]
> I would have liked to include a potential failure case test for the token refresh, but due to how it is baked into the existing system, this would require some substantial changes; however, a good consideration for future internal feature flagging or mocks

Breaking changes:

`csp::systems::UserSystem::Login(...)` has been updated to include an optional `TokenOptions` 

- Before: `CSP_ASYNC_RESULT void Login(const csp::common::String& UserName, const csp::common::String& Email, const csp::common::String& Password, const csp::common::Optional<bool>& UserHasVerifiedAge,  LoginStateResultCallback Callback);`
- After: `CSP_ASYNC_RESULT void Login(const csp::common::String& UserName, const csp::common::String& Email, const csp::common::String& Password, const csp::common::Optional<bool>& UserHasVerifiedAge, const csp::common::Optional<TokenOptions>& TokenOptions, LoginStateResultCallback Callback);`

`csp::systems::UserSystem::LoginWithRefreshToken(...)` has been updated to include an optional `TokenOptions` 

- Before: `CSP_ASYNC_RESULT void LoginWithRefreshToken(const csp::common::String& UserId, const csp::common::String& RefreshToken, LoginStateResultCallback Callback);`
- After: `CSP_ASYNC_RESULT void LoginWithRefreshToken(const csp::common::String& UserId, const csp::common::String& RefreshToken, const csp::common::Optional<TokenOptions>& TokenOptions, LoginStateResultCallback Callback);`

`csp::systems::UserSystem::LoginAsGuest(...)` has been updated to include an optional `TokenOptions` 

- Before: `CSP_ASYNC_RESULT void LoginAsGuest(const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);`
- After: `CSP_ASYNC_RESULT void LoginAsGuest(const csp::common::Optional<bool>& UserHasVerifiedAge, const csp::common::Optional<TokenOptions>& TokenOptions, LoginStateResultCallback Callback);`

`csp::systems::UserSystem::LoginToThirdPartyAuthenticationProvider(...)` has been updated to include an optional `TokenOptions` 

- Before: `CSP_ASYNC_RESULT void LoginToThirdPartyAuthenticationProvider(const csp::common::String& ThirdPartyToken, const csp::common::String& ThirdPartyStateId, const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);`
- After: `CSP_ASYNC_RESULT void LoginToThirdPartyAuthenticationProvider(const csp::common::String& ThirdPartyToken, const csp::common::String& ThirdPartyStateId, const csp::common::Optional<bool>& UserHasVerifiedAge, const csp::common::Optional<TokenOptions>& TokenOptions, LoginStateResultCallback Callback);`

> [!TIP]
> The interfaces introduce the same pattern for each of the interfaces, the documentation here is for completeness.

### DIFF
--- a/Library/include/CSP/Systems/Users/Authentication.h
+++ b/Library/include/CSP/Systems/Users/Authentication.h
@@ -152,7 +152,7 @@ class CSP_API TokenOptions
 {
 public:
     /// @brief The length of time for a token to expire formatted as "HH:MM:SS", must be between "00:00:01" and "00:30:00"
-    /// The default token expiry length is configured by MCS and defaults to 40 minutes. Value must be less than the default expiry length, or it will
+    /// The default token expiry length is configured by MCS and defaults to 30 minutes. Value must be less than the default expiry length, or it will
     /// be ignored.
     csp::common::String ExpiryLength;
 };
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -98,7 +98,7 @@ public:
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
     /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
-    /// The default token expiry length is configured by MCS and defaults to 40 minutes. Value must be less than the default expiry length, or it will
+    /// The default token expiry length is configured by MCS and defaults to 30 minutes. Value must be less than the default expiry length, or it will
     /// be ignored.
     /// @param Callback LoginStateResultCallback : callback to call when a response is received
     /// @pre One of either UserName or Email must not be empty.
@@ -115,7 +115,7 @@ public:
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
-    /// The default token expiry length is configured by MCS and defaults to 40 minutes. Value must be less than the default expiry length, or it will
+    /// The default token expiry length is configured by MCS and defaults to 30 minutes. Value must be less than the default expiry length, or it will
     /// be ignored.
     /// @param Callback LoginStateResultCallback : Callback when asynchronous task finishes
     /// @pre UserId must not be empty.
@@ -128,7 +128,7 @@ public:
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
     /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
-    /// The default token expiry length is configured by MCS and defaults to 40 minutes. Value must be less than the default expiry length, or it will
+    /// The default token expiry length is configured by MCS and defaults to 30 minutes. Value must be less than the default expiry length, or it will
     /// be ignored.
     /// @param Callback LoginStateResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void LoginAsGuest(bool CreateMultiplayerConnection, const csp::common::Optional<bool>& UserHasVerifiedAge,
@@ -162,7 +162,7 @@ public:
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
     /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
-    /// The default token expiry length is configured by MCS and defaults to 40 minutes. Value must be less than the default expiry length, or it will
+    /// The default token expiry length is configured by MCS and defaults to 30 minutes. Value must be less than the default expiry length, or it will
     /// be ignored.
     /// @param Callback LoginStateResultCallback : callback that contains the result of the Magnopus Connected Services Authentication operation
     CSP_ASYNC_RESULT void LoginToThirdPartyAuthenticationProvider(const csp::common::String& ThirdPartyToken,


MAG-ChristopherAtkinson:
During development, myself and @MAG-mv I discussed the potential of adding a multiplayer test to verify the behaviour of the related UE bug; however, after spending some time with were unable to find a suitable solution, and it would be best to have QA test the changes once adopted

MAG-ChristopherAtkinson:
@MAG-mv I know we discussed the possibility of this being an internal configuration on the user system; however, from Jerome's comments in the sprint review, I feel that having this exposed to the client is an important feature, especially when it comes to their own QA efforts

MAG-ChristopherAtkinson:
> [!IMPORTANT]
> The test duration has been reduced from 40min to 10sec on local MCS

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -147,6 +147,14 @@ class CSP_API TokenInfoParams
     csp::common::Map<csp::common::String, csp::common::String> Parameters;
 };
 
+/// @brief Data structure for overrides to the default token token options
```
MAG-mv: [fix] token token

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -147,6 +147,14 @@ class CSP_API TokenInfoParams
...
+class CSP_API TokenOptions
+{
+public:
+    /// @brief The length of time for a token to expire formatted as "HH:MM:SS", must be between "00:00:01" and "00:30:00"
```
MAG-mv, line 154: [Comment] Did you say chs has a 15-minute default? If so, it's worth mentioning that.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -96,32 +97,36 @@ class CSP_API UserSystem : public SystemBase
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
```
MAG-mv: [Fix] Should be "override"

MAG-mv: Also token token

Library/include/CSP/Systems/Users/Authentication.h
MAG-AdamThorn: [fix] "token token options"

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Also, I think the first override is redundant, maybe "An optional value to specify overrides"?

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -96,32 +97,36 @@ class CSP_API UserSystem : public SystemBase
...
+    CSP_ASYNC_RESULT void Login(const csp::common::String& UserName, const csp::common::String& Email, const csp::common::String& Password, bool CreateMultiplayerConnection,
+        const csp::common::Optional<bool>& UserHasVerifiedAge, const csp::common::Optional<TokenOptions>& TokenOptions,
+        LoginStateResultCallback Callback);
 
     /// @brief Resume a previous session for the associated user ID using a refresh token
     /// The refresh token can be obtained after registering a callback with `SetNewLoginTokenReceivedCallback` and logging in regularly.
     /// @param UserId csp::common::String : User ID for the previous session
     /// @param RefreshToken csp::common::String : Refresh token to be used for refreshing the authentication token
-    /// @param CreateMultiplayerConnection bool : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
+    /// /// @param CreateMultiplayerConnection bool : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
```
MAG-mv: [Fix] added an additional ///

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [fix] As above "token token"

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -96,32 +97,36 @@ class CSP_API UserSystem : public SystemBase
...
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
```
MAG-mv: [Fix] Same issues as above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -96,32 +97,36 @@ class CSP_API UserSystem : public SystemBase
...
     /// @param Callback LoginStateResultCallback : Callback when asynchronous task finishes
     /// @pre UserId must not be empty.
     CSP_ASYNC_RESULT void LoginWithRefreshToken(const csp::common::String& UserId, const csp::common::String& RefreshToken,
```
MAG-mv, line 122: [Fix] Same issues as above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -96,32 +97,36 @@ class CSP_API UserSystem : public SystemBase
...
-        bool CreateMultiplayerConnection, LoginStateResultCallback Callback);
+        bool CreateMultiplayerConnection, const csp::common::Optional<TokenOptions>& TokenOptions, LoginStateResultCallback Callback);
 
     /// @brief Log in to Magnopus Connected Services as a guest.
     /// @param CreateMultiplayerConnection bool : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
```
MAG-mv: [Fix] Same issues as above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -150,9 +155,11 @@ class CSP_API UserSystem : public SystemBase
...
```
MAG-mv: [Fix] Same issues as above

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [Consider] Also the description is a little hard to parse with the double 'override'. Consider rewording.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -226,6 +230,13 @@ void UserSystem::Login(const csp::common::String& UserName, const csp::common::S
             Request->SetVerifiedAgeEighteen(*UserHasVerifiedAge);
         }
 
+        if (TokenOptions.HasValue())
+        {
+            auto Options = std::make_shared<chs_user::TokenOptions>();
```
MAG-AdamThorn, line 255: [Fix] We should verify that that expiry length has been specified in the correct format.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -430,24 +437,30 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LoginErrorTest)
     LogOut(UserSystem);
 }
 
-// This will be updated and re-instated in OF-1533
-CSP_PUBLIC_TEST(DISABLED_CSPEngine, UserSystemTests, RefreshTest)
+CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, RefreshTest)
```
MAG-AdamThorn, line 442: [fix] I don't see any new tests specifically for the new Token Options. So for example what happens if they pass an expiry time that is outside the allowed bounds. Or if they pass it in an invalid format?

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: Also you should state what happens with token expiration if the TokenOptions argument is not provided.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [fix] Double '///'

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [fix] As above.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [fix] as above

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: [fix] As above

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -430,24 +437,30 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LoginErrorTest)
...
 {
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto* UserSystem = SystemsManager.GetUserSystem();
 
     csp::common::String UserId;
 
     // Log in
-    LogInAsNewTestUser(UserSystem, UserId);
+    auto TokenOptions = csp::systems::TokenOptions();
```
MAG-mv, line 450: [Consider] Having a constructor which takes the string would be nice here

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -450,10 +450,17 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, RefreshTest)
 
     LogInAsNewTestUser(UserSystem, UserId, true, true, TokenOptions);
 
+    bool TokenHasBeenRefresh = false;
```
MAG-AdamThorn: [fix] 'TokenHasBeenRefreshed'

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -450,10 +450,17 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, RefreshTest)
...
+    // Ensure that the token is refresh when attempting to make a authenticated call after expiry
+    UserSystem->SetNewLoginTokenReceivedCallback(
+        [&TokenHasBeenRefresh](const csp::systems::LoginTokenInfoResult& /*Result*/) { TokenHasBeenRefresh = true; });
+
```
MAG-AdamThorn, line 464: [Question] Is the 'SetNewLoginTokenReceivedCallback' callback only used to communicate successful refresh of a token?

MAG-AdamThorn (changes requested):

Tests/src/PublicAPITests/UserSystemTestHelpers.h
```diff
@@ -24,19 +24,24 @@ const char GeneratedTestAccountPassword[] = "3R{d2}3C<x[J7=jU";
 // The implementation for all this is in UserSystemTests.cpp, if you can believe it.
 
 void LogIn(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, const csp::common::String& Email,
-    const csp::common::String& Password, bool CreateMultiplayerConnection = true, bool AgeVerified = true,
+    const csp::common::String& Password, bool CreateMultiplayerConnection = true, bool AgeVerified = true, 
+    const csp::systems::TokenOptions& TokenOptions = csp::systems::TokenOptions(),
```
MAG-AdamThorn, line 28: [Question] Token options is optional so why pass an empty object here and not nullopt?
I have made a request elsewhere that we should be verifying that tokenoptions is valid which would require this change (to nullopt) anyway.

MAG-AdamThorn: Same with LogSystemTests above and login methods below.

MAG-mv (changes requested):

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -89,7 +90,8 @@ void LogIn(csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId,
 void LogInAsGuest(
     csp::systems::UserSystem* UserSystem, csp::common::String& OutUserId, bool CreateMultiplayerConnection, csp::systems::EResultCode ExpectedResult)
 {
-    auto [Result] = Awaitable(&csp::systems::UserSystem::LoginAsGuest, UserSystem, CreateMultiplayerConnection, true).Await(RequestPredicate);
+    auto [Result]
+        = Awaitable(&csp::systems::UserSystem::LoginAsGuest, UserSystem, CreateMultiplayerConnection, true, nullptr).Await(RequestPredicate);
```
MAG-AdamThorn: [fix] LoginAsGuest takes a const std::optional ref, you should pass nullopt here rather than nullptr.

MAG-AdamThorn: [Question] Why does the LogInAsGuest helper function not have an argument for tokenoptions?

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -298,7 +302,7 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, BadTokenLogInTest)
     LogOut(UserSystem);
 
     // Log in
-    auto [Result] = AWAIT_PRE(UserSystem, LoginWithRefreshToken, RequestPredicate, UserId, "badtoken", true);
+    auto [Result] = AWAIT_PRE(UserSystem, LoginWithRefreshToken, RequestPredicate, UserId, "badtoken", true, nullptr);
```
MAG-AdamThorn, line 307: [fix] As mentioned above this should be nullopt. Same with below.

MultiplayerTestRunner/src/LoginRAII.cpp
```diff
@@ -33,7 +33,7 @@ csp::common::String LogIn(csp::systems::UserSystem& UserSystem, const csp::commo
     std::promise<csp::systems::LoginStateResult> ResultPromise;
     std::future<csp::systems::LoginStateResult> ResultFuture = ResultPromise.get_future();
 
-    UserSystem.Login("", Email, Password, true, AgeVerified,
+    UserSystem.Login("", Email, Password, true, AgeVerified, nullptr,
```
MAG-AdamThorn, line 36: [fix] As mentioned elsewhere, login takes a const ref so this should be nullopt.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: the use of nullptr is correct in this case as this is csp::common::Optional, and we do not guarantee nullopt across the dll boundary

MultiplayerTestRunner/src/LoginRAII.cpp
MAG-ChristopherAtkinson: same as above

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: This was initially done as age verification was also not passed; however, thinking on this, it still makes sense due to a restful endpoint to include the token information, will update

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: I suggest that we exclude it here; these classes should realistically be structs, as they act as models rather than first-class objects.  (Granted wrapper gen limitation as usual)

Tests/src/PublicAPITests/UserSystemTestHelpers.h
MAG-ChristopherAtkinson: The test helper generally simplifies the interface, removing the need for csp::common::Optional altogether, in this case we adopt that mentality and allow the token options to default construct

MAG-ChristopherAtkinson: as for std::nullopt, the use of nullptr is correct in this case as this is csp::common::Optional, and we do not guarantee nullopt across the dll boundary

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: I believe this can be triggered on failure as well. I can update the test to ensure that the result is successful in the context

Library/include/CSP/Systems/Users/Authentication.h
MAG-ChristopherAtkinson: I addressed these as a bulk change due to their relation; changes can be found here: [843ca22](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/843ca22af521056c6891cebe44870b6fef09a9c0)

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/Authentication.h
MAG-ChristopherAtkinson: I addressed these as a bulk change due to their relation; changes can be found here: [843ca22](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/843ca22af521056c6891cebe44870b6fef09a9c0)

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: This has now been changed: [6fe6c7d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/6fe6c7d8916f23bb2d4b9d9d6c1194ad10c32e16)

Library/include/CSP/Systems/Users/Authentication.h
MAG-ChristopherAtkinson: MCS provides a default of 40 minutes, which has now been added. I do wonder if we want this, however, as this could be stale if changes on their end [843ca22](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/843ca22af521056c6891cebe44870b6fef09a9c0)

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: Fixed here: [f23ecc7](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/f23ecc702451a1ced1e55eed5cbe41e01768ae96)

Library/src/Systems/Users/UserSystem.cpp
MAG-ChristopherAtkinson: I have added a check to ensure it is formatted as expected, and also provided logging to the user, granted if you think the message could be improved please let me know: [1ae8f1f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/1ae8f1f85db5868c5e811ae4a176f3d6e04ecbcc)

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: I have re-implemented the test that I lost during rebase, and also added some additional tests around expected MCS behaviour for clamping the value: [90567ec](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/90567ecef175c615d8efa6d22e2aa236c63047e3)

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: Included change here to ensure the result is successful: [90567ec](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/90567ecef175c615d8efa6d22e2aa236c63047e3)

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -147,11 +147,13 @@ class CSP_API TokenInfoParams
     csp::common::Map<csp::common::String, csp::common::String> Parameters;
 };
 
-/// @brief Data structure for overrides to the default token token options
+/// @brief Data structure for overrides to the default token options
```
MAG-AdamThorn, line 150: [fix] I don't think this comment makes sense. You are overriding the default token expiration.

Library/include/CSP/Systems/Users/Authentication.h
MAG-AdamThorn: I agree, as it is my understanding that the default is 30 mins, hence why only the range 00:00:01 to 00:00:30 is valid.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -97,22 +97,26 @@ class CSP_API UserSystem : public SystemBase
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
-    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
```
MAG-AdamThorn, line 100: [fix] As above, I think this should be: "override for default token expiration length." or similar.

[Question] Are you certain it is 40 mins? It is my understanding it is 30 mins, hence the range available for the token options.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -97,22 +97,26 @@ class CSP_API UserSystem : public SystemBase
...
     /// The refresh token can be obtained after registering a callback with `SetNewLoginTokenReceivedCallback` and logging in regularly.
     /// @param UserId csp::common::String : User ID for the previous session
     /// @param RefreshToken csp::common::String : Refresh token to be used for refreshing the authentication token
-    /// /// @param CreateMultiplayerConnection bool : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
+    /// @param CreateMultiplayerConnection bool : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
-    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
```
MAG-AdamThorn, line 117: [fix] As above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -123,10 +127,12 @@ class CSP_API UserSystem : public SystemBase
     /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
     /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
     /// @param UserHasVerifiedAge csp::common::Optional<bool> : An optional bool to specify whether or not the user has verified that they are over 18
-    /// @param TokenOptions csp::common::Optional<TokenOptions> : An optional token overide to specify overrides to the default token token options
+    /// @param TokenOptions csp::common::Optional<TokenOptions> : Optional override for default token options.
```
MAG-AdamThorn, line 130: [fix] As above

MAG-AdamThorn, line 130: [fix] As above

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -155,12 +161,13 @@ class CSP_API UserSystem : public SystemBase
...
```
MAG-AdamThorn, line 164: [fix] As above

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -232,9 +249,12 @@ void UserSystem::Login(const csp::common::String& UserName, const csp::common::S
 
         if (TokenOptions.HasValue())
         {
-            auto Options = std::make_shared<chs_user::TokenOptions>();
-            Options->SetExpiryLength(TokenOptions->ExpiryLength);
-            Request->SetTokenOptions(Options);
+            if (CheckExpiryLengthFormat(TokenOptions->ExpiryLength))
```
MAG-AdamThorn: [Question] What is the value of sending the ExpiryLength if it is an empty string? In that case MCS will just set it to the default so why both sending it at all.

MAG-AdamThorn: This question applies to the below as well.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: I have added a comment, I do not think the result of that check is correct.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -465,6 +472,127 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, RefreshTest)
...
+    auto* UserSystem = SystemsManager.GetUserSystem();
+
+    SystemsManager.GetLogSystem()->SetSystemLevel(csp::common::LogLevel::Warning);
+    auto TokenOptions = csp::systems::TokenOptions();
+
+    // Ensure MCS clamps the value of expiry length when above bounds
+    {
+        TokenOptions.ExpiryLength = "12:00:00";
+
+        // Ensure that the token expiry time matched the provided token options
```
MAG-AdamThorn: [fix] Looks to be a copy/paste error

Tests/src/PublicAPITests/UserSystemTestHelpers.h
MAG-AdamThorn: Ah right yes, our optional accepts a nullptr.

Library/src/Systems/Users/UserSystem.cpp
MAG-ChristopherAtkinson: In the future, this may be expanded, even if 'ExpiryLength' is ignored other properties should still be applied

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: The default expiration time is 30 minutes; however, MCS will still accept a new token for an additional 10 minutes. However, it would make sense not to include the additional behaviour

Library/include/CSP/Systems/Users/Authentication.h
MAG-ChristopherAtkinson: Currently, we only have the means to override the token expiration. However, this may not always be the case; therefore, this will likely hold the related token properties

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Are there plans to add other properties? I am not aware of any not that I have asked CHS that question.

Library/include/CSP/Systems/Users/Authentication.h
MAG-AdamThorn: Yes that makes sense

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: Yeah was supposed to reference default properties: [a3d2b27](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/a3d2b27cd2dcf89d1f3fc799d80953408c8c8efb)

Library/src/Systems/Users/UserSystem.cpp
MAG-ChristopherAtkinson: No, but consider CHS structure supports more properties, I would not rule this out. However, based on discussion, I have opted to go with coupling the structure with the property. The change prevents the result from being populated when provided an empty string: [45d93cf](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/45d93cf62aeb2f831c2db1c0a14325430938c25b)

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: Updated this to reflect the 30-minute expectation from MCS: [d417e4b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/d417e4b0f04a62affb69bd6835b4c2b3aa6d8dce)

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

MAG-ChristopherAtkinson: same as above ^

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ChristopherAtkinson: same as above ^

Library/src/Systems/Users/UserSystem.cpp
MAG-ChristopherAtkinson: I am unsure what comment this refers to; however, this has now been updated with the following behaviour: [45d93cf](https://github.com/magnopus-opensource/connected-spaces-platform/pull/785/commits/45d93cf62aeb2f831c2db1c0a14325430938c25b)

The change now returns false on an empty string, preventing the request from being populated with token options

PR #786, opened 2025-09-05
[nt-0] fix: Protect against null pointers

magnopus-swifty:
Attempt to prevent the class of native level crash caused by accessing invalid pointers.

* `Dispose()` now sets the internal pointer to zero
* Changes `_ptr` to a property
* Accessing `_ptr` when it's `IntPtr.Zero` throws a NullReferenceException (rather than segfaulting)
* Prevent modification of the underlying pointer for `NativeClassWrapper`

This will catch two common causes of segfaults and instead throw a managed `NullReferenceException`.

* Accessing native code when the pointer is `null_ptr`
* Accessing fields and properties on a `NativeClassWrapper` after `Dispose()` has been called.

### DIFF
--- a/Library/CSharpWrapper/src/NativeClassWrapper.cs
+++ b/Library/CSharpWrapper/src/NativeClassWrapper.cs
@@ -39,13 +39,41 @@ namespace Csp
 
     public class NativeClassWrapper
     {
-        internal IntPtr _ptr;
+        private IntPtr _ptrValue = IntPtr.Zero;
         internal bool _ownsPtr;
         internal bool _disposed = false;
 
         internal virtual string _safeTypeName { get; }
 
-        public bool PointerIsValid => _ptr != IntPtr.Zero;
+        public bool PointerIsValid => _ptrValue != IntPtr.Zero;
+
+        /// <summary>
+        /// Pointer to the native object.
+        /// </summary>
+        /// <exception cref="NullReferenceException">Thrown if the pointer is null.</exception>
+        /// <exception cref="InvalidOperationException">Thrown if attempting to change the pointer once it's set to a non-null value.</exception>
+        internal IntPtr _ptr
+        {
+            get
+            {
+                // Prevent accessing the pointer if it's null
+                if (_ptrValue == IntPtr.Zero)
+                {
+                    throw new NullReferenceException($"Attempting to access a null pointer for {_safeTypeName}");
+                }
+                return _ptrValue;
+            }
+            set
+            {
+                // Prevent changing the pointer once it's set to a non-null value
+                if (_ptrValue != IntPtr.Zero && value != IntPtr.Zero)
+                {
+                    throw new InvalidOperationException($"Attempting to change the native pointer for {_safeTypeName} from {_ptrValue} to {value}");
+                }
+
+                _ptrValue = value;
+            }
+        }
 
         public NativeClassWrapper() { }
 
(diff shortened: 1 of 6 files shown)


magnopus-swifty:
Revisiting this PR as it may assist in fixing [OB-4695: OKO iOS client randomly crashes](https://magnopus.atlassian.net/browse/OB-4695). While it won't fix bugs it may turn segfaults into more diagnosable `NullReferenceException`s.

Verified that this change appears to work as expected by running the full suite of tests in https://github.com/magnopus/oly-foundation-unity/

Library/CSharpWrapper/src/NativeClassWrapper.cs
```diff
@@ -37,18 +37,92 @@ internal NativePointer(IntPtr pointer, byte ownsOwnData)
...
+        /// This can be used to prevent a <see cref="NullReferenceException"/> when calling <see cref="_ptr"/>.
+        /// </remarks>
+        public bool PointerIsValid => _ptrValue != IntPtr.Zero;
+
+        /// <summary>
+        /// Pointer to the native object.
+        /// </summary>
+        /// <exception cref="NullReferenceException">Thrown if the pointer is null.</exception>
+        /// <exception cref="InvalidOperationException">Thrown if attempting to change the pointer once it's set to a non-null value.</exception>
+        internal IntPtr _ptr
```
magnopus-swifty, line 86: I considered renaming this (since it's now a property) but decided to keep it as `_ptr` both to keep the change localized to this file and to indicate that it's still an internal implementation detail.

UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
```diff
@@ -63,5 +69,17 @@ public void MultipleDisposes()
             simpleClass.Dispose();
             simpleClass.Dispose();
         }
+
+        [Test]
+        public void AccessAfterDisposeThrows()
```
magnopus-swifty, line 74: There isn't a nice way to do a before/after comparison with this test because prior to this PR it will literally segfault and take out the entire process. It's fairly easy to reproduce this behaviour if you care to.

Tools/WrapperGenerator/Templates/CSharp/Partials/Template/Method/Destructor.mustache
```diff
@@ -8,9 +8,9 @@ void RealDispose() {
     if (_ownsPtr && !_disposed)
     {
         {{ unique_name }}(_ptr);
-        _disposed = true;
     }
 
+    _ptr = IntPtr.Zero;
```
MAG-ElliotMorris, line 13: [Question] Is the move of this assignment outside the conditional significant, or merely a safety thing (since this is nulling out the ptr, is shouldn't really matter if `_disposed` is true or not, might as well set it to zero in either branch.)? I'm presuming the latter.

MAG-ElliotMorris (approved):
Approving based mostly on faith, the fact that you've run a suite of manual tests, and the fact that the `_ptr` property logic reads as eminently reasonable. ... As well as this only really being able to hurt C#, which is your domain :P

I'd definitely get a proper C# dev's eyeball before merge, if you haven't yet.

MAG-AlessioRegalbuto (approved):
LGTM

MAG-mv (approved):

Tools/WrapperGenerator/Templates/CSharp/Partials/Template/Method/Destructor.mustache
magnopus-swifty: Yeah, this was just a safety thing - we weren't nulling the pointer at all before and we always want to null it whether it's owned or not.

The two lines doing the same thing (`_disposed = true`) made me scratch my head and question whether it was significant, which seemed enough reason to make it crystal clear.

PR #787, opened 2025-09-08
[OF-1754] test: Skip application settings tests when no admin login is set

MAG-ElliotMorris:
These tests fail out of the box and should not run in contexts where they cannot pass.

We set the admin email/password on teamcity, so this effectively makes these tests run on CI, or locally only if you have set the environment variable (which we probably will never do, but that's fine).


MAG-SamBirley:
[Suggestion] Should we make a utility function called in these tests? Might be helpful for other/future tests which similarly depend upon valid admin accounts being set.

PR #788, opened 2025-09-09
[OF-1738] feat: Implementation of the AnalyticsSystem

MAG-AdamThorn:
This work involves the introduction of a new Analytics System, which wraps a single MCS Analytics endpoint on the User Service - `POST /analytics/bulk`.

This PR introduces a single public system method - AnalyticsSystem::SendAnalyticsEvent().

A subsequent PR will add a second method to this system for the batching of analytics events - AnalyticsSystem::BatchAnalyticsEvent().

### DIFF
--- a/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
+++ b/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/CancellationToken.h"
+#include "CSP/Common/Optional.h"
+#include "CSP/Common/String.h"
+#include "CSP/Systems/SystemBase.h"
+
+namespace csp
+{
+class ClientUserAgent;
+} // namespace csp
+
+namespace csp::services
+{
+class ApiBase;
+} // namespace csp::services
+
+namespace csp::web
+{
+class WebClient;
+} // namespace csp::web
+
+namespace csp::systems
+{
+
+/// @ingroup Analytics System
+/// @brief Public facing system that allows AnalyticsRecords to be sent to MCS.
+class CSP_API AnalyticsSystem : public SystemBase
+{
+    CSP_START_IGNORE
+    /** @cond DO_NOT_DOCUMENT */
+    friend class SystemsManager;
+    /** @endcond */
+    CSP_END_IGNORE
+
+public:
+    /**
+     * @brief Sends an analytics event to the Analytics service.
+     * @details Please note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
+     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * Constructs an Analytics Record with the following format:
+     *      {ProductIdentifer}.{ProductContext}.{Tenant}.{ProductContextSection}.{Category}.{InteractionType}.{SubCategory}
+     *      csp.web-client-x.CLIENT_TENANT_X.UI.menus.click
+     * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context - 'UI' in the example provided above.
+     * @param Category const csp::common::String& : Categorization field - 'menus' in the example provided above.
+     * @param InteractionType const csp::common::String& : The interaction that occurred - 'click' in the example provided above.
+     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field - not used in the example provided above.
+     * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata -
+     * not used in the example provided above.
+     */
+    void SendAnalyticsEvent(const csp::common::String& ProductContextSection, const csp::common::String& Category,
+        const csp::common::String& InteractionType, const csp::common::Optional<csp::common::String>& SubCategory,
+        const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& Metadata, NullResultCallback Callback);
+
+private:
+    AnalyticsSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
+    CSP_NO_EXPORT AnalyticsSystem(csp::web::WebClient* InWebClient, const csp::ClientUserAgent* AgentInfo, common::LogSystem& LogSystem);
+    ~AnalyticsSystem();
+
+    csp::services::ApiBase* AnalyticsApi;
+
+    const csp::ClientUserAgent* UserAgentInfo;
+};
+
+} // namespace csp::systems
--- a/Library/src/Systems/Analytics/AnalyticsSystem.cpp
+++ b/Library/src/Systems/Analytics/AnalyticsSystem.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CSP/Systems/Analytics/AnalyticsSystem.h"
+
+#include "CSP/CSPFoundation.h"
+#include "CallHelpers.h"
+#include "Services/UserService/Api.h"
+#include "Services/UserService/Dto.h"
+#include "Systems/ResultHelpers.h"
+
+#include <fmt/format.h>
+
+using namespace csp;
+using namespace csp::common;
+
+namespace chs = csp::services::generated::userservice;
+
+namespace csp::systems
+{
+
+AnalyticsSystem::AnalyticsSystem()
+    : SystemBase(nullptr, nullptr, nullptr)
+    , AnalyticsApi(nullptr)
+    , UserAgentInfo(nullptr)
+{
+}
+
+AnalyticsSystem::AnalyticsSystem(csp::web::WebClient* InWebClient, const csp::ClientUserAgent* AgentInfo, csp::common::LogSystem& LogSystem)
+    : SystemBase(InWebClient, nullptr, &LogSystem)
+{
+    AnalyticsApi = new chs::AnalyticsApi(InWebClient);
+    UserAgentInfo = AgentInfo;
+}
+
+AnalyticsSystem::~AnalyticsSystem() { delete (AnalyticsApi); }
+
+void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, const String& Category, const String& InteractionType,
+    const Optional<String>& SubCategory, const Optional<Map<String, String>>& Metadata, NullResultCallback Callback)
+{
+    if (ProductContextSection.IsEmpty() || Category.IsEmpty() || InteractionType.IsEmpty())
+    {
+        CSP_LOG_MSG(common::LogLevel::Error, "Missing the required fields for the Analytics Event.");
+        INVOKE_IF_NOT_NULL(Callback, MakeInvalid<NullResult>());
+
+        return;
+    }
+
+    // Construct an AnalyticsRecord object
+    auto Record = std::make_shared<chs::AnalyticsRecord>();
+    Record->SetProductIdentifier("csp");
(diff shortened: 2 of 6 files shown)


Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -0,0 +1,82 @@
...
+    CSP_START_IGNORE
+    /** @cond DO_NOT_DOCUMENT */
+    friend class SystemsManager;
+    /** @endcond */
+    CSP_END_IGNORE
+
+public:
+    /**
+     * @brief Sends an analytics event to the Analytics service.
+     * @details Please note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
```
MAG-AdamThorn, line 56: [Comment] I will be adding this `BatchAnalyticsEvent` method as part of my next ticket which I will be picking up as soon as this PR is merged.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -0,0 +1,82 @@
...
+     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * Constructs an Analytics Record with the following format:
+     *      {ProductIdentifer}.{ProductContext}.{Tenant}.{ProductContextSection}.{Category}.{InteractionType}.{SubCategory}
+     *      csp.web-client-x.CLIENT_TENANT_X.UI.menus.click
```
MAG-ElliotMorris: [nit] Love the example, perhaps prefix this with `example:` just to be clear that this is an example of the above schema.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -0,0 +1,82 @@
...
+    /** @endcond */
+    CSP_END_IGNORE
+
+public:
+    /**
+     * @brief Sends an analytics event to the Analytics service.
+     * @details Please note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
+     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * Constructs an Analytics Record with the following format:
+     *      {ProductIdentifer}.{ProductContext}.{Tenant}.{ProductContextSection}.{Category}.{InteractionType}.{SubCategory}
```
MAG-ElliotMorris: [Consider] Torn on this. I like giving context to what we're doing ... but this feels like internal behaviour that consumers of this method should be hidden from. Think about just not specifying the specific format.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -0,0 +1,82 @@
...
+     *      csp.web-client-x.CLIENT_TENANT_X.UI.menus.click
+     * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context - 'UI' in the example provided above.
+     * @param Category const csp::common::String& : Categorization field - 'menus' in the example provided above.
+     * @param InteractionType const csp::common::String& : The interaction that occurred - 'click' in the example provided above.
+     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field - not used in the example provided above.
+     * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata -
+     * not used in the example provided above.
+     */
```
MAG-ElliotMorris: [Comment] This all feels rather overdesigned to me, but I understand it's probably gone through a PRD process and whatnot. Would bet you anything that category and subcategory will get used strangely though, just like how no-one ever _really_ uses the extended message in exceptions.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-ElliotMorris: [Consider] If we are going to have this. Perhaps mention which one of these are populated from internal state. Just a note maybe like "@ note Product Identifier, Product Context and Tenant are set from pre-existing CSP state".

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -0,0 +1,133 @@
...
+    : SystemBase(nullptr, nullptr, nullptr)
+    , AnalyticsApi(nullptr)
+    , UserAgentInfo(nullptr)
+{
+}
+
+AnalyticsSystem::AnalyticsSystem(csp::web::WebClient* InWebClient, const csp::ClientUserAgent* AgentInfo, csp::common::LogSystem& LogSystem)
+    : SystemBase(InWebClient, nullptr, &LogSystem)
+{
+    AnalyticsApi = new chs::AnalyticsApi(InWebClient);
```
MAG-ElliotMorris: [Consider] I know this is probably following a pattern, but could this could trivially be a `unique_ptr`?

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -0,0 +1,133 @@
...
+    Record->SetVersionMatrix(VersionsMatrix);
+
+    if (Metadata.HasValue())
+    {
+        std::map<String, String> AnalyticsRecordMetadata;
+
+        auto* Keys = Metadata->Keys();
+
+        for (size_t idx = 0; idx < Keys->Size(); ++idx)
+        {
```
MAG-ElliotMorris: [fix] This leaks cause of the dumb keys allocating thing.

There's a method on `Map` now, `GetUnderlying`, which lets you turn it into a `std::map` and do a regular `for(const std::pair<const ...,...> MetadataEntry : MetaData->GetUnderlying()){}`

MAG-ElliotMorris: Saying that, you're trying to transform a `csp::Map` to a `std::map`, you might just be able to do `std::map<String, String> AnalyticsRecordMetadata = Metadata->GetUnderlying();`

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -0,0 +1,133 @@
...
+    {
+        if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
+        {
+            return;
+        }
+
+        if (Result.GetResultCode() == csp::systems::EResultCode::Failed)
+        {
+            CSP_LOG_FORMAT(common::LogLevel::Error, "Failed to send Analytics Event. ResCode: %d, HttpResCode: %d",
+                static_cast<int>(Result.GetResultCode()), Result.GetHttpResultCode());
```
MAG-ElliotMorris: [nit] Despite the macros, I feel `LogSystem->LogMsg(common::LogLevel::Error, fmt::format("...", ...,...,));` reads a tad clearer, although there's not much in it. We should think about preferring that form going forward, because it might lead to us being able to delete macros which is always nice.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [nit] 2025

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -0,0 +1,108 @@
...
+    const auto TestCategory = String("Event_Category");
+    const auto TestInteractionType = String("Event_InteractionType");
+    const auto TestSubCategory = String("Event_SubCategory");
+    const Map<String, String> TestMetadata = { { "Key1", "Value1" }, { "Key2", "Value2" } };
+
+    auto [Result] = AWAIT_PRE(AnalyticsSystem, SendAnalyticsEvent, RequestPredicate, TestProductContextSection, TestCategory, TestInteractionType,
+        TestSubCategory, TestMetadata);
+
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+    EXPECT_EQ(Result.GetHttpResultCode(), 201);
```
MAG-ElliotMorris: [Comment] It's a shame we can't validate this works end-to-end. Might be worth checking with CHS/Datadog that they actually received the data you sent. Not that I think we need such a test in our suite, for our part, the return codes should be enough for integration.

MAG-AdamThorn: I had a meet with Jake and confirmed that the events were being sent correctly via datadog - link [here](https://app.datadoghq.com/logs?query=env%3Ao2dev%2A%20service%3Amag-user%20Analytics&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1757430290127&to_ts=1757431190127&live=true).

image

Jake was also able to confirm that he could see the event data in AWS.

MAG-ChristopherAtkinson: [Nit] I would expect the use of `EResponseCodes` here for clarity against what is being evaluated `static_cast<uint16_t>(csp::web::EResponseCodes::ResponseCreated)`

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    // Ensure the required fields error message is logged when we try to send an analytics event with a required field missing
+    const csp::common::String LockErrorMsg = "Missing the required fields for the Analytics Event.";
+    EXPECT_CALL(MockLogger.MockLogCallback, Call(LockErrorMsg)).Times(1);
+
+    auto [Result] = AWAIT_PRE(AnalyticsSystem, SendAnalyticsEvent, RequestPredicate, TestProductContextSection, TestCategory, TestInteractionType,
+        TestSubCategory, TestMetadata);
+
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
+    EXPECT_EQ(Result.GetHttpResultCode(), 0);
```
MAG-ChristopherAtkinson, line 104: [Question] On failure, I would expect a return in the 400 range. I assume this due to the default invalid null result as we did not actually make any request?

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: I have rewritten the comment to be clearer as I don't think that example was really helping any. Additionally I have documented what data is captured internally as well as expanding on the proposed use of Metadata. 

Update: [7d6f75d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/7d6f75df120376e03b95278991d57c6d1b77bafb)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed, see the updated documentation in the change linked above - 7d6f75d.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: I have updated documentation, see the change linked above - 7d6f75d.

There I include a bit more information around the intended use of Metadata. Speaking to Andy I also got this [documentation](https://magnopus.atlassian.net/wiki/spaces/OL/pages/3697934339/Dakoko+Sample+Data+Reports) which goes into further detail.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed, see the updated documentation in the change linked above - 7d6f75d.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Good shout.

Update: [75bd8e9](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/75bd8e9d71e8b6848e68d4d77d2dcadd4829fc20)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Thanks, I hadn't noticed that.

Fix as you suggested: [cc0d1df](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/cc0d1dfacf407fa075f657f6e544b35b647fde4c)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Included in the following commit: [cc0d1df](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/cc0d1dfacf407fa075f657f6e544b35b647fde4c)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Update: [5ad29cb](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/5ad29cbfc4bc644527213b455377383a35a1985f)

MAG-ChristopherAtkinson (approved):

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Yes that's correct. This test validates correct behaviour when you do not provide the required fields. As such it returns 0.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Good shout, updated here: [f1887e6](https://github.com/magnopus-opensource/connected-spaces-platform/pull/788/commits/f1887e631224dc25b265d92d223ce83e600ab6b9)

PR #789, opened 2025-09-09
[OF-1750] feat: LoginAsGuestWithDeferredProfileCreation

MAG-ElliotMorris:
Provide the special case API for logging into MCS in an "optimized" manner.

This is a unique endpoint for specific projects, and should not be called in the general case. A large warning is added to the documentation to that effect.

The endpoint has an undefined stretch of time before other profile related methods will become capable of returning non 404 responses. This is a shame, but it's been documented.

Need to figure this out more generally, I understand that in distributed systems, a 404 is always a possibility, but there's a difference between an error always being possible vs a success never being possible. These are two categories imo and I would like to be able to know when endpoints are capable of returning success results vs when they are not. Otherwise you've just gotta sleep, loop, and hope.


Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -123,6 +123,18 @@ class CSP_API UserSystem : public SystemBase
     CSP_ASYNC_RESULT void LoginAsGuest(
         bool CreateMultiplayerConnection, const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
+    /// @brief Log in to Magnopus Connected Services as a guest, allowing the backend to defer profile creation and perform other optimizations.
```
MAG-ChristopherAtkinson: [Fix] Magnopus Connected Services -> Magnopus Cloud Services

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -284,6 +297,24 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LogInAsGuestTest)
     LogOut(UserSystem);
 }
 
+CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LogInAsGuestDeferredProfileCreationTest)
+{
+    // We can't really test more than this. This deferred endpoint explicitly provides
+    // no guarentees of actually creating a profile in any sort of timely manner.
+    // Any subsequent failures to query endpoints would be within the bounds of permitted behaviour.
```
MAG-ChristopherAtkinson, line 310: [Comment] I would have liked to see some additional testing here, but I agree, due to the deferred creation we would likely see a bunch of false positives or require long wait delays.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -391,6 +391,66 @@ void UserSystem::LoginAsGuest(
...
+                csp::multiplayer::MultiplayerConnection::ErrorCodeCallbackHandler ConnectionCallback
+                    = [Callback, LoginStateRes](csp::multiplayer::ErrorCode ErrCode)
+                {
+                    if (ErrCode != csp::multiplayer::ErrorCode::None)
+                    {
+                        // It would be extremely strange to hit this branch, but it remains here just in case.
+                        CSP_LOG_ERROR_FORMAT("Unexpected error connecting MultiplayerConnection. This is strange! : %s",
+                            csp::multiplayer::ErrorCodeToString(ErrCode).c_str());
+                        Callback(LoginStateRes);
+                        return;
```
MAG-ChristopherAtkinson, line 467: [Nit] these can both be removed as it will trigger the callback after the conditional anyway

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -391,6 +391,66 @@ void UserSystem::LoginAsGuest(
...
+                Callback(LoginStateRes);
+            }
+        };
+
+        csp::services::ResponseHandlerPtr ResponseHandler
+            = AuthenticationAPI->CreateHandler<LoginStateResultCallback, LoginStateResult, csp::common::LoginState, chs_user::AuthDto>(
+                LoginStateResCallback, &CurrentLoginState);
+
+        // Despite the naming, "login-guest" is the deferred, optimized, non-standard guest login.
+        // The regular login endpoint that "loginAsGuest" uses is the "real" one.
```
MAG-ChristopherAtkinson, line 488: [Comment] not a fan of the naming here either, does not really capture intended use, thank you for providing a more adapt method name and explicit comments

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -391,6 +391,66 @@ void UserSystem::LoginAsGuest(
...
+        static_cast<chs_user::AuthenticationApi*>(AuthenticationAPI)->usersLoginGuestPost(Request, ResponseHandler);
+    }
+    else
+    {
+        csp::systems::LoginStateResult BadResult;
+        BadResult.SetResult(csp::systems::EResultCode::Failed, (uint16_t)csp::web::EResponseCodes::ResponseBadRequest);
+        Callback(BadResult);
```
MAG-ChristopherAtkinson, line 495: [Consider] I know we do this is a few places, but wonder if we should use the request body to provide additional information

MAG-ChristopherAtkinson (changes requested):

Library/src/Systems/Users/UserSystem.cpp
MAG-ElliotMorris: I tried to address this upstream as best I could. 🤷 

https://github.com/magnopus/Magnopus.Services/pull/1121

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ElliotMorris: I agree. If you can think of any valid tests let me know. Not really sure what to do, tried to extract some postconditions from MCS but couldn't get anything. "This is a special case just for this project" vibes were strong.

Library/src/Systems/Users/UserSystem.cpp
MAG-ElliotMorris: I guess. I think there's an argument for both forms really, and since this is the pattern from the existing functions, going to keep it.

Library/src/Systems/Users/UserSystem.cpp
MAG-ElliotMorris: I don't disagree, but I'm not going to do that here, would imply changing every login function, but more importantly, I can't see a trivial way to add decent information.

I am on the fence with it though, it's a fair point. Frankly I'm short on energy right this second so am going to resolve.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-ElliotMorris: Updated every occurrence in this file. I'm assuming Cloud is correct now, I've not heard "Connected Services" around. An old name?

[b6803ea](https://github.com/magnopus-opensource/connected-spaces-platform/pull/789/commits/b6803ea5164b46cce2e9f51c220b03fc40cd2397)

MAG-ChristopherAtkinson: Not sure, likely a coupling of connected spaces platform and magnopus cloud services; likely why we normally MCS

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ChristopherAtkinson: Yeah I am in the same boat here, not sure I can think of any worth while test.

PR #790, opened 2025-09-11
[OF-1759] fix!: Workaround not having heap allocatable strings in wrapper generator

MAG-ElliotMorris:
Fixes an issue with large checkpoint files (ie, files larger than the stack size, common on web as they have smaller stacks, but a problem everywhere) causing stack overflows when constructing scene description objects.

Does this by using a container type, which is heap allocated. This is unfortunate, and not at all the best API. Nonetheless, changing the wrapper generator to allow heap allocated strings is very fundamental, and despite my best efforts, I couldn't do it in reasonable time/risk.

**Wasm Example**
```js
  // Get the json string and construct the scene representation objects
  const jsonString = await getCheckpointFileJsonString();
  const jsonStringList = Common.List.ofString();
  jsonStringList.append(jsonString); // Append into a list to deal with annoying wrapper gen issue.

  let sceneData = Systems.CSPSceneData.create_sceneDescriptionJson(jsonStringList);
  let sceneDescription = Multiplayer.CSPSceneDescription.create_sceneDescriptionJson(jsonStringList);
```
_This snippet taken from a wasm test introduced in this PR. See below for the full example_

**Breaking Changes**
- Constructors for `CspSceneDescription` and `CSPSceneData` now take `List<String>` rather than `String` The split into this list is arbitrary, you may split by character or by token.

### DIFF
--- a/Library/include/CSP/Multiplayer/CSPSceneDescription.h
+++ b/Library/include/CSP/Multiplayer/CSPSceneDescription.h
@@ -28,8 +28,10 @@ class CSP_API CSPSceneDescription
 {
 public:
     /// @brief Constructor for CSPSceneDescription by deserializing a SceneDescription json file.
-    /// The specific packing of the JSON string into this list is not specific, you may pack by character
-    /// or by token, so long as when naively concatenated, the original string is reproduced.
+    /// @param SceneDescriptionJson csp::common::String : The SceneDescription to parse.
+    /// @warning The expression of this interface as a list is a wrapper generator workaround, so
+    /// whilst you may split your string into many elements, it is not advisable. You should prefer
+    /// inserting your entire string as the first, single, and only element of the list.
     CSPSceneDescription(const csp::common::List<csp::common::String>& SceneDescriptionJson);
 
     CSPSceneDescription() { }
--- a/Library/include/CSP/Systems/CSPSceneData.h
+++ b/Library/include/CSP/Systems/CSPSceneData.h
@@ -31,9 +31,10 @@ class CSP_API CSPSceneData
 {
 public:
     /// @brief Constructor for CSPSceneData by deserializing a SceneDescription json file.
-    /// @param SceneDescriptionJson csp::common::List<csp::common::String> : The SceneDescription JSON.
-    /// The specific packing of the JSON string into this list is not specific, you may pack by character
-    /// or by token, so long as when naively concatenated, the original string is reproduced.
+    /// @param SceneDescriptionJson csp::common::String : The SceneDescription to parse.
+    /// @warning The expression of this interface as a list is a wrapper generator workaround, so
+    /// whilst you may split your string into many elements, it is not advisable. You should prefer
+    /// inserting your entire string as the first, single, and only element of the list.
     CSPSceneData(const csp::common::List<csp::common::String>& SceneDescriptionJson);
 
     /// @brief The space the scene data represents.
--- a/Tests/src/InternalTests/SceneDescriptionTests.cpp
+++ b/Tests/src/InternalTests/SceneDescriptionTests.cpp
@@ -454,6 +454,8 @@ CSP_INTERNAL_TEST(CSPEngine, SceneDescriptionTests, SceneDescriptionDeserializeB
 // The same test as above, but test that when we split the input, everything still works
 // The interface that forces us to pass a split array rather than a string is a wrapper gen constraint
 // rather than the true form of the, but lets still test it
+// WARNING, this isn't the best way to be doing this, just put the whole string as the first element of the list,
+// we support this behaviour because we are forced by the wrapper gen to provide this non-optimal interface expression.
 CSP_INTERNAL_TEST(CSPEngine, SceneDescriptionTests, SceneDescriptionDeserializeBasicSplitInputTest)
 {
     InitialiseFoundationWithUserAgentInfo(EndpointBaseURI());
--- a/WasmTesting/html_tests/EnterSpaceFromCheckpoint.html
+++ b/WasmTesting/html_tests/EnterSpaceFromCheckpoint.html
@@ -34,10 +34,12 @@
           return await response.text(); 
         }
 
-        // Split the json string and construct the scene representation objects
+        // Get the json string and construct the scene representation objects
         const jsonString = await loadJsonString();
         const jsonStringList = Common.List.ofString()
-        for (const ch of jsonString){ jsonStringList.append(ch); }
+
+        // Workaround for wrapper generator, this needs to be passed as a list, despite only being one element.
+        jsonStringList.append(jsonString)
 
         let sceneData = Systems.CSPSceneData.create_sceneDescriptionJson(jsonStringList);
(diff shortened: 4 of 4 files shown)


WasmTesting/html_tests/EnterSpaceFromCheckpoint.html
```diff
@@ -0,0 +1,102 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Cross-Origin-Opener-Policy" content="same-origin">
+    <meta http-equiv="Cross-Origin-Embedder-Policy" content="require-corp">
+    <meta charset="utf-8">
+    <title>CSP WASM Browser Test, EnterSpace</title>
```
MAG-ElliotMorris: [fix] Bad name

WasmTesting/html_tests/EnterSpaceFromCheckpoint.html
```diff
@@ -0,0 +1,102 @@
...
+          const response = await fetch('./shared/testdata/checkpoint-large.json');  
+          if (!response.ok) {
+            throw new Error(`Failed to load file: ${response.status}`);
+          }
+          return await response.text(); 
+        }
+
+        // Split the json string and construct the scene representation objects
+        const jsonString = await loadJsonString();
+        const jsonStringList = Common.List.ofString()
```
MAG-ElliotMorris, line 39: [Consider] Do you need to explicitly delete this?

MAG-ElliotMorris: [f85980c](https://github.com/magnopus-opensource/connected-spaces-platform/pull/790/commits/f85980c0f45c452602535c798fb7bfb4e5c3774d)

WasmTesting/html_tests/EnterSpaceFromCheckpoint.html
MAG-ElliotMorris: [ad87891](https://github.com/magnopus-opensource/connected-spaces-platform/pull/790/commits/ad87891eb22f029312f90be158670f3ba22d18f1)

Library/include/CSP/Systems/CSPSceneData.h
```diff
@@ -31,8 +31,10 @@ class CSP_API CSPSceneData
 {
 public:
     /// @brief Constructor for CSPSceneData by deserializing a SceneDescription json file.
-    /// @param SceneDescriptionJson csp::common::String : The SceneDescription to deserialize.
-    CSPSceneData(const csp::common::String& SceneDescriptionJson);
+    /// @param SceneDescriptionJson csp::common::List<csp::common::String> : The SceneDescription JSON.
```
MAG-AdamThorn: [Consider] Adding the rationale for this ctor taking a list here as well.

Tests/src/InternalTests/SceneDescriptionTests.cpp
```diff
@@ -296,10 +296,10 @@ CSP_INTERNAL_TEST(CSPEngine, SceneDescriptionTests, SceneDescriptionDeserializeE
 
     csp::multiplayer::OfflineRealtimeEngine RealtimeEngine(LogSystem, ScriptRunner);
 
-    CSPSceneDescription SceneDescription { Json.c_str() };
+    CSPSceneDescription SceneDescription { csp::common::List<csp::common::String> { Json.c_str() } };
```
MAG-AdamThorn, line 299: [fix] Add a test that passes a list split by character and/or token to check the list unpacking.

MAG-ElliotMorris: Aye. The wasm test does exercise this, but a cpp one can't hurt.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/790/commits/3c03437294c9d999b290fb78c87cd7ba62e74d7e

Library/include/CSP/Systems/CSPSceneData.h
MAG-ElliotMorris: I'm not sure about this. Is the fact that this API is this weird way because of the wrapper gen a concern for the user, thinking of the user more broadly than just folks who work at magnopus? Seems like irrelevant context that doesn't help in actually using the method. I did consider this but ultimately decided against making it public docs.

Will change it if you still think it'd be good to add after reading this, was on the fence already.

MAG-AdamThorn: I was imagining a user armed with a single string representing their scenedescription seeing this and being confused as to why it takes a list. I too am on the fence though, so happy if you elect to leave it as is.

MAG-ElliotMorris: 🤔 Still struggling over this one, but think I'll leave it on reflection.

The thing that convinced me is that you can extend this thinking to many of our interfaces. Why do things take pointers to arrays sometimes ... why do we return raw pointers and not smart pointers, the answer is always "wrapper generator" but it seems wrong to call it out on a method by method basis.

Library/include/CSP/Multiplayer/CSPSceneDescription.h
```diff
@@ -28,8 +28,9 @@ class CSP_API CSPSceneDescription
 {
 public:
     /// @brief Constructor for CSPSceneDescription by deserializing a SceneDescription json file.
-    /// @param SceneDescriptionJson csp::common::String : The SceneDescription to deserialize.
-    CSPSceneDescription(const csp::common::String& SceneDescriptionJson);
+    /// The specific packing of the JSON string into this list is not specific, you may pack by character
```
MAG-mv: [Fix] @param doc has been removed

Library/include/CSP/Multiplayer/CSPSceneDescription.h
```diff
@@ -28,8 +28,9 @@ class CSP_API CSPSceneDescription
...
+    /// or by token, so long as when naively concatenated, the original string is reproduced.
+    CSPSceneDescription(const csp::common::List<csp::common::String>& SceneDescriptionJson);
```
MAG-mv, line 35: [Fix] I think it's worth mentioning the reason this is a list, for visibility. Also, I think we should recommend putting the entire string in the first element. I imagine it would be slow if we were making a separate string allocation for each character or token. These shouldn't be treated as individual characters

Library/include/CSP/Multiplayer/CSPSceneDescription.h
MAG-ElliotMorris: [29914d3](https://github.com/magnopus-opensource/connected-spaces-platform/pull/790/commits/29914d3e36f476c31ddbbb6b4b10a481195318ad)

Library/include/CSP/Multiplayer/CSPSceneDescription.h
MAG-ElliotMorris: [29914d3](https://github.com/magnopus-opensource/connected-spaces-platform/pull/790/commits/29914d3e36f476c31ddbbb6b4b10a481195318ad)

PR #791, opened 2025-09-12
[OF-1739] test: Add batched analytics events

MAG-AdamThorn:
Added two new methods to the AnalyticsSystem; `BatchAnalyticsEvent` and `FlushAnalyticsEvents`.

The `BatchAnalyticsEvent` method makes it possible to add an analytic event to a queue to be sent to MCS later as part of a batch.
The queue will be processed when one of the following conditions is met:
1. The time since the last batch was sent reaches the batch rate (default 60 seconds).
2. The number of events in the queue reaches the maximum batch size (default 25 events).
3. The client application calls FlushAnalyticsEvents() as part of their log out or shut down procedure to force the batch to be sent.

The `FlushAnalyticsEvents` method will trigger immediate dispatch of the Analytics Records queue to MCS. This method is designed to be called as part of a client applications log out and shutdown procedures to ensure that any queued analytics records are flushed and sent to MCS before the user is logged out or the application is shut down.

### DIFF
--- a/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
+++ b/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
@@ -75,13 +75,6 @@ class CSP_API AnalyticsSystem : public SystemBase
 
 public:
     /**
-     * @brief Sets a callback to be executed when a queue of Analytics Records has been sent to the backend services.
-     * This callback will be executed in response both to a call to QueueAnalyticsEvent() as well as to FlushAnalyticsEventsQueue().
-     * @param Callback NullResultCallback : the callback to execute.
-     */
-    CSP_EVENT void SetQueueAnalyticsEventCallback(NullResultCallback Callback);
-
-    /**
      * @brief Constructs an Analytics Record which is added to a queue to be sent to the backend services in a single batch.
      * @details The queue will be sent when one of the following conditions are met:
      * 1. The time since the last batch was sent reaches the AnalyticsQueueSendRate (default 60 seconds).
@@ -146,9 +139,10 @@ public:
      * @brief Trigger immediate dispatch of the Analytics Records queue to the backend services.
      * @note This method should be called as part of client log out or shut down procedure to ensure that any queued analytics records are flushed and
      * sent to the backend services before the user is logged out or the application is shut down.
+     * @param Callback NullResultCallback : the callback to execute on completion of the flush operation.
      * @pre The user must be logged in to send an Analytics Record to the backend services.
      */
-    void FlushAnalyticsEventsQueue();
+    CSP_EVENT void FlushAnalyticsEventsQueue(NullResultCallback Callback);
 
     /**
      * @brief Retrieves the time since the queue was last sent.
@@ -190,8 +184,6 @@ private:
     std::unique_ptr<class AnalyticsQueueEventHandler> EventHandler;
     std::mutex AnalyticsQueueLock;
     std::vector<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>> AnalyticsRecordQueue;
-
-    NullResultCallback SendAnalyticsEventQueueCallback;
 
     const csp::ClientUserAgent* UserAgentInfo;
     std::chrono::milliseconds AnalyticsQueueSendRate;
--- a/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
+++ b/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
@@ -79,19 +79,19 @@ public:
     /**
      * @brief Constructs an Analytics Record which is added to a queue to be sent to MCS in a single batch.
      * @details The batch will be sent when one of the following conditions are met:
-     * 1) The time since the last batch was sent reaches the batch rate (default 60 seconds).
-     * 2) The number of events in the queue reaches the maximum batch size (default 25 events).
-     * 3) The client application calls FlushAnalyticsEvents() as part of their log out or shut down procedure to force the batch to be sent.
-     * For more information about FlushAnalyticsEvents(), see the method documentation @ref AnalyticsSystem::FlushAnalyticsEvents().
+     * 1. The time since the last batch was sent reaches the batch rate (default 60 seconds).
+     * 2. The number of events in the queue reaches the maximum batch size (default 25 events).
+     * 3. The client application calls FlushAnalyticsEvents() as part of their log out or shut down procedure to force the batch to be sent.
+     * For more information about flushing events see the method documentation @ref AnalyticsSystem::FlushAnalyticsEvents().
      *
      * Example: Consider the following user action that is to be captured as an analytics event:
-     *      A [web client] user [clicks] on a [menu] item in the [UI].
+     * A [web client] user [clicks] on a [menu] item in the [UI].
      *
      * In this example:
-     *      [web client] is captured internally.
-     *      [clicks] is the InteractionType.
-     *      [menu] is the Category.
-     *      [UI] is the ProductContextSection.
+     * - [web client] is captured internally.
+     * - [clicks] is the InteractionType.
+     * - [menu] is the Category.
+     * - [UI] is the ProductContextSection.
      *
      * @note The following data is captured internally and included in the analytics record: tenant name, client sku, client version and client build
      * environment.
@@ -102,7 +102,7 @@ public:
      * @param InteractionType const csp::common::String& : The interaction that occurred.
      * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field.
      * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata.
-     * @note Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
+     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
      * relevant device specs.
      */
     void BatchAnalyticsEvent(const csp::common::String& ProductContextSection, const csp::common::String& Category,
@@ -111,12 +111,10 @@ public:
 
     /**
      * @brief Constructs an Analytics Record which is immediately sent to MCS.
-     * @note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
+     * @note The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
      * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
      *
-     * For more information about how the Analytics Record is constructed, see the documentation for @ref AnalyticsSystem::BatchAnalyticsEvent(const
-     csp::common::String&, const csp::common::String&, const csp::common::String&, const csp::common::Optional<csp::common::String>&, const
-     csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>&).
+     * For more information about how the Analytics Record is constructed, see the documentation for @ref AnalyticsSystem::BatchAnalyticsEvent().
      *
      * @pre The user must be logged in to send an Analytics Record to MCS.
      * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context.
@@ -124,7 +122,7 @@ public:
      * @param InteractionType const csp::common::String& : The interaction that occurred.
      * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field.
      * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata.
-     * @note Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
+     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
      * relevant device specs.
      */
     void SendAnalyticsEvent(const csp::common::String& ProductContextSection, const csp::common::String& Category,
--- a/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
+++ b/Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
@@ -23,7 +23,6 @@
 #include "CSP/Systems/SystemBase.h"
 
 #include <chrono>
-#include <deque>
 #include <mutex>
 
 CSP_START_IGNORE
@@ -58,8 +57,8 @@ namespace csp::systems
 {
 
 /// @ingroup Analytics System
-/// @brief Public facing system that allows AnalyticsRecords to be sent to MCS.
-/// @invariant Users must be logged in to send AnalyticsRecords to MCS.
+/// @brief Public facing system that allows AnalyticsRecords to be sent to the backend services.
+/// @invariant Users must be logged in to send AnalyticsRecords to the backend services.
 class CSP_API AnalyticsSystem : public SystemBase
 {
     CSP_START_IGNORE
@@ -76,22 +75,23 @@ class CSP_API AnalyticsSystem : public SystemBase
 
 public:
     /**
-     * @brief Sets a callback to be executed when a queue of Ananlytics Records has been sent to backend services.
-     * This callback will be executed in response to both a call to QueueAnalyticsEvent() as well as FlushAnalyticsEventsQueue().
+     * @brief Sets a callback to be executed when a queue of Analytics Records has been sent to the backend services.
+     * This callback will be executed in response both to a call to QueueAnalyticsEvent() as well as to FlushAnalyticsEventsQueue().
      * @param Callback NullResultCallback : the callback to execute.
      */
     CSP_EVENT void SetQueueAnalyticsEventCallback(NullResultCallback Callback);
 
     /**
-     * @brief Constructs an Analytics Record which is added to a queue to be sent to MCS in a single batch.
-     * @details The batch will be sent when one of the following conditions are met:
-     * 1. The time since the last batch was sent reaches the batch rate (default 60 seconds).
-     * 2. The number of events in the queue reaches the maximum batch size (default 25 events).
-     * 3. The client application calls FlushAnalyticsEventsQueue() as part of their log out or shut down procedure to force the batch to be sent.
-     * For more information about flushing events see the method documentation @ref AnalyticsSystem::FlushAnalyticsEventsQueue().
+     * @brief Constructs an Analytics Record which is added to a queue to be sent to the backend services in a single batch.
+     * @details The queue will be sent when one of the following conditions are met:
+     * 1. The time since the last batch was sent reaches the AnalyticsQueueSendRate (default 60 seconds).
+     * 2. The number of events in the queue reaches the MaxQueueSize threshhold (default 25 events).
+     * 3. The client application calls FlushAnalyticsEventsQueue(). Clients should call this as part of their log out or shut down procedure to force
+     * the queue to be sent. For more information about flushing events see the method documentation @ref
+     * AnalyticsSystem::FlushAnalyticsEventsQueue().
      *
      * Example: Consider the following user action that is to be captured as an analytics event:
-     * A [web client] user [clicks] on a [menu] item in the [UI].
+     * - A [web client] user [clicks] on a [menu] item in the [UI].
      *
      * In this example:
      * - [web client] is captured internally.
@@ -99,16 +99,19 @@ public:
      * - [menu] is the Category.
      * - [UI] is the ProductContextSection.
      *
-     * @note The following data is captured internally and included in the analytics record: tenant name, client sku, client version and client build
-     * environment.
+     * @note The following data is captured internally and included in the analytics record:
+     * - tenant name, client sku, client version and client build environment.
      *
-     * @pre The user must be logged in to send Analytics Records to MCS.
-     * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context.
-     * @param Category const csp::common::String& : Categorization field.
-     * @param InteractionType const csp::common::String& : The interaction that occurred.
-     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field.
+     * @pre The user must be logged in to send Analytics Records to the backend services.
+     * @param ProductContextSection const csp::common::String& : The specific, high-level functional area or context within the product where the
+     * event occurred. This field acts as a primary identifier for the part of the application or system the user is interacting with.
+     * @param Category const csp::common::String& : Categorization field which acts as a namespace for the InteractionType. It provides a means of
+     * grouping similar events, which makes it easier to analyze and filter analytics data.
+     * @param InteractionType const csp::common::String& : Describes the precise and specific interaction that is being tracked. This field identifies
+     * what the user did or what happened within the product at a specific moment in time.
+     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field to provide additional context if required.
      * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata.
-     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
+     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region, as well as
      * relevant device specs.
      */
     void QueueAnalyticsEvent(const csp::common::String& ProductContextSection, const csp::common::String& Category,
@@ -116,41 +119,45 @@ public:
         const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& Metadata);
 
     /**
-     * @brief Constructs an Analytics Record which is immediately sent to MCS.
-     * @note The QueueAnalyticsEvent method should be used by default as it will batch events before sending them.
-     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * @brief Constructs an Analytics Record which is immediately sent to the backend services.
+     * @note The QueueAnalyticsEvent method should be used by default as it will queue events before sending them. This method will immediately send
+     * the analytics event and should therefore only be used when this behaviour is required.
      *
      * For more information about how the Analytics Record is constructed, see the documentation for @ref AnalyticsSystem::QueueAnalyticsEvent().
      *
-     * @pre The user must be logged in to send an Analytics Record to MCS.
-     * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context.
-     * @param Category const csp::common::String& : Categorization field.
-     * @param InteractionType const csp::common::String& : The interaction that occurred.
-     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field.
+     * @pre The user must be logged in to send Analytics Records to the backend services.
+     * @param ProductContextSection const csp::common::String& : The specific, high-level functional area or context within the product where the
+     * event occurred. This field acts as a primary identifier for the part of the application or system the user is interacting with.
+     * @param Category const csp::common::String& : Categorization field which acts as a namespace for the InteractionType. It provides a means of
+     * grouping similar events, which makes it easier to analyze and filter analytics data.
+     * @param InteractionType const csp::common::String& : Describes the precise and specific interaction that is being tracked. This field identifies
+     * what the user did or what happened within the product at a specific moment in time.
+     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field to provide additional context if required.
      * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata.
-     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
+     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region, as well as
      * relevant device specs.
+     * @param Callback NullResultCallback : the callback to execute on completion of the send operation.
      */
     void SendAnalyticsEvent(const csp::common::String& ProductContextSection, const csp::common::String& Category,
         const csp::common::String& InteractionType, const csp::common::Optional<csp::common::String>& SubCategory,
         const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& Metadata, NullResultCallback Callback);
 
     /**
-     * @brief Trigger immediate dispatch of the Analytics Records queue to MCS.
-     * @note This method is designed to be called as part of a client applications log out and shutdown procedures to ensure that any queued analytics
-     * records are flushed and sent to MCS before the user is logged out or the application is shut down.
-     * @pre The user must be logged in to send an Analytics Record to MCS.
+     * @brief Trigger immediate dispatch of the Analytics Records queue to the backend services.
+     * @note This method should be called as part of client log out or shut down procedure to ensure that any queued analytics records are flushed and
+     * sent to the backend services before the user is logged out or the application is shut down.
+     * @pre The user must be logged in to send an Analytics Record to the backend services.
      */
     void FlushAnalyticsEventsQueue();
 
     /**
-     * @brief Retrieves the time the last batch was sent.
+     * @brief Retrieves the time since the queue was last sent.
      * @return std::chrono::milliseconds : time since epoch in milliseconds.
      */
-    CSP_NO_EXPORT std::chrono::milliseconds GetTimeOfLastBatchSend() const { return TimeOfLastBatchSend; }
+    CSP_NO_EXPORT std::chrono::milliseconds GetTimeSinceLastQueueSend() const { return TimeSinceLastQueueSend; }
 
     /**
-     * @brief Retrieves the rate at which the queued Analytics Records are sent as a batch.
+     * @brief Retrieves the rate at which the queued Analytics Records are sent.
      * @return std::chrono::milliseconds : send rate in milliseconds.
      */
     CSP_NO_EXPORT std::chrono::milliseconds GetQueueSendRate() const { return AnalyticsQueueSendRate; }
@@ -159,12 +166,12 @@ public:
      * @brief Retrieves the current size of the Analytics Records queue.
      * @return size_t : the queue size.
      */
-    CSP_NO_EXPORT size_t GetCurrentQueueSize() const { return AnalyticsRecordBatch.size(); }
+    CSP_NO_EXPORT size_t GetCurrentQueueSize() const { return AnalyticsRecordQueue.size(); }
 
     /**
      * @brief Retrieves the max permitted size of the Analytics Records queue.
-     * If the queue size reaches this value, the queue will be sent as a batch to the backend services.
-     * @return size_t : the queue size at which a bacth will be sent.
+     * If the queue size reaches this value, the queue will be sent as a single batch to the backend services.
+     * @return size_t : the queue size at which a batch will be sent.
      */
     CSP_NO_EXPORT size_t GetMaxQueueSize() const { return MaxQueueSize; }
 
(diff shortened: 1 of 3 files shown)


Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+{
+class AnalyticsRecord;
+}
+
 namespace csp::systems
 {
 
 /// @ingroup Analytics System
 /// @brief Public facing system that allows AnalyticsRecords to be sent to MCS.
+/// @invariant Users must be logged in to send AnalyticsRecords to MCS.
```
MAG-ElliotMorris: [comment] I didn't know doxygen had an `invariant` tag. neat, and I think a pretty appropriate usage of it.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
 public:
     /**
-     * @brief Sends an analytics event to the Analytics service.
-     * @details Please note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
-     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * @brief Constructs an Analytics Record which is added to a queue to be sent to MCS in a single batch.
+     * @details The batch will be sent when one of the following conditions are met:
+     * 1. The time since the last batch was sent reaches the batch rate (default 60 seconds).
+     * 2. The number of events in the queue reaches the maximum batch size (default 25 events).
+     * 3. The client application calls FlushAnalyticsEvents() as part of their log out or shut down procedure to force the batch to be sent.
```
MAG-ElliotMorris: [consider] Technically this is incorrect, even if clients call `Flush` _not_ in their log out of shut down procedure, the batch will still be sent. Maybe reframe this as :

> The client application calls FlushAnalyticsEvents(). Clients should call this as part of their log out or shut down procedure to force the batch to be sent.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+     * - [clicks] is the InteractionType.
+     * - [menu] is the Category.
+     * - [UI] is the ProductContextSection.
      *
      * @note The following data is captured internally and included in the analytics record: tenant name, client sku, client version and client build
      * environment.
      *
+     * @pre The user must be logged in to send Analytics Records to MCS.
+     * @param ProductContextSection const csp::common::String& : Section of the client or runtime-context.
+     * @param Category const csp::common::String& : Categorization field.
```
MAG-ElliotMorris: [Consider] We don't explain what a category is. You can infer it from the example, but tbh I'm still unsure.

 I can see this is probably tricky, because I can't really think of what i'd write here, from the example i'd almost rather call this `InteractionLocation` or something. `InteractionElementOfInterest` ... if I just had the example to go off.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+     * @param InteractionType const csp::common::String& : The interaction that occurred.
+     * @param SubCategory const csp::common::Optional<csp::common::String>& : Optional sub-category field.
+     * @param Metadata const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& : Optional analytics event metadata.
+     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region as well as
```
MAG-ElliotMorris: [nit] Oxford comma.
```suggestion
     * Metadata is the event payload. It may be used to store such information as the space the user is in, their geographical region, as well as
```

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+#endif
     /** @endcond */
     CSP_END_IGNORE
 
 public:
     /**
-     * @brief Sends an analytics event to the Analytics service.
-     * @details Please note: The BatchAnalyticsEvent method should be used by default as it will batch events before sending them.
-     * This method will immediately send the analytics event and should therefore only be used when this behaviour is required.
+     * @brief Constructs an Analytics Record which is added to a queue to be sent to MCS in a single batch.
```
MAG-ElliotMorris: [Consider] I might frame this as 
```suggestion
     * @brief Constructs an Analytics Record which is added to a queue to be sent to the backend services in a single batch.
```

Spoken to people about this MCS backend thing recently, and I think our stance should be :
- Make no formal effort right now to support non MCS services.
- Don't make it literally impossible to use non MCS services.

That is, unless the analytics system is _specifically_ Magnopus MCS only.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+
+    // This is a utility function to allow us to test the batching functionality in a reasonable time frame.
+    void SetBatchRateAndSize(std::chrono::milliseconds NewBatchRate, size_t NewBatchSize);
+
+    class AnalyticsBatchEventHandler* EventHandler;
+    std::recursive_mutex* AnalyticsBatchLock;
+
     std::unique_ptr<csp::services::ApiBase> AnalyticsApi;
 
+    std::deque<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>>* AnalyticsRecordBatch;
```
MAG-ElliotMorris: [consider]
```suggestion
    std::deque<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>>* AnalyticsRecordQueue;
```

Just more accurate to what this is, makes it a bit clearer to see why we're doing array operations on this thing in the .cpp

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+     */
+    void FlushAnalyticsEvents();
+
 private:
     AnalyticsSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
     CSP_NO_EXPORT AnalyticsSystem(csp::web::WebClient* InWebClient, const csp::ClientUserAgent* AgentInfo, common::LogSystem& LogSystem);
     ~AnalyticsSystem();
 
+    // Method called on Tick. This will send batched Analytics Records if the batch rate or size have been met.
+    void ProcessBatchedAnalyticsRecords();
```
MAG-ElliotMorris: [fix] This is exploiting friendship, the type doesn't make sense without the event handler. Make this public and NO_EXPORT imo.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+
+    std::chrono::milliseconds GetTimeOfLastBatch() const;
```
MAG-ElliotMorris: [Question] Does a private getter make sense? I'm unsure, one could argue you should just access the data directly. Willing to admit there's arguments on both sides though.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+    void SetTimeOfLastBatch(std::chrono::milliseconds NewTimeOfLastBatch);
+
+    // This is a utility function to allow us to test the batching functionality in a reasonable time frame.
+    void SetBatchRateAndSize(std::chrono::milliseconds NewBatchRate, size_t NewBatchSize);
```
MAG-ElliotMorris: [comment] We should maybe consider eventually having a standard at some point to mark these sorts of functions, like `__SetBatchRateAndSize` or something.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+    std::recursive_mutex* AnalyticsBatchLock;
+
     std::unique_ptr<csp::services::ApiBase> AnalyticsApi;
 
+    std::deque<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>>* AnalyticsRecordBatch;
     const csp::ClientUserAgent* UserAgentInfo;
+    std::chrono::milliseconds AnalyticsBatchRate;
+    std::chrono::milliseconds TimeOfLastBatch;
+    size_t MaxBatchSize;
+    bool IsFlushingAnalyticsEvents;
```
MAG-ElliotMorris: [Question] Is `IsFlushingAnalyticsEvents` used for anything? I can't spot it. If not, [fix] via removal.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
+
+    class AnalyticsBatchEventHandler* EventHandler;
+    std::recursive_mutex* AnalyticsBatchLock;
+
     std::unique_ptr<csp::services::ApiBase> AnalyticsApi;
 
+    std::deque<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>>* AnalyticsRecordBatch;
     const csp::ClientUserAgent* UserAgentInfo;
+    std::chrono::milliseconds AnalyticsBatchRate;
+    std::chrono::milliseconds TimeOfLastBatch;
```
MAG-ElliotMorris: [consider] 
There's three potential things this could mean, the time the last batch was created, the time it was last updated, or the time it was last sent.

```suggestion
    std::chrono::milliseconds TimeOfLastBatchSend;
```

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -89,6 +67,103 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    , MaxBatchSize(25)
+    , IsFlushingAnalyticsEvents(false)
+{
+}
+
+AnalyticsSystem::AnalyticsSystem(csp::web::WebClient* InWebClient, const csp::ClientUserAgent* AgentInfo, csp::common::LogSystem& LogSystem)
+    : SystemBase(InWebClient, nullptr, &LogSystem)
+    , EventHandler(new AnalyticsBatchEventHandler(this))
+    , AnalyticsBatchLock(new std::recursive_mutex)
+    , AnalyticsRecordBatch(new(std::deque<std::shared_ptr<chs::AnalyticsRecord>>))
```
MAG-ElliotMorris: [fix] I don't believe any of these need to be heap allocated.

If some do, perhaps some or all could at least be unique_ptrs.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -89,6 +67,103 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    , UserAgentInfo(AgentInfo)
+    , AnalyticsBatchRate(std::chrono::seconds(60))
+    , TimeOfLastBatch(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()))
+    , MaxBatchSize(25)
+    , IsFlushingAnalyticsEvents(false)
+{
+    AnalyticsApi = std::unique_ptr<csp::services::ApiBase>(new chs::AnalyticsApi(InWebClient));
```
MAG-ElliotMorris: [Consider] Construct this in the initializer list.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
     static_cast<chs::AnalyticsApi*>(AnalyticsApi.get())->analyticsBulkPost(Records, ResponseHandler);
 }
 
+void AnalyticsSystem::BatchAnalyticsEvent(const String& ProductContextSection, const String& Category, const String& InteractionType,
```
MAG-ElliotMorris: [Question] What do you think about the term batch, in general for this api? It's servicable, but I'm wondering if `Queue` is a better term, as it's slightly more accurate. This method could be `QueueAnalyticsEvent`. Not sure.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    const Optional<String>& SubCategory, const Optional<Map<String, String>>& Metadata)
+{
+    if (ProductContextSection.IsEmpty() || Category.IsEmpty() || InteractionType.IsEmpty())
+    {
+        CSP_LOG_MSG(common::LogLevel::Error, "Missing the required fields for the Analytics Event.");
```
MAG-ElliotMorris: [Fix] State which fields are required.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    {
+        FlushAnalyticsEvents();
+        return;
+    }
+
+    if (AnalyticsRecordBatch->size() >= MaxBatchSize)
+    {
+        FlushAnalyticsEvents();
+        return;
+    }
```
MAG-ElliotMorris: [nit] Slightly fragile under change with the returns. Maybe an `else if` or an `||` statement to make it a single case.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+}
+
+void AnalyticsSystem::ProcessBatchedAnalyticsRecords()
+{
+    std::scoped_lock AnalyticsBatchLocker(*AnalyticsBatchLock);
+
+    if (AnalyticsRecordBatch->empty())
+    {
+        return;
+    }
```
MAG-ElliotMorris: [Consider] I think this introduces a fair bit of pernicious complexity. We do this lock, and the check for empty, later on in `Flush`. It's like there's 2 points of synchronisation, but this one is redundant. I'd just remove this.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    const std::chrono::milliseconds CurrentTime
+        = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
+
+    if (CurrentTime - GetTimeOfLastBatch() >= AnalyticsBatchRate)
+    {
+        FlushAnalyticsEvents();
+        return;
+    }
+
+    if (AnalyticsRecordBatch->size() >= MaxBatchSize)
```
MAG-ElliotMorris: [Consider] If you exposed the method to get the time of last batch and the batch size, you could remove this entire function and do the implementation entirely in the tick handler.

Sort of a philosophical api design question, I can understand wanting to encapsulate, but on the other hand it means all the logic concerned with responding to ticks is seperated to the object that is _about_ that, it makes the actual analytics system agnostic to whether its run in a tick-context or not.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -122,4 +197,92 @@ void AnalyticsSystem::SendAnalyticsEvent(const String& ProductContextSection, co
...
+    Records.reserve(AnalyticsRecordBatch->size());
+
+    while (AnalyticsRecordBatch->empty() == false)
+    {
+        auto Record = AnalyticsRecordBatch->front();
+
+        Records.push_back(Record);
+
+        AnalyticsRecordBatch->pop_front();
+    }
```
MAG-ElliotMorris: [Consider] This makes me wonder why we're using a `deque` at all. If we used a `vector`, you could just use/copy it directly without having to do this loop.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -106,3 +111,276 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, SendAnalyticsEventMissingFields
...
+
+    // Log in
+    LogInAsNewTestUser(UserSystem, UserId);
+
+    // Ensure the MockLogger will ignore all logs except the one we care about
+    EXPECT_CALL(MockLogger.MockLogCallback, Call(::testing::_)).Times(::testing::AnyNumber());
+
+    // We can confirm that the batch has been successfully sent by checking for this success log message
+    const csp::common::String BatchAnalyticsSuccessMsg = "Batched Analytics Events sent.";
+    EXPECT_CALL(MockLogger.MockLogCallback, Call(BatchAnalyticsSuccessMsg)).Times(1);
```
MAG-ElliotMorris: [Question] You think there will ever be any situation where a client needs to react to a batch being successfully sent or not? If so, this isn't good enough, we'll want to expose an error callback.

If not though, then it's fine.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -106,3 +111,276 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, SendAnalyticsEventMissingFields
...
+    // We also check that the batch has been sent as a result of flush rather than batch rate by checking that
+    // the elapsed test time is less than the batch rate of 20 seconds
+    const csp::common::String BatchAnalyticsSuccessMsg = "Batched Analytics Events sent.";
+    EXPECT_CALL(MockLogger.MockLogCallback, Call(BatchAnalyticsSuccessMsg))
+        .Times(1)
+        .WillOnce(::testing::Invoke(
+            [&](const csp::common::String&)
+            {
+                LogTime = std::chrono::steady_clock::now();
+                long long ElapsedTestTime = std::chrono::duration_cast<std::chrono::seconds>(LogTime - SendBatchStart).count();
```
MAG-ElliotMorris: [nit] Use sized types when possible. `long` should be banished to hades for how many problems it can cause.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -106,3 +111,276 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, SendAnalyticsEventMissingFields
...
+    const auto TestCategory = String("Event_Category");
+    const auto TestInteractionType = String("Event_InteractionType");
+    const auto TestSubCategory = String("Event_SubCategory");
+    const Map<String, String> TestMetadata = { { "Key1", "Value1" }, { "Key2", "Value2" } };
+
+    // Send two analytics events to be queued for sending later as a batch
+    AnalyticsSystem->BatchAnalyticsEvent(TestProductContextSection, TestCategory, TestInteractionType, TestSubCategory, TestMetadata);
+    AnalyticsSystem->BatchAnalyticsEvent(TestProductContextSection, TestCategory, TestInteractionType, TestSubCategory, TestMetadata);
+
+    // The batch rate has been set to 20 seconds so we need to tick foundation for at least that long to allow the batch to be sent
```
MAG-ElliotMorris: [Question] Confused on this. This test calls FlushAnalyticsEvent manually, so isn't this irrelevent?

... or is this more, "So long as we don't tick for more than 20 seconds, then we wont have sent a batch so we can test flush"?

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -106,3 +111,276 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, SendAnalyticsEventMissingFields
...
+        TestTime = std::chrono::duration_cast<std::chrono::seconds>(Current - Start).count();
+
+        if (TestTime == 3)
+        {
+            // Flush the batch of events after 3 seconds
+            AnalyticsSystem->FlushAnalyticsEvents();
+        }
+
+        csp::CSPFoundation::Tick();
+    }
```
MAG-ElliotMorris: [Fix] This seems quite fragile, assuming that 2 will always come after 1 when dealing with system clocks isn't something i'd be comfortable with. If CSP is descheduled by the OS for a second at the right time this will break.

I believe you can rewrite this linearly, if you change to using a future as the notifier rather than a bool.

```
std::this_thread::sleep_for(3s); //Is this delay even neccesary?
csp::CSPFoundation::Tick();
ASSERT_NEQ(BatchFlushedFuture.wait_for(0s), std::future_status::ready) << "analytics batch should not yet have sent";
AnalyticsSystem->FlushAnalyticsEvents();
ASSERT_EQ(BatchFlushedFuture.wait_for(5s), std::future_status::ready) << "analytics batch should have sent";
// Assert system clock here if you want to validate this wasn't sent from a time-trigger, 
// although a real callback, or even just different logging depending on what sends the batch,
// could also tell you that in a more direct way.
```

This again makes me think about [whether a proper callback is required](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791#discussion_r2348651360), although I can see this _is_ just for testing, and using the log to drive the behaviour is sort of clever, albeit fragile.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +49,119 @@ namespace csp::web
...
         const csp::common::String& InteractionType, const csp::common::Optional<csp::common::String>& SubCategory,
         const csp::common::Optional<csp::common::Map<csp::common::String, csp::common::String>>& Metadata, NullResultCallback Callback);
 
+    /**
+     * @brief Trigger immediate dispatch of the Analytics Records queue to MCS.
+     * @note This method is designed to be called as part of a client applications log out and shutdown procedures to ensure that any queued analytics
+     * records are flushed and sent to MCS before the user is logged out or the application is shut down.
+     * @pre The user must be logged in to send an Analytics Record to MCS.
+     */
+    void FlushAnalyticsEvents();
```
MAG-ElliotMorris: [fix] Coming back to this after doing the whole review. I've convinced myself that this at least needs a callback.

Consider that a client must delay shutting down CSP for the course of this network send/response if they are going to call this in shutdown. The natural and simple way to do this will be `await FlushAnalyticsEvents()`, which you won't be able to do without a notifying callback.

MAG-AdamThorn: I went back-and-forth on this. My initial thinking was that clients will not be able to do anything in response to a callback telling them the batch of analytics failed to send.
However you are right, with the introduction of the flush method and the way in which clients will use it, means a callback is likely now required.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed.

Fixed here as one of many documentation changes: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed. I have added further detail for all of the method arguments. Hopefully this is clearer.

Fixed here as one of many documentation changes: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed.

Fixed here as one of many documentation changes: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Excellent point. I have updated all occurrences of MCS with backend services

Fixed here as one of many documentation changes: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed. I have replaced nearly all usage of the word batch with queue.

This specific instance is addressed here: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed. I have made this change as part of a broader refactor.

Change here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4).

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed. I have made this change as part of a broader refactor.

Change here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4).

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: I agree. I liked your idea of prefixing the method name with two underscores.

Fix as part of: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: It is not.

Removed: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Agreed. I have renamed most of the members and corresponding methods.

Fix: [14ee13](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Excellent suggestion. I have made them all unique ptrs.

Fix: [14ee13](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Fix: [14ee13](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Agreed. I have removed almost all occurrences of batch now and replaced with queue.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Fix as part of change here: [14ee13d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/14ee13d68a6e64d84b6539fb18c88acacd1c9036#diff-911c841fb91ee658f4333c3eb57c7ded53407ae59c0106494deecc11de7af7b3)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Fixed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Agreed. Fixed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: This was a great suggestion, thank you.

Fixed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Agreed 100%: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Introduced a callback as part of the refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Done as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: The idea of the test was to ensure that the batch was being sent as a result of the call to flush, rather than due to the batch rate. The 3 second delay was arbitrary.

This has all been addressed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: Agreed, addressed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Addressed as part of refactor here: [f68026f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/f68026f6c3be201cea81d936490fb94339e0d6a4#diff-66381206ad90eecd78fb1bc8fed0c320df08a4959ae2a9bc7ecd977e6e1a513d)

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
```diff
@@ -37,57 +48,155 @@ namespace csp::web
...
 
+    void SetTimeSinceLastQueueSend(std::chrono::milliseconds NewTimeSinceLastQueueSend);
+
+    // This is a utility function to allow us to test the queueing functionality in a reasonable time frame.
+    void __SetQueueSendRateAndMaxSize(std::chrono::milliseconds NewSendRate, size_t NewQueueSize);
+
     std::unique_ptr<csp::services::ApiBase> AnalyticsApi;
 
+    std::unique_ptr<class AnalyticsQueueEventHandler> EventHandler;
+    std::recursive_mutex AnalyticsQueueLock;
```
MAG-ChristopherAtkinson: [Question] Does this need to be a recursive_mutex? feels like this should be a std::mutex in the context

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-ElliotMorris: Lets mention this at the next sync then, double underscore is a pretty standard way of saying "Not for external use"

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-ElliotMorris: [Question] Even so though, you could merely have them as direct members, rather than pointers _at all_. Are you trying to limit includes in the header?

MAG-ChristopherAtkinson (approved):
A good amount of feedback has already been address, the inline documentation is amazing great use of pre conditions and other tags.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-ElliotMorris: [consider] By doing it this way, our wrapper gen won't generate methods that can be `await`'d, which is probably especially annoying for `Flush`

However, I understand that this lets you know when a timed event has been sent, which is nice. So I can see why the system-wide callback has been chosen. Will leave this comment, but both options work.

MAG-AdamThorn: I had considered the above. Currently the Flush method can be called both directly by the user, and internally as part of the tick.

If the Flush method takes a callback, then as things stand, we would also need to add one as an argument to the QueueAnalyticsEvent method (which is responsible for adding events to the queue). The issue with that is that multiple systems may be calling QueueAnalyticsEvent, but only the last one to call (and pass a callback) would be alerted to the fact that all events have been sent.

What I could do is separate the flush method so that it is not called internally by QueueAnalyticsEvent. That would mean that the flush method could take a callback, and the system-wide callback would only be used for queued callbacks sent as a result of the batch rate or size limits being reach.

I like that Idea, but will think on it and make any changes in the morning.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Apologies, I misspoke above. EventHandler is a unique_ptr, whereas AnalyticsQueueLock and AnalyticsRecordQueue are now direct class members.

MAG-ElliotMorris: 👍 (consequence of me being lazy in the larger diffs, didn't actually look)

MAG-ElliotMorris (approved):
This should have been at approve, none of my extent comments are blockers.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-ElliotMorris: 🤔 Yeah, I see the complexity. Perhaps a system wide one is best ... , it's certainly a cleaner interface when you don't consider the translation to `awaitable` languages. I'm not sure what the pattern ends up being in those languages, but it's definately doable because we do it elsewhere.

Library/include/CSP/Systems/Analytics/AnalyticsSystem.h
MAG-AdamThorn: Ah yes excellent spot.

Updated: [38a5553](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/38a55539cabece8a67df7c45f72d7a5b44b19878)

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
MAG-AdamThorn: So ultimately I elected to remove the system-wide callback and to make FlushAnalyticsEventsQueue awaitable.

I did play around with keeping the system-wide callback and passing that to the FlushAnalyticsEventsQueue method when called on Tick, but ultimately decided that there was little value in a callback related to the queue being processed based on rate or size.

Update: [44b864b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/791/commits/44b864b8c7ad9b937ed22c6434118495110b9030)

PR #792, opened 2025-09-16
[OF-1761] feat: Added single-point property registration

MAG-mv:
Adds a single-point replicated property registration, meaning we can define all properties that should be replicated in one place, without modifying serialization/deserialization logic every time we add new properties.

I also found a bug in one of our tests, which was causing a failure due to not de-initializing CSP.

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -411,7 +411,8 @@ public:
     // The state patcher. This is the object that handles dirty/pending properties,
     // another way of thinking about this is the "network patch manager" or something like that.
     // If this is null, then the space entity does immediate updates without any deferred patching.
-    CSP_NO_EXPORT const std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher();
+    CSP_NO_EXPORT const std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher() const;
+    CSP_NO_EXPORT std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher();
 
     /// @brief Update after the property of a component was changed
     /// @param DirtyComponent ComponentBase* : the dirty component to update
@@ -424,23 +425,23 @@ public:
     /// @brief Sets the internal ParentId to nullptr
     CSP_NO_EXPORT void RemoveParentId();
 
-    // Direct setters that bypass any patching behaviour or conditionals
-    // Non-ideal, here because the responsibility for managing network behaviour is massively munged into SpaceEntity, so it's doing two things at
-    // once. At least with these, we can move the logic out to another type and quarantine the complexity.
-    // These should really be the only things SpaceEntity has. Neither the network management, nor really the locking, should matter to what is
-    // essentially a data-type.
-    CSP_NO_EXPORT void SetNameDirect(const csp::common::String& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetPositionDirect(const csp::common::Vector3& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetRotationDirect(const csp::common::Vector4& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetScaleDirect(const csp::common::Vector3& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetThirdPartyRefDirect(const csp::common::String& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetThirdPartyPlatformDirect(const csp::systems::EThirdPartyPlatform Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetEntityLockDirect(LockType Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetSelectedIdDirect(uint64_t Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback);
+    // Direct setters that bypass any patching behaviour or conditionals.
+    // SetPropertyDirect allows us to set all of our replicated property values, without the need for individual setters.
+    // We still have to handle ParentId separately, as this is a required MCS object property, and not an MCS component
+    // like the rest of our properties.
+    // We also have to handle CSP components separately, as CSP currently replicates them by using the whole component as a data container, preventing
+    // us from buffering updated state in a patch as you'd expect, as we can't copy whole components. This manifests especially in
+    // `UpdateComponentDirect` where the update sequencing happens too early, and is in many ways a bug.
+    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool AddComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool UpdateComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool RemoveComponentDirect(uint16_t ComponentKey, bool CallNotifyingCallback = false);
+
+    CSP_START_IGNORE
+    template <typename P, typename V>
+    void SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback = false);
+    CSP_END_IGNORE
+
     /// @brief Setter for the owner ID
     /// @param InOwnerId uint64_t : the owner ID to set
     CSP_NO_EXPORT void SetOwnerId(const uint64_t InOwnerId);
@@ -515,5 +516,23 @@ private:
     /// @return bool : the selection state of the entity
     bool InternalSetSelectionStateOfEntity(const bool SelectedState);
 };
+
+CSP_START_IGNORE
+template <typename P, typename V>
+void SpaceEntity::SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback)
+{
+    std::scoped_lock PropertiesLocker(PropertiesLock);
+
+    // We cast the value to the property type to get around issues with type <-> ReplicatedValue conversions,
+    // as ReplicatedValues can only hold specific types.
+    // This is quite brittle, so we are finding a better way to handle this.
+    Property = static_cast<P>(Value);
+    if (CallNotifyingCallback && EntityUpdateCallback)
+    {
+        csp::common::Array<ComponentUpdateInfo> Empty;
+        EntityUpdateCallback(this, Flag, Empty);
+    }
+}
+CSP_END_IGNORE
 
 } // namespace csp::multiplayer
--- a/Library/src/Multiplayer/SpaceEntity.cpp
+++ b/Library/src/Multiplayer/SpaceEntity.cpp
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 #include "CSP/Multiplayer/SpaceEntity.h"
-
 #include "CSP/Common/Interfaces/IRealtimeEngine.h"
 #include "CSP/Common/StringFormat.h"
 #include "CSP/Common/Systems/Log/LogSystem.h"
@@ -49,6 +48,7 @@
 #include "CSP/Multiplayer/Script/EntityScript.h"
 #include "Multiplayer/MCS/MCSTypes.h"
 #include "Multiplayer/MCSComponentPacker.h"
+#include "Multiplayer/PatchUtils.h"
 #include "Multiplayer/Script/EntityScriptBinding.h"
 #include "Multiplayer/Script/EntityScriptInterface.h"
 #include "Multiplayer/SpaceEntityKeys.h"
@@ -176,25 +176,7 @@ const csp::common::String& SpaceEntity::GetName() const { return Name; }
 
 bool SpaceEntity::SetName(const csp::common::String& Value)
 {
-    if (!IsModifiable())
-    {
-        if (LogSystem != nullptr)
-        {
-            LogSystem->LogMsg(csp::common::LogLevel::Error,
-                fmt::format("Entity is not modifiable, you can only modify entities that have transferable ownership, or which you already are the "
-                            "owner of. Entity name: {}",
-                    Name)
-                    .c_str());
-        }
-        return false;
-    }
-
-    return StatePatcher != nullptr ? StatePatcher->SetDirtyProperty(SpaceEntityComponentKey::Name, GetName(), Value)
-                                   : [this, &Value]()
-    {
-        SetNameDirect(Value, true);
-        return true;
-    }();
+    return SetProperty(*this, Name, Value, SpaceEntityComponentKey::Name, UPDATE_FLAGS_NAME, LogSystem);
 }
 
 const SpaceTransform& SpaceEntity::GetTransform() const { return Transform; }
@@ -230,25 +212,7 @@ csp::common::Vector3 SpaceEntity::GetGlobalPosition() const
 
 bool SpaceEntity::SetPosition(const csp::common::Vector3& Value)
 {
-    if (!IsModifiable())
-    {
-        if (LogSystem != nullptr)
-        {
-            LogSystem->LogMsg(csp::common::LogLevel::Error,
-                fmt::format("Entity is not modifiable, you can only modify entities that have transferable ownership, or which you already are the "
-                            "owner of. Entity name: {}",
-                    Name)
-                    .c_str());
-        }
-        return false;
-    }
-
-    return StatePatcher != nullptr ? StatePatcher->SetDirtyProperty(SpaceEntityComponentKey::Position, GetPosition(), Value)
-                                   : [this, &Value]()
-    {
-        SetPositionDirect(Value, true);
-        return true;
-    }();
+    return SetProperty(*this, Transform.Position, Value, SpaceEntityComponentKey::Position, UPDATE_FLAGS_POSITION, LogSystem);
 }
 
 const csp::common::Vector4& SpaceEntity::GetRotation() const { return Transform.Rotation; }
@@ -270,25 +234,7 @@ csp::common::Vector4 SpaceEntity::GetGlobalRotation() const
 
 bool SpaceEntity::SetRotation(const csp::common::Vector4& Value)
 {
-    if (!IsModifiable())
-    {
-        if (LogSystem != nullptr)
-        {
-            LogSystem->LogMsg(csp::common::LogLevel::Error,
-                fmt::format("Entity is not modifiable, you can only modify entities that have transferable ownership, or which you already are the "
-                            "owner of. Entity name: {}",
-                    Name)
-                    .c_str());
-        }
-        return false;
-    }
-
-    return StatePatcher != nullptr ? StatePatcher->SetDirtyProperty(SpaceEntityComponentKey::Rotation, GetRotation(), Value)
-                                   : [this, &Value]()
-    {
-        SetRotationDirect(Value, true);
-        return true;
-    }();
+    return SetProperty(*this, Transform.Rotation, Value, SpaceEntityComponentKey::Rotation, UPDATE_FLAGS_ROTATION, LogSystem);
 }
 
 const csp::common::Vector3& SpaceEntity::GetScale() const { return Transform.Scale; }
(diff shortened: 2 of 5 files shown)


Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -356,55 +325,48 @@ SpaceEntity* SpaceEntityStatePatcher::NewFromObjectMessage(const mcs::ObjectMess
...
 
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_ENTITYNAME, Name);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_POSITION, Transform.Position);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_ROTATION, Transform.Rotation);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_SCALE, Transform.Scale);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_SELECTEDCLIENTID, SelectedId);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_THIRDPARTYPLATFORM, ThirdPartyPlatform);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_THIRDPARTYREF, ThirdPartyRef);
-        ComponentUnpacker.TryReadValue(COMPONENT_KEY_VIEW_LOCKTYPE, EntityLock);
+        // It's unfortunate we have to break the usual pattern of getting the registered properties from the state patcher here,
```
MAG-mv, line 343: This case made me consider just having the entity returning the properties in all cases. But this is a bit of an odd ball, as this functions scope extends past just patching, . However, I  understand why this function exists here

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -450,12 +451,17 @@ class CSP_API SpaceEntity
     CSP_NO_EXPORT ComponentUpdateInfo AddComponentFromItemComponentDataPatch(
         uint16_t ComponentId, const csp::multiplayer::mcs::ItemComponentData& ComponentData);
 
+    // Creates an array of entity properties which should be replicated
+    CSP_NO_EXPORT csp::common::Array<EntityProperty> CreateProperties();
```
MAG-ElliotMorris: [nit]

```suggestion
    // Creates the array of entity properties which should be replicated.
    CSP_NO_EXPORT csp::common::Array<EntityProperty> CreateReplicatedProperties();
```

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1082,6 +1087,8 @@ ComponentBase* SpaceEntity::FindFirstComponentOfType(ComponentType FindType) con
 
 void SpaceEntity::AddChildEntity(SpaceEntity* ChildEntity) { ChildEntities.Append(ChildEntity); }
 
+void SpaceEntity::RegisterReplicatedProperties() { StatePatcher->RegisterProperties(CreateProperties()); }
```
MAG-ElliotMorris: [fix] This method seems redundant, just do this directly in the constructor. It's only ever used once, and it means there's less stuff on SpaceEntity that is about replication.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1256,6 +1263,55 @@ ComponentUpdateInfo SpaceEntity::AddComponentFromItemComponentDataPatch(uint16_t
...
 }
 
+csp::common::Array<EntityProperty> SpaceEntity::CreateProperties()
+{
+    /* clang-format off */
+    return { 
+        { 
+            COMPONENT_KEY_VIEW_ENTITYNAME, UPDATE_FLAGS_NAME, 
+            [&Name = Name]() { return csp::common::ReplicatedValue { Name }; },
+            [this](const csp::common::ReplicatedValue& Value) { SetNameDirect(Value.GetString()); }
```
MAG-ElliotMorris: [Consider] Keep in mind that capturing like this means that these `EntityProperty` objects are only valid so long as the SpaceEntity in question remains at the same pointer address. Think about documenting that. Any function that uses this method is not thread safe due to this.

I'm trying to think if there's a way to do this that wouldn't be vulnerable to that... would probably involve shared pointers to bake in lifetime extension.

Yeah, like
```
class SpaceEntity
{
  ///@brief Builds an array of property objects intended for replication. Lifetime of Entity is extended to the lifetime of EntityProperty
  static csp::common::Array<EntityProperty> ReplicatedPropertiesFromEntity(std::shared_ptr<SpaceEntity> Entity);
}
```

Then you capture the entity shared_ptr by value in the callbacks, extending the lifetime formally. That would also force folk to think about this when calling which is a good thing. This is the primary category of bug that Rust is designed to fix :P. That sort of lifetime relationship is built into method/type sigs.

**I doubt you can move to this interface until we make SpaceEntity's actually live inside owned objects, which is a damn shame.**

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -82,6 +81,27 @@ std::pair<SpaceEntityUpdateFlags, csp::common::Array<ComponentUpdateInfo>> Space
 
     auto UpdateFlags = static_cast<SpaceEntityUpdateFlags>(0);
 
+    for (const auto& DirtyProperty : DirtyProperties)
```
MAG-ElliotMorris, line 83: [Question] This is the main property update code right? Any particular reason it's moved above the component processing? (Not sure if it matters or not)

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -41,6 +42,37 @@ namespace csp::multiplayer
...
+        , ToReplicatedValue { ToReplicatedValue }
+        , FromReplicatedValue { FromReplicatedValue }
+    {
+    }
+
+    void Set(const csp::common::ReplicatedValue& RepValue);
+    csp::common::ReplicatedValue Get() const;
+
+    uint16_t GetKey() const { return Key; }
+    SpaceEntityUpdateFlags GetUpdateFlag() const { return UpdateFlag; }
```
MAG-ElliotMorris, line 84: [Fix] Explain what these do. What's the key for? What data does Set and Get address? What's an update flag?

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -41,6 +42,37 @@ namespace csp::multiplayer
...
+        std::function<void(const csp::common::ReplicatedValue&)> FromReplicatedValue)
+        : Key { Key }
+        , UpdateFlag { UpdateFlag }
+        , ToReplicatedValue { ToReplicatedValue }
+        , FromReplicatedValue { FromReplicatedValue }
+    {
+    }
+
+    void Set(const csp::common::ReplicatedValue& RepValue);
+    csp::common::ReplicatedValue Get() const;
```
MAG-ElliotMorris, line 75: [nit] Almost seems like redundant indirection. You could just expose reference to the `To/From` function objects

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -41,6 +42,37 @@ namespace csp::multiplayer
 class SpaceEntity;
 class ComponentBase;
 
+/*
+    Object used to define information needed for a property to be replicated.
+    This creates a single-point property registration to remove the need to update replication logic in multiple locations.
```
MAG-ElliotMorris, line 48: [Consider] Can you mention how it manages getting/setting data, and how that/this object is related to a "source" SpaceEntity.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -82,6 +81,27 @@ std::pair<SpaceEntityUpdateFlags, csp::common::Array<ComponentUpdateInfo>> Space
...
+    {
+        const uint16_t PropertyKey = DirtyProperty.first;
+        auto Prop = RegisteredProperties.find(PropertyKey);
+
+        if (Prop != RegisteredProperties.end())
+        {
+            UpdateFlags = static_cast<SpaceEntityUpdateFlags>(UpdateFlags | Prop->second.GetUpdateFlag());
+            Prop->second.Set(DirtyProperty.second);
+        }
```
MAG-ElliotMorris: [Fix] By compressing this code, you've also made it harder to read. Some explanatory comments will help. (Also think about taking `Prop->second` and putting it into a named variable, that, combined with the use of auto, makes it somewhat impenetrable.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -356,55 +325,48 @@ SpaceEntity* SpaceEntityStatePatcher::NewFromObjectMessage(const mcs::ObjectMess
...
-    csp::common::String ThirdPartyRef = "";
-    LockType EntityLock = LockType::None;
-
     auto MessageComponents = Message.GetComponents();
 
     std::vector<std::pair<uint16_t, mcs::ItemComponentData>> ComponentsToAdd;
 
+    csp::multiplayer::SpaceEntity* NewEntity = new csp::multiplayer::SpaceEntity(
+        &RealtimeEngine, ScriptRunner, &LogSystem, Type, Id, "", SpaceTransform {}, OwnerId, ParentId, IsTransferable, IsPersistent);
+
```
MAG-ElliotMorris: [Consider] Now that there's a fair bit more code that might fail between here and the return, you might want to create this as a `unique_ptr` and then release at the end. Any throwing code that exists now or in the future between here and the return will leak otherwise.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -356,55 +325,48 @@ SpaceEntity* SpaceEntityStatePatcher::NewFromObjectMessage(const mcs::ObjectMess
...
+                {
+                    csp::common::ReplicatedValue Value;
+                    ComponentUnpacker.TryReadValue(ComponentDataPair.first, Value);
+                    Prop->Set(Value);
+                }
+                else
+                {
+                    LogSystem.LogMsg(csp::common::LogLevel::Error, "NewFromObjectMessage: No Property with the specified key found!");
+                }
+            }
```
MAG-ElliotMorris, line 375: [Fix] Please explain what this is doing, it's denser now so harder to grok. It's much easier to see a `//Found a property, now set it from the service data" to describe a branch, rather than having to understand the code

[Consider] Also pointing the reader to where this data is sourced, as they won't be able to see what the expected default propety set of a SpaceEntity is now, compared to the prior code, and they might want to know.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -135,6 +168,9 @@ class SpaceEntityStatePatcher
     SpaceEntityStatePatcher::PatchSentCallback GetEntityPatchSentCallback();
     void CallEntityPatchSentCallback(bool Success);
 
+    void RegisterProperty(const EntityProperty& Property);
+    void RegisterProperties(const csp::common::Array<EntityProperty>& Properties);
```
MAG-ElliotMorris, line 195: [fix] Explain what registration means.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -146,6 +182,8 @@ class SpaceEntityStatePatcher
     csp::common::List<uint16_t> TransientDeletionComponentIds;
     std::chrono::milliseconds TimeOfLastPatch;
 
+    std::unordered_map<uint16_t, EntityProperty> RegisteredProperties;
```
MAG-ElliotMorris: [consider] Explain what the key is, and what this container is expected to contain (I know prior code hasn't done this, but this is new)

MAG-ElliotMorris (changes requested):

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/8d4fb064f3fb5b6a35cf15a35e0db6176c96fb7f

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/a7b655cbda61efaab8ba0b0379adc06ce130de78

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: I did think about this. It's not something I'm currently not worried about, as this is only used in the state patcher, which will never exist longer than the SpaceEntity.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: I'd rather keep as-is as it makes the interfaces a big nicer to use.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: This is the order it's done in other functions, so I wanted to make it the same. Theres no reason this would cause issues.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/1e666afa594bc8f860a5af9a1ff401134d4a7410

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-AdamThorn: Some docs around usage of the RegisteredProperties collection would be much appreciated. I have had the PR open in two windows to make it easier to reason about what is going on.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/f1688096c836a3a62003c4d0636e862e1fe3aa7b

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -356,55 +325,48 @@ SpaceEntity* SpaceEntityStatePatcher::NewFromObjectMessage(const mcs::ObjectMess
...
-    // This stuff read from the component unpacker.
-    // The above is first-class, mandatory data for MCS objects. All else are optional components, hence the split.
-    csp::common::String Name = "";
-    SpaceTransform Transform {};
-    uint64_t SelectedId = 0;
-    csp::systems::EThirdPartyPlatform ThirdPartyPlatform = csp::systems::EThirdPartyPlatform::NONE;
-    csp::common::String ThirdPartyRef = "";
-    LockType EntityLock = LockType::None;
-
     auto MessageComponents = Message.GetComponents();
```
MAG-AdamThorn: [fix] Can we change this from an auto to the type. I had to go check ObjectMessage to see what this returned to help me parse the changes below.

MAG-AdamThorn (changes requested):

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/4a26489f76b778041f50f187138a957f156a20bd

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/9cb1d244d5509ac1097229fbf0097eb0615bf7e1

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/aebefa37a6d5f42d762d9671f9887616fa3ecf36

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/4680a0794634980fe6c8b6e456cec66440b509a7

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: I used this as a chance to refactor. Enjoy: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/5870ef595d1f5f837af1eb003ff44250b5ddec91

MAG-mv: Just realized this is going to make the rebase into my child branch not very fun...

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/050baf900cae623f5dbb39955db3e52f26d28051

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-ElliotMorris: When you force push, remember these links can get weird. The actual thing for this is https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/cf166773969d7fd40b69b00e6003a19f2b2c2063

This is important because I can't comment on the above link.

MAG-ElliotMorris: Love the refactor.

[nit]

I could make the argument that just because `SpaceEntityComponentKey` is docced, and `EntityProperty` is docced, dosen't mean

`std::unordered_map<SpaceEntityComponentKey, EntityProperty> RegisteredProperties;`

Is something that can be inherently understood. There's still the matter of why it exists in this type and how it's used.

What'd i'd write is something like :

```cpp
// Container of EntityProperties, which are proxy types that allow the get and set of a specific replicatable
// values on a SpaceEntity. Populated via RegisterProperty/RegisterProperties.
std::unordered_map<SpaceEntityComponentKey, EntityProperty> RegisteredProperties;
```

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -627,12 +627,14 @@ SpaceEntity* OnlineRealtimeEngine::CreateRemotelyRetrievedEntity(const signalr::
...
+    auto NewEntity = SpaceEntityStatePatcher::NewFromObjectMessage(Message, *this, *ScriptRunner, *LogSystem);
 
     std::scoped_lock EntitiesLocker(*EntitiesLock);
-    PendingAdds->emplace_back(NewEntity);
 
-    return NewEntity;
+    SpaceEntity* Entity = NewEntity.release();
+    PendingAdds->emplace_back(Entity);
+
+    return Entity;
```
MAG-ElliotMorris: [nit] The most minor of nits. Technically safer :

```cpp
PendingAdds->emplace_back(NewEntity.release())
return PendingAdds.back(); //Not sure the exact api
```
(I'd probably reject this comment myself as not important enough to bother with a fixup, more wanted to run it past you to see what you thought about this sort of thing)

MAG-ElliotMorris: Apparently since c++17 you get a [reference to the inserted element](https://en.cppreference.com/w/cpp/container/vector/emplace_back)

Which might make `return PendingAdds.emplace_back(NewEntity.release());` possible, which would be ideal.

MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/8f0f1d1026bff3492620a06d883cde7d1b3f6cd7

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/792/commits/8f0f1d1026bff3492620a06d883cde7d1b3f6cd7

PR #794, opened 2025-09-16
[OB-4678] Stop SpaceEntity::UpdateCallback firing wven when the value is the same for OfflineEngine.

MAG-mv:
I would advise reviewing [OF-1761](https://github.com/magnopus-opensource/connected-spaces-platform/pull/792) first, as this merges into that branch.

We were only checking if the values were the same in our StatePatcher::SetDirtyProperty, which isn't used on the offline version.
Instead of adding a conditional check in every setter, I created a generic, templated setter with this extra conditional in. This cuts out a lot of code from SpaceEntity.

[OF-1761]: https://magnopus.atlassian.net/browse/OF-1761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -429,7 +429,7 @@ public:
     // SetPropertyDirect allows us to set all of our replicated property values, without the need for individual setters.
     // We still have to handle ParentId separately, as this is a required MCS object property, and not an MCS component
     // like the rest of our properties.
-    // We also have to handle CSP components separately, as CSP currently replicates them using a different method.
+    // We also have to handle CSP components separately, as CSP currently replicates them by using the whole component as a data container, preventing us from buffering updated state in a patch as you'd expect, as we can't copy whole components. This manifests especially in `UpdateComponentDirect` where the update sequencing happens too early, and is in many ways a bug.
     CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool AddComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool UpdateComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -429,7 +429,9 @@ public:
     // SetPropertyDirect allows us to set all of our replicated property values, without the need for individual setters.
     // We still have to handle ParentId separately, as this is a required MCS object property, and not an MCS component
     // like the rest of our properties.
-    // We also have to handle CSP components separately, as CSP currently replicates them by using the whole component as a data container, preventing us from buffering updated state in a patch as you'd expect, as we can't copy whole components. This manifests especially in `UpdateComponentDirect` where the update sequencing happens too early, and is in many ways a bug.
+    // We also have to handle CSP components separately, as CSP currently replicates them by using the whole component as a data container, preventing
+    // us from buffering updated state in a patch as you'd expect, as we can't copy whole components. This manifests especially in
+    // `UpdateComponentDirect` where the update sequencing happens too early, and is in many ways a bug.
     CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool AddComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool UpdateComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
@@ -520,6 +522,10 @@ template <typename P, typename V>
 void SpaceEntity::SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback)
 {
     std::scoped_lock PropertiesLocker(PropertiesLock);
+
+    // We cast the value to the property type to get around issues with type <-> ReplicatedValue conversions,
+    // as ReplicatedValues can only hold specific types.
+    // This is quite brittle, so we are finding a better way to handle this.
     Property = static_cast<P>(Value);
     if (CallNotifyingCallback && EntityUpdateCallback)
     {
--- a/Library/include/CSP/Multiplayer/SpaceEntity.h
+++ b/Library/include/CSP/Multiplayer/SpaceEntity.h
@@ -411,7 +411,8 @@ public:
     // The state patcher. This is the object that handles dirty/pending properties,
     // another way of thinking about this is the "network patch manager" or something like that.
     // If this is null, then the space entity does immediate updates without any deferred patching.
-    CSP_NO_EXPORT const std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher();
+    CSP_NO_EXPORT const std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher() const;
+    CSP_NO_EXPORT std::unique_ptr<SpaceEntityStatePatcher>& GetStatePatcher();
 
     /// @brief Update after the property of a component was changed
     /// @param DirtyComponent ComponentBase* : the dirty component to update
@@ -424,23 +425,21 @@ public:
     /// @brief Sets the internal ParentId to nullptr
     CSP_NO_EXPORT void RemoveParentId();
 
-    // Direct setters that bypass any patching behaviour or conditionals
-    // Non-ideal, here because the responsibility for managing network behaviour is massively munged into SpaceEntity, so it's doing two things at
-    // once. At least with these, we can move the logic out to another type and quarantine the complexity.
-    // These should really be the only things SpaceEntity has. Neither the network management, nor really the locking, should matter to what is
-    // essentially a data-type.
-    CSP_NO_EXPORT void SetNameDirect(const csp::common::String& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetPositionDirect(const csp::common::Vector3& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetRotationDirect(const csp::common::Vector4& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetScaleDirect(const csp::common::Vector3& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetThirdPartyRefDirect(const csp::common::String& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetThirdPartyPlatformDirect(const csp::systems::EThirdPartyPlatform Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetEntityLockDirect(LockType Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetSelectedIdDirect(uint64_t Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback);
+    // Direct setters that bypass any patching behaviour or conditionals.
+    // SetPropertyDirect allows us to set all of our replicated property values, without the need for individual setters.
+    // We still have to handle ParentId separately, as this is a required MCS object property, and not an MCS component
+    // like the rest of our properties.
+    // We also have to handle CSP components separately, as CSP currently replicates them using a different method.
+    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool AddComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool UpdateComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool RemoveComponentDirect(uint16_t ComponentKey, bool CallNotifyingCallback = false);
+
+    CSP_START_IGNORE
+    template <typename P, typename V>
+    void SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback = false);
+    CSP_END_IGNORE
+
     /// @brief Setter for the owner ID
     /// @param InOwnerId uint64_t : the owner ID to set
     CSP_NO_EXPORT void SetOwnerId(const uint64_t InOwnerId);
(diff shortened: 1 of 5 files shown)


Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -424,23 +425,21 @@ class CSP_API SpaceEntity
...
-    CSP_NO_EXPORT void SetThirdPartyRefDirect(const csp::common::String& Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetThirdPartyPlatformDirect(const csp::systems::EThirdPartyPlatform Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetEntityLockDirect(LockType Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetSelectedIdDirect(uint64_t Value, bool CallNotifyingCallback = false);
-    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback);
+    // Direct setters that bypass any patching behaviour or conditionals.
+    // SetPropertyDirect allows us to set all of our replicated property values, without the need for individual setters.
+    // We still have to handle ParentId separately, as this is a required MCS object property, and not an MCS component
+    // like the rest of our properties.
+    // We also have to handle CSP components separately, as CSP currently replicates them using a different method.
```
MAG-ElliotMorris: [Consider]
```suggestion
    // We also have to handle CSP components separately, as CSP currently replicates them by using the whole component as a data container, preventing us from buffering updated state in a patch as you'd expect, as we can't copy whole components. This manifests especially in `UpdateComponentDirect` where the update sequencing happens too early, and is in many ways a bug.
```

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -424,23 +425,21 @@ class CSP_API SpaceEntity
...
+    CSP_NO_EXPORT void SetParentIdDirect(csp::common::Optional<uint64_t> Value, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool AddComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool UpdateComponentDirect(uint16_t ComponentKey, ComponentBase* Component, bool CallNotifyingCallback = false);
     CSP_NO_EXPORT bool RemoveComponentDirect(uint16_t ComponentKey, bool CallNotifyingCallback = false);
+
+    CSP_START_IGNORE
+    template <typename P, typename V>
+    void SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback = false);
+    CSP_END_IGNORE
```
MAG-ElliotMorris, line 443: [Question] Why are you declaring and defining this in different places? Seems redundant.

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -518,4 +517,18 @@ class CSP_API SpaceEntity
     bool InternalSetSelectionStateOfEntity(const bool SelectedState);
 };
 
+CSP_START_IGNORE
+template <typename P, typename V>
+void SpaceEntity::SetPropertyDirect(P& Property, const V& Value, SpaceEntityUpdateFlags Flag, bool CallNotifyingCallback)
+{
+    std::scoped_lock PropertiesLocker(PropertiesLock);
+    Property = static_cast<P>(Value);
```
MAG-ElliotMorris, line 529: [fix] Document why this cast is here, (it would seem obvious to someone just reading this code that you might as well just have a single template type, we should explain why there are two)

Library/include/CSP/Multiplayer/SpaceEntity.h
```diff
@@ -518,4 +517,18 @@ class CSP_API SpaceEntity
...
+    if (CallNotifyingCallback && EntityUpdateCallback)
+    {
+        csp::common::Array<ComponentUpdateInfo> Empty;
```
MAG-ElliotMorris, line 532: [Comment] I suppose this has to be extended when we centralise component updates through this same thing.

Library/src/Multiplayer/PatchUtils.h
```diff
@@ -0,0 +1,44 @@
...
+#include "CSP/Multiplayer/SpaceEntity.h"
+#include "SpaceEntityStatePatcher.h"
+
+namespace csp::multiplayer
+{
+// Helper function to generically set entity properties without duplicating a bunch of logic.
+// This needs to live in a private include, as it relies on other non-exported includes.
+template <typename P, typename V>
+static bool SetProperty(
+    SpaceEntity* Entity, P& Property, const V& Value, uint16_t Key, SpaceEntityUpdateFlags Flag, csp::common::LogSystem* LogSystem)
```
MAG-ElliotMorris: [Consider] Does `SpaceEntity&` communicate your intent better? Unless you expect `Entity` to be nullable or reseatable ... or some other pointer behavioural property, you should use refs.

Library/src/Multiplayer/PatchUtils.h
```diff
@@ -0,0 +1,44 @@
...
+                fmt::format("Entity is not modifiable, you can only modify entities that have transferable ownership, or which you already are the "
+                            "owner of. Entity name: {}",
+                    Entity->GetName())
+                    .c_str());
+        }
+
+        return false;
+    }
+
+    if (Entity->GetStatePatcher())
```
MAG-ElliotMorris: [Comment] This being a non OO method is cool, cause you can start to see how the State Patcher can be further decoupled from entity (ie provided as an argument to this method)

Remember I mentioned the idea of a local state patcher related via inheritance, rather than going through direct sets on the entity ... that's probably what this will end up being if it keeps evolving imo.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -230,25 +212,7 @@ csp::common::Vector3 SpaceEntity::GetGlobalPosition() const
...
-        return false;
-    }
-
-    return StatePatcher != nullptr ? StatePatcher->SetDirtyProperty(COMPONENT_KEY_VIEW_POSITION, GetPosition(), Value)
-                                   : [this, &Value]()
-    {
-        SetPositionDirect(Value, true);
-        return true;
-    }();
+    return SetProperty(this, Transform.Position, Value, COMPONENT_KEY_VIEW_POSITION, UPDATE_FLAGS_POSITION, LogSystem);
```
MAG-ElliotMorris: [Comment] Gosh, we're almost at `SetData(PROPERRTY_KEY)`, which would be great if we _really_ cared about performance cause we could start processing stuff in data-oriented-design like buffers. Not saying that should be the goal though ... you do start to lose interface clarity on that route.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -811,17 +681,16 @@ bool SpaceEntity::Unlock()
...
-    // old lock is the same before setting the val again.
-    const auto OldLockReplicatedVal = csp::common::ReplicatedValue(static_cast<int64_t>(LockType::UserAgnostic));
-    const auto NewLockReplicatedVal = csp::common::ReplicatedValue(static_cast<int64_t>(LockType::None));
-
-    return StatePatcher != nullptr ? StatePatcher->SetDirtyProperty(COMPONENT_KEY_VIEW_LOCKTYPE, OldLockReplicatedVal, NewLockReplicatedVal)
-                                   : [this]()
+    // We don't call "SetProperty" here, because the internal IsModifiable check will always fail due to the entity being locked.
+    if (StatePatcher)
+    {
+        return StatePatcher->SetDirtyProperty(COMPONENT_KEY_VIEW_LOCKTYPE, EntityLock, static_cast<int64_t>(LockType::None));
```
MAG-ElliotMorris: [Consider] You've lost the prior explanation about why we have to provide the old value as a directly constructed thing.

You can probably condense it, but it's worth mentioning still imo as it is a pattern break.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -1270,42 +1053,42 @@ csp::common::Array<EntityProperty> SpaceEntity::CreateProperties()
         { 
             COMPONENT_KEY_VIEW_ENTITYNAME, UPDATE_FLAGS_NAME, 
             [&Name = Name]() { return csp::common::ReplicatedValue { Name }; },
-            [this](const csp::common::ReplicatedValue& Value) { SetNameDirect(Value.GetString()); } 
+            [this](const csp::common::ReplicatedValue& Value) { SetPropertyDirect(Name, Value.GetString(), UPDATE_FLAGS_NAME); }
```
MAG-ElliotMorris: [Comment] Just a little bit away from

```
{
 MakeProperty(UPDATE_FLAGS_NAME, this->Name),
 MakeProperty(UPDATE_FLAGS_POSITION, this->Position)
 ...
}
```

Not saying we should go that route, might be too abstract, and I know we'll still have to solve the type casting thing (at this point I wonder if instead of dealing with it we should just solve it ... make our types match up root and stem.)

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -400,10 +400,33 @@ void SpaceEntityStatePatcher::ApplyPatchFromObjectPatch(const mcs::ObjectPatch&
         {
             if (ComponentDataPair.first < COMPONENT_KEY_END_COMPONENTS)
             {
-                ComponentUpdateInfo UpdateInfo
-                    = SpaceEntity.AddComponentFromItemComponentDataPatch(ComponentDataPair.first, ComponentDataPair.second);
-                ComponentUpdates[ComponentIndex] = UpdateInfo;
-                ComponentIndex++;
+                if (ComponentDataPair.first < COMPONENT_KEY_END_COMPONENTS)
+                {
```
MAG-ElliotMorris: [fix] You're doing this check twice, almost seems like a merge error.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
```diff
@@ -400,10 +400,33 @@ void SpaceEntityStatePatcher::ApplyPatchFromObjectPatch(const mcs::ObjectPatch&
...
+                    auto Prop = RegisteredProperties.find(ComponentDataPair.first);
+
+                    if (Prop != RegisteredProperties.end())
+                    {
+                        UpdateFlags = SpaceEntityUpdateFlags(UpdateFlags | Prop->second.GetUpdateFlag());
+
+                        csp::common::ReplicatedValue Value;
+                        ComponentUnpacker.TryReadValue(ComponentDataPair.first, Value);
+                        Prop->second.Set(Value);
+                    }
```
MAG-ElliotMorris: [fix] As usual, add a human-readable note saying what this branch is doing, as it's harder to know behind all the abstraction.

Library/src/Multiplayer/SpaceEntityStatePatcher.h
```diff
@@ -118,6 +118,8 @@ class SpaceEntityStatePatcher
     {
         std::scoped_lock<std::mutex> PropertiesLocker(DirtyPropertiesLock);
 
+        // This erase is useful for a very specific case where a value was changed, and before a patch is sent, the value is set back to its original
+        // value. This will prevent a redundant patch from being sent.
```
MAG-ElliotMorris: [Consider] Huh, at least, that's what we _think_. Think about adding a note to the end saying that this is supposition, and we're not _certain_ this is neccesary to do, or if it was designed this way intentionally.

Tests/src/PublicAPITests/RealtimeEngineEntityTests.cpp
```diff
@@ -1603,6 +1673,8 @@ INSTANTIATE_TEST_SUITE_P(RealtimeEngineEntityTests, EntityLockAddComponent,
     testing::Values(csp::common::RealtimeEngineType::Offline, csp::common::RealtimeEngineType::Online));
 INSTANTIATE_TEST_SUITE_P(RealtimeEngineEntityTests, EntityLockRemoveComponent,
     testing::Values(csp::common::RealtimeEngineType::Offline, csp::common::RealtimeEngineType::Online));
+INSTANTIATE_TEST_SUITE_P(RealtimeEngineEntityTests, EntityUpdateCallback,
+    testing::Values(csp::common::RealtimeEngineType::Offline, csp::common::RealtimeEngineType::Online));
```
MAG-ElliotMorris, line 1677: [Comment] Thanks for adding this, surprised it didn't already exist.

MAG-ElliotMorris (changes requested):

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-ElliotMorris: [Consider] I've noticed this sort of pattern exists a lot, I wonder if there's scope to make this a method in `PatchUtils` (or even just here in `StatePatcher`) Then the function name can serve as the call-site documentation.

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: Added your suggestion: https://github.com/magnopus-opensource/connected-spaces-platform/pull/794/commits/8df322d477520afdc8e2b8788d693d889981fe82

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: I prefer defining it below so we don't bloat the class declaration.

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: Yep. I have correctly rebased now

Library/src/Multiplayer/SpaceEntityStatePatcher.cpp
MAG-mv: This was a merge error and has been fixed in the parent PR

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: We will be able to do this once we sort out our type conversions to/from ReplicatedValues in a streamlined manner. We have solved it to a degree is our MCSComponentPacker using templates and assumptions about what certain types cast to, but it's currently too brittle to introduce here. I am actively thinking about this problem

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/794/commits/b4456ef54e3bd6f1d42169a1470283f0dfa32a01

Library/src/Multiplayer/PatchUtils.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/794/commits/2273e05945020bab54609c51e1f6027be65cd444

Library/src/Multiplayer/SpaceEntityStatePatcher.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/794/commits/2694f855ffd0a41db34275bd6dda61a7b0586fc5

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: In this change, I am checking the old value by passing "EntityLock". It's technically redundant, as we do it above, but it's less confusing than before

Library/include/CSP/Multiplayer/SpaceEntity.h
MAG-ElliotMorris: Fair enough.

Library/src/Multiplayer/SpaceEntity.cpp
MAG-ElliotMorris: Ah right I see, I wonder why I didn't do that initially ... i'm sure there was a reason? 🤔

MAG-ElliotMorris: Can't see any reason why this isn't correct now though...

PR #795, opened 2025-09-17
[NT-0] fix: Resolve initialization ordering issue

MAG-AdamThorn:
We had a circular dependency issue in the SystemsManager due to SignalR requiring the AuthContext from the UserSystem for construction. To get around this we pass nullptr to the UserSystem ctor for the NetworkEventBus, and then call a newly introduced  method to set it after the NetworkEventBus has been constructed.


MAG-ChristopherAtkinson (approved):
I would normally have concerns around dependency resolutions in delayed expectations such as this changes, however as this is all constructed within the initialisation flow it mitigates substantial risk, as a result these changes seem reasonable given the issue being fixed.

PR #796, opened 2025-09-17
[OF-1664] feat: mock the service layer

MAG-ChristopherAtkinson:
This PR included changes to allow for mocking of our services, as well as an example of a service being mocked using the new GMOCK abstractions. The Pr consists of the following changes:

- Services not inherited from a specific IServiceBase
- Parameters for each service function are now abstracted into a structure to avoid GMOCK param limitations 
- Mock API are generated for each service 
- Application setting test has been updated to use the new mock functionality alongside standard testing 
- JsonSerializer has been updated to allow for ApplicationSettings to populate Json for tests 

> [!IMPORTANT]
> There is a lot of noise in this PR due to the nature of the changes. I would recommend looking at the subset of the changes from the generation, and then focus on the use case [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/796/files#diff-61a8b1007bae175868cd544746ca98d0b28d046c4e97cec232aa06aa1ccb530c)

> [!NOTE]
> This PR requires changes to services generation; those changes can be found here: https://github.com/magnopus-opensource/csp-services/pull/17


Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -455,3 +459,285 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidTentantSett
         EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
     }
 }
+
+namespace chs = csp::services::generated::userservice;
+
+CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemMockTests, WhenApplicationSettingsPutResponseCreatedThenRecieveSuccessResponseTests)
+{
+    if (std::string(AdminAccountEmail()).empty())
```
MAG-AdamThorn: [fix] You mentioned removing these as not required when we ran through it in our call earlier.

MAG-ChristopherAtkinson: They have already been removed from mock tests, see: [45df99b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/796/commits/45df99be7cbde96fde303611ca86159bd7b48a2e)

MAG-AdamThorn: Apologies, I was looking at the wrong commit list

Library/src/Json/JsonSerializer.h
```diff
@@ -231,6 +233,21 @@ template <typename T> inline void JsonSerializer::SerializeValue(const csp::comm
     Writer.EndArray();
 }
 
+template <typename T> inline void JsonSerializer::SerializeValue(const csp::common::Map<csp::common::String, T>& Value)
+{
+    Writer.StartObject();
+
+    std::unique_ptr<common::Array<common::String>> Keys(const_cast<common::Array<common::String>*>(Value.Keys()));
```
MAG-ElliotMorris: [fix]

Two issues:
- I don't like to see `const_cast`, it should be avoided at all costs
- ~~This leaks because `Keys` allocates~~

It looks like you just want to iterate over the map, use `GetUnderlying` to get a std::map and use a range based for on that instead (or add iterators to the type itself)

Tests/src/PublicAPITests/ApplicationSettingsSystemTests.cpp
```diff
@@ -455,3 +459,284 @@ CSP_PUBLIC_TEST(CSPEngine, ApplicationSettingsSystemTests, GetInvalidTentantSett
...
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+    EXPECT_EQ(Result.GetHttpResultCode(), static_cast<uint16_t>(csp::web::EResponseCodes::ResponseNotFound));
+    EXPECT_EQ(Result.GetFailureReason(), csp::systems::ERequestFailureReason::None);
+
+    const auto ApplicationSettingsResult = Result.GetApplicationSettings();
+    EXPECT_EQ(ApplicationSettingsResult.ApplicationName, ApplicationSettings.ApplicationName);
+    EXPECT_EQ(ApplicationSettingsResult.Context, ApplicationSettings.Context);
+    EXPECT_EQ(ApplicationSettingsResult.AllowAnonymous, ApplicationSettings.AllowAnonymous);
+    EXPECT_EQ(ApplicationSettingsResult.Settings.Size(), ApplicationSettings.Settings.Size());
+}
```
MAG-ElliotMorris, line 742: [Comment] Nice looking tests. (Referring to the whole file)

Library/src/Json/JsonSerializer.h
MAG-ElliotMorris: Edit : Oh no, it's a unique ptr ... okay only one issue.

MAG-ChristopherAtkinson: I have still updated it, to resolve`const_cast`, please see here: [836b4e1](https://github.com/magnopus-opensource/connected-spaces-platform/pull/796/commits/836b4e16fca4e6203c300e68418266bb93e67ecb)

MAG-ChristopherAtkinson: Also, improve readability as a by-product

MAG-ElliotMorris: [fix] Needs to be a ref in the loop or you'll copy and potentially be very slow, as you don't know how big `T` is.

```for (const std::pair<csp::common::String, T>& Pair : Value.GetUnderlying())```

modules/csp-services
MAG-ElliotMorris, line 1: [fix] By @MAG-ChristopherAtkinson 's request, blocking this PR until underlying service wrapper PR has gone in.

MAG-ElliotMorris (changes requested):

Library/src/Json/JsonSerializer.h
MAG-ChristopherAtkinson: Apologies, missed this when I auto-pilot the change: [7d42575](https://github.com/magnopus-opensource/connected-spaces-platform/pull/796/commits/7d42575d21e229ce322af49593643308cf096be3)

PR #798, opened 2025-09-19
[OF-1511] Update ReplicatedValue to use a variant internally

MAG-mv:
This refactors our ReplicatedValue implementation use a std::variant instead of a union. This gives us a lot more power when using this type from within CSP, as we can use visit/compile-time type checking instead of runtime conditionals. An example of this exists in this PR when updating MCSComponentPacker::CreateItemComponentData to use std::visit. 

This change will also allow us to further streamline our ReplicatedValue conversions, as we now have access to templated setters and getters.

### DIFF
--- a/Library/include/CSP/Common/ReplicatedValue.h
+++ b/Library/include/CSP/Common/ReplicatedValue.h
@@ -21,10 +21,16 @@
 #include "CSP/Common/String.h"
 #include "CSP/Common/Vector.h"
 
-#include <functional>
+#include <variant>
 
 namespace csp::common
 {
+
+CSP_START_IGNORE
+class ReplicatedValue;
+using ReplicatedValueImplType = std::variant<bool, float, int64_t, csp::common::String, csp::common::Vector2, csp::common::Vector3,
+    csp::common::Vector4, csp::common::Map<csp::common::String, ReplicatedValue>>;
+CSP_END_IGNORE
 
 /// @brief Enum representing the type of a replicated value.
 /// These values are serialized and stored as integers.
@@ -48,9 +54,27 @@ class CSP_API ReplicatedValue
 {
 public:
     /// @brief A default ReplicatedValue will not have a valid type ("ReplicatedValueType::InvalidType"), and will have no internal value associated.
-    ///
-    /// Do not use this constructor unless you know what you are doing!
+    /// This constuctor will create the value in an invalid state. Do not use unless you know what you are doing!
     ReplicatedValue();
+
+    CSP_START_IGNORE
+
+    // Internal templated constructor
+    CSP_NO_EXPORT template <class T>
+    inline ReplicatedValue(const T& InValue)
+        : Value { InValue }
+    {
+    }
+
+    // Internal templated setter.
+    CSP_NO_EXPORT template <class T> inline void Set(const T& InValue) { Value = InValue; }
+    // Internal templated setter.
+    CSP_NO_EXPORT template <class T> inline const T& Get() const { return std::get<T>(Value); }
+
+    // Internal getter for variant.
+    CSP_NO_EXPORT const ReplicatedValueImplType& GetValue() const { return Value; };
+
+    CSP_END_IGNORE
 
     /// @brief Construct a ReplicatedValue based on a bool type.
     /// @param InBoolValue bool : Initial value.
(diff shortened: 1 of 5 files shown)


Library/src/Common/ReplicatedValue.cpp
```diff
@@ -24,303 +24,149 @@ static const csp::common::Map<csp::common::String, ReplicatedValue> InvalidStrin
 
 ReplicatedValue::ReplicatedValue() { ReplicatedType = ReplicatedValueType::InvalidType; }
 
-ReplicatedValue::~ReplicatedValue()
-{
-    if (ReplicatedType == ReplicatedValueType::String)
```
MAG-mv, line 29: We can get rid of this monstrosity!

Library/include/CSP/Common/ReplicatedValue.h
```diff
@@ -48,10 +54,28 @@ class CSP_API ReplicatedValue
...
+    {
+    }
+
+    // Internal templated setter.
+    CSP_NO_EXPORT template <class T> inline void Set(const T& InValue) { Value = InValue; }
+    // Internal templated setter.
+    CSP_NO_EXPORT template <class T> inline const T& Get() const { return std::get<T>(Value); }
+
+    // Internal getter for variant.
+    CSP_NO_EXPORT const ReplicatedValueImplType& GetValue() const { return Value; };
```
MAG-ElliotMorris, line 76: [Consider] Explaining why these are internal and templated. Just describing the method signatures doesn't help anyone.

Library/include/CSP/Common/ReplicatedValue.h
```diff
@@ -85,35 +109,35 @@ class CSP_API ReplicatedValue
...
 
     /// @brief Destroys the replicated value instance.
     ~ReplicatedValue();
 
-    /// @brief Assignment operator overload.
-    /// @param InValue ReplicatedValue : Other replicated value to set this one to.
-    ReplicatedValue& operator=(const ReplicatedValue& InValue);
+    /// @brief Copy constructor
+    /// @param Other csp::common::ReplicatedValue& : The value to copy.
+    CSP_NO_EXPORT ReplicatedValue(const ReplicatedValue& Other);
```
MAG-ElliotMorris, line 121: [Comment] I want to think about this sort of stuff when I get a moment. This sort of documentation is the most redundant thing in the world, and I half-suspect doing this is making us thing this sort of documentation descriptive of only the method signature is acceptable elsewhere. I wonder if it's necessary, or even if doxygen can automatically label the special constructors and such.

Library/include/CSP/Common/ReplicatedValue.h
```diff
@@ -85,35 +109,35 @@ class CSP_API ReplicatedValue
...
+    /// @brief Equality operator overload.
     /// @param ReplicatedValue : Other value to compare to.
-    bool operator<(const ReplicatedValue& OtherValue) const;
+    CSP_NO_EXPORT bool operator==(const ReplicatedValue& OtherValue) const;
 
-    /// @brief Greater than operator overload.
+    /// @brief Inequality operator overload.
     /// @param ReplicatedValue : Other value to compare to.
-    bool operator>(const ReplicatedValue& OtherValue) const;
+    CSP_NO_EXPORT bool operator!=(const ReplicatedValue& OtherValue) const;
```
MAG-ElliotMorris, line 141: [Question] Does losing the equality (and greater/less than) operators have client implications?

Library/include/CSP/Common/ReplicatedValue.h
```diff
@@ -226,31 +226,12 @@ class CSP_API ReplicatedValue
...
-        int64_t Int;
-        csp::common::String String;
-        csp::common::Vector2 Vector2;
-        csp::common::Vector3 Vector3;
-        csp::common::Vector4 Vector4;
-        csp::common::Map<csp::common::String, ReplicatedValue> StringMap;
-    };
-
-    InternalValue Value;
+    ReplicatedValueImplType Value;
```
MAG-ElliotMorris, line 232: [Comment] ❤️ Lovely.

Library/src/Common/ReplicatedValue.cpp
```diff
@@ -24,303 +24,149 @@ static const csp::common::Map<csp::common::String, ReplicatedValue> InvalidStrin
...
-        Value.StringMap.~Map();
-    }
-}
+ReplicatedValue::~ReplicatedValue() { }
 
 ReplicatedValue::ReplicatedValue(bool InValue)
-    : ReplicatedType(ReplicatedValueType::Boolean)
+    : Value { InValue }
+    , ReplicatedType(ReplicatedValueType::Boolean)
+
```
MAG-ElliotMorris, line 32: [Question] What's the point of the templated constructor if you have to do this anyway?

Library/src/Common/ReplicatedValue.cpp
```diff
@@ -24,303 +24,149 @@ static const csp::common::Map<csp::common::String, ReplicatedValue> InvalidStrin
...
-        }
-        }
-    }
-    return IsEqual;
+ReplicatedValue::ReplicatedValue(ReplicatedValue&& Other)
+{
+    this->Value = std::move(Other.Value);
+
+    this->ReplicatedType = std::move(Other.ReplicatedType);
+    Other.ReplicatedType = ReplicatedValueType::InvalidType;
```
MAG-ElliotMorris, line 95: [Comment] Very friendly (invalidating moved-from object), I almost never do this but I probably should.

Library/src/Common/ReplicatedValue.cpp
```diff
@@ -24,303 +24,149 @@ static const csp::common::Map<csp::common::String, ReplicatedValue> InvalidStrin
...
-        break;
-    }
-    }
-
-    return false;
+ReplicatedValue& ReplicatedValue::operator=(const ReplicatedValue& Other)
+{
+    this->Value = Other.Value;
+    this->ReplicatedType = Other.ReplicatedType;
+    return *this;
```
MAG-ElliotMorris, line 102: [Nit] You can likely implement your copy/move assignment operators in terms of your constructors, or vice versa, to avoid duplication.

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -174,48 +174,12 @@ mcs::ItemComponentData MCSComponentPacker::CreateItemComponentData(ComponentBase
     return mcs::ItemComponentData { ComponentPacker.GetComponents() };
 }
 
-// TODO: We can make a safer version of this function when we convert our ReplicatedValue to use a variant,
-// as we can create compile-time checking by using std::visit and function overloads.
-// This will prevent us forgetting to update this when we add new types.
-// https://magnopus.atlassian.net/browse/OF-1511
```
MAG-ElliotMorris, line 180: [Comment] Nice! TODO complete +50xp +2 quest points.

Library/src/Multiplayer/MCSComponentPacker.cpp
```diff
@@ -174,48 +174,12 @@ mcs::ItemComponentData MCSComponentPacker::CreateItemComponentData(ComponentBase
...
-    {
-        return CreateItemComponentData(Value.GetStringMap());
-    }
-    else
-    {
-        throw std::runtime_error("Invalid ReplicatedValue property");
-    }
+    mcs::ItemComponentData Data;
+    // Extract the internal type from the variant and parse using its corrosponding typed CreateItemComponentData.
+    std::visit([&Data, this](auto&& InternalType) { Data = CreateItemComponentData(InternalType); }, Value.GetValue());
```
MAG-ElliotMorris: [Consider] `CreateItemComponentData` only ever takes `const&` as a value category, so I think auto&& is redundant here (and potentially leaves you open for copies you don't want). Might as well be simple and just do `const auto&`

MAG-ElliotMorris (approved):
Love to see it. Nice one adding additional unit tests too.

Library/src/Common/ReplicatedValue.cpp
```diff
@@ -331,64 +177,60 @@ const csp::common::String& ReplicatedValue::GetDefaultString()
...
 const csp::common::Vector4& ReplicatedValue::GetVector4() const
 {
     assert(ReplicatedType == ReplicatedValueType::Vector4);
-    return Value.Vector4;
+    return Get<csp::common::Vector4>();
 }
 
 const csp::common::Vector4& ReplicatedValue::GetDefaultVector4() { return InvalidVector4; }
 
+csp::common::Map<csp::common::String, ReplicatedValue> m;
```
MAG-AdamThorn: [fix] this can be removed.

Library/include/CSP/Common/ReplicatedValue.h
MAG-mv: Nope. These were only used internally

Library/src/Common/ReplicatedValue.cpp
MAG-mv: Yeah, I removed this in my child PR

Library/src/Common/ReplicatedValue.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/798/commits/fa20b71d6b864a7c08cdb94ef023f15af9aa1920

Library/src/Multiplayer/MCSComponentPacker.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/798/commits/fa20b71d6b864a7c08cdb94ef023f15af9aa1920

Library/include/CSP/Common/ReplicatedValue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/798/commits/fa20b71d6b864a7c08cdb94ef023f15af9aa1920

Library/src/Common/ReplicatedValue.cpp
MAG-AdamThorn: [This](https://github.com/magnopus-opensource/connected-spaces-platform/pull/799) PR?

MAG-ElliotMorris (approved):

PR #800, opened 2025-09-22
[NT-0] fix: Crash on duplication of asset collections

MAG-SamBirley:
Following an update to the endpoints for duplicating prototypes we updated the corresponding CSP abstraction.

However, as part of that we forgot to allocate memory when declaring shared pointers that we then go on to derefrence, causing a crash.

Fixed by allocating memory and assigning it to the shared pointers when invoking their constructors.

### DIFF
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -571,14 +571,15 @@ void AssetSystem::CopyAssetCollectionsToSpace(csp::common::Array<AssetCollection
         = PrototypeAPI->CreateHandler<AssetCollectionsResultCallback, AssetCollectionsResult, void, csp::services::DtoArray<chs::PrototypeDto>>(
             Callback, nullptr);
 
-    auto PrototypeFilters = std::shared_ptr<chs::PrototypeFilters>();
+    auto PrototypeFilters = std::shared_ptr<chs::PrototypeFilters>(new chs::PrototypeFilters);
     PrototypeFilters->SetHasGroup(true);
 
-    auto DuplicateGroupPrototypesOptions = std::shared_ptr<chs::DuplicateGroupPrototypesOptions>();
+    auto DuplicateGroupPrototypesOptions = std::shared_ptr<chs::DuplicateGroupPrototypesOptions>(new chs::DuplicateGroupPrototypesOptions);
     DuplicateGroupPrototypesOptions->SetOriginalGroupId(SourceSpaceId);
     DuplicateGroupPrototypesOptions->SetNewGroupId(DestSpaceId);
     DuplicateGroupPrototypesOptions->SetAdditionalFilters(PrototypeFilters);
     DuplicateGroupPrototypesOptions->SetAsyncCall(CopyAsync);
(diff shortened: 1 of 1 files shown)


Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -571,14 +571,15 @@ void AssetSystem::CopyAssetCollectionsToSpace(csp::common::Array<AssetCollection
         = PrototypeAPI->CreateHandler<AssetCollectionsResultCallback, AssetCollectionsResult, void, csp::services::DtoArray<chs::PrototypeDto>>(
             Callback, nullptr);
 
-    auto PrototypeFilters = std::shared_ptr<chs::PrototypeFilters>();
+    auto PrototypeFilters = std::shared_ptr<chs::PrototypeFilters>(new chs::PrototypeFilters);
```
MAG-ElliotMorris, line 574: [Nit] This is completely fine for now, just a note that best practice for smart pointer construction is `auto PrototypeFilters = std::make_shared<chs::PrototypeFilters>();`, for the exact reason of avoiding this mistake. 

It's also:
- Safer if the ownership-transfer constructor (should be move constructor, but probably copy constructor for us as we rarely define them) of `PrototypeFilters` throws, which would leak in this form. ie, the control block allocation in the shared-ptr fails. I admit this is extremely arcane.
- Slightly more efficient, performing one allocation rather than two.

MAG-ElliotMorris:
Am concerned that this got through initially without adequate exercise in tests.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -571,14 +571,15 @@ void AssetSystem::CopyAssetCollectionsToSpace(csp::common::Array<AssetCollection
...
     PrototypeFilters->SetHasGroup(true);
 
-    auto DuplicateGroupPrototypesOptions = std::shared_ptr<chs::DuplicateGroupPrototypesOptions>();
+    auto DuplicateGroupPrototypesOptions = std::shared_ptr<chs::DuplicateGroupPrototypesOptions>(new chs::DuplicateGroupPrototypesOptions);
     DuplicateGroupPrototypesOptions->SetOriginalGroupId(SourceSpaceId);
     DuplicateGroupPrototypesOptions->SetNewGroupId(DestSpaceId);
     DuplicateGroupPrototypesOptions->SetAdditionalFilters(PrototypeFilters);
     DuplicateGroupPrototypesOptions->SetAsyncCall(CopyAsync);
+    DuplicateGroupPrototypesOptions->SetIncludeMusubiGeneratedAssets(true);
```
MAG-mv, line 582: [Question] Ticket implies we are just fixing the crash. What do we need this line for?

MAG-SamBirley: Thanks for spotting this. I had forgotten I had also made that change. The idea is that:
* We don't expose/abstract over the concept of Musubi currently in the public API.
* So the user can only assume that when they copy an asset collection, all asset details created under that asset collection must come along for the ride.

I think it's the right call given the current API, to do otherwise would lead to assets appearing to be missing their LODs etc when duplicated, but it's also true that its not in the spirit with the rest of this change.

What are your thoughts?

PR #801, opened 2025-09-23
[OF-1649] fix and re enable the poi tests

MAG-ChristopherAtkinson:
The previous failing test worked out of the box on local CHS; it was likely that, due to confusion around local CHS at the time, this was not supported in the local version on the developer machine, but worked as intended on CI due to the latest version of the services being pulled. 

Following this, I took the opportunity to evaluate the current state of the test and coverage. This highlighted a few separate problems that have been fixed in this PR:
 - Previously disabled tests have been reinstated 
 - We did not have a direct test for the delete operation; this has been added 
 - There were POI artefacts that persisted from the test runs; tests have been updated to remove the created POIs
 - An additional test has been added for testing the negative of `GetPOIInsideCircularAreaTest`

Overall, these changes are relatively minor to improve our current coverage and expectations around POI. 

> [!NOTE]
> During this investigation, I have found some inconsistencies with the API I plan to document and share these; however, they will be external to the current scope. 

> [!IMPORTANT]
> I considered mocking the service at this point; however, due to possible discussions around inconsistencies with API I decided to avoid this at the current time. However happy to reconsider if the team feels this would be a valuable change, considering the API is opinionated.

### DIFF
--- a/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
+++ b/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
@@ -454,22 +454,28 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, QuerySpacePOITest)
         const csp::common::Array<csp::systems::PointOfInterest>& POIs = GetPOIsResult.GetPOIs();
         EXPECT_GE(POIs.Size(), 1);
 
-        // There may be more than one POI at this location, so we search for the one we have created and expect to find it.
-        bool FoundSpacePOI = false;
-        bool FoundDefaultPOI = false;
+        // Multiple POIs have been created at the same location, so we need to validate that the expected POIs are found.
+        csp::systems::PointOfInterest POITypeDefault;
+        csp::systems::PointOfInterest POITypeSpace;
+
         for (uint32_t i = 0; i < POIs.Size(); i++)
         {
             if (POIs[i].Id == DefaultPOI.Id)
             {
-                FoundDefaultPOI = true;
+                POITypeDefault = POIs[i];
             }
 
             if (POIs[i].SpaceId == Space.Id)
             {
-                FoundSpacePOI = true;
+                POITypeSpace = POIs[i];
             }
         }
-        EXPECT_TRUE(FoundDefaultPOI && FoundSpacePOI);
+
+        EXPECT_NE(POITypeDefault.Id, "");
+        DeletePointOfInterest(POISystem, POITypeDefault);
+
+        EXPECT_NE(POITypeSpace.Id, "");
+        DeletePointOfInterest(POISystem, POITypeSpace);
     }
 
     DeleteSpace(SpaceSystem, Space.Id);
--- a/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
+++ b/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
@@ -208,8 +208,6 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaT
     CreatePointOfInterest(POISystem, nullptr, POILocation, nullptr, PointOfInterest);
 
     // Search for the newly created POI inside a circular area
-    csp::common::Array<csp::systems::PointOfInterest> POICollection;
-
     csp::systems::GeoLocation SearchLocationOrigin;
     SearchLocationOrigin.Latitude = 44.0;
     SearchLocationOrigin.Longitude = 160.0;
@@ -221,16 +219,7 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaT
 
     EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
 
-    if (Result.GetResultCode() == csp::systems::EResultCode::Success)
-    {
-        const auto& ResultPOIs = Result.GetPOIs();
-        POICollection = csp::common::Array<csp::systems::PointOfInterest>(ResultPOIs.Size());
-
-        for (size_t idx = 0; idx < ResultPOIs.Size(); ++idx)
-        {
-            POICollection[idx] = ResultPOIs[idx];
-        }
-    }
+    const csp::common::Array<csp::systems::PointOfInterest>& POICollection = Result.GetPOIs();
 
     // we should have at least the POI we've created
     EXPECT_TRUE(POICollection.Size() > 0);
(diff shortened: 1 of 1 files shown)


Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -254,6 +253,42 @@ CSP_PUBLIC_TEST(DISABLED_CSPEngine, PointOfInterestSystemTests, GetPOIInsideCirc
     LogOut(UserSystem);
 }
 
+CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIOutsideCircularAreaTest)
```
MAG-AdamThorn, line 245: [fix] I think it would be worth having a test where two pois are created at different locations, so that you can confirm that you only retrieve one when calling GetPOIsInArea with an input location and radius that overlaps one but not the other.

MAG-AdamThorn:
[fix] I was reviewing the existing tests (which you have not updated so I can't comment on them directly) and had the following comment relating to the `QuerySpacePOITest` test:

The test currently checks that at least one poi is returned and that the first item is a space, but it checks that the POIs.Size() is >= to one rather than checking that it **only** returns one poi. It could be returning two, with the first poi being the space and still pass.

MAG-AdamThorn:
[Consider] If you are making further updates to the tests it would be worth migrating any that create spaces over to using the CreateDefaultTestSpace() helper function rather than CreateSpace() - where possible.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -420,25 +454,57 @@ CSP_PUBLIC_TEST(DISABLED_CSPEngine, PointOfInterestSystemTests, QuerySpacePOITes
...
+                POITypeSpace = POIs[i];
             }
         }
-        EXPECT_TRUE(FoundDefaultPOI && FoundSpacePOI);
+
+        EXPECT_NE(POITypeDefault.Id, "");
+        DeletePointOfInterest(POISystem, POITypeDefault);
+
+        EXPECT_NE(POITypeSpace.Id, "");
+        DeletePointOfInterest(POISystem, POITypeSpace);
```
MAG-ChristopherAtkinson, line 501: > [fix] I was reviewing the existing tests (which you have not updated so I can't comment on them directly) and had the following comment relating to the `QuerySpacePOITest` test:
> 
> The test currently checks that at least one poi is returned and that the first item is a space, but it checks that the POIs.Size() is >= to one rather than checking that it **only** returns one poi. It could be returning two, with the first poi being the space and still pass.

I think this may have been the case of the viewing old state; this was updated to test explicitly for the POI being created in the tests.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-ChristopherAtkinson: Yeah, that was a great idea, I have included a new test here: [e112c81](https://github.com/magnopus-opensource/connected-spaces-platform/pull/801/commits/e112c81f9cdeb01d296eea271e60b423e3c0cab6)

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -289,6 +289,81 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIOutsideCircularArea
...
+
+    auto [Result] = Awaitable(&csp::systems::PointOfInterestSystem::GetPOIsInArea, POISystem, SearchLocationOrigin, SearchRadius,
+        csp::systems::EPointOfInterestType::DEFAULT)
+                        .Await(RequestPredicate);
+
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+    if (Result.GetResultCode() == csp::systems::EResultCode::Success)
+    {
+        const auto& ResultPOIs = Result.GetPOIs();
```
MAG-AdamThorn: [fix] I appreciate that there are only two elements here, but we could avoid an element by element copy here by using the array copy ctor. Or perhaps even better would be to use this as an opportunity to add move semantics to our common Array class.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -289,6 +289,81 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIOutsideCircularArea
...
+        POICollection = csp::common::Array<csp::systems::PointOfInterest>(ResultPOIs.Size());
+
+        for (size_t idx = 0; idx < ResultPOIs.Size(); ++idx)
+        {
+            POICollection[idx] = ResultPOIs[idx];
+        }
+    }
+
+    bool InsidePointOfInterestFound = false;
```
MAG-AdamThorn: [Comment] This approach felt odd to me but works. Personally I would have checked there was only a single element in the array and then confirmed it was the expected inside element.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-AdamThorn: I have added comments to that commit. Looks like they are going to appear disconnected at the bottom.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-AdamThorn: [Question] Rereviewing the code I think maybe the comments threw me. You only create a single POI of type default via this call: 
`CreatePointOfInterest(POISystem, nullptr, SpaceGeolocation, nullptr, DefaultPOI);`

However you then query for POIs of type Space and according to the comments expect more than one result. Do points on the geofence counts as Space POIs?

If so:
[Consider] making the documentation clearer.

MAG-AdamThorn (changes requested):

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-ChristopherAtkinson: Yeah, we can create points using two methods, the space system and POI system. These systems are referenced when the points are created at the start of the test, as well as the expected POI

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-ChristopherAtkinson: great spot, I have updated this here to just capture a reference from the result for the duration: [834209a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/801/commits/834209acfd5234b0e06d055069de557f22fdab6d)

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-ChristopherAtkinson: Updated for readability, initially we had an issue with test persisting previous state, but no that has been fixed this read much better: [1044551](https://github.com/magnopus-opensource/connected-spaces-platform/pull/801/commits/10445510ef225973321bfc98c81cd136af6c05bf)

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-ChristopherAtkinson: not seeing any comment around GetPOIOutsideCircularAreaTest?

MAG-AdamThorn (approved):

PR #803, opened 2025-09-25
[OF-1733] feat!: Read default settings from login

MAG-ElliotMorris:
Read the default settings now returned from login, and provide them in the LoginState return.

As LoginState is common, had to move existing settings structures to common also, which is what makes this a breaking change.
This entailed the removal/splitting up of the `ApplicationSettings.h` file, I've just put the stuff that remained from there in `ApplicationSettingsSystem`

**Breaking Changes**
- `csp::systems::SettingsCollection` relocated to `csp::common::SettingsCollection`.
- `csp::systems::ApplicationSettings` relocated to `csp::common::ApplicationSettings`.


Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
```diff
@@ -50,6 +50,38 @@ class WebClient;
 namespace csp::systems
 {
 
+/// @ingroup Application Settings System
+/// @brief Represents the result of a request for application settings.
+class CSP_API ApplicationSettingsResult : public csp::systems::ResultBase
```
MAG-ElliotMorris: [Comment] I'm still unsure if this should go here. Maybe I should have kept the `ApplicationSettings.h` file around, despite it only containing results now. I did that for `SettingsCollection` but that had more stuff in it (and I think we want to rename/reframe that anyhow)

Library/src/Common/LoginState.cpp
```diff
@@ -49,9 +49,11 @@ void LoginState::CopyStateFrom(const LoginState& OtherState)
     RefreshToken = OtherState.RefreshToken;
     UserId = OtherState.UserId;
     DeviceId = OtherState.DeviceId;
+    DefaultApplicationSettings = OtherState.DefaultApplicationSettings;
+    DefaultSettings = OtherState.DefaultSettings;
 
     // Must reallocate the access token when copying otherwise destructor of
-    // copied state will delete the original memory pointer potentially causing corruption
+    // copied state will delete the original memory pointer potentially causing corruption ... (Have you heard of copy constructors my friend?)
```
MAG-AdamThorn: [fix] Reviewing this comment in isolation, or the part you have added at the end, it isn't clear what point you are making. Are saying that they should have added a copy ctor to the Datetime type or that one exists and is being used.
Please update the overall comment to make it clearer.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-AdamThorn: [Consider] Personally I think moving result types to a centralised namespace is a good idea, but in the meantime I would favour symmetry and reinstate the ApplicationSettings.h as it could confuse developers.

MAG-AdamThorn: [Consider] Another option could be to move both sets of result objects to Settings.h and then to remove both ApplicationSettings.h and SettingsCollection.h.

MAG-AdamThorn (changes requested):

Library/src/Common/LoginState.cpp
MAG-ElliotMorris: Ach, normally I get a clear head and remove these little sarcastic notes I keep to myself during implementation, will zap this.

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ChristopherAtkinson: I would agree with @MAG-AdamThorn here, +1 for migrating everything to Settings.h, at the time I did not give this much thought, in retrospect as these systems both use the same underlying service, just expose separate API, would be a good opportunity o consolidate

MAG-ChristopherAtkinson (commented):
Overall the changes look good, @MAG-AdamThorn raise a good point around result objects that I have commented on, will leave this unapproved until thread are resolved.

MAG-ElliotMorris:
@MAG-AdamThorn @MAG-ChristopherAtkinson 

We can't move the Result objects themselves to `Settings.h` because they are not common types, so they can't be in the include set of that file. I will have to re-instate the deleted file.

Library/src/Common/LoginState.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/803/commits/31e978c7541b694076acd1287970388080407c20

Library/include/CSP/Systems/Settings/ApplicationSettingsSystem.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/803/commits/73e5fc5cefd6a75a735e122f833eeea2cef920e3

MAG-ElliotMorris: Ah, here's the comment thread, got confused cause it's mine.

I can't move this stuff to `Settings.h` because of the `systems`/`common` dependency issue with result types. The stuff in `Settings.h` needs to be common because it's included in `LoginState`, which is common.

MAG-ElliotMorris:
I have also added a commit that disables the test on local mcs, hopefully temporarily. CHS have tried to ship changes that seed local MCS but it hasn't taken properly as yet.

MAG-AdamThorn (approved):

MAG-ChristopherAtkinson (approved):

PR #805, opened 2025-09-26
[OB-4673] fix: csp disconnect callback is only firing whilst the user is in a space

MAG-ChristopherAtkinson:
The event was being processed as part of the `OnlineRealtimeEngine.cpp` which lifetime is coupled to a user entering an exiting a space, please see here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/755

 - Existing behaviour has been refactored to the `MultiplayerConnection.cpp`, meaning it can be processed outside of space interactions.
 - Removed related Event as this is no longer being processed within `OnlineRealtimeEngine.cpp`
 - Added a simple mock to dummy the response from CHS and test ON_REQUEST_TO_DISCONNECT

### DIFF
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -680,18 +680,13 @@ void MultiplayerConnection::BindOnRequestToDisconnect()
         OnRequestToDisconnect,
         [&](const signalr::value& Params)
         {
-            auto Done = false;
+            std::promise<bool> Promise;
+            std::future<bool> Future = Promise.get_future();
+
             const std::string Reason = Params.as_array()[0].as_string();
-            DisconnectWithReason(Reason.c_str(), [&Done](ErrorCode /*Error*/) { Done = true; });
+            DisconnectWithReason(Reason.c_str(), [&Promise](ErrorCode /*Error*/) { Promise.set_value(true); });
 
-            int TimeoutCounter = 2000;
-
-            while (!Done && TimeoutCounter > 0)
-            {
-                std::this_thread::sleep_for(1ms);
-
-                --TimeoutCounter;
-            }
+            Future.wait_for(2000ms);
         },
         LogSystem);
 }


MAG-ChristopherAtkinson:
I initially posted this PR as a draft with concerns about the threading raised by @MAG-mv, however, this makes more sense to start the review while I investigate, as likely this would delay further.

MAG-ChristopherAtkinson:
> > I initially posted this PR as a draft with concerns about the threading raised by @MAG-mv, however, this makes more sense to start the review while I investigate, as likely this would delay further.
> 
> Based on this comment, I have converted the PR to a draft. Let us know when it is ready for review, or ping the team if there is some aspect of it you wish to discuss.

this is related to the off-thread callback issue: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806

The changes in this PR do not introduce any additional. However, did highlight that the current and new implementations should suffer from the off-thread call-back issue

I attempted to put together a WASM test to confirm this behaviour to no avail, promoting from draft as this should be resolved with the changes here https://github.com/magnopus-opensource/connected-spaces-platform/pull/806, and currently is not showing any symptoms

Happy to discuss further

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -678,13 +678,20 @@ void MultiplayerConnection::BindOnRequestToDisconnect()
     const std::string OnRequestToDisconnect = GetMultiplayerHubMethods().Get(MultiplayerHubMethod::ON_REQUEST_TO_DISCONNECT);
     GetSignalRConnection()->On(
         OnRequestToDisconnect,
-        [](const signalr::value& Params)
+        [&](const signalr::value& Params)
         {
+            auto Done = false;
             const std::string Reason = Params.as_array()[0].as_string();
+            DisconnectWithReason(Reason.c_str(), [&Done](ErrorCode /*Error*/) { Done = true; });
```
MAG-ElliotMorris: ~~[Question] You are no longer firing the Disconnect eventsystem event, does `DisconnectWithReason` also do that? I didn't think it did. Seems like an observable behaviour change? There's a couple of places in the codebase that reference it from a skim.~~

~~They might be doing nothing, but you should delete them if so.~~

You do this in the next commit no bother.

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -678,13 +678,20 @@ void MultiplayerConnection::BindOnRequestToDisconnect()
...
-            DisconnectEvent->AddString("Reason", Reason.c_str());
-            csp::events::EventSystem::Get().EnqueueEvent(DisconnectEvent);
+            int TimeoutCounter = 2000;
+
+            while (!Done && TimeoutCounter > 0)
+            {
+                std::this_thread::sleep_for(1ms);
+
+                --TimeoutCounter;
+            }
```
MAG-ElliotMorris: [Consider] This smells. I'm not sure why it was in the previous place, and it's a different context here, being in a signal r callback. Think about changing this to a future/promise pair, this is a source of undefined behaviour otherwise. You can still maintain the timeout behaviour, in fact futures have it built in :)

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1590,4 +1590,36 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, MultiplayerTestsMock, TestNoSignalRCommuni
...
+
+                return true;
+            });
+
+    // Then the error callback we be called
+    MockMultiplayerErrorCallback MockErrorCallback;
+
+    // And the disconnection callback will be called with a message
+    MockConnectionCallback MockDisconnectionCallback;
+    EXPECT_CALL(MockDisconnectionCallback, Call(csp::common::String("New Multiplayer Session Initiated")));
```
MAG-ElliotMorris, line 1621: [Question] Why is the Disconnection callback called with a message that implies we're starting a new session?

MAG-ChristopherAtkinson: This is the reason MCS provides when another client starts a session and the previous is receives the disconnect

Library/src/Multiplayer/MultiplayerConnection.cpp
MAG-ChristopherAtkinson: This has been updated here: [4e0a981](https://github.com/magnopus-opensource/connected-spaces-platform/pull/805/commits/4e0a981c9a02b13748d5facdfa0ba067121c4780)

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-ElliotMorris: Ah yeah, that makes sense.

MAG-AdamThorn (approved):

PR #806, opened 2025-09-26
[OF-1524] Cross-thread Callbacks

MAG-mv:
CSP sometimes fires certain callbacks on a different thread the function was invoked. This causes crashes in WASM builds due to it not supporting this behaviour.

I decided to fix this at the binding level (wrapper generator), as we only need to fix this on wasm, and not anywhere else. Also, creating a callback queue that needs to be polled may change expected behavior, as clients would need to call CSPFoundation::Tick before logging in to get the callback.

I used the emscripten api to create a callback queue, which pushes callbacks onto the main thread if we are not on main. Doing it here also means we don't have to poll, as Emscripten resolved the inter-thread messaging for us using worker-thread messaging.

I also fixed a small logging issue in the NetworkInterruptionCallback, as we were always logging, even if it was a regular disconnect.

I was unfortunatly unable to reproduce the NetworkInterruptionCallback crash, both via testing and using a web build. Our testing framework crashes when trying to block to wait for the callback, so it isn't easily testable. Also, using a web build showed the callback logs, but no crash.

I also tested my new changes on a local web build to make sure there were no obvious issues/

### DIFF
--- a/Library/src/Multiplayer/MultiplayerConnection.cpp
+++ b/Library/src/Multiplayer/MultiplayerConnection.cpp
@@ -203,6 +203,7 @@ namespace
         Connection->SetDisconnected(
             [&NetworkInterruptionCallback, &LogSystem](const std::exception_ptr& Except)
             {
+                // We currently detect a connection interrupt if the disconnected callback contains an exception.
                 if (Except)
                 {
                     try
@@ -212,10 +213,9 @@ namespace
                     catch (const std::exception& e)
                     {
                         INVOKE_IF_NOT_NULL(NetworkInterruptionCallback, e.what());
+                        LogSystem.LogMsg(csp::common::LogLevel::Log, "Connection Interrupted.");
                     }
                 }
-
-                LogSystem.LogMsg(csp::common::LogLevel::Log, "Connection Interrupted.");
             });
     }
 }
--- a/Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
+++ b/Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
@@ -12,6 +12,86 @@
 
 #pragma endregion Generated Includes
 
+/*
+    Callback queue designed to push callbacks onto the main thread using the emscripten api.
+    This works because emscriptens threads are just web workers, which allows us to push messages to different workers.
+
+    This is implemented to work around the limitation of callbacks needing to fire on the same thread they were created on.
+*/
+#pragma region Emscripten Callback Queue
+
+#if CSP_WASM
+#include <emscripten.h>
+#include <emscripten/proxying.h>
+#include <emscripten/threading.h>
+#include <pthread.h>
+#include <tuple>
+
+static em_proxying_queue* ProxyQueue = nullptr;
+static pthread_t MainThread = 0;
+
+// Creates an emscripten queue and sets the main thread if we haven't already initialized.
+void Emscripten_InitializeProxyQueue()
+{
+    if (ProxyQueue != nullptr)
+    {
+        return;
+    }
+
+    ProxyQueue = em_proxying_queue_create();
+    MainThread = emscripten_main_runtime_thread_id();
+}
+
+// Structure to hold the callback and arguments.
+template <typename ...T>
+struct CallbackData
+{
+    void (*Callback)(T...);
+    std::tuple<T...> Args;
+ };
+
+// This function is called internally from emscripten_proxy_sync on the main thread.
+template <typename ...T>
+static void Emscripten_CallbackWrapper(void* InData) 
+{
+    auto* Data = static_cast<CallbackData<T...>*>(InData);
+    std::apply(Data->Callback, Data->Args);
+}
+
+// Function that will push the callback to the main thread if we are not on it.
+template <typename ...T>
+static void Emscripten_CallbackOnThread(void (*Callback)(T...), T... Args) 
+{
+    bool OnMainThread = pthread_equal(pthread_self(), emscripten_main_runtime_thread_id());
+    if (OnMainThread)
+    {
+        // We're on the main thread already, just call normally
+        Callback(Args...);
+    }
+    else
+    {
+        Emscripten_InitializeProxyQueue();
+
+        // Pack our data and send to the emscripten queue.
+        CallbackData<T...> Data{Callback, std::make_tuple(Args...)};
+        emscripten_proxy_sync(ProxyQueue, MainThread, Emscripten_CallbackWrapper<T...>, static_cast<void*>(&Data));
+    }
+}
+#endif
+#pragma endregion Emscripten Callback Queue
+
+// Generic callback function which will either call the callback directly, or, 
+// if we're on wasm, push to the main thread if we're not on the main thread.
+template <typename ...T>
+static void CallCallback(void (*Callback)(T...), T... Args)
+{
+#ifdef CSP_WASM
+    Emscripten_CallbackOnThread(Callback, Args...);
+#else
+    Callback(Args...);
+#endif
+}
+
 extern "C"
 {
     typedef struct {
--- a/Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
+++ b/Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
@@ -239,12 +239,10 @@
             {{/ type.is_string }}
           {{/ parameters }}
 
-          {{ name }}(
-            {{ name }}_StateObject,
-
-            {{# parameters }}
+          CallCallback({{ name }}, {{ name }}_StateObject,
+              {{# parameters }}
               {{# type.is_string }}
-                _{{ name }}
+                (const char*)_{{ name }}
               {{/ type.is_string }}
 
               {{^ type.is_string }}
--- a/WasmTesting/csp-wasm.test.ts
+++ b/WasmTesting/csp-wasm.test.ts
@@ -50,9 +50,7 @@ test('Cross Thread Callbacks From Log Callback, OB-3782', async () => {
   console.log(errors);
 
   assert.ok(
-    errors.some(e => e.message.includes('table index is out of bounds')),
-    'Expected cross-thread `table index is out of bounds` error. Message not found, did you fix the bug (OB-3782)? Nice job! ',
-  );
+    !errors.some(e => e.message.includes('table index is out of bounds')));
   
 });
 


Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -212,10 +213,9 @@ namespace
                     catch (const std::exception& e)
                     {
                         INVOKE_IF_NOT_NULL(NetworkInterruptionCallback, e.what());
+                        LogSystem.LogMsg(csp::common::LogLevel::Log, "Connection Interrupted.");
```
MAG-mv, line 218: Log should be in here, as this is when we call the NetworkInterruptionCallback

Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
```diff
@@ -239,12 +239,10 @@
...
 
-          {{ name }}(
-            {{ name }}_StateObject,
-
-            {{# parameters }}
+          CallCallback({{ name }}, {{ name }}_StateObject,
+              {{# parameters }}
               {{# type.is_string }}
-                _{{ name }}
+                (const char*)_{{ name }}
```
MAG-mv, line 245: I was getting compile errors for strings, as were were creating char*s. but we are expected cont char*

Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
```diff
@@ -239,12 +239,10 @@
             {{/ type.is_string }}
           {{/ parameters }}
 
-          {{ name }}(
-            {{ name }}_StateObject,
-
-            {{# parameters }}
+          CallCallback({{ name }}, {{ name }}_StateObject,
```
MAG-mv: Now calls the wrapper function instead

WasmTesting/csp-wasm.test.ts
```diff
@@ -50,9 +50,7 @@ test('Cross Thread Callbacks From Log Callback, OB-3782', async () => {
   console.log(errors);
 
   assert.ok(
-    errors.some(e => e.message.includes('table index is out of bounds')),
-    'Expected cross-thread `table index is out of bounds` error. Message not found, did you fix the bug (OB-3782)? Nice job! ',
-  );
+    !errors.some(e => e.message.includes('table index is out of bounds')));
```
MAG-mv, line 53: We can flip this test logic now its fixed!

MAG-ElliotMorris:
> I was unfortunatly unable to reproduce the NetworkInterruptionCallback crash, both via testing and using a web build. Our testing framework crashes when trying to block to wait for the callback, so it isn't easily testable. Also, using a web build showed the callback logs, but no crash.

I don't think we should merge this until we do this. It's arguably the main point of solving this, and simply not verifying that this fix works for it doesn't seem acceptable.

MAG-ElliotMorris:
I don't think this PR is complete, there is not enough testing. We need at least one test for each of the known cases. And given that this is an emscripten implementation, I was also expecting tests demonstrating calling functions from web workers. 

> I was unfortunatly unable to reproduce the NetworkInterruptionCallback crash, both via testing and using a web build. Our testing framework crashes when trying to block to wait for the callback, so it isn't easily testable. Also, using a web build showed the callback logs, but no crash.

I'm don't think this is acceptable, the way the testing framework or CSP is setup isn't immutable, did you experiment with a proxying server & changing the interruption timeout time to be much lower/instant?



_The below is ideological critique, and wouldn't neccesarily block a merge, that would be up to @MAG-AdamThorn as tech lead_

Primarily though, looking at this implementation I struggle to agree with the tradeoffs of using the emscripten api for this, rather than a hand-rolled queue. It seems almost equally large in terms of code footprint, with the disadvantages of :

- Using unfamiliar and arcane emscripten apis
- Existing within mustache files which are very hard to read and write
- Existing within the wrapper generator itself, which isn't an area we really have a handle on, it's going to make migrations and refactors harder as you've gotta keep this in mind.

Clients needing to call `Tick` to trigger certain events seems like a better tradeoff than the imposition of this sort of wrapper generator infrastructure.

MAG-mv:
@MAG-ElliotMorris

> Using unfamiliar and arcane emscripten apis
It's pretty well documented, with 1 emscripten api call, and a few small functions

> Existing within mustache files which are very hard to read and write
Agreed. However, if we move to embind, it would be a better fit for this, and would be easy to move


> Clients needing to call Tick to trigger certain events seems like a better tradeoff than the imposition of this sort of wrapper generator infrastructure.

The tick solution is limited, as it will push callbacks onto the main thread, even if the functions onto the main thread, even if emscripten calls these things on a worker, which will also cause the crash.

Witht he emscripten soilution, we can update this to call back to the thread the client originally called it on.

Also, I think the tick solution creates a confusing api, where certain functions will only callback if you are calling tick, especially since this affects functions during the csp initialization process (Login).  This will break all of our tests and likely client implementations.

MAG-ElliotMorris:
> The tick solution is limited, as it will push callbacks onto the main thread, , even if clients call these functions on a worker, which will also cause the crash.

Why is that true? The example I showed you only pushed data back onto the main thread via a copy, callbacks never moved. It was a simpler model imo which completely sidestepped the problem of moving capture state.

>This will break all of our tests and likely client implementations.

I think this is a vast overstatement.

Nonetheless, as we disagree on this there's not much to be done. I'll reluctantly review the existing implementation presuming @MAG-AdamThorn wants to press ahead with this, although my critiques around the level of test coverage still stands.

I wish there was a way of moving this out of the mustache files still ... it's bound to just be forgotten about.

MAG-mv:
> Why is that true? The example I showed you only pushed data back onto the main thread via a copy, callbacks never moved. It was a simpler model imo, which completely sidestepped the problem of moving capture state and should have worked from any calling environment.

The issue we are experiencing is that the callbacks are called on a _different_ thread from which it was created, not that its called on a  non-main thread. So, if clients call these functions on a worker thread, we will push to the main thread, which will crash again.

This current solution doesn't support this. However, it can be updated to support if needed.

MAG-ElliotMorris:
> The issue we are experiencing is that the callbacks are called on a different thread from which it was created, not that its called on a non-main thread. So, if clients call these functions on a worker thread, we will push to the main thread, which will crash again.

Ah, I see. Yeah, the tick solution suffers from the same issue as this implementation then without additional work, I'll admit this weakens my argument somewhat.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,86 @@
...
+#include <emscripten/proxying.h>
+#include <emscripten/threading.h>
+#include <pthread.h>
+#include <tuple>
+
+static em_proxying_queue* ProxyQueue = nullptr;
+static pthread_t MainThread = 0;
+
+// Creates an emscripten queue and sets the main thread if we haven't already initialized.
+void Emscripten_InitializeProxyQueue()
```
MAG-ElliotMorris: [Question] I don't understand why this logic is in a mustache file, and not in a proper .cpp with `#EMSCRIPTEN` guards like all of the rest of our emscripten specific code.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,86 @@
...
+{
+    if (ProxyQueue != nullptr)
+    {
+        return;
+    }
```
MAG-ElliotMorris: [Question] Does this imply this `Initialize` method is called lazily? Is this a necessity, [consider] if it could be called during the standard initialize flow.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,86 @@
...
+    MainThread = emscripten_main_runtime_thread_id();
+}
+
+// Structure to hold the callback and arguments.
+template <typename ...T>
+struct CallbackData
+{
+    void (*Callback)(T...);
+    std::tuple<T...> Args;
+ };
```
MAG-ElliotMorris: [Question] How does captured state play into this, seeing that this is a function pointer? Normally function pointer based callback approaches have a `*UserContext` sort of thing to work as a shim for that.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,86 @@
...
+
+// This function is called internally from emscripten_proxy_sync on the main thread.
+template <typename ...T>
+static void Emscripten_CallbackWrapper(void* InData) 
+{
+    auto* Data = static_cast<CallbackData<T...>*>(InData);
+    std::apply(Data->Callback, Data->Args);
```
MAG-ElliotMorris: [Comment] Pretty neat, I almost never see `std::apply` used. I _think_ its only real purpose is to do this sort of parameter unpack.

MAG-ElliotMorris: [Nit] You might want to consider moving/forwarding the args, just for a bit of optimality for callbacks with value signatures. I struggle to trace the `InData` ownership however, so this might not be the thing to do.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,86 @@
...
+        emscripten_proxy_sync(ProxyQueue, MainThread, Emscripten_CallbackWrapper<T...>, static_cast<void*>(&Data));
+    }
+}
+#endif
+#pragma endregion Emscripten Callback Queue
+
+// Generic callback function which will either call the callback directly, or, 
+// if we're on wasm, push to the main thread if we're not on the main thread.
+template <typename ...T>
+static void CallCallback(void (*Callback)(T...), T... Args)
```
MAG-ElliotMorris: [Consider] Naming this more specifically so in the C-api it's known why it exists. `CallCallbackOnMainThread` or something.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-ElliotMorris: [Consider] Ah I see, the StateObject is always the first arg of the pack. Perhaps make that explicit and having the parameter pack only apply to user arguments in order to reduce arcanity, it's not possible to know how this works currently without also exploring the C-wrapper.

MAG-ElliotMorris (commented):
Despite my hesitation around solving the problem in this area at the moment, given the roughness of the terrain, I think the implementation is rather good, albeit underdocumented for the level of complexity on display.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: That's a good point. It would be a lot nicer in a cpp

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: That's correct. I was worried about doing it when we first created the variable because its global. Maybe it's ok to do so.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-ElliotMorris: [Comment] This makes me worry about the "This will be trivial in embind" thing, we'll have to re-solve this. I suppose it's not the hugest thing.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/fa27cbd8efa5127eeb4251b8ea39dd1a5cd90ecd

I left the "CallCallback" function in the generated code, as this isnt related to cross-thread callback and is used by everything to prevent repeated wrapper template code

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: That function is used for all callbacks to prevent duplicate wrapper template code, not just for callbacks pushed onto the main thread. I think the name captures what its doing imo. And the comment gives more detail

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: Also fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/commit/fa27cbd8efa5127eeb4251b8ea39dd1a5cd90ecd

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-ElliotMorris: I'll likely have something to say on the "CallCallback" function, but resolving here in favour of a fresh review, given the time delay.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-ElliotMorris: May have more to say on this also, but deferring to fresh review

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
```diff
@@ -209,6 +209,9 @@ class CSP_API MultiplayerConnection
     /// exiting a space.
     csp::multiplayer::OnlineRealtimeEngine* GetOnlineRealtimeEngine() const;
 
+    // Use for testing to send an error back to SignalR.
+    void __CauseFailure();
```
MAG-ElliotMorris: [consider] Be more specific with the comment here, it's an unusual function guidance would be appreciated as to how it goes about this and why we need to take this sort of approach. Remember that you're writing for someone who might need to move, replace, or extend this function in a refactoring effort, not just a general reader.

Library/src/EmscriptenBindings/CallbackQueue.h
```diff
@@ -0,0 +1,58 @@
+/*
+    Callback queue designed to push callbacks onto the main thread using the emscripten api.
+    This works because emscriptens threads are just web workers, which allows us to push messages to different workers.
+
+    This is implemented to work around the limitation of callbacks needing to fire on the same thread they were created on.
```
MAG-ElliotMorris: [Consider] Explain/link out to more context on this. Folk are gonna wonder why they need to fire on the same thread they were created on, and what the consequences might be otherwise. We don't want this to become cargo-cult knowledge that folk down-the-line just have to trust. Consider that it's completely possible for this restriction to be removed under CSP in a later WASM/Emscripten version, folks need to know where the restriction comes from during refactoring efforts.

Library/src/EmscriptenBindings/CallbackQueue.h
```diff
@@ -0,0 +1,58 @@
...
+namespace csp
+{
+// Setup our proxy queue to which sends callbacks to the main thread
+static em_proxying_queue* ProxyQueue = em_proxying_queue_create();
+// Get a reference to the main thread to send callbacks to
+static pthread_t MainThread = emscripten_main_runtime_thread_id();
+
+// Structure to hold the callback and arguments.
+// With the current implementation, the first argument will always be the callback context.
+template <typename... T> struct CallbackData
```
MAG-ElliotMorris, line 65: > // With the current implementation, the first argument will always be the callback context.

[fix] You should make the template arguments reflect that then. Documentation is a weak interface guarentee. 
`template <typename CallbackContext, typename... T>`

Library/src/EmscriptenBindings/CallbackQueue.h
```diff
@@ -0,0 +1,58 @@
...
+{
+    void (*Callback)(T...);
+    std::tuple<T...> Args;
+};
+
+// This function is called internally from emscripten_proxy_sync on the main thread.
+template <typename... T> static void Emscripten_CallbackWrapper(void* InData)
```
MAG-ElliotMorris: [Consider] Explain what this function does and why, as well as where it's called. I cannot tell the point of having a CallbackWrapper just from reading the code.

Library/src/EmscriptenBindings/CallbackQueue.h
```diff
@@ -0,0 +1,58 @@
...
+{
+    auto* Data = static_cast<CallbackData<T...>*>(InData);
+    std::apply(Data->Callback, Data->Args);
+}
+
+// Function that will push the callback to the main thread if we are not on it.
+template <typename... T> static void Emscripten_CallbackOnThread(void (*Callback)(T...), T... Args)
```
MAG-ElliotMorris: [fix] Presumably the first T... is the calling context here as well? Same comment as above, make it explicit otherwise the reader just has to guess.

Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
MAG-ElliotMorris: [Question] Can you be more specific. Where does the non-const char creation source from? Trying to find this in my local `generated_wrapper.h` and strugging to find a non-const char*.

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
```diff
@@ -12,6 +12,20 @@
...
+{
+#ifdef CSP_WASM
+    csp::Emscripten_CallbackOnThread(Callback, Args...);
+#else
+    Callback(Args...);
+#endif
+}
+
```
MAG-ElliotMorris: [Consider] I won't put this as [fix] because you've already made a point on this. However, I still think you should consider moving this to a regular `.h` and not be in a mustache file. Primarily because the mustache files are debt, and we should do what we can to make it a clean delete when we need to migrate away, which is soon if we manage to stay on track. I don't want to have to reimplement all this work in a few months.

Even without that, it seems inappropriate. This code does no templating, and is not relevent to any templating logic, seems inappropriate to be here.

.github/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.
   In the continuation flow, we would return an invalid result and therefore lose the associated result data. This change ensures the result data from the exception is preserved.
 - [NT-0] fix: Use const string ref for return type by MAG-AlessioRegalbuto in https://github.com/magnopus-opensource/connected-spaces-platform/pull/856
   As part of ongoing Unity interop API work, we updated the 'MimeTypeHelper' to return a const ref string for the mime type to ensure correct SWIG code generation.
+- [OF-1524] fix: Wasm callbacks are now automatically forwarded to the main thread, prevent crashes due to Wasm limitations. This fixes crahes with the log callback if UserSystem::Login fails, and also the MultiplayerConnection::ConnectionInterrupted callback.
```
MAG-AdamThorn: [fix] can you follow the convention set by other changelog updates. So it should be split over two lines like so:

[ticket] category: PR title by MAG-mv in [link to PR]
Concise description

Library/src/EmscriptenBindings/CallbackQueue.h
```diff
@@ -0,0 +1,58 @@
...
+#include <emscripten.h>
+#include <emscripten/proxying.h>
+#include <emscripten/threading.h>
+#include <pthread.h>
+#include <tuple>
+#include <utility>
+
+namespace csp
+{
+// Setup our proxy queue to which sends callbacks to the main thread
```
MAG-AdamThorn: [consider] rewording this comment, it does not make sense to me.

Library/src/Multiplayer/WebSocketClient.h
```diff
@@ -35,6 +35,9 @@ class IWebSocketClient
     virtual void Stop(CallbackHandler Callback) = 0;
     virtual void Send(const std::string& Message, CallbackHandler Callback) = 0;
     virtual void Receive(ReceiveHandler Callback) = 0;
+
+    // Use for testing to send an error back to SignalR.
```
MAG-AdamThorn: [consider] expanding on this comment slightly.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1623,3 +1623,60 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, TestMultiplayerDisconnectionWhenNew
     Connection.SetDisconnectionCallback(std::bind(&MockConnectionCallback::Call, &MockDisconnectionCallback, std::placeholders::_1));
     Connection.Connect(std::bind(&MockMultiplayerErrorCallback::Call, &MockErrorCallback, std::placeholders::_1), "", "", "");
 }
+
+CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, ConnectionInterrupedTest)
```
MAG-AdamThorn: [fix] spelling: 'ConnectionInterrupedTest' -> 'ConnectionInterruptedTest'

WasmTesting/csp-wasm.test.ts
MAG-AdamThorn: Whoop whoop!

WasmTesting/html_tests/CrossThreadConnectionInterrupted.html
```diff
@@ -0,0 +1,82 @@
...
+        true,
+        Multiplayer.AvatarState.Idle,
+        authState.userId,
+        Multiplayer.AvatarPlayMode.Default
+      );
+
+      // Set interrupted callback
+      let interrupted = false;
+      const connection = systemsManager.getMultiplayerConnection();
+      connection.setNetworkInterruptionCallback((message) => { console.log("Connection inturrupted: true"); interrupted = true });
```
MAG-AdamThorn: [fix] spelling: 'Connection inturrupted' -> 'Connection interrupted'

WasmTesting/csp-wasm.test.ts
```diff
@@ -44,27 +44,42 @@ test('EnterSpace', async() => {
...
   assert.ok(
-    errors.some(e => e.message.includes('table index is out of bounds')),
-    'Expected cross-thread `table index is out of bounds` error. Message not found, did you fix the bug (OB-3782)? Nice job! ',
-  );
-  
+    !errors.some(e => e.message.includes('table index is out of bounds')));
+
+});
+
+test.only('Cross Thread Callbacks From NetworkInterrupted Callback, OB-1524', async () => {
```
MAG-ElliotMorris: [fix] Remember to take this `.only` out.

Tools/WrapperGenerator/Templates/C/Partials/Class/Method/Method.mustache
MAG-mv: We construct a new char buffer within the generated body
image

Library/include/CSP/Multiplayer/MultiPlayerConnection.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/dcd01bfa62e040c1bd6a785c1c56faec2bb1727a

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/857c85fff41415a8e051864241241561cb549071

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/bca9c89c219b6f99dcf20427e5a4d868f09f026b

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/bca9c89c219b6f99dcf20427e5a4d868f09f026b

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/8736c389e4c02c41458a2cb2f3381728b5844e63

Tools/WrapperGenerator/Templates/C/GeneratedWrapper.mustache
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/5dc90681f6d49c1daef50d86743f641edddb8bd8

.github/CHANGELOG.md
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/8339f11bd973f37a2502a91926758e45f4b7fd81

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/0a9a20340407c9f4287e39da6a76337a9247ca5c

Library/src/Multiplayer/WebSocketClient.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/0a9a20340407c9f4287e39da6a76337a9247ca5c

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/0a9a20340407c9f4287e39da6a76337a9247ca5c

WasmTesting/html_tests/CrossThreadConnectionInterrupted.html
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/0a9a20340407c9f4287e39da6a76337a9247ca5c

Library/src/Multiplayer/WebSocketClient.h
MAG-mv: Also adding more docs in calling function: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/dcd01bfa62e040c1bd6a785c1c56faec2bb1727a

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-ElliotMorris: Really good

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-ElliotMorris: Still dosen't really get to the root of it, but I think I can deduce now that it's there because `emscripten proxy api` requires a function of that signiature? It's an adapter?

WasmTesting/csp-wasm.test.ts
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/806/commits/2085d1602043dcf1b1751ff7f2cd5eb061a0f977

Library/src/EmscriptenBindings/CallbackQueue.h
MAG-mv: Pretty much. Emscripten requires a function that passed user data as a voi*

PR #807, opened 2025-09-26
Fix leaking allocations & thrown exception in constructors with C# wrapper generator 

magnopus-swifty:
Addresses an issue raised by #786 where `NativeWrapper`-derived classes with base classes (e.g. `Space` which derives from `BasicSpace`) would throw an exception when the default constructor is called. This was causing some OKO Unity unit tests to fail and thus could eventually pose issues for production code.

`TestDerivedType()` in UnitTesting\tests\csharp\Assets\Tests\ClassTests.cs tests this behaviour. I've verified before and after the change that it reproduces the error.

Without this change, `Space.Space()` would call the default constructor `BasicSpace.BasicSpace()`:

```csharp
        public BasicSpace()
        {
            var np = csp_systems_BasicSpace_Ctor();
            _ptr = np.Pointer;
            _ownsPtr = np.OwnsOwnData;
        }
```

This sets the initial value of `_ptr` to a freshly allocated `csp::systems::BasicSpace`. Then the constructor for `Space` is called:

```csharp
        public Space()
        {
            var np = csp_systems_Space_Ctor();
            _ptr = np.Pointer;
            _ownsPtr = np.OwnsOwnData;
        }
```

This constructs an instance of `csp::systems::Space` and then _throws away_ the previous value, leaking memory.

#786 explicitly forbids changing the value of `_ptr` (throws an `InvalidOperationException`) and highlighted this problem.

### DIFF
--- a/UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
+++ b/UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
@@ -15,6 +15,7 @@
  */
 
 using NUnit.Framework;
+using System;
 
 namespace Csp.Tests
 {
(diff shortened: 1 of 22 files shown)


UnitTesting/tests/csharp/Assets/Tests/ClassTests.cs
```diff
@@ -81,5 +81,15 @@ public void AccessAfterDisposeThrows()
             var ex = Assert.Throws<NullReferenceException>(() => { int value = simpleClass.GetValue(); });
             StringAssert.Contains("Attempting to access a null pointer", ex.Message);
         }
+
+        [Test]
+        public void TestDerivedType()
```
magnopus-swifty, line 86: I've confirmed that without the fix this will throw an exception: `System.InvalidOperationException : Attempting to change the native pointer for csp_Tests_DerivedClass from 2686249035216 to 2686249035136`

and with the fix will succeed as expected.

MAG-AdamThorn (approved):

MAG-SangeetShah (approved):
Code changes look good. Tested locally and failing Foundation unit tests now pass:
Screenshot 2025-09-29 at 14 51 50

PR #808, opened 2025-09-30
[OPE-2742] fix: Use LIB_NAME for Android symbol resolution

MAG-AlessioRegalbuto:
Resolving a runtime failure on Android (Unity) Quest 3 (see [this Slack thread](https://magnopus.slack.com/archives/C02TKEGRJE6/p1759241197504239?thread_ts=1759159460.239089&cid=C02TKEGRJE6) for more info).

This change allows our exported functions to be correctly found by Unity, since the previous behavior was limiting the search to the running executable rather than the library, causing a nullptr due to not found signatures.

### DIFF
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -51,8 +51,8 @@
 // For fstat and mkdir
 #include <sys/stat.h>
 
-#define LOAD_OWN_MODULE() dlopen(nullptr, RTLD_LAZY)
-#define GET_FUNCTION_ADDRESS(mod, name) dlsym((RTLD_DEFAULT), name)
+#define LOAD_OWN_MODULE() dlopen(LIB_NAME, RTLD_LAZY)
+#define GET_FUNCTION_ADDRESS(mod, name) dlsym((mod), name)
 #elif defined(CSP_MACOSX) || defined(CSP_IOS)
 // For dlopen and dlsym
 #include <dlfcn.h>


Library/src/CSPFoundation.cpp
```diff
@@ -52,7 +52,7 @@
 #include <sys/stat.h>
 
 #define LOAD_OWN_MODULE() dlopen(nullptr, RTLD_LAZY)
-#define GET_FUNCTION_ADDRESS(mod, name) dlsym((mod), name)
+#define GET_FUNCTION_ADDRESS(mod, name) dlsym((RTLD_DEFAULT), name)
```
magnopus-swifty: Would it not be better to query the CSP library directly, the way it works on Windows? Would look something like:

```c++
#define LOAD_OWN_MODULE() dlopen(LIB_NAME, RTLD_LAZY)
#define GET_FUNCTION_ADDRESS(mod, name) dlsym((mod), name)
```

There's a (very small) chance that you have a symbol collision between two loaded libraries and you could end up looking up the wrong one.

OTOH this is exactly how it works on iOS so maybe this is perfectly fine. I'd hate to have to diagnose this going wrong though!

MAG-ElliotMorris:
Looks like this was the correct diagnosis.

... Is there _any_ way to test this currently in an automated manner 🤔. 
This is a Linux problem really isn't it, not an android problem. I don't relish the idea of needing actual android hardware hooked into a CI environment. S'pose there's emulators.

Library/src/CSPFoundation.cpp
MAG-SamBirley: @MAG-AlessioRegalbuto would you mind testing using `mod` via another local build? Agreed with @magnopus-swifty it'd close down a possible source of gnarly bugs if we can.

MAG-AlessioRegalbuto: Thanks for the suggestion, I made a new commit to address that: https://github.com/magnopus-opensource/connected-spaces-platform/pull/808/commits/16549a564093d6979ecce93466eaedfd538a50c0

I also rebased the branch on top of release tag 6.6.0 instead of 6.4.4 since I also rebased the Quest 3 on latest Unity dev branch (which uses 6.6.0 instead of 6.4.0), so that I can test the change on device.

I will report back with the result from my testing shortly.

MAG-AlessioRegalbuto: I confirm that the suggested change also works, so this PR is ready to be merged!

For reference, I used on Quest 3 the following build Android release library generated by this PR: https://magnopus.teamcity.com/buildConfiguration/Olympus_OlympusFoundation_Android_ReleaseDLL_2/302098

It was a success, no more binding issues for any CSP function required for our Unity app flow to correctly work.

Small cookie for your eyes:

https://github.com/user-attachments/assets/ffbd1ee8-c1c7-4d85-a002-97ce0cd009db

This concludes the work for the spike (which I fully documented on Confluence: https://magnopus.atlassian.net/wiki/spaces/OL/pages/3715366922/Spike+Create+a+VR+Quest+build+for+OKO).
Now we can have fun making an awesome VR SKU!

MAG-SamBirley: Thanks @MAG-AlessioRegalbuto. @magnopus-swifty would you mind approving? I'd like to get this one in now that we've done some manual validation.

magnopus-swifty (approved):

PR #811, opened 2025-10-01
[OF-1718] chore: Remove out of date library examples

MAG-AdamThorn:
Our current CSP examples have not been updated in 6-10 months. In that period there have been some major changes to the codebase, a trend that will continue for the foreseeable future. As such the examples are no longer helpful to those looking to adopt CSP and will likely create a negative impression to those evaluating our library.

The decision has been taken to remove our examples entirely. In addition to removing the `Examples` folder, the documentation has also been updated to remove an reference to them.


MAG-SamBirley (approved):
Nice., thanks!

PR #812, opened 2025-10-01
[NT-0] feat!: Log levels associated with messages are returned to the client application

MAG-SamBirley:
## The Problem
Client apps currently have affordances for controlling CSP's verbosity. What they don't have is a way to know the log level of any given message. Consequently, they themselves cannot filter log messages retroactively. 

## The Solution
Return the log level associated with any given CSP log message in the callback bound by the client app in `LogSystem::SetLogCallback`.

## Where's the tests?
Initially I planned to write dedicated tests, but quickly found that our current tests exercise the change incredibly well as they are.
Moreover, what's neat about this change (and something I didn't have in mind at the start) is that this change allows us to validate our log messages use an appropriate log level.

To that end, part of this change also involves reacting to the discovery that `LogErrorAndCancelContinuation` was just emitting `Log` level error messages. Now it emits those as `Error` logs. Much better.

## Breaking Changes
The  signature of the callback expected by `LogSystem::SetLogCallback` now includes a `csp::common::LogLevel` parameter, describing the verbosity/log level of the message.

In client code, this means code that looked like this:

```cpp
auto& logCallback = [](const csp::common::String& logMessage)
{
    /// Handle log message
};

auto& systemsManager = csp::systems::SystemManager::Get();
auto& logSystem = systemsManager.GetLogSystem();
logSystem->SetLogCallback(logCallback);
```

Now looks like this:

```cpp
auto& logCallback = [](csp::common::LogLevel logLevel, const csp::common::String& logMessage)
{
    /// Handle log message, taking the log level into account!
};

auto& systemsManager = csp::systems::SystemManager::Get();
auto& logSystem = systemsManager.GetLogSystem();
logSystem->SetLogCallback(logCallback);
```


MAG-AdamThorn:
@MAG-SamBirley - wherever possible/appropriate we like to keep a clean commit history which can be viewed by the reviewer in sequence, isolating changes and guiding the reviewer through the PR. We have found that it makes them far easier to consume. Please see guidance [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/PR_GUIDELINES.md).

Library/src/Common/Systems/Log/LogSystem.cpp
```diff
@@ -35,7 +35,7 @@ struct LogCallbacks
     }
 
     LogSystem::LogCallbackHandler LogCallback;
-    LogSystem::LogCallbackHandler EventCallback;
+    LogSystem::EventCallbackHandler EventCallback;
```
MAG-ElliotMorris, line 38: [Question] What a strange oversight to have been in the codebase, considering all the infrastructure is hooked up around it.

I'm presuming because the signatures were identical before this doesn't actually change any observable behaviour?

Library/src/Common/Systems/Log/LogSystem.cpp
```diff
@@ -75,7 +75,7 @@ void LogSystem::LogMsg(const csp::common::LogLevel Level, const csp::common::Str
     if (Callbacks->LogCallback != nullptr)
     {
         // Send message to clients to display the log on the client side.
-        Callbacks->LogCallback(InMessage);
+        Callbacks->LogCallback(Level, InMessage);
```
MAG-AdamThorn, line 78: [Question] This question relates to line 73 above, which I can't comment on directly.
Do we also want to capture the log level when logging to file?

MAG-ElliotMorris (approved):
Really like the effect this has on our log testing.

1. If you were a regular contributor I'd encourage you to move to a rebase workflow to clean up this changeset, there's 4 or 5 commits that could just not be here. But it's just a review hygiene thing and not that important for drive by PR's.
2. [Consider] Calling out the change to default loglevel behaviour in the PR notes + the commit message. This isn't technically breaking, but I would imagine some clients could need to react to this.

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-AdamThorn, line 38: [consider] this comment relates to lines 51 - 57 below.
This feels like a good opportunity to update the SetXCallback methods to std::move the callbacks and avoid unnecessary copies.

MAG-AdamThorn (approved):

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-SamBirley: Yep that's correct. No functional difference between the two meant it had gone unnoticed all this time!

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-SamBirley: Yeah that would be good. Happy to make that a part of this PR but won't get to it till next week.

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-SamBirley: Addressed in https://github.com/magnopus-opensource/connected-spaces-platform/pull/812/commits/54aa52981b707853b4cfa4bb7f02c727339c0472.

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-SamBirley: Addressed in https://github.com/magnopus-opensource/connected-spaces-platform/pull/812/commits/cbc810bf10b42c1e6e76f99d9d5298539d6aa6bf.

MAG-ChristopherAtkinson: Reviewing on behalf of @MAG-AdamThorn, this has been addressed

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-ChristopherAtkinson: Reviewing on behalf of @MAG-AdamThorn, this has also been addressed, and great change

PR #817, opened 2025-10-07
[OF-1796] feat: Generic ExternalService result processing

MAG-SamBirley:
## The Problem
It was pointed to me that we cannot assume any schema in the Json data returned via the external service proxy services.

This is because the proxy service is assumed to simply acts as a router for invoking operations on external services, and it is intended to feed the results of those operations directly back to the caller.

## The Solution
To account for that, this change refactors the result class used by the external service system into two classes. One generic class for returning the result payload with no tampering, and another for the Agora specialization.

### Result Classes

The generic result class is used by `ExternalServiceProxySystem::InvokeOperation`, and returns the full string value of the `operationResult` Json field in the returned payload.

The specialized class is used by `ExternalServiceProxySystem::GetAgoraUserToken`. Since we _can_ assume the schema in this specialized instance, this class takes the raw `operationResult` value, and parses it for the token (this is what the previous implementation did for all operations).

The mock test has been updated to not assume any particular Json format. Instead, it now simply stress tests whether the Json data under the `operationResult` is parsable, and that the string held in the result class is the expected full string.

### ExteralServiceSystem Refactor

One thing that was nice about the previous implementation was how the Agora specialisation of `InvokeOperation` fell through into the generalized function. Since we now need to use different types for handling results, and using templated public API functions is messy/has ABI risks, we now do the next best thing and have both calls go through a templated internal-only function.

> [!NOTE]
> This is not a breaking change, as it does not change the public API, nor the data passed back when retrieving Agora tokens. It _does_ change the data returned by the generic `InvokeOperation` endpoint, but that function has yet to be included in a release.

### DIFF
--- a/Library/src/Systems/ExternalServices/ExternalServiceProxySystem.cpp
+++ b/Library/src/Systems/ExternalServices/ExternalServiceProxySystem.cpp
@@ -35,6 +35,24 @@ namespace csp::systems
 
 inline const char* BoolToApiString(bool Val) { return Val ? "true" : "false"; }
 
+/// @brief Generic function which will make a post request to the services proxy endpoint to trigger some specified operation of specified
+/// service. The class is templated to allow specialized functions to customize the result object used to process the response sent back from
+/// services.
+template <class ResultType>
+void InvokeOperationWithResult(
+    csp::services::ApiBase* InExternalServiceProxyApi, const ExternalServicesOperationParams& Params, StringResultCallback Callback)
+{
+    auto TokenInfo = std::make_shared<chs_aggregation::ServiceRequest>();
+    TokenInfo->SetServiceName(Params.ServiceName);
+    TokenInfo->SetOperationName(Params.OperationName);
+    TokenInfo->SetHelp(Params.SetHelp);
+    TokenInfo->SetParameters(Convert(Params.Parameters));
+
+    csp::services::ResponseHandlerPtr ResponseHandler
+        = InExternalServiceProxyApi->CreateHandler<StringResultCallback, ResultType, void, chs_aggregation::ServiceResponse>(Callback, nullptr);
+    static_cast<chs_aggregation::ExternalServiceProxyApi*>(InExternalServiceProxyApi)->service_proxyPost({ TokenInfo }, ResponseHandler);
+}
+
 ExternalServiceProxySystem::ExternalServiceProxySystem()
     : SystemBase(nullptr, nullptr, nullptr)
     , ExternalServiceProxyApi(nullptr)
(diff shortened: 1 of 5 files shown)


MAG-ChristopherAtkinson (approved):
The changes look good, glad to see this being address so quickly was a good catch. 

I like the separation between generic and opinionated systems, I hope we can carry this practice forward.

Library/src/Systems/ExternalServices/ExternalServiceProxySystem.cpp
```diff
@@ -35,6 +35,24 @@ namespace csp::systems
...
+    csp::services::ApiBase* InExternalServiceProxyApi, const ExternalServicesOperationParams& Params, StringResultCallback Callback)
+{
+    auto TokenInfo = std::make_shared<chs_aggregation::ServiceRequest>();
+    TokenInfo->SetServiceName(Params.ServiceName);
+    TokenInfo->SetOperationName(Params.OperationName);
+    TokenInfo->SetHelp(Params.SetHelp);
+    TokenInfo->SetParameters(Convert(Params.Parameters));
+
+    csp::services::ResponseHandlerPtr ResponseHandler
+        = InExternalServiceProxyApi->CreateHandler<StringResultCallback, ResultType, void, chs_aggregation::ServiceResponse>(Callback, nullptr);
```
MAG-ElliotMorris, line 52: [Comment] This sort of dispatch could become a neat pattern.

Library/src/Systems/ExternalServices/ExternalServiceProxySystem.cpp
```diff
@@ -35,6 +35,24 @@ namespace csp::systems
...
 inline const char* BoolToApiString(bool Val) { return Val ? "true" : "false"; }
 
+/// @brief Generic function which will make a post request to the services proxy endpoint to trigger some specified operation of specified
+/// service. The class is templated to allow specialized functions to customize the result object used to process the response sent back from
+/// services.
+template <class ResultType>
+void InvokeOperationWithResult(
+    csp::services::ApiBase* InExternalServiceProxyApi, const ExternalServicesOperationParams& Params, StringResultCallback Callback)
+{
+    auto TokenInfo = std::make_shared<chs_aggregation::ServiceRequest>();
```
MAG-ElliotMorris, line 45: [Consider] Constraining the template type some, more for documentation than anything else. A (psuedocode) `static_assert(<type_is<ExternalServiceInvocationResult> || type_is<GetAgoraTokenResult>>, "Type not allowed it should be...")`.


On the other hand, this specific method would work with any sort of dispatch like that, it's just where the method has been implemented that makes me think to constrain it, but I could go either way really. If you don't wish to constrain it, does that imply this method should live elsewhere?

MAG-ElliotMorris (approved):
I am impressed that you discovered this prior to it breaking in prod. Sort of thing that we need well managed specs for.

Library/src/Systems/ExternalServices/ExternalServiceProxySystem.cpp
MAG-SamBirley: I think it could be constrained, but I also imagine a future where we find clients often using some other external service proxy, and we decide to bake it into CSP for a QoL improvement/consistency across clients.  Which would lead to an increasing number of conditions in the constraint.

To me, it feels like keeping it colocated with the system itself is a strong enough expression of it only being intended for this system, whilst keeping it flexible for the future.

PR #818, opened 2025-10-07
[OB-4737] fix: Script execution by elected leader

MAG-SamBirley:
## The Bug
With the realtime engine refactor, we introduced a bug where scripts would not be executed by the script leader unless they were also the owner of the entity hosting the script.

This would manifest to users as scripts not working, until after they touched any state on the entity (at which point CSP claims ownership).

However scripts would then appear to stop work again later, when another user enters the space, becomes the script leader but once again does not have ownership.

## The Fix
The fix is to restore the logic gating script execution; rather than checking for entity ownership, we check to see if we are the elected leader.

Non-online engines, however, should still continue to always run scripts.

Please enjoy this video of scripts once again working as expected (running local CSP changes in OKO Unreal), despite the local user not being the owner of the scripts when first entering.

https://github.com/user-attachments/assets/f1fc495e-123d-4b61-af7a-2a9b7f3f5f0f

### DIFF
--- a/Library/src/Multiplayer/RealtimeEngineUtils.cpp
+++ b/Library/src/Multiplayer/RealtimeEngineUtils.cpp
@@ -242,13 +242,21 @@ std::chrono::system_clock::time_point TickEntityScripts(std::recursive_mutex& En
 
     for (size_t i = 0; i < Entities.Size(); ++i)
     {
-        // Ownership is not a concern for offline realtime engines
-        const bool IsNotOnlineEngine = RealtimeEngineType != csp::common::RealtimeEngineType::Online;
+        bool CanRunScripts = true;
 
-        // The script leader is always responsible for running scripts
-        bool IsLocalClientScriptLeader = ElectionManager.HasValue() && (*ElectionManager)->IsLocalClientLeader();
+        // Note that offline realtime engines may always run scripts.
+        if (RealtimeEngineType == csp::common::RealtimeEngineType::Online)
+        {
+            // If this is an online engine with leadership election enabled, then only the script leader may run scripts.
+            // If there is no leadership election, then we assume all clients may run scripts.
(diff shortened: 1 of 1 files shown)


Library/src/Multiplayer/OfflineRealtimeEngine.cpp
```diff
@@ -58,8 +58,8 @@ void OfflineSpaceEntityEventHandler::OnEvent(const csp::events::Event& InEvent)
 {
     if (InEvent.GetId() == csp::events::FOUNDATION_TICK_EVENT_ID)
     {
-        LastTickTime = RealtimeEngineUtils::TickEntityScripts(EntitySystem->GetEntitiesLock(), EntitySystem->GetRealtimeEngineType(),
-            OfflineRealtimeEngine::LocalClientId(), *EntitySystem->GetAllEntities(), LastTickTime);
+        LastTickTime = RealtimeEngineUtils::TickEntityScripts(
+            EntitySystem->GetEntitiesLock(), EntitySystem->GetRealtimeEngineType(), *EntitySystem->GetAllEntities(), LastTickTime, nullptr);
```
MAG-ElliotMorris, line 62: [fix] As this is an optional to a pointer, this is bug prone, as nullptr might be a valid optional to null, or an invalid optional, depends on how the constructors are written.

Replace with an explicitly constructed empty optional, or just a raw pointer and document nullability is fine, but I prefer optionals for conveyence that emptyness is an expected state.

MAG-SamBirley: Good call. The tests also agree with you 😅 Will fix.

MAG-SamBirley: Fixed in https://github.com/magnopus-opensource/connected-spaces-platform/pull/818/commits/77675f364de2b2a6fa168f4256d32708c45440fc.

MAG-ElliotMorris (approved):

PR #819, opened 2025-10-08
[OB-4420] Add catch-all to allow Unity iOS to load old incompatible spaces

MAG-ElliotMorris:
This is not at all a real fix, it's just a patch we can choose to apply that will allow Unity ios to load old spaces in the same way the other clients currently can.

Testing this would imply either threading the log system down, inventing a whole new callback pattern, or even just moving this catch, which isn't something I'm willing to do for an off book fix like this. 

We should discuss this in standup, not sure if we should merge this or not, just putting it up to have it ready.

See [here ](https://magnopus.atlassian.net/browse/OB-4420?focusedCommentId=412476)for more.

### DIFF
--- a/Library/src/Multiplayer/MCS/MCSTypes.cpp
+++ b/Library/src/Multiplayer/MCS/MCSTypes.cpp
@@ -147,6 +147,15 @@ namespace
                 catch (const std::exception&)
                 {
                 }
+                catch(...)
+                {
+                    /* This sort of exception management isn't really advisable, we need to emit a callback when we find old data, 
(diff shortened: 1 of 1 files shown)


MAG-ElliotMorris:
Hi @SandyMagnopus , sort of surprised to see you auto-assigned. We're you expecting to be a code-owner on CSP files, or is this a configuration error somewhere?

MAG-AdamThorn (approved):

MAG-ElliotMorris:
The ticket to make this actually do error handling is here, this "fix" isn't the end.
https://magnopus.atlassian.net/browse/OF-1781

PR #826, opened 2025-10-09
[NT-0] fix: Update CSP C++ Build documentation

SandyMagnopus:
This PR comes as a follow-up of [this PR](https://github.com/magnopus-opensource/connected-spaces-platform/pull/822), which modified the default Microsoft SDK used to build CSP in C++. It mostly proposes a fresh update to the "Building CSP for C++" user-facing document:
- Mentioned the tutorial was predominantly for Win users explicitly (_Visual Studio_ being taken as an example by default, as specific dependencies need to be installed),
- Updated the screenshots to be C++-specific and for VS users,
- Specified some instructions (_Clang_ and VS Solution configurations in particular),
- Fixed a couple of mistakes (e.g. still mentions of the deprecated "Foundation" instead of "CSP" - which hurts a bit, as I really admire Isaac Asimov),
- Added interactive links to external websites/internal subsections of the tutorial,
- Organised the _Prerequisites_ section into several subsections,
- Bolded the main components to keep in mind while going through the tutorial steps,
- Gave a formatting pass to the whole document.

--------------------------------------------------------------------------------------------------------------------------------------------

Tested using _Markdown Live Preview_ (hopefully the internal interactive links and the screenshot links work!):

**SCREENSHOT 1:**
> [!NOTE]
>image

**SCREENSHOT 2:**
> [!NOTE]
>image

**SCREENSHOT 3:**
> [!NOTE]
>image

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -320,7 +320,7 @@ if not Project then
             linkoptions {
                 "-gdwarf-5",
                 "-gseparate-dwarf", -- preserve debug information (DWARF)
-                "-sSEPARATE_DWARF_URL=../debug/ConnectedSpacesPlatform_WASM.wasm.debug.wasm"
+                "-sSEPARATE_DWARF_URL=../Debug/ConnectedSpacesPlatform_WASM.wasm.debug.wasm"
             }
         filter { "platforms:wasm", "configurations:*Release*" }
             -- We want to reduce the size of Release builds as much as possible


Library/docs/source/manual/building/windows.md
```diff
@@ -1,68 +1,77 @@
...
-Check the last option for `Use custom path to clang-format.exe` and then click browse to locate the executable in your llvm installation folder.
+### SDK versions
 
-Additionally, you will also need two specific SDK versions:
-- Windows SDK version 2004 (10.0.19041.0) - required to build the C++ code.
-- .Net Developer Pack 4.7.1 - required to load and build the C# tests.
+Additionally, you will need three specific Individual Components from the Visual Studio Installer, including two **SDKs**:
+- MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest) - required to compile the C++ code.
+- Windows 10 SDK (10.0.19041.0) - required to build the C++ code.
+- .NET Framework 4.7.1 SDK and targeting pack - both required to load and build the C# tests.
```
MAG-ElliotMorris, line 31: [Comment] I don't think we even have C# tests in the solution anymore. There is still some remnant c# code though so this is probably still a technical requirement, if not a practical one. :(

(I wouldn't bother going to the effort to investigate and rewrite this copy right now)

Library/docs/source/manual/building/cpp.md
```diff
@@ -1,68 +1,77 @@
 # Building CSP for C++
```
MAG-ElliotMorris: [consider] Since this is now explicitly windows (it was before, you've just made it clearer ... or the original authors thought that being vague meant it was a cross-platform guide but no.... it doesn't)

... anyways, since this is now explicitly windows, maybe call this `Building CSP on Windows`. (What does "For C++ mean do you think? I guess the C++ api, but that's rather implied...🤔 )

SandyMagnopus: Yes, I was about to do this change to be fair, as it's more consistent with the other CSP building tutorials we have (as "Building CSP for a **specific platform**"), but the existence of a "Building CSP for C#" tutorial [here](https://connected-spaces-platform.net/manual/building/csharp.html) made me doubt my understanding... Should we remove the latter document, and name this tutorial "Building CSP for Windows" then?

MAG-ElliotMorris: 🤔 There's not really much consistency to be found here. This hasn't been very well thought through. I'd make the change if it were me because at least it makes sense local to this document.

SandyMagnopus: Done in [e0efb6f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/826/commits/e0efb6fea55f3baf010dc32ba4cbca9c8d78a0ee), [17ac06b](https://github.com/magnopus-opensource/connected-spaces-platform/pull/826/commits/17ac06bd2735035bd098aef0c151353bf606aaf7), and [79b3231](https://github.com/magnopus-opensource/connected-spaces-platform/pull/826/commits/79b3231294242a3c8911ef400eabfed9f4b93a92).

PR #829, opened 2025-10-10
[NT-0] test: Mock the Web Client

MAG-AdamThorn:
Our original WebClient Unit Tests were using a third party backend service to validate our requests. However we have found this service to be unstable. 3-6 months ago tests stopped working as the service changed to require an auth token, and last week the service fell over entirely.

As we now have Google Mock setup, we decided to mock the Web Client.

While the WebClient has five public methods, the tests were only really exercising one of them, `SendRequest`. Originally there were many tests, but they went beyond the scope of testing core WebClient functionality, for example also testing polling behaviour of the derived WebClient types - SignalRWebClient and EmscriptenWebClient.

These additional tests have therefore been stripped out, leaving us with two tests that I believe give us the same coverage.

The diff will be very hard to view so probably best to select to 'View file'.


MAG-mv (approved):
All good!

PR #832, opened 2025-10-13
[OB-4711] fix: Asset duplication

MAG-SamBirley:
## The Problem
The services endpoint function signature changed recently, and whilst we have updated CSP to reason about the new API, we had missed some parts, namely:

* The IDs of assets to be duplicated are now expected via a Filters object in the request. We had missed this, so were effectively invoking duplications with _no_ specified assets. (Note: This is assumed to mean the client wishes to copy _all_ assets in a space, which is definitely not intended).
* The DTO returned in responses is no longer arrays of prototypes, but a `CopyPrototypesResult` type - which _is_ effectively still just an array of prototypes, but in a different form, which we had nothing in place to parse.

## The Fix
* Provide the list of asset IDs to duplicate when duplication is invoked https://github.com/magnopus-opensource/connected-spaces-platform/commit/2ca27bfefd8fda66abaf46a09cfdfa188d7aea15
* Parse the `CopyPrototypesResult` on response and populate the array of copied asset collections to be passed to the user https://github.com/magnopus-opensource/connected-spaces-platform/commit/b047eade3c26ff394653e5915b47294d7e6dde14

This change also re-enables the asset duplication test, which was part of why we missed that this issue was occurring.

### DIFF
--- a/Tests/assets/checkpoint-basic.json
+++ b/Tests/assets/checkpoint-basic.json
@@ -3,6 +3,7 @@
     "checkpointTimestamp": "2025-08-27T14:33:22.7204028+00:00",
     "spaceId": "68af162f015bb6793cacf4a2",
     "tenantName": "OKO_TESTS",
+    "exportId": "679aba06b50e18a31a050532",
     "data": {
         "group": {
             "id": "68af162f015bb6793cacf4a2",
--- a/Tests/assets/checkpoint-empty.json
+++ b/Tests/assets/checkpoint-empty.json
@@ -3,6 +3,7 @@
     "checkpointTimestamp": "2025-08-26T16:15:24.6191891+00:00",
     "spaceId": "68addce4985d7612f76b9461",
     "tenantName": "OKO_TESTS",
+    "exportId": "679aba06b50e18a31a050532",
     "data": {
         "group": {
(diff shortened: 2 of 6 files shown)


Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2151,7 +2151,7 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, DownloadAssetDataInvalidURLTest)
     LogOut(UserSystem);
 }
 
-CSP_PUBLIC_TEST(DISABLED_CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
+CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
```
MAG-AdamThorn, line 2154: [fix] Can we add a test/s to validate the unhappy path. What happens if we specify an invalid source asset collection when trying to copy?

MAG-SamBirley: Good call. Tracking this work via a separate ticket to address: https://magnopus.atlassian.net/browse/OF-1792

PR #833, opened 2025-10-13
[OB-4757] fix: Ensure AnalyticsRecord queue is cleared after batch sent

MAG-AdamThorn:
The AnalyticsRecord queue was not being cleared after each batch of records was sent. This has been addressed and tests updated/added to verify that the behaviour is correct.

### DIFF
--- a/Library/src/Systems/Analytics/AnalyticsSystem.cpp
+++ b/Library/src/Systems/Analytics/AnalyticsSystem.cpp
@@ -221,7 +221,7 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
 
     SetTimeSinceLastQueueSend(CurrentTime);
 
-    NullResultCallback SendBatchAnalyticsCallback = [Callback, LogSystem = this->LogSystem](const NullResult& Result)
+    NullResultCallback SendBatchAnalyticsCallback = [this, Callback](const NullResult& Result)
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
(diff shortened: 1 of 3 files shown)


Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -230,16 +230,20 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
...
-            LogSystem->LogMsg(common::LogLevel::Error,
+            this->LogSystem->LogMsg(common::LogLevel::Error,
                 fmt::format("Failed to send Analytics Event. ResCode: {}, HttpResCode: {}", static_cast<int>(Result.GetResultCode()),
                     Result.GetHttpResultCode())
                     .c_str());
         }
 
+        // Note: We may in the future wish to consider a retry mechanism when there is a failure to send the AnalyticsRecordQueue.
+        // For now we are just clearing the queue in either case.
+        this->AnalyticsRecordQueue.clear();
```
MAG-ElliotMorris, line 245: [Comment] Referencing @magnopus-swifty's crusade on error handling, if we'd exposed callbacks (or had an interface that could support exceptions) rather than used logs for this, it would have opened up the possibility of retry handling at the client level. Which perhaps isn't ideal, but is flexible.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -327,6 +327,107 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, FlushAnalyticsEventsQueueTest)
...
+            = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
+
+        AnalyticsSystem->SetTimeSinceLastQueueSend(CurrentTime);
+
+        // Send two analytics events to be queued for sending later as a batch
+        AnalyticsSystem->QueueAnalyticsEvent(TestProductContextSection, TestCategory, TestInteractionType, TestSubCategory, TestMetadata);
+        AnalyticsSystem->QueueAnalyticsEvent(TestProductContextSection, TestCategory, TestInteractionType, TestSubCategory, TestMetadata);
+
+        // Wait for 4 seconds before calling Tick to send the queued events
+        ASSERT_NE(QueueSentFuture1.wait_for(4s), std::future_status::ready) << "Analytics queue should not yet have been sent.";
```
MAG-ElliotMorris, line 388: [Comment] This is a neat inverted use of a future to check something hasn't happened, I like it.

Tests/src/PublicAPITests/AnalyticsSystemTests.cpp
```diff
@@ -327,6 +327,107 @@ CSP_PUBLIC_TEST(CSPEngine, AnalyticsSystemTests, FlushAnalyticsEventsQueueTest)
...
+        csp::CSPFoundation::Tick();
+
+        // Wait for the callback to be received
+        QueueSentFuture1.wait();
+        EXPECT_TRUE(QueueSentFuture1.get() == true);
+
+        // Wait a second to ensure the internal call to FlushAnalyticsEventsQueue() has completed before checking the queue size
+        std::this_thread::sleep_for(1s);
```
MAG-ElliotMorris, line 396: [Consider] Dosen't `FlushAnalyticsEventQueue` take a callback? Can we use that instead of a fragile sleep?

MAG-AdamThorn: It does, but we are not calling it directly. It is being called on Tick. We do not have a 'global' callback.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: Just discussed, I will add it to our CSP Weekly Sync for further discussion.

PR #834, opened 2025-10-14
[OF-1668] chore!: Remove SpaceEntity::ThirdPartyPlatform

MAG-ElliotMorris:
Remove SpaceEntity::ThirdParty platform, a redundant property.

Does not remove the third party platform concept entirely, as that's required for assets (for some reason, this truth sourced from the ticket).

Was quick, so just pulled this out of the backlog and did it in a timeslot I had.

**Breaking Changes**
Removed `SpaceEntity::SetThirdPartyPlatformType`, `SpaceEntity::GetThirdPartyPlatformType`

### DIFF
--- a/Library/include/CSP/Multiplayer/PatchTypes.h
+++ b/Library/include/CSP/Multiplayer/PatchTypes.h
@@ -30,9 +30,8 @@ enum SpaceEntityUpdateFlags
     UPDATE_FLAGS_COMPONENTS = 16,
     UPDATE_FLAGS_SELECTION_ID = 32,
     UPDATE_FLAGS_THIRD_PARTY_REF = 64,
-    UPDATE_FLAGS_THIRD_PARTY_PLATFORM = 128,
-    UPDATE_FLAGS_PARENT = 256,
-    UPDATE_FLAGS_LOCK_TYPE = 512,
+    UPDATE_FLAGS_PARENT = 128,
+    UPDATE_FLAGS_LOCK_TYPE = 256,
 };
 
 /// @brief This Enum should be used to determine what kind of operation the component update represents.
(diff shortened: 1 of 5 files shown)


Library/include/CSP/Multiplayer/PatchTypes.h
```diff
@@ -30,9 +30,8 @@ enum SpaceEntityUpdateFlags
     UPDATE_FLAGS_COMPONENTS = 16,
     UPDATE_FLAGS_SELECTION_ID = 32,
     UPDATE_FLAGS_THIRD_PARTY_REF = 64,
-    UPDATE_FLAGS_THIRD_PARTY_PLATFORM = 128,
-    UPDATE_FLAGS_PARENT = 256,
-    UPDATE_FLAGS_LOCK_TYPE = 512,
+    UPDATE_FLAGS_PARENT = 128,
```
MAG-SamBirley, line 33: [Question] Won't changing the value of `UPDATE_FLAGS_PARENT` (and any subsequent flags) cause serialization issues with older spaces?

MAG-ElliotMorris: I didn't think this was a serialized enum, this is the enum used to tell clients which property has updated in the entity updated callbacks. Is it?

MAG-ElliotMorris: ... This is where that small test suite of old spaces would be helpful, wouldn't need to actually think about this.

MAG-AdamThorn: I have just had a look to confirm. In [SpaceEntityStatePatcher::ApplyPatchFromObjectPatch](https://github.com/magnopus-opensource/connected-spaces-platform/blob/4156af936edc9910b828e6d85224f08999643198/Library/src/Multiplayer/SpaceEntityStatePatcher.cpp#L397) we determine the update flag type based on the information the Patch contains and do not serialise the flag when sending patches/object messages.

So this change looks correct to me.

MAG-SamBirley: Ok, thanks for checking!

PR #835, opened 2025-10-16
[OF-1794] test: support disconnect callback in WASM tests 

MAG-ChristopherAtkinson:
This PR introduces two new tests into the WASM framework for OnRequestToDisconnect from MCS. Several methods can invoke this behavior, but these tests utilize a second connection with a SignalR connection, causing the first to receive the OnRequestToDisconnect event. 

> [!CAUTION]
> These tests use a delay between the main test and the additional test to invoke the OnRequestToDisconnect event; as a result, these could be flaky depending on the length of time needed for CSP to initialize.

### DIFF
--- a/WasmTesting/csp-wasm.test.ts
+++ b/WasmTesting/csp-wasm.test.ts
@@ -154,6 +154,7 @@ test('DisconnectWhenNewMultiplayerSessionInitiatedInSpace', async() => {
     return values;
   });
 
+  assert.ok(consoleMessages.some(e => e.includes('Successfully entered space')));
   assert.ok(consoleMessages.some(e => e.includes('Disconnected from server! Reason: New Multiplayer Session Initiated')));
 
   //Cleanup


WasmTesting/csp-wasm.test.ts
```diff
@@ -126,6 +126,74 @@ test('EnterSpaceFromCheckpoint', async() => {
...
+  function delay(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
+  // Start 2 new signalR connections with the same user without either entering a space.
+  // Delay the second one so we know the first one will be disconnected
+  const [{errors, consoleMessages}] = await Promise.all([
+    LaunchTestPage('http://127.0.0.1:8888/DisconnectWhenNewMultiplayerSessionInitiatedInSpace.html', USE_DEBUG_CSP, { email: user.getProfile().email, password: TEST_ACCOUNT_PASSWORD }, spaceId),
+    (async () => {
+      await delay(3000); // 3 second delay
```
MAG-ElliotMorris, line 143: [Comment] I think this is less a potential flake, and more a false positive scenario.

Consider that what will happen is that when the time between logging in and entering a space takes to long, the other window will have connected, thus sending the disconnect event before the first window has entered the space. This makes this test exactly the same as the out-of-space disconnect test.

This is why I was asking after knowing if people were in a space before entering it yesterday, you could conceivably poll here, in the typescript, for when someone has entered the space, and only then start the second puppeteer window, which would remove the false positive chance.

To prove this test is _actually_ working, you'll need to `assert.ok` on the entered space success message as well, (and probably also check that entering a space comes before the disconnect) I suspect your 100% reliability is mostly down to this.

MAG-ChristopherAtkinson: That is a great point. Did you get a response in the end to how they were querying users in a space? 

Can also add the `assert.ok` on enter space, and retest later today

MAG-ElliotMorris: Yep. https://magnopus.slack.com/archives/C037VC24UE7/p1760526978272179

MAG-ChristopherAtkinson: Hmm interesting. If the quota system is not returning the space owner as in the space, then this will remain at 0 indefinitely

MAG-ChristopherAtkinson: Fixed here: [84dbd4c](https://github.com/magnopus-opensource/connected-spaces-platform/pull/835/commits/84dbd4c814c2c534f90abd7bec617314ea88bd75)

The order is not dependent. However, ran the test 10 times and always received the correct order. I want to revisit this tomorrow and do some additional testing

MAG-ChristopherAtkinson:
Ran the test in the background for a while, the test completed 73 times with 13 failure cases, indicating an error rate of 17.8% in the limited sample. Looking into the coverage of the resulting failure, we are receiving the following in all 13 cases: 

`Blocking on the main thread is very dangerous, see https://emscripten.org/docs/porting/pthreads.html#blocking-on-the-main-browser-thread`

I am uncertain if the blocking is occurring because of the usage of `setTimeout` or if we have some blocking behavior in the `entered space` flow. I took a quick parse over the flow nothing jumped out. It is possible that looping over the test is also adding to this, rather than running them in isolation; e.g. `for /L %%x in (1, 1, 250) do CALL yarn test` in run-csp-wasm-tests.cmd

Following this, I would like to do some further testing once the [Cross-thread Callbacks Issue](https://github.com/magnopus-opensource/connected-spaces-platform/pull/806) has been merged, as this could be potentially adjacent.

PR #837, opened 2025-10-16
[OB-4773] fix: Fix Analytics queue not being cleared correctly

MAG-AdamThorn:
The Analytics Queue is currently being cleared via a callback which is passed to the CHS analyticsBulkPost endpoint.

The issue is that on Tick, we check to see if the queue size has reached the MaxQueueSize. This means that there is a period of time between the queue size reaching the MaxQueueSize and the callback returning and clearing the queue.

During this period the tick method will continue to evaluate the queue size as being greater than the MaxQueueSize, which will result in the queue being sent each tick.

This ticket fixes the issue by creating a local moved copy of the queue, which is then passed to the MCS analyticsBulkPost endpoint. The tests have also been updated to mimic the issue that the Dakoko team were seeing, to ensure that the queue is being correctly cleared after each batch has been sent.

@MAG-ME has verified that this fix is working via a local build.

### DIFF
--- a/Library/src/Systems/Analytics/AnalyticsSystem.cpp
+++ b/Library/src/Systems/Analytics/AnalyticsSystem.cpp
@@ -216,9 +216,6 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
         return;
     }
 
-    // Create a local copy of the queue to be sent so that the original queue can be cleared.
-    std::vector<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>> AnalyticsRecordQueueCopy(std::move(AnalyticsRecordQueue));
-
     const std::chrono::milliseconds CurrentTime
         = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
 
@@ -252,7 +249,11 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
     csp::services::ResponseHandlerPtr ResponseHandler
         = AnalyticsApi->CreateHandler<NullResultCallback, NullResult, void, chs::AnalyticsRecord>(SendBatchAnalyticsCallback, nullptr);
 
-    static_cast<chs::AnalyticsApi*>(AnalyticsApi.get())->analyticsBulkPost({ AnalyticsRecordQueueCopy }, ResponseHandler);
+    static_cast<chs::AnalyticsApi*>(AnalyticsApi.get())->analyticsBulkPost({ AnalyticsRecordQueue }, ResponseHandler);
+
+    // Clear the analytics record queue.
+    // The async analyticsBulkPost endpoint serializes the records data to json so it is safe to clear here.
+    AnalyticsRecordQueue.clear();
 }
 
 } // namespace csp::systems


Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -230,20 +233,16 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
...
-            this->LogSystem->LogMsg(common::LogLevel::Error,
+            LogSystem->LogMsg(common::LogLevel::Error,
                 fmt::format("Failed to send Analytics Event. ResCode: {}, HttpResCode: {}", static_cast<int>(Result.GetResultCode()),
                     Result.GetHttpResultCode())
                     .c_str());
         }
 
-        // Note: We may in the future wish to consider a retry mechanism when there is a failure to send the AnalyticsRecordQueue.
-        // For now we are just clearing the queue in either case.
-        this->AnalyticsRecordQueue.clear();
```
MAG-mv, line 245: [Question] What made you move into another vector instead of putting the clear after the analyticsBulkPost call?

MAG-AdamThorn: That is an excellent point, it is an async call, but the analyticsBulkPost method will serialise the vector of records to a string before sending (and the vector going out of scope). Let me confirm that change works and make the change.

Library/src/Systems/Analytics/AnalyticsSystem.cpp
```diff
@@ -216,12 +216,15 @@ void AnalyticsSystem::FlushAnalyticsEventsQueue(NullResultCallback Callback)
         return;
     }
 
+    // Create a local copy of the queue to be sent so that the original queue can be cleared.
+    std::vector<std::shared_ptr<csp::services::generated::userservice::AnalyticsRecord>> AnalyticsRecordQueueCopy(std::move(AnalyticsRecordQueue));
```
MAG-ChristopherAtkinson: [Question] won't this go out of scope, as we capture this in the async call as a reference, this should be a problem

MAG-ChristopherAtkinson (changes requested):

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: I have made that change: [f1f42f0](https://github.com/magnopus-opensource/connected-spaces-platform/pull/837/commits/f1f42f00e4d399a88766945d8e605337b0b90d9a)

Library/src/Systems/Analytics/AnalyticsSystem.cpp
MAG-AdamThorn: @MAG-ChristopherAtkinson the async analyticsBulkPost endpoint serializes the data to json before returning so it was safe to do it that way. However, I have made a change as per Matt's suggestion above and have removed that local copy in favour of clearing the queue after the call to analyticsBulkPost.

Change: [f1f42f0](https://github.com/magnopus-opensource/connected-spaces-platform/pull/837/commits/f1f42f00e4d399a88766945d8e605337b0b90d9a)

MAG-ChristopherAtkinson: That is great, thank you for the clarification

PR #838, opened 2025-10-20
[OF-1774] feat: stereo video component data

MAG-ChristopherAtkinson:
The video component has been updated to include a new property to allow clients to reason about if a video supports stereo. The changes introduce the following:

 - New enum  `StereoVideoType` introduce 
 - The property has been exposed through the `VideoPlayerSpaceComponentScriptInterface`
 - Test have been updated to reflect the property and script interface

### DIFF
--- a/Library/include/CSP/Multiplayer/Components/VideoPlayerSpaceComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/VideoPlayerSpaceComponent.h
@@ -236,11 +236,11 @@ public:
     void SetVideoPlayerSourceType(VideoPlayerSourceType Value);
 
     /// @brief Gets the type of stereo the video of this component uses.
-    /// @return The type of video stereo used by this component.
+    /// @return The type of stereo used by this video.
     StereoVideoType GetStereoVideoType() const;
 
     /// @brief Sets the type of stereo the video of this component uses.
-    /// @param Value The type of video stereo used by this component.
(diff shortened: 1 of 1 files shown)


Library/include/CSP/Multiplayer/Components/VideoPlayerSpaceComponent.h
```diff
@@ -225,6 +235,14 @@ class CSP_API VideoPlayerSpaceComponent : public ComponentBase, public IEnableab
     /// @param Value The type of video source used by this component.
     void SetVideoPlayerSourceType(VideoPlayerSourceType Value);
 
+    /// @brief Gets the type of stereo the video of this component uses.
+    /// @return The type of video stereo used by this component.
```
MAG-AdamThorn: [Consider] I found this slightly hard to parse, consider "The type of stereo used by this video."

MAG-ChristopherAtkinson: Updated here: [55e39e4](https://github.com/magnopus-opensource/connected-spaces-platform/pull/838/commits/55e39e43d0a808b09011570100f693f89fdc7311)

Also updated the set function to match.

MAG-mv (commented):
[Question] Do we need to add support for scripting?

MAG-mv (approved):

PR #839, opened 2025-10-21
[OF-1792] test: Asset duplication unnhappy paths

MAG-SamBirley:
Adding three further tests to validate CSP behavior when provided with unexpected parameters (or combinations thereof):

* What happens when the user provides malformed asset collection IDs.
* What happens when the provided destination space ID is malformed.
* What happens if the user attempts to invoke a copy with the same space as the target.


Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2298,6 +2294,35 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
...
+    {
+        printf("Validating we encounter failures when providing malformed asset collection IDs...\n");
+
+        csp::systems::AssetCollection FirstSpaceAssetCollection;
+        FirstSpaceAssetCollection.SpaceId = SourceSpace.Id;
+        FirstSpaceAssetCollection.Id = "AnInvalidlId";
+
+        const csp::common::Array<csp::systems::AssetCollection> AssetCollections = { FirstSpaceAssetCollection };
+        auto [Result] = AWAIT_PRE(AssetSystem, CopyAssetCollectionsToSpace, RequestPredicate, AssetCollections, DestSpace.Id, false);
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
```
MAG-ChristopherAtkinson, line 2332: [Comment] Shame we don't have a failure reason from CHS about why the return failed, would be good to capture to ensure we cover all potential failure paths

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2298,6 +2294,35 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
...
+        printf("Validating we encounter failures when a malformed destination space ID is provided...\n");
+
+        const csp::common::String InvalidDestSpaceId = "AnInvalidlId";
+        const csp::common::Array<csp::systems::AssetCollection> AssetCollections = { SourceAssetCollection };
+        auto [Result] = AWAIT_PRE(AssetSystem, CopyAssetCollectionsToSpace, RequestPredicate, AssetCollections, InvalidDestSpaceId, false);
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
+    }
+
+    {
+        printf("Validating that we cannot copy an asset to the same space...\n");
```
MAG-ChristopherAtkinson, line 2354: [Question] can we guard against this at the CSP level to prevent needing to make a RESTful API call

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2298,6 +2294,35 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
...
+
+        const csp::common::Array<csp::systems::AssetCollection> AssetCollections = { SourceAssetCollection };
+        auto [Result] = AWAIT_PRE(AssetSystem, CopyAssetCollectionsToSpace, RequestPredicate, AssetCollections, SourceSpace.Id, false);
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
+    }
+
```
MAG-ChristopherAtkinson, line 2365: [Question] What happens if we duplicate an asset into a space when the asset or equivalent already exists in the space? Worth checking that CHS will not perform the operation or replace the existing asset?

MAG-ChristopherAtkinson (commented):
Overall looking great, just a few minor questions and comments.

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: I wouldn't, as there's a high change in the near future that [we will start supporting this](https://magnopus.atlassian.net/browse/CHS-5370).

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Yes, that's been discussed with our services team. Their implementation guarantees that the copies will lead to new prototypes being generated (which cannot clash with another).

MAG-ChristopherAtkinson: Thank you for confirming that

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2298,6 +2294,35 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
         EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
     }
 
+    {
```
MAG-ElliotMorris, line 2322: [Comment] Don't change as it is the pattern here, but using scope blocks rather than declaring new tests is a poor decision. You lose observability in what particular thing is failing from dashboards and the like, for little positive tradeoff that I can tell.

Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2298,6 +2294,35 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, CopyAssetCollectionTest)
...
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
+    }
+
+    {
+        printf("Validating we encounter failures when a malformed destination space ID is provided...\n");
+
+        const csp::common::String InvalidDestSpaceId = "AnInvalidlId";
+        const csp::common::Array<csp::systems::AssetCollection> AssetCollections = { SourceAssetCollection };
+        auto [Result] = AWAIT_PRE(AssetSystem, CopyAssetCollectionsToSpace, RequestPredicate, AssetCollections, InvalidDestSpaceId, false);
+        EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Failed);
```
MAG-ElliotMorris, line 2346: [Question] Do any of these failures log anything specific (ie, not the generic HTTP loggers)? If so, [Consider] testing that with the mocklogger.

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-ElliotMorris, line 2332: [consider] I'd like to see more assertions here. Things that come to mind.

- Behaviour of "AssetCollectionResult.GetAssetCollection" in failure cases. (I'm presuming a default constructed AssetCollection, given the signiature?)
- Demonstrating the values of the other Result base properties, particularly HttpResultCode and FailureReason.

Tests are more than just exercising a behaviour path, they're also the mechanism by which we understand and document .

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-ElliotMorris: Be good to put that in a docstring if it isn't already. Think about how anyone but us could ever learn that.

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Agreed.

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Great catch, thanks. The answer is yes, there are CSP-emitted logs we can validate.

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Agreed. Will do so.

MAG-SamBirley: Ok, addressed (see over [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/839#issuecomment-3431363047) for more info).

Tests/src/PublicAPITests/AssetSystemTests.cpp
MAG-SamBirley: Ok, addressed (see over [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/839#issuecomment-3431363047) for more info).

PR #840, opened 2025-10-22
[OF-1736] feat: Expose AsyncCall to DuplicateSpace

MAG-AdamThorn:
The current `DuplicateSpace` method is a synchronous operation that can timeout and fail if the Space is overly complex or the services are under excessive load.

A new generic `AsyncCallCompleted` event has been added by CHS to alert the user when a specific async operation has completed. Operations such as DuplicateSpace can take a `AsyncCall` bool argument, which determines whether the operation should be performed synchronously or asynchronously.
If performed asynchronously the call will immediately return a result object, which in the case of Duplicate Space will contain a `204 - No content` response. Then, once the actual duplication operation has been completed, the `AsyncCallCompleted` event will fire, passing a `AsyncCallCompletedEventData` object.

As part of this work the current `DuplicateSpace` method has been marked deprecated and a new `DuplicateSpaceAsync` method has been introduced. This approach has been taken for two reasons: 

1. 
CHS intend to remove the synchronous flow.
2. The async flow returns a different result object - SpaceResult vs NullResult.

A new `AsyncCallCompletedCallbackHandler` has been introduced along with a public setter to allow clients to register for the AsyncCallCompleted event: `SetAsyncCallCompletedCallback()`.

As noted above, this event will return a generic 'AsyncCallCompletedEventData` object, which in the case of the duplicate Space operation, will contain the following data:

- OperationName: "DuplicateSpaceAsync"
- ReferenceId: "Id of the new Space"
- ReferenceType: "GroupId"

### DIFF
--- a/Library/src/Systems/Spaces/SpaceSystem.cpp
+++ b/Library/src/Systems/Spaces/SpaceSystem.cpp
@@ -108,7 +108,16 @@ SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient, multiplayer::NetworkE
     SpaceAPI = new chsaggregation::SpaceApi(InWebClient);
 }
 
-SpaceSystem::~SpaceSystem() { delete (GroupAPI); }
+SpaceSystem::~SpaceSystem()
+{
+    delete (GroupAPI);
+
+    if (EventBusPtr)
+    {
+        EventBusPtr->StopListenNetworkEvent(csp::multiplayer::NetworkEventRegistration("CSPInternal::SpaceSystem",
+            csp::multiplayer::NetworkEventBus::StringFromNetworkEvent(csp::multiplayer::NetworkEventBus::NetworkEvent::AsyncCallCompleted)));
+    }
+}
 
 /* CreateSpace Continuations */
 async::task<SpaceResult> SpaceSystem::CreateSpaceGroupInfo(


Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -344,21 +344,55 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
...
     /// @param NewAttributes csp::systems::SpaceAttributes : Attributes to apply to the duplicated space.
-    /// @param MemberGroupIds csp::common::Array<csp::common::String> : An optional array of group (space) IDs to copy users from.
+    /// @param MemberGroupIds csp::common::Array<csp::common::String> : An optional array of group (space) IDs. Members of these groups will be added
+    /// to the duplicated space with the same roles.
     /// @param ShallowCopy bool : If true, the duplicated space will reference the assets of the original space. Otherwise, all assets will be
     /// duplicated.
-    /// @param Callback NullResultCallback : callback when asynchronous task finishes
-    CSP_ASYNC_RESULT void DuplicateSpace(const csp::common::String& SpaceId, const csp::common::String& NewName, SpaceAttributes NewAttributes,
+    /// @param Callback SpaceResultCallback : callback when asynchronous task finishes.
+    [[deprecated(
```
MAG-AdamThorn: The wrapper generator didn't like this. I have instead adopted doxygens style of marking methods as deprecated - result of this shown in the image below.

Change here: [1995cdd](https://github.com/magnopus-opensource/connected-spaces-platform/pull/840/commits/1995cdd318fe3ef7f41f71c8ba0ffc5640388d67)

image

PR #841, opened 2025-10-24
[NT-0] fix: Removing an asset collection assert

MAG-SamBirley:
## The Issue
A CSP client app developer recently reported a crash when attempting to enter a space. Turns out it was due to there being an asset collection/prototype in the space that did not conform to the types CSP expects, which caused CSP to fire an assert.

## The Fix
To fix, we follow the same pattern as with [Asset.cpp](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/src/Systems/Assets/Asset.cpp#L60) and emit an `Error` log message instead of an assert (see [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/841/commits/701d334a0d6f611bf35b77fa2f2715ec81838357)).

## The Tests
Initial test implementation to validate type parsing for the DTO came in [here](https://github.com/magnopus-opensource/connected-spaces-platform/commit/f01ef4fcc59598299133f71dad4887bf2670c94d). Then refined that to use the test matrix approach [here](https://github.com/magnopus-opensource/connected-spaces-platform/commit/5736afed9187702957321968b5e4494d8b5c5c2e#diff-6f68134c97a4a18072df386d3506e61b6665f0825b0275f45182b7b44e392964).

## Bonus
The tests highlighted that CSP was (not unreasonanbly) assuming certain fields would **always** exist on prototype DTOs. Problem was if they didn't exist, it would lead to an uncaught exception. To fix, we now follow the same pattern for all prototype DTO fields, first testing for then existence, and only then retrieving their value.

Added bonus, whilst in the area I elected to nuke the `Charity` DTO type CSP was checking for, as it was agreed we should remove it in [a recent CSP working group session](https://docs.google.com/document/d/16Atmb4M-cJrfvQ8dfcbxjFHSeFqbyGwj_zczqxMGGWA/edit?tab=t.0#heading=h.y8827ltlxcqu). 

Both bonus changes can be seen [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/841/commits/d8a8e80e0f8e3c3c1bde85e9aff376ba82279233).

### DIFF
--- a/Tests/src/PublicAPITests/AssetSystemTests.cpp
+++ b/Tests/src/PublicAPITests/AssetSystemTests.cpp
@@ -2488,7 +2488,8 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, GetAssetCollectionCountTest)
     LogOut(UserSystem);
 }
 
-class AssetCollectionTypeDtoMock : public PublicTestBaseWithParam<std::tuple<csp::common::String, csp::systems::EAssetCollectionType>>
+class AssetCollectionTypeDtoMock
+    : public PublicTestBaseWithParam<std::tuple<csp::common::String, csp::systems::EAssetCollectionType, csp::common::String>>
 {
 };
 
@@ -2501,6 +2502,7 @@ TEST_P(AssetCollectionTypeDtoMock, AssetCollectionTypeDtoMockTest)
     // Test parameters
     const csp::common::String& DtoTypeString = std::get<0>(GetParam());
     const csp::systems::EAssetCollectionType ExpectedAssetCollectionType = std::get<1>(GetParam());
+    const csp::common::String& ExpectedLogMessage = std::get<2>(GetParam());
 
     const csp::common::String& MockAssetCollectionId = "1234";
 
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/AssetSystemTests.cpp
```diff
@@ -2481,4 +2486,70 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, GetAssetCollectionCountTest)
...
+    EXPECT_EQ(Result.GetAssetCollection().Type, ExpectedAssetCollectionType);
+}
+
+INSTANTIATE_TEST_SUITE_P(AssetSystemTests, AssetCollectionTypeDtoMock,
+    testing::Values(std::make_tuple("Default", csp::systems::EAssetCollectionType::DEFAULT),
+        std::make_tuple("FoundationInternal", csp::systems::EAssetCollectionType::FOUNDATION_INTERNAL),
+        std::make_tuple("CommentContainer", csp::systems::EAssetCollectionType::COMMENT_CONTAINER),
+        std::make_tuple("Comment", csp::systems::EAssetCollectionType::COMMENT),
+        std::make_tuple("SpaceThumbnail", csp::systems::EAssetCollectionType::SPACE_THUMBNAIL),
+        std::make_tuple("NotARealType", csp::systems::EAssetCollectionType::DEFAULT)));
```
MAG-AdamThorn: [fix] For this last one we would expect the unknown prototype error to be logged, is that correct? I think it would be worth adding a test that uses the RAIIMockLogger to confirm that.

MAG-SamBirley: Yep, I can do that.

MAG-SamBirley: Addressed in https://github.com/magnopus-opensource/connected-spaces-platform/pull/841/commits/9b3605aa16a5374fc98013fe901c8c2b70a61361.

MAG-mv (approved):

PR #842, opened 2025-10-27
[OF-1784] feat: Multiplayer System

MAG-mv:
This is part 1 of the server-side leader election work, which contains a new MultiplayerSystem which exposes the necessary functionality for leader election from the mcs MultiplayerServices api.

This work allows us to:
- GetScopesInSpce - This is needed to get the current scope information
- UpdateScope - This is needed to be able to turn on leader election
- PerformLeaderElectionInScope-  This is added for testing, as leader election is done automatically

While setting up the MultiplayerSystem, I ran into an issue with our URIs, which is fixed within this ticket. We were conflating the MultiplayerConnection Uri and the MultiplayerServices uri, which should be separate. This has been fixed by creating 2 separate uris. and also applying the root uri transform on both uris, as they both need this.

### DIFF
--- a/Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
+++ b/Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
@@ -144,7 +144,7 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerSystemTests, GetScopeByInvalidSpaceTest)
     // Get the default scope
     auto [GetScopesResult] = AWAIT_PRE(MultiplayerSystem, GetScopesBySpace, RequestPredicate, "INVALID_SPACE_ID");
     EXPECT_EQ(GetScopesResult.GetResultCode(), csp::systems::EResultCode::Failed);
-    EXPECT_EQ(GetScopesResult.GetHttpResultCode(), 400);
+    EXPECT_EQ(GetScopesResult.GetHttpResultCode(), 0);
     EXPECT_EQ(GetScopesResult.GetScopes().Size(), 0);
 
     // Clean up


Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
```diff
@@ -0,0 +1,89 @@
...
+class CSP_API CSP_NO_DISPOSE MultiplayerSystem : public SystemBase
+{
+public:
+    CSP_NO_EXPORT MultiplayerSystem(csp::web::WebClient* WebClient, csp::common::LogSystem& LogSystem);
+    MultiplayerSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
+    ~MultiplayerSystem();
+
+    /// @brief Gets all scopes associated with the given space id.
+    /// @param SpaceId const csp::common::String& : The id of the space we want to get scopes for.
+    /// @param Callback csp::systems::ScopesResultCallback : Callback when asynchronous task finishes.
```
MAG-ElliotMorris: [Consider] "Callback when asynchronous task finishes." is almost useless doc. I know it's everywhere but can we make this even slightly helpful?

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
```diff
@@ -0,0 +1,89 @@
...
+    /// @pre Must already have entered the space of the SpaceId parameter.
```
MAG-ElliotMorris, line 58: [Consider] Stating what happens when this precondition is violated.

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
```diff
@@ -0,0 +1,89 @@
...
+    CSP_ASYNC_RESULT void GetScopesBySpace(const csp::common::String& SpaceId, ScopesResultCallback Callback);
+
+    /// @brief Updates Data on a scope
+    /// @param ScopeId const csp::common::String& : The id of the scope we want to update.
+    /// @param Scope const csp::systems::Scope& : Scope containing new values for the given id.
+    /// @param Callback csp::systems::ScopesResultCallback : Callback when asynchronous task finishes.
+    /// @pre Must already have entered the space of the SpaceId parameter.
+    CSP_ASYNC_RESULT void UpdateScopeById(const csp::common::String& ScopeId, const csp::systems::Scope& Scope, ScopeResultCallback Callback);
+
+    /// @brief Gets details about a scope leader.
```
MAG-ElliotMorris, line 69: [Consider] Can you be more specific? What sort of details? Why might I want to do this? This wouldn't help me very much if I'm browsing the docs as it stands.

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
```diff
@@ -0,0 +1,89 @@
...
+    MultiplayerSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
+    ~MultiplayerSystem();
+
+    /// @brief Gets all scopes associated with the given space id.
+    /// @param SpaceId const csp::common::String& : The id of the space we want to get scopes for.
+    /// @param Callback csp::systems::ScopesResultCallback : Callback when asynchronous task finishes.
+    /// @pre Must already have entered the space of the SpaceId parameter.
+    CSP_ASYNC_RESULT void GetScopesBySpace(const csp::common::String& SpaceId, ScopesResultCallback Callback);
+
+    /// @brief Updates Data on a scope
```
MAG-ElliotMorris: [Consider] Explaining why anyone might ever want to do this.

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
```diff
@@ -0,0 +1,89 @@
...
+
+    /// @brief Gets details about a scope leader.
+    /// @param ScopeId const csp::common::String& : The id of the scope we want to get scope leader details about.
+    /// @param Callback csp::systems::ScopesResultCallback : Callback when asynchronous task finishes.
+    /// @pre "ManagedLeaderElection" should be set to true on the scope, otherwise this function will fail with a EResultCode::Failed response.
+    /// @pre Must already have entered the space of the SpaceId parameter.
+    CSP_ASYNC_RESULT void GetScopeLeader(const csp::common::String& ScopeId, ScopeLeaderResultCallback Callback);
+
+    /// @brief Starts leader election for the given scope.
+    /// This should not need to be called outside of testing, as leader election is done automatically.
```
MAG-ElliotMorris, line 79: [Consider] Maybe we should do the `__` thing for the naming that we've been talking about for this? Heck, i'd even feature flag it.

Library/include/CSP/Systems/Multiplayer/Scope.h
```diff
@@ -0,0 +1,124 @@
...
+class CSP_API Scope
+{
+public:
+    csp::common::String Id;
+    csp::common::String ReferenceId;
+    csp::common::String ReferenceType;
+    csp::common::String Name;
+    PubSubModelType PubSubType = PubSubModelType::Global;
+    double SolveRadius = 0.0;
+    bool ManagedLeaderElection = false;
```
MAG-ElliotMorris, line 81: [Fix] You need to doc what these values are as this is public API. `ReferenceId` and `ReferenceType` especially are hard to deduce. `SolveRadius` could really do with elaboration too.

Library/include/CSP/Systems/Multiplayer/Scope.h
```diff
@@ -0,0 +1,124 @@
...
+};
+
+void DtoToScope(const csp::services::generated::multiplayerservice::ScopeDto& Dto, csp::systems::Scope& ScopeLeader);
```
MAG-ElliotMorris, line 84: [Question] This the normal pattern of just having a free method like this? Seems messy.

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
```diff
@@ -0,0 +1,81 @@
...
+/// @ingroup Multiplayer System
+/// @brief Data representation for a scope leader.
+/// A scope leader represents a user which owns a specific scope in a space.
+/// The scope leader will run scripts and other operations for the scope.
+class CSP_API ScopeLeader
+{
+public:
+    csp::common::String ScopeId;
+    csp::common::String ScopeLeaderUserId;
+    bool ElectionInProgress = false;
```
MAG-ElliotMorris, line 54: [fix] I won't markup every undocumented property.

I'm not trying to be docs mad, I understand in these DTO types some stuff is obvious, but here for example, `ElectionInProgress` demands explanation. I'm presuming it's in-progress at the time the DTO was generated, folks might mistake it for a live marker property. Does this being `true` mean the user should assume the ScopeLeader is about to change and should re-query? etc, etc.

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+MultiplayerSystem::MultiplayerSystem()
+    : SystemBase(nullptr, nullptr, nullptr)
+{
+}
+
+MultiplayerSystem::~MultiplayerSystem()
+{
+    delete ScopeLeaderApi;
+    delete ScopesApi;
```
MAG-ElliotMorris: [fix] You can store these as unique_ptrs, no need to propogate this old pattern.

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+void MultiplayerSystem::GetScopesBySpace(const csp::common::String& SpaceId, ScopesResultCallback Callback)
+{
+    constexpr const char* ReferenceType = "GroupId";
```
MAG-ElliotMorris, line 34: [Consider] Thinking of/Finding a central location for magic strings like this (Unless it is specific to _just_ this method). At least doc why this is what it is/is it arbitrary or magic, if nothing else.

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
```diff
@@ -0,0 +1,91 @@
...
+    auto PubSubModelDto = std::make_shared<chs::multiplayerservice::PubSubModel>();
+
+    if (Scope.PubSubType == PubSubModelType::Global)
+    {
+        PubSubModelDto->SetValue(chs::multiplayerservice::PubSubModel::ePubSubModel::GLOBAL);
+    }
+    else if (Scope.PubSubType == PubSubModelType::Object)
+    {
+        PubSubModelDto->SetValue(chs::multiplayerservice::PubSubModel::ePubSubModel::OBJECT);
+    }
```
MAG-ElliotMorris: [nit] else-if without a final else is a bit of an unstable pattern under change. (Modern languages error in these sorts of cases).

Library/src/Systems/SystemsManager.cpp
```diff
@@ -151,6 +154,8 @@ SystemsManager::SystemsManager()
     , HotspotSequenceSystem(nullptr)
     , AnalyticsSystem(nullptr)
     , ExternalServiceProxySystem(nullptr)
+    , ConversationSystem(nullptr)
```
MAG-ElliotMorris: [Comment] Oops...

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
```diff
@@ -0,0 +1,277 @@
...
+    EXPECT_EQ(GetScopeLeaderResult.GetResultCode(), csp::systems::EResultCode::Failed);
+    EXPECT_EQ(GetScopeLeaderResult.GetHttpResultCode(), 400);
+
+    // Clean up
+    auto [ExitResult] = AWAIT_PRE(SpaceSystem, ExitSpace, RequestPredicate);
+    DeleteSpace(SpaceSystem, Space.Id);
+    LogOut(UserSystem);
+}
+
+// TODO: test out of space behaviour
```
MAG-ElliotMorris: [Consider] Just marking this up in case you forgot. I don't have a good sense for how much testing this system demands to be confident in correctness.

Library/include/CSP/Systems/Multiplayer/Scope.h
```diff
@@ -0,0 +1,124 @@
...
+/// @brief Enum representing the scopes pub/sub model type
+/// Object: used in object scopes, each object is published to its own channel, and
+/// client subscribes to the channels of only the objects they can see
+/// Global: used in global scopes, all objects are published to a single channel,
+/// client subscribes to the channel and can see everything in the channel/scope
+enum class PubSubModelType
+{
+    Object,
+    Global
+};
```
MAG-ChristopherAtkinson, line 52: [Question] is this limited to the multiplayer API? I had a quick search over the Swagger documentation and seems to be at first glance.

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
```diff
@@ -0,0 +1,81 @@
...
+};
+
+void DtoToScopeLeader(const csp::services::generated::multiplayerservice::ScopeLeaderDto& Dto, csp::systems::ScopeLeader& ScopeLeader);
```
MAG-ChristopherAtkinson, line 57: [Consider] moving to a anonymous namespace function

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
MAG-ChristopherAtkinson, line 74: [Consider] would like to see a switch statement here, which would allow for warning as errors to flag if all potential cases are not stratified

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
```diff
@@ -0,0 +1,277 @@
...
+    auto [GetScopesResult] = AWAIT_PRE(MultiplayerSystem, GetScopesBySpace, RequestPredicate, Space.Id);
+    EXPECT_EQ(GetScopesResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+    if (GetScopesResult.GetScopes().Size() != 1)
+    {
+        FAIL();
+    }
+
+    // Ensure default scope has correct default values.
+    const csp::systems::Scope& DefaultScope = GetScopesResult.GetScopes()[0];
```
MAG-ChristopherAtkinson, line 68: [Question] Does CHS guarantee the order scopes are returned in, e.g. these are not ordered by last modified

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
```diff
@@ -0,0 +1,277 @@
...
+    EXPECT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+    // Get the default scope
+    auto [GetScopesResult] = AWAIT_PRE(MultiplayerSystem, GetScopesBySpace, RequestPredicate, Space.Id);
+    EXPECT_EQ(GetScopesResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+    if (GetScopesResult.GetScopes().Size() != 1)
+    {
+        FAIL();
+    }
```
MAG-ChristopherAtkinson, line 299: [Question] This assumes that the only scope being returned is the default; it is possible that in the future we may have several scopes by default, which would cause this test to be invalid. Do we know if any additional scope is intended to be introduced? (Thinking that POI could have an impact here)

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309 by setting these types NO_EXPORT, until we understand how scopes are going to be used in CSP

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309 by setting these types NO_EXPORT, until we understand how scopes are going to be used in CSP

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309 by setting these types NO_EXPORT, until we understand how scopes are going to be used in CSP

Library/include/CSP/Systems/Multiplayer/Scope.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309

Library/include/CSP/Systems/Multiplayer/Scope.h
MAG-mv: Yes, we do this for all dtos. We previously had them in an unnamed namespace, be we needed to use them for the Checkpoint deserialization, so they were moved to header files.

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/2553f80b02ba6265ed9fa12fe0f3883c2cc1e309

Library/include/CSP/Systems/Multiplayer/Scope.h
MAG-mv: Yes, scopes are only used in multiplayer

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
MAG-mv: This is how other dto conversion functions were done previously. However, we needed to change this because they were needed for checkpoint file deserialization. This may be needed in the future, as well

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-mv: Unsure, but there is always 1 default scope, currently.

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-mv: None that I know of. However, I think this is useful to keep here so we know that the behaviour has been changed

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/3a56cf42cc3a7f0c2fdac51ce6701928ec910a6c

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/c5e5ae65afed9fe7c28d5e996b70810e4e465ad5

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/eb537eebbc7497ebfca281571092426dc22e1465

MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/eb537eebbc7497ebfca281571092426dc22e1465

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
MAG-mv: Added docs. I can't think of any other magic string, and I dont ant to create a file for one string at this point. https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/f99c3dd903ae7149a4dfce10d69c24d12fa8f7b8

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-ElliotMorris: [Comment] I still think this erroring behaviour is poor, an empty array of scopes is not a good marker for "You were not in the space", it could be caused by anything. I'm presuming we don't get a helpful error from CHS telling us why the request failed, which is what I'd ideally like.

[Question[ To give better errors, would we have to query CHS again to check if we're in the space beforehand? Seems like context the forwarded CHS response should be able to provide.

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-ElliotMorris: [Comment] I suppose this is technically acceptable. I suspect what's going to happen though is when we find we do need this functionality, someone is just going to toggle this and we won't actually improve the documentation to be good enough :(.

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-ChristopherAtkinson: Can we check with MCS as a precaution?

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-ChristopherAtkinson: Can we check with MCS as a precaution?

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-ElliotMorris: [Question] I was under the impression we agreed on the `__` naming convention for methods that only make sense in a test context going forwards. Am I mistaken in my understanding of either our new standard or what this method is? We can't just agree to do things and then not do them. I'm fine with this code, but you'll need to bring a reason why you didn't want to use the naming convention to the next sync.

Library/include/CSP/Systems/Multiplayer/Scope.h
MAG-ElliotMorris: 👍 Probably want to organize this better in the future, it's the same for callbacks, all scattered about. We should be able to know what symbols are in what namespace more easily.

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
MAG-ElliotMorris: [Consider] This still isn't good enough. 

> /// @brief Whether there is a server side election currently in progress when this object is received.

and 

> bool ElectionInProgress = false;

mean practically the same thing.

Think about if you were integrating against our Api. Think of a reason why you would call `GetScopeLeader`, and you got a result with `ElectionInProgress = true` on it, what would that mean to you, how are you supposed to react? Is the scope leader you have queried about imminently about to be dethroned? Is it guarenteed they will be removed, or just a possibility? Will there be an event when the election is finished, what should I do about that?

I suspect the answer is that you just can't know, as it is with all the other docs, and the NO_EXPORT "strategy".

@MAG-SamBirley @MAG-AdamThorn It is quite frustrating to have to direct my issues at Matt here, it's not his failing. All this is a consequence from CHS not having adequate api documentation that considers workflows, as they are the only people who can answer these sorts of questions. There's only so far we can go with documenting Api's like this and it's difficult to impossible to maintain quality standards in this environment. 😢 

I'm tempted to just give up on documentation for the sinkhole REST endpoints, or at least the stuff that CSP dosen't insert opinions on (most of it). It's a losing battle.

Library/src/Systems/Multiplayer/MultiplayerSystem.cpp
MAG-ElliotMorris: [Consider] The most important piece of information about this string is whether it is magic or not, ie, could you put anything here, or does it relate to other data? Ie is it something that CHS or we populate so it _has_ to say exactly "GroupId". That's not expressed in the comment.

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-mv: @jeromagnopus How are the scopes retrieved from `/scopes/referenceType/{referenceType}/referenceId/{referenceId}` ordered?

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
MAG-mv: @jeromagnopus Do you know of any additional default scopes that will be created?

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/16e37678de81d4a8b4ef2dfef6a7fe82e6f70041

Library/include/CSP/Systems/Multiplayer/MultiplayerSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/842/commits/9860b3d889a1eaad0c3e1f434f1bd165a7c6368b

Used the space system to check if we're in the space. We do this for HotspotSequenceSystem, as well.

Library/include/CSP/Systems/Multiplayer/ScopeLeader.h
MAG-mv: Yeah, I can't answer the questions above. What makes it more difficult is that CHS are still making more changes to this on the backend

Tests/src/PublicAPITests/MultiplayerSystemTests.cpp
jeromagnopus: I don't believe there are any intended until CSP has onboarded to AOI and scopes in general.  Although it has come up to use a second scope to manage some things on the server-side, however, nothing concrete or planned.

PR #843, opened 2025-10-27
[OF-1763] feat!: Runtime feature flags

MAG-AdamThorn:
This work introduces the concept of runtime feature flags to CSP. Feature flags can be defined by CSP developers when required for the feature they are developing.

Feature flag state can be set as part of the CSPFoundation initialization flow and cannot be updated at any other time.

An optional array of FeatureFlag objects can be passed on Initialization to override the default state of the developer defined flags.

The following methods have been added to CSPFoundation:

```
- static bool IsCSPFeatureEnabled(EFeatureFlag Flag);
- static const csp::common::Array<FeatureFlag>& GetFeatureFlags();
- static csp::common::String GetCSPFeatureFlagDescription(EFeatureFlag Flag);
```

In addition, two function have been added to make it easier to test the functionality. As per our convention, there methods are prefixed with a double `_`. They are:
 
```
- CSP_NO_EXPORT static void __AddFeatureFlagForTesting(EFeatureFlag Type, bool IsEnabled, const csp::common::String Description);
- CSP_NO_EXPORT static void __ResetFeatureFlagsForTesting();
```

**Breaking change:**
This is a breaking change.
The CSPFoundation::Initialise() method has been updated to take an additional `CSPFeatureFlags` optional argument.

Old:
```
static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant,
    const csp::ClientUserAgent& ClientUserAgentHeader);
```

New:
```
static bool Initialise(const csp::common::String& EndpointRootURI, const csp::common::String& Tenant,
    const csp::ClientUserAgent& ClientUserAgentHeader, const csp::common::Optional<csp::common::Array<FeatureFlag>>& CSPFeatureFlags);
```

### DIFF
--- a/Library/include/CSP/CSPFoundation.h
+++ b/Library/include/CSP/CSPFoundation.h
@@ -61,7 +61,7 @@ public:
     FeatureFlag() = default;
 
     // Constructor for client usage where no description is needed
-    FeatureFlag(const EFeatureFlag Type, bool IsEnabled)
+    FeatureFlag(EFeatureFlag Type, bool IsEnabled)
         : Type(Type)
         , Enabled(IsEnabled)
         , Description("")
@@ -69,7 +69,7 @@ public:
     }
 
     // Constructor for internal library usage with description
-    CSP_NO_EXPORT FeatureFlag(const EFeatureFlag Type, bool IsEnabled, const csp::common::String& Description)
+    CSP_NO_EXPORT FeatureFlag(EFeatureFlag Type, bool IsEnabled, const csp::common::String& Description)
         : Type(Type)
         , Enabled(IsEnabled)
         , Description(Description)
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -384,14 +384,14 @@ namespace
             return;
         }
 
-        for (size_t i = 0; i < (*FeatureFlagOverrides).Size(); ++i)
+        for (const auto& OverrideFlag : *FeatureFlagOverrides)
         {
             bool FeatureFlagFound = false;
-            for (size_t j = 0; j < FeatureFlags.Size(); j++)
+            for (auto& Flag : FeatureFlags)
             {
-                if (FeatureFlags[j].Type == (*FeatureFlagOverrides)[i].Type)
+                if (Flag.Type == OverrideFlag.Type)
                 {
-                    FeatureFlags[j].Enabled = (*FeatureFlagOverrides)[i].Enabled;
+                    Flag.Enabled = OverrideFlag.Enabled;
                     FeatureFlagFound = true;
                     break;
                 }
@@ -399,7 +399,7 @@ namespace
             if (!FeatureFlagFound)
             {
                 CSP_LOG_MSG(csp::common::LogLevel::Warning,
-                    fmt::format("Unknown feature flag passed with integer value: {}", static_cast<int>((*FeatureFlagOverrides)[i].Type)).c_str());
+                    fmt::format("Unknown feature flag passed with integer value: {}", static_cast<int>(OverrideFlag.Type)).c_str());
             }
         }
     }
@@ -539,34 +539,38 @@ const csp::common::String& CSPFoundation::GetTenant() { return *Tenant; }
 
 bool CSPFoundation::IsFeatureEnabled(EFeatureFlag Flag)
 {
-    for (size_t i = 0; i < FeatureFlags.Size(); i++)
-    {
-        if (FeatureFlags[i].Type == Flag)
-        {
-            return FeatureFlags[i].Enabled;
-        }
-    }
+    auto it = std::find_if(FeatureFlags.begin(), FeatureFlags.end(), [Flag](const FeatureFlag& FeatureFlag) { return FeatureFlag.Type == Flag; });
 
-    CSP_LOG_MSG(csp::common::LogLevel::Warning, fmt::format("Unknown feature flag queried with integer value: {}", static_cast<int>(Flag)).c_str());
-    return false;
+    if (it != FeatureFlags.end())
+    {
+        return it->Enabled;
+    }
+    else
+    {
+        CSP_LOG_MSG(
+            csp::common::LogLevel::Warning, fmt::format("Unknown feature flag queried with integer value: {}", static_cast<int>(Flag)).c_str());
+
+        return false;
+    }
 }
 
 const csp::common::Array<FeatureFlag>& CSPFoundation::GetFeatureFlags() { return FeatureFlags; }
 
 csp::common::String CSPFoundation::GetFeatureFlagDescription(EFeatureFlag Flag)
 {
-    for (size_t i = 0; i < FeatureFlags.Size(); i++)
+    auto it = std::find_if(FeatureFlags.begin(), FeatureFlags.end(), [Flag](const FeatureFlag& FeatureFlag) { return FeatureFlag.Type == Flag; });
+
+    if (it != FeatureFlags.end())
     {
-        if (FeatureFlags[i].Type == Flag)
-        {
-            return FeatureFlags[i].GetDescription();
-        }
+        return it->GetDescription();
     }
+    else
+    {
+        CSP_LOG_MSG(csp::common::LogLevel::Warning,
+            fmt::format("Unknown feature flag description requested with integer value: {}", static_cast<int>(Flag)).c_str());
 
-    CSP_LOG_MSG(csp::common::LogLevel::Warning,
-        fmt::format("Unknown feature flag description requested with integer value: {}", static_cast<int>(Flag)).c_str());
-
-    return "";
+        return "";
+    }
 }
 
 void CSPFoundation::__AddFeatureFlagForTesting(EFeatureFlag Type, bool IsEnabled, const csp::common::String Description)


Library/include/CSP/CSPFoundation.h
```diff
@@ -189,6 +253,26 @@ class CSP_API CSPFoundation
...
     static const csp::common::String& GetTenant();
 
+    /// @brief Checks if a given feature flag is enabled.
+    /// @param Flag EFeatureFlag : The feature flag to check
+    /// @return bool : true if the feature flag is enabled, false otherwise
+    static bool IsCSPFeatureEnabled(EFeatureFlag Flag);
+
+    /// @brief Returns an array of feature flags.
+    /// @return const csp::common::Array<FeatureFlag>& : An array of the defined feature flags
+    static const csp::common::Array<FeatureFlag>& GetFeatureFlags();
```
MAG-ChristopherAtkinson: [Consider] consistency in function naming; we specify CSP in the enabled check, but not in the get. The same applies to other functions here; I'm happy for this to go either way, but it should maintain a level of consistency.

Library/src/CSPFoundation.cpp
```diff
@@ -368,6 +375,34 @@ namespace
...
+    {
+        if (!FeatureFlagOverrides.HasValue())
+        {
+            return;
+        }
+
+        for (size_t i = 0; i < (*FeatureFlagOverrides).Size(); ++i)
+        {
+            bool FeatureFlagFound = false;
+            for (size_t j = 0; j < FeatureFlags.Size(); j++)
```
MAG-ChristopherAtkinson: [Nit] it would be more optimal to use the feature flag Enum as the key to have direct access on the array rather than nested loops, granted this approach is reasonable given the potential scope and size of the feature

Library/src/CSPFoundation.cpp
```diff
@@ -368,6 +375,34 @@ namespace
         }
         return false;
     }
+
+    void ApplyFeatureFlagOverrides(
```
MAG-ChristopherAtkinson, line 379: [Consider] Do we want to rename this to `ApplyCSPFeatureFlagOverrides` due to the multiple definitions, ours and client for clarity? Personally, this is scoped to CSP, so not needed, but in other areas, we specify `CSP` is naming

Library/include/CSP/CSPFoundation.h
```diff
@@ -28,10 +41,42 @@ class ServicesDeploymentStatus;
...
+    // Constructor for client usage where no description is needed
+    FeatureFlag(const EFeatureFlag Type, bool IsEnabled)
+        : Type(Type)
+        , Enabled(IsEnabled)
+        , Description("")
+    {
+    }
+
+    // Constructor for internal library usage with description
+    FeatureFlag(const EFeatureFlag Type, bool IsEnabled, const csp::common::String& Description)
```
MAG-ChristopherAtkinson: [Fix] should make this `CSP_NO_EXPORT` as the client should not have means to change a flag's description

MAG-ChristopherAtkinson (changes requested):

Library/include/CSP/CSPFoundation.h
```diff
@@ -28,10 +41,42 @@ class ServicesDeploymentStatus;
 namespace csp::multiplayer
 {
 class ISignalRConnection;
-}
+} // namespace csp::multiplayer
 
 namespace csp
 {
+// New developer feature flags should be added here. Please leave 'Invalid' as the first element.
```
MAG-mv, line 48: [Consider] If Invalid should be the first element, we can set its value to 0

Library/include/CSP/CSPFoundation.h
```diff
@@ -28,10 +41,42 @@ class ServicesDeploymentStatus;
...
+{
+public:
+    EFeatureFlag Type;
+    bool Enabled;
+    csp::common::String Description;
+
+    FeatureFlag() = default;
+
+    // Constructor for client usage where no description is needed
+    FeatureFlag(const EFeatureFlag Type, bool IsEnabled)
```
MAG-mv: [Nit] const enum in parameter is pointless

Library/src/CSPFoundation.cpp
```diff
@@ -368,6 +375,34 @@ namespace
...
+        const csp::common::Optional<csp::common::Array<FeatureFlag>>& FeatureFlagOverrides, csp::common::Array<csp::FeatureFlag>& FeatureFlags)
+    {
+        if (!FeatureFlagOverrides.HasValue())
+        {
+            return;
+        }
+
+        for (size_t i = 0; i < (*FeatureFlagOverrides).Size(); ++i)
```
MAG-mv: [Consider] Our containers now have iterators. You can sue a ranged for loop

Library/src/CSPFoundation.cpp
MAG-mv: [Consider] Same here

Library/src/CSPFoundation.cpp
```diff
@@ -498,6 +537,54 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
 
 const csp::common::String& CSPFoundation::GetTenant() { return *Tenant; }
 
+bool CSPFoundation::IsCSPFeatureEnabled(EFeatureFlag Flag)
+{
+    for (size_t i = 0; i < FeatureFlags.Size(); i++)
```
MAG-mv: [Consider] Same here

MAG-mv: You could also use std::find_if here

Library/src/CSPFoundation.cpp
```diff
@@ -498,6 +537,54 @@ const csp::common::String& CSPFoundation::GetClientUserAgentString() { return *C
...
+
+    CSP_LOG_MSG(csp::common::LogLevel::Warning, fmt::format("Unknown feature flag queried with integer value: {}", static_cast<int>(Flag)).c_str());
+    return false;
+}
+
+const csp::common::Array<FeatureFlag>& CSPFoundation::GetFeatureFlags() { return FeatureFlags; }
+
+csp::common::String CSPFoundation::GetCSPFeatureFlagDescription(EFeatureFlag Flag)
+{
+    for (size_t i = 0; i < FeatureFlags.Size(); i++)
```
MAG-mv: [Consider] Same here

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: Excellent suggestion. Addressed naming here: [03fde53](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/03fde5321987e6c09812c1d570f03ab88f52d4c3)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: Addressed naming inconsistency in this commit here: [03fde53](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/03fde5321987e6c09812c1d570f03ab88f52d4c3)

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: Yes, good idea. I also made the description member private and added a getter for it here: [03fde53](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/03fde5321987e6c09812c1d570f03ab88f52d4c3)

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: Agreed good idea.

Update: [03fde53](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/03fde5321987e6c09812c1d570f03ab88f52d4c3)

Library/include/CSP/CSPFoundation.h
MAG-AdamThorn: Agreed, fix: [75f342d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/75f342d52787b33e3ea2e909dbf76bb02caacdd0)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: Agreed fix: [75f342d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/75f342d52787b33e3ea2e909dbf76bb02caacdd0)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: fix: [75f342d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/75f342d52787b33e3ea2e909dbf76bb02caacdd0)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: Agreed. Fix: [75f342d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/75f342d52787b33e3ea2e909dbf76bb02caacdd0)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: Fix: [75f342d](https://github.com/magnopus-opensource/connected-spaces-platform/pull/843/commits/75f342d52787b33e3ea2e909dbf76bb02caacdd0)

Library/src/CSPFoundation.cpp
MAG-AdamThorn: I did originally do this, but ultimately I went with an array as I wanted to be able to return a const ref to it, which would have been difficult in cases where there were no feature flags to return.

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Great, marking this as resolved

Library/include/CSP/CSPFoundation.h
MAG-ChristopherAtkinson: Even better, really rounds out our access policy

PR #844, opened 2025-10-28
[OF-1736] feat: Expose AsyncCall to DuplicateSpace

MAG-AdamThorn:
The current DuplicateSpace method is a synchronous operation that can timeout and fail if the Space is overly complex or the backend services are under excessive load.

A new generic `AsyncCallCompleted` event has been added by the backend services to alert the user when a specific async operation has completed. Operations such as DuplicateSpace can take a AsyncCall bool argument, which determines whether the operation should be performed synchronously or asynchronously.
If performed asynchronously the call will immediately return a result object, which in the case of Duplicate Space will contain a 204 - No content response on success. Then, once the actual duplication operation has been completed, the `AsyncCallCompleted` event will fire, passing a `AsyncCallCompletedEventData` object.

As part of this work the current `DuplicateSpace` method has been marked deprecated and a new `DuplicateSpaceAsync` method has been introduced. This approach has been taken for two reasons:

1. CHS intend to remove the synchronous flow.
2. The async flow returns a different result object - SpaceResult vs NullResult.

A new `AsyncCallCompletedCallbackHandler` has been introduced along with a public setter to allow clients to register for the `AsyncCallCompleted` event: `SetAsyncCallCompletedCallback()`.

As noted above, this event will return a generic 'AsyncCallCompletedEventData` object, which in the case of the duplicate Space operation, will contain the following data:

- OperationName: "DuplicateSpaceAsync"
- ReferenceId: "Id of the new Space"
- ReferenceType: "GroupId"

### DIFF
--- a/Library/include/CSP/Systems/SystemsManager.h
+++ b/Library/include/CSP/Systems/SystemsManager.h
@@ -171,7 +171,7 @@ public:
 
     csp::multiplayer::MultiplayerConnection* GetMultiplayerConnection();
 
-    csp::multiplayer::NetworkEventBus& GetEventBus();
+    csp::multiplayer::NetworkEventBus* GetEventBus();
 
     // Convenience methods for the moment. This will need to be broken at formal modularization, but the standard pattern it creates throughout
     // integrations/tests will no doubt be helpful in doing that anyhow, rather than having big constructors everywhere.
--- a/Library/src/Systems/SystemsManager.cpp
+++ b/Library/src/Systems/SystemsManager.cpp
@@ -102,11 +102,11 @@ ExternalServiceProxySystem* SystemsManager::GetExternalServicesProxySystem() { r
 
 csp::multiplayer::MultiplayerConnection* SystemsManager::GetMultiplayerConnection() { return MultiplayerConnection; }
 
-csp::multiplayer::NetworkEventBus& SystemsManager::GetEventBus() { return MultiplayerConnection->GetEventBus(); }
+csp::multiplayer::NetworkEventBus* SystemsManager::GetEventBus() { return &MultiplayerConnection->GetEventBus(); }
 
 csp::multiplayer::OnlineRealtimeEngine* SystemsManager::MakeOnlineRealtimeEngine()
 {
-    return new csp::multiplayer::OnlineRealtimeEngine { *GetMultiplayerConnection(), *GetLogSystem(), GetEventBus(), *GetScriptSystem() };
+    return new csp::multiplayer::OnlineRealtimeEngine { *GetMultiplayerConnection(), *GetLogSystem(), *GetEventBus(), *GetScriptSystem() };
 }
 
 csp::multiplayer::OfflineRealtimeEngine* SystemsManager::MakeOfflineRealtimeEngine()
--- a/MultiplayerTestRunner/src/RunnableTests/EventBusPing.cpp
+++ b/MultiplayerTestRunner/src/RunnableTests/EventBusPing.cpp
@@ -28,10 +28,10 @@ namespace EventBusPing
 void RunTest()
 {
     // Listen for an event, then ping it back to the client that sent it.
-    csp::systems::SystemsManager::Get().GetEventBus().ListenNetworkEvent(csp::multiplayer::NetworkEventRegistration("Receiver", "EventPingRequest"),
+    csp::systems::SystemsManager::Get().GetEventBus()->ListenNetworkEvent(csp::multiplayer::NetworkEventRegistration("Receiver", "EventPingRequest"),
         [](const csp::common::NetworkEventData& NetworkEventData)
         {
-            csp::systems::SystemsManager::Get().GetEventBus().SendNetworkEventToClient(
+            csp::systems::SystemsManager::Get().GetEventBus()->SendNetworkEventToClient(
                 "EventPingResponse", {}, NetworkEventData.SenderClientId, [](csp::multiplayer::ErrorCode) {});
         });
 }
--- a/Tests/src/PublicAPITests/AssetSystemTests.cpp
+++ b/Tests/src/PublicAPITests/AssetSystemTests.cpp
@@ -2114,7 +2114,7 @@ CSP_PUBLIC_TEST(CSPEngine, AssetSystemTests, AssetProcessGracefulFailureCallback
     csp::common::ReplicatedValue Param4 = "";
     csp::common::ReplicatedValue Param5 = "";
 
-    SystemsManager.GetEventBus().SendNetworkEventToClient(
+    SystemsManager.GetEventBus()->SendNetworkEventToClient(
         NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::AssetDetailBlobChanged), { Param1, Param2, Param3, Param4, Param5 },
         Connection->GetClientId(), [](ErrorCode Error) { EXPECT_EQ(Error, ErrorCode::None); });
 
--- a/Tests/src/PublicAPITests/ConversationSystemTests.cpp
+++ b/Tests/src/PublicAPITests/ConversationSystemTests.cpp
@@ -117,7 +117,7 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationSystemTests, ConversationSystemEventTest)
 
         bool EventSent = false;
 
-        SystemsManager.GetEventBus().SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
+        SystemsManager.GetEventBus()->SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
             csp::systems::ConversationSystemHelpers::MessageInfoToReplicatedValueArray(Params.MessageType, Params.MessageInfo),
             Connection->GetClientId(), [&EventSent](csp::multiplayer::ErrorCode) { EventSent = true; });
 
@@ -152,7 +152,7 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationSystemTests, ConversationSystemEventTest)
 
         bool EventSent = false;
 
-        SystemsManager.GetEventBus().SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
+        SystemsManager.GetEventBus()->SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
             csp::systems::ConversationSystemHelpers::MessageInfoToReplicatedValueArray(Params.MessageType, Params.MessageInfo),
             Connection->GetClientId(), [&EventSent](csp::multiplayer::ErrorCode) { EventSent = true; });
 
@@ -208,7 +208,7 @@ CSP_PUBLIC_TEST(CSPEngine, ConversationSystemTests, ConversationSystemEventDelay
 
     bool EventSent = false;
 
-    SystemsManager.GetEventBus().SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
+    SystemsManager.GetEventBus()->SendNetworkEventToClient(NetworkEventBus::StringFromNetworkEvent(NetworkEventBus::NetworkEvent::Conversation),
         csp::systems::ConversationSystemHelpers::MessageInfoToReplicatedValueArray(Params.MessageType, Params.MessageInfo), Connection->GetClientId(),
         [&EventSent](csp::multiplayer::ErrorCode) { EventSent = true; });
 
--- a/Tests/src/PublicAPITests/EventBusTests.cpp
+++ b/Tests/src/PublicAPITests/EventBusTests.cpp
@@ -72,17 +72,17 @@ CSP_PUBLIC_TEST(CSPEngine, EventBusTests, RegisterDeregister)
     const char* ReceiverId = "TestReceiverId";
     const char* EventName = "TestEventName";
 
-    const csp::common::Array<NetworkEventRegistration> InitialRegisteredEvents = SystemsManager.GetEventBus().AllRegistrations();
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    const csp::common::Array<NetworkEventRegistration> InitialRegisteredEvents = SystemsManager.GetEventBus()->AllRegistrations();
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId, EventName }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    const csp::common::Array<NetworkEventRegistration> AddedRegistration = SystemsManager.GetEventBus().AllRegistrations();
+    const csp::common::Array<NetworkEventRegistration> AddedRegistration = SystemsManager.GetEventBus()->AllRegistrations();
 
     EXPECT_TRUE(AddedRegistration.Size() == InitialRegisteredEvents.Size() + 1);
     EXPECT_TRUE(AddedRegistration.ToList().Contains(NetworkEventRegistration { ReceiverId, EventName }));
 
-    SystemsManager.GetEventBus().StopListenNetworkEvent(NetworkEventRegistration { ReceiverId, EventName });
+    SystemsManager.GetEventBus()->StopListenNetworkEvent(NetworkEventRegistration { ReceiverId, EventName });
 
-    const csp::common::Array<NetworkEventRegistration> RemovedRegistration = SystemsManager.GetEventBus().AllRegistrations();
+    const csp::common::Array<NetworkEventRegistration> RemovedRegistration = SystemsManager.GetEventBus()->AllRegistrations();
     EXPECT_TRUE(RemovedRegistration.Size() == InitialRegisteredEvents.Size());
     EXPECT_FALSE(RemovedRegistration.ToList().Contains(NetworkEventRegistration { ReceiverId, EventName }));
 }
@@ -100,20 +100,20 @@ CSP_PUBLIC_TEST(CSPEngine, EventBusTests, RegisterDeregisterMulti)
 
     const char* EventName3 = "TestEventName3";
 
-    const csp::common::Array<NetworkEventRegistration> InitialRegisteredEvents = SystemsManager.GetEventBus().AllRegistrations();
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    const csp::common::Array<NetworkEventRegistration> InitialRegisteredEvents = SystemsManager.GetEventBus()->AllRegistrations();
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId, EventName }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId, EventName2 }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId, EventName3 }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId2, EventName }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId2, EventName2 }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    SystemsManager.GetEventBus().ListenNetworkEvent(
+    SystemsManager.GetEventBus()->ListenNetworkEvent(
         NetworkEventRegistration { ReceiverId2, EventName3 }, [](const csp::common::NetworkEventData& /*NetworkEventData*/) {});
-    const csp::common::Array<NetworkEventRegistration> AddedRegistration = SystemsManager.GetEventBus().AllRegistrations();
+    const csp::common::Array<NetworkEventRegistration> AddedRegistration = SystemsManager.GetEventBus()->AllRegistrations();
 
     EXPECT_TRUE(AddedRegistration.Size() == InitialRegisteredEvents.Size() + 6);
     EXPECT_TRUE(AddedRegistration.ToList().Contains(NetworkEventRegistration { ReceiverId, EventName }));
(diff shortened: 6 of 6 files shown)


Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -343,22 +343,55 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
     /// @param Callback NullResultCallback : callback when asynchronous task finishes
     CSP_ASYNC_RESULT void DeleteSpaceGeoLocation(const csp::common::String& SpaceId, NullResultCallback Callback);
 
-    /// @brief Duplicate an existing space and assign it to the current user
+    /// @brief Duplicate an existing space and assign it to the current user.
+    /// This is a synchronous operation and can have a high execution time for complex spaces.
+    /// \deprecated Use DuplicateSpaceAsync() instead. This method performs a synchronous duplication of a Space which can timeout and fail for
```
MAG-AdamThorn, line 351: The wrapper generator didn't me marking this function deprecated using the `[[deprecated]]` attribute.
I have instead adopted doxygens style of marking methods as deprecated - result of this shown in the image below.

Change here: [1995cdd](https://github.com/magnopus-opensource/connected-spaces-platform/pull/840/commits/1995cdd318fe3ef7f41f71c8ba0ffc5640388d67)

PR_Documentation_Screenshot

Library/include/CSP/Common/NetworkEventData.h
```diff
@@ -152,6 +152,21 @@ class CSP_API SequenceChangedNetworkEventData : public NetworkEventData
...
 };
 
+class CSP_API AsyncCallCompletedEventData : public NetworkEventData
+{
+public:
+    /// @brief The name of the async operation that has been completed.
+    csp::common::String OperationName;
+
+    /// @brief An Id related to the async operation that has been completed.
+    /// This could for example be a groud Id, if this were an async duplicate group operation.
```
MAG-ElliotMorris: [nit] groud -> group

Library/include/CSP/Common/NetworkEventData.h
```diff
@@ -152,6 +152,21 @@ class CSP_API SequenceChangedNetworkEventData : public NetworkEventData
     csp::common::Optional<csp::common::HotspotSequenceChangedNetworkEventData> HotspotData = nullptr;
 };
 
+class CSP_API AsyncCallCompletedEventData : public NetworkEventData
+{
+public:
+    /// @brief The name of the async operation that has been completed.
+    csp::common::String OperationName;
```
MAG-ElliotMorris, line 161: [Question] How can we tell our users what they should expect here (and for the other values)? To my eye this isn't really usable unless you have a series of strings (or strongly typed values) to expect. Along with formalish descriptions on what actions cause the events to be fired and when.

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-ElliotMorris: [Comment] Ah neat. At least we can do something, damn wrapper gen.

Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -343,22 +343,55 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
...
     /// @param ShallowCopy bool : If true, the duplicated space will reference the assets of the original space. Otherwise, all assets will be
     /// duplicated.
-    /// @param Callback NullResultCallback : callback when asynchronous task finishes
+    /// @param Callback SpaceResultCallback : callback when asynchronous task finishes.
     CSP_ASYNC_RESULT void DuplicateSpace(const csp::common::String& SpaceId, const csp::common::String& NewName, SpaceAttributes NewAttributes,
         const csp::common::Optional<csp::common::Array<csp::common::String>>& MemberGroupIds, bool ShallowCopy, SpaceResultCallback Callback);
 
+    /// @brief Duplicate an existing space and assign it to the current user.
+    /// This is an asynchronous operation. Please subcribe to the AsyncCallCompletedCallback via @ref SpaceSystem::SetAsyncCallCompletedCallback() to
+    /// be notified when the duplication operation is complete.
```
MAG-ElliotMorris: [Fix] Say what strings they will be getting back from AsyncCallCompletedCallback. How are they meant to know otherwise?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -70,15 +100,24 @@ SpaceSystem::SpaceSystem()
 {
 }
 
-SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient, csp::common::LogSystem& LogSystem)
-    : SystemBase(InWebClient, nullptr, &LogSystem)
+SpaceSystem::SpaceSystem(csp::web::WebClient* InWebClient, multiplayer::NetworkEventBus* InEventBus, csp::common::LogSystem& LogSystem)
+    : SystemBase(InWebClient, InEventBus, &LogSystem)
```
MAG-ElliotMorris: [Nit] We should stop doing this silly `In` naming, and start passing by ref when appropriate. The question here for that is, can `EventBus` be null when constructing a `SpaceSystem`. I'm guessing not, so I'd advocate for reference semantics here if possible.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -70,15 +100,24 @@ SpaceSystem::SpaceSystem()
...
-SpaceSystem::~SpaceSystem() { delete (GroupAPI); }
+SpaceSystem::~SpaceSystem()
+{
+    delete (GroupAPI);
+
+    if (EventBusPtr)
+    {
+        EventBusPtr->StopListenNetworkEvent(csp::multiplayer::NetworkEventRegistration("CSPInternal::SpaceSystem",
+            csp::multiplayer::NetworkEventBus::StringFromNetworkEvent(csp::multiplayer::NetworkEventBus::NetworkEvent::AsyncCallCompleted)));
+    }
```
MAG-ElliotMorris: [Comment] I'm torn on this.

On this one hand, this is an end-of-life cleanup, so it makes me want to put it in the NetworkEventBus destructor itself. Have it handle its own resources and whatnot. Could iterate through all registered events and stop listening (although it's somewhat redundant, since the application is closing anyway...)

On the otherhand, this implementation gestures at good decoupling. Implies that an EventBus may live longer than a SpaceSystem which isn't currently possible, but may be in the future.

🤔 No suggestion here, just musing.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -1925,37 +1964,72 @@ void SpaceSystem::DeleteSpaceGeoLocation(const csp::common::String& SpaceId, Nul
...
     );
 }
+
+void SpaceSystem::SetAsyncCallCompletedCallback(AsyncCallCompletedCallbackHandler Callback)
+{
+    AsyncCallCompletedCallback = std::move(Callback);
+
+    if (!EventBusPtr)
+    {
+        CSP_LOG_ERROR_MSG("Error: Failed to register SpaceSystem. NetworkEventBus must be instantiated in the MultiplayerConnection first.");
```
MAG-ElliotMorris: [fix] "Failed to register SpaceSystem" isn't honest about what we're actually doing here. "Instantiated in the MultiplayerConnection" is also a bit unhelpful, it implies that the user has done something wrong and needs to go instantiate something.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -1925,37 +1964,72 @@ void SpaceSystem::DeleteSpaceGeoLocation(const csp::common::String& SpaceId, Nul
...
+        return;
+    }
+
+    if (!AsyncCallCompletedCallback)
+    {
+        CSP_LOG_ERROR_MSG("Error: AsyncCallCompletedCallback not set. SpaceSystem has not been registered with the AsyncCallCompleted event.");
```
MAG-ElliotMorris: [Consider] Telling the user how to do this, or pointing them at the right method.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -3247,6 +3258,100 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, DuplicateSpaceTest)
...
+        SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+        auto [Result]
+            = AWAIT_PRE(SpaceSystem, DuplicateSpaceAsync, RequestPredicate, Space.Id, UniqueSpaceName, SpaceAttributes::Private, nullptr, true);
+
+        EXPECT_EQ(Result.GetResultCode(), EResultCode::Success);
+
+        // Wait for the callback to be received
+        AsyncCallCompletedFuture.wait();
+        EXPECT_TRUE(AsyncCallCompletedFuture.get() == true);
```
MAG-ElliotMorris: [Consider] Here and elsewhere. Do you mean `ASSERT_EQ` here? `EXPECT` lets the test keep running on a failure.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -3247,6 +3258,100 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, DuplicateSpaceTest)
...
+
+    // Attempt to duplicate space asynchrously
+    {
+        std::promise<bool> AsyncCallCompletedPromise;
+        std::future<bool> AsyncCallCompletedFuture = AsyncCallCompletedPromise.get_future();
+
+        auto AsyncCallCompletedCallback = [&](const csp::common::AsyncCallCompletedEventData& NetworkEventData)
+        {
+            EXPECT_EQ(NetworkEventData.OperationName, "DuplicateSpaceAsync");
+            EXPECT_EQ(NetworkEventData.ReferenceType, "GroupId");
```
MAG-ElliotMorris: [Nit] I'm a little negative towards assertions inside callbacks, because they may not run if the callback isn't called and I've seen bugs from that.

In this case, you're asserting that the future has a correct value so I think these must have been run. Nonetheless, you might consider a style of bundling all the data into the future (as a tuple) and asserting everything linearly.

Tests/src/PublicAPITests/SpaceSystemTests.cpp
```diff
@@ -3247,6 +3258,100 @@ CSP_PUBLIC_TEST(CSPEngine, SpaceSystemTests, DuplicateSpaceTest)
...
+
+        SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
+
+        auto [Result]
+            = AWAIT_PRE(SpaceSystem, DuplicateSpaceAsync, RequestPredicate, Space.Id, UniqueSpaceName, SpaceAttributes::Private, nullptr, true);
+
+        EXPECT_EQ(Result.GetResultCode(), EResultCode::Success);
+
+        // Wait for the callback to be received
+        AsyncCallCompletedFuture.wait();
```
MAG-ElliotMorris: [Question] The expected time of this duplication is always going to be lower than the 30 seconds gtest timeout? Thinking forward to potential flakes.

Library/include/CSP/Common/NetworkEventData.h
MAG-ElliotMorris: @MAG-SamBirley This is the sort of thing I mean when I say we need to focus on specifying the multiplayer connection interface.
@jeromagnopus You spoke yesterday about wanting to reduce communicative necessity, and this is one of those things. This interface is specified so weakly (whether that's from there not being a specification, or it just not being enough in the zeitgeist for us to know about), it creates ongoing communicative needs around what these strings are and mean.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: I agree re the naming and will remove the 'In' prefix.
Regarding your second point I did take a look at what could be done but unfortunately the wrapper gen is forcing our hand once more.

In summary in NetworkEventBus is owned by the MultiplayerConnection and is currently stored as a regular raw ptr. I looked at converting that to a unique_ptr and changing the MultiplayerConnection::GetEventBusPtr() method to return a ref. Unfortunately the NetworkEventBus is then passed into a number of System ctors, just as it is here with the SpaceSystem. It is then in turn passed into the SystemBase class ctor. The issue is that all of these systems require a default ctor to appease the wrapper gen which cannot take a ref.

This is one of those areas we will have to revisit once the wrapper generator has been updated.

Library/include/CSP/Common/NetworkEventData.h
```diff
@@ -152,6 +152,21 @@ class CSP_API SequenceChangedNetworkEventData : public NetworkEventData
     csp::common::Optional<csp::common::HotspotSequenceChangedNetworkEventData> HotspotData = nullptr;
 };
```
MAG-mv, line 154: [Fix] Doc explaining what this class does

Library/include/CSP/Systems/Spaces/SpaceSystem.h
```diff
@@ -343,22 +343,55 @@ class CSP_API CSP_NO_DISPOSE SpaceSystem : public SystemBase
...
+    /// @param MemberGroupIds csp::common::Array<csp::common::String> : An optional array of group (space) IDs. Members of these groups will be added
+    /// to the duplicated space with the same roles.
     /// @param ShallowCopy bool : If true, the duplicated space will reference the assets of the original space. Otherwise, all assets will be
     /// duplicated.
-    /// @param Callback NullResultCallback : callback when asynchronous task finishes
+    /// @param Callback SpaceResultCallback : callback when asynchronous task finishes.
     CSP_ASYNC_RESULT void DuplicateSpace(const csp::common::String& SpaceId, const csp::common::String& NewName, SpaceAttributes NewAttributes,
         const csp::common::Optional<csp::common::Array<csp::common::String>>& MemberGroupIds, bool ShallowCopy, SpaceResultCallback Callback);
 
+    /// @brief Duplicate an existing space and assign it to the current user.
```
MAG-mv, line 364: [Fix] It would be good to understand this behaviour in different circumstances. Does it work if the client is out of the space? What happens if they shutdown their connection and rejoin?

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: @MAG-ElliotMorris Currently the Network Event Bus (via the MultiplayerConnection which owns it) is one of the last systems to be destroyed by the Systems Manager. I would therefore be in favour of having the Network Event Bus clean up all registered listeners in it's destructor as you suggested above.

@MAG-mv @MAG-ChristopherAtkinson any concerns with me making that change?

MAG-ChristopherAtkinson: I am in favour of this change, it makes sense that Network Event Bus handles safe clean-up of any registered event that it manages

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -58,6 +58,36 @@ namespace
...
 constexpr const int MAX_SPACES_RESULTS = 100;
 
+// Construct a new DuplicateSpaceOptions dto request object. This function is called by both DuplicateSpace and DuplicateSpaceAsync methods.
+// The only difference is in the value they pass for the AsyncCall parameter.
+std::shared_ptr<chsaggregation::DuplicateSpaceOptions> ConstructDuplicateSpaceOptions(const String& SpaceId, const String& NewName,
+    csp::systems::SpaceAttributes NewAttributes, const Optional<Array<String>>& MemberGroupIds, bool ShallowCopy, bool AsyncCall)
+{
+    auto Request = std::make_shared<chsaggregation::DuplicateSpaceOptions>();
+    Request->SetSpaceId(SpaceId);
+    Request->SetNewGroupOwnerId(csp::systems::SystemsManager::Get().GetUserSystem()->GetLoginState().UserId);
```
MAG-mv: [Consider] Injecting UserSystem into SpaceSystem at constuction time to remove the singleton dependency

Library/include/CSP/Common/NetworkEventData.h
MAG-AdamThorn: Fix: [64f863f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/64f863f7d3c1bc27efc763dddd6252076f843118)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: Fix: [64f863f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/64f863f7d3c1bc27efc763dddd6252076f843118)

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-AdamThorn: Fix: [924779e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/924779e0006c9c82e6b256a05bd31d9e9b34d3b2)

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-AdamThorn: Fix: [e6d8ab2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/e6d8ab21c72f7e5a7999dc6b50f11ebde8332e88)

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-AdamThorn: I like that approach, updated: [e6d8ab2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/e6d8ab21c72f7e5a7999dc6b50f11ebde8332e88)

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-AdamThorn: The space is simple but it could take a while to complete if the backend services are under load.

I liked the idea of adding a 30 second timeout which I have done here: [e6d8ab2](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/e6d8ab21c72f7e5a7999dc6b50f11ebde8332e88)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: Following our sync @MAG-ElliotMorris, the MultiplayerConnection::GetEventBus() method has been updated to return a reference to the EventBus. This ref is then passed to the systems, which dereferences it before passing it to the SystemBase class. This is similar to what is done with the LogSystem.

Update: [a69b403](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/a69b4031c30f3fe7da37a2d0d9b0804f3e36fb8a)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: This has been removed entirely as part of this change: [a69b403](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/a69b4031c30f3fe7da37a2d0d9b0804f3e36fb8a)

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -58,6 +58,36 @@ namespace
...
+    Request->SetNewUniqueName(NewName);
+    Request->SetDiscoverable(HasFlag(NewAttributes, csp::systems::SpaceAttributes::IsDiscoverable));
+    Request->SetRequiresInvite(HasFlag(NewAttributes, csp::systems::SpaceAttributes::RequiresInvite));
+    Request->SetShallowCopy(ShallowCopy);
+    Request->SetAsyncCall(AsyncCall);
+
+    if (MemberGroupIds.HasValue())
```
MAG-mv, line 76: [Consider] csp::Convert will convert an optional csp array to optional vector in one line of code

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: Yep. I see no issues

Library/include/CSP/Common/NetworkEventData.h
MAG-AdamThorn: @MAG-ElliotMorris currently we can not tell them. We only have a single flow making use of the AsyncCallCompleted event but I know that others are planned. I suggest that separately to this ticket we work out how best to capture that information so that it can be surfaced to users.

@jeromagnopus I will start a thread to discuss this and we can also discuss as a team in our next sync.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: I will make that change then.

MAG-AdamThorn: Change here: [4124379](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/41243791d38a784c5caeba8c92b7f64a4c2a590d)

Library/include/CSP/Common/NetworkEventData.h
MAG-AdamThorn: fix: [0d8959f](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/0d8959f4cbc855aeff63d0bd037d30e4bd3f65e7)

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-AdamThorn: Updated here: [e000f11](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/e000f110a722e67722b1d5130eb05b2fcd7d7ef1)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: Good idea. Update: [ce05c62](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/ce05c62cee81ad5609d4c4fa637170e2b91816c4)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-AdamThorn: Update: [ce05c62](https://github.com/magnopus-opensource/connected-spaces-platform/pull/844/commits/ce05c62cee81ad5609d4c4fa637170e2b91816c4)

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: Love to see the extra-mile of propagating into all the systems 👍

Tests/src/PublicAPITests/SpaceSystemTests.cpp
MAG-ElliotMorris: > I liked the idea of adding a 30 second timeout which I have done here

[Comment] I was pretty sure that gtest has a default 30 second timeout (which would imply this condition can never be hit because gtest will terminate first). However, I can't find any documentation showing that. Perhaps I hallucinated.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: [Nit] Pretty sure you dont even need to do: if (MemberGroupIds.HasValue()), you can convert the entire array

Library/include/CSP/Systems/Spaces/SpaceSystem.h
MAG-mv: [Comment] I'm surprised it the request would be cancelled if they disconnect!

PR #846, opened 2025-11-04
[OB-4723] fix: exceptions in continuations return incorrect result

MAG-ChristopherAtkinson:
In the continuation flow, we would commonly return an invalid result, which resulted in the loss of data from the result that triggered the failure path. This PR promotes the result from the exception to preserve the information in the event of `ResultException`, and introduces a utility function to prevent code duplication.

### DIFF
--- a/Library/include/CSP/Common/ContinuationUtils.h
+++ b/Library/include/CSP/Common/ContinuationUtils.h
@@ -83,7 +83,8 @@ template <typename T> const inline T GetResultExceptionOrInvalid(const csp::comm
     if (Exception.GetExceptionType() == csp::common::continuations::ExceptionType::Result)
     {
         const auto ResultException = static_cast<const csp::common::continuations::ResultException*>(&Exception);
-        return *static_cast<const T*>(&ResultException->GetResult());
+        const auto Result = ResultException->GetResult();
+        return T(Result.GetResultCode(), static_cast<csp::web::EResponseCodes>(Result.GetHttpResultCode()), Result.GetFailureReason());
     }
 
     return T(csp::systems::EResultCode::Failed, 0);


Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -72,9 +72,22 @@ class ResultException : public ExpectedExceptionBase
     const csp::systems::ResultBase& GetResult() const { return Result; }
 
 private:
-    const csp::systems::ResultBase& Result;
+    const csp::systems::ResultBase Result;
```
MAG-ElliotMorris: [Fix] This slices, leaving you unable to use any of the derived data to construct your errors. This doesn't seem like the intent given the how the name is `Base`. Is this what you were trying to do?

Library/include/CSP/Common/ContinuationUtils.h
```diff
@@ -72,9 +72,22 @@ class ResultException : public ExpectedExceptionBase
...
+    // Check that the exception base is of type result and cast to derived type.
+    if (Exception.GetExceptionType() == csp::common::continuations::ExceptionType::Result)
+    {
+        const auto ResultException = static_cast<const csp::common::continuations::ResultException*>(&Exception);
+        const auto Result = ResultException->GetResult();
+        return T(Result.GetResultCode(), static_cast<csp::web::EResponseCodes>(Result.GetHttpResultCode()), Result.GetFailureReason());
+    }
+
+    return MakeInvalid<T>();
+}
```
MAG-ElliotMorris, line 91: [fix] Quite confusing, seems like an innaprioriate shortcut.

As far as I can tell :
- `T` is always a `ResultBase` derived object
- We get a `ResultBase` out of `GetResult`, and then use that to manually construct a new `ResultBase` derived object. Using only the base state, thus throwing away any derived context.
- We never use the `MakeInvalid` branch of this `GetXorY` function

This seems to me like a confusion caused by the inappropriate slicing of the `Result` object in `ResultException`

It strikes me that in the below examples, `exception.GetResult()` should be able to be used directly as an argument to the callback. This may require that you ensure the correct result exceptions are thrown throughout the chain, or write converting constructors & operators and such. Ideally the former.

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: This caused a dangling reference issue where the result object has been deallocated; as a result, we need to store a copy to ensure that the data is present outside of the calling scope

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: - `T` could be any derived object from `ExpectedExceptionBase`, not just `ResultException`. 
- `GetResult` returns a reference to a `ResultBase` and is used to construct a new derived object 
- `MakeInvalid` will trigger in the event `ExpectedExceptionBase` is not a `ResultException`
- The callbacks being triggered from these flows do not have explicit converters from `ResultBase` to type `T`, resulting in C2440

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: Revisiting I realised I missed the point here, Yes you are correct and the copy by value is truncating the object, and yeah should be changed.

MAG-ChristopherAtkinson: Thanks for the chat to help narrow down the issue resolving around throwing an exception copy-initializes a temporary object, this has been resolved here: [06a9de5](https://github.com/magnopus-opensource/connected-spaces-platform/pull/846/commits/06a9de5ed78c7d2c4ce532eb10e187fa30d3604f)

Library/include/CSP/Common/ContinuationUtils.h
MAG-ChristopherAtkinson: prevent unnecessary construction of derived object but kept the general utility to reduce code duplication 

fixed here: [6b1a166](https://github.com/magnopus-opensource/connected-spaces-platform/pull/846/commits/6b1a166e959169da16d07446c525d8f2c04d693b)

MAG-ChristopherAtkinson: Following the failing tests, I realised that the result exception being passed is not always the same type as the expected return of the continuations being used. This results in an unsafe pointer cast leading to undefined behaviour.

MAG-AdamThorn (approved):

Library/include/CSP/Common/ContinuationUtils.h
MAG-AdamThorn: @MAG-ChristopherAtkinson I misread the above comment from you and approved the PR, but rereading I would like to clarify what you mean by the above as it is not clear.

Are you saying that as things stand we are at risk of undefined behaviour? It sounds like there is more work to be done or is that not the case.

MAG-ElliotMorris: > This results in an unsafe pointer cast leading to undefined behaviour.

[Question]
I don't understand what you mean here.

The core issue is, you have just gone to the work to make the exception object be able to own and pass a polymorphic `ResultBase` type through the continuation chain, and yet you are throwing that away and reconstructing a new object from only the base data. These two changes don't make sense with eachother, you may as well have just stored `ResultCode`, `HttpResultCode` and `FailureReason` on  the exception as that is the only data that can possibly make it back to the user in the error case. Can you explain why both these are neccesary:

- The storage of a Base result type on the exception object
- That we discard that object and construct a new one from its properties

MAG-ChristopherAtkinson: The fixed that was reverted [92bd274](https://github.com/magnopus-opensource/connected-spaces-platform/pull/846/commits/92bd274249e08511aa040199f1d3cc8f4c0b8e2b) was at risk of undefined behaviour

The construction of the derived type, even though not optimal, will ensure the return matches the expectation and no longer attempts a potentially unsafe static cast while persisting the required data from the result object

MAG-ElliotMorris: > The fixed that was reverted [92bd274](https://github.com/magnopus-opensource/connected-spaces-platform/pull/846/commits/92bd274249e08511aa040199f1d3cc8f4c0b8e2b) was at risk of undefined behaviour

Do you not think this calls the entire method into question?

 I think that you're saying here that uses of this method get down the `if (exceptionType == Result)` path whilst calling the method such that the type of `ResultException->GetResult()` is not derived from `ResultsBase`, _despite_ that being the type `GetResult` stores, _and_ the cast to `ResultException` from `ExpectedExceptionBase` being valid and not unsafe. Can you make this make sense? If there's something the both of us are missing and you can't communicate it, you may need to setup a call.
 
 If you respond in writing, please provide explicit concrete examples of what you mean. Why specifically is there undefined behaviour. What specifically is "the expectation".

MAG-ChristopherAtkinson: For clarity, the continuation chain may throw any derived result exception; however, callbacks expect an explicit type:
 - Due to WASM not supporting dynamic casting, we need to ensure the type is known when static casting to prevent unsafe pointer casts.
 - Alongside that, we need to guarantee that the return type matches the expectation of the callback.
 - Conversation system are a prime example of this behaviour, where the callback will expect a AnnotationResult, but the exception could be triggered as AssetCollectionResult, AssetResult, UriResult etc... 
https://github.com/magnopus-opensource/connected-spaces-platform/blob/860b2e2e79e5087285c1f5e1fe05e05391ae28f9/Library/src/Systems/Conversation/ConversationSystemInternal.cpp#L944-L980
 - In the event that we cast an incompatible derived object, this would cause the memory allocation to be incorrect and potentially cause corrupt memory due to the mismatch in memory allocation size

*Note: I am taking in terms of result exceptions here, but there is also consideration for multiplayer exceptions as well. However, wanted to provide a simplified context.

MAG-ChristopherAtkinson: > > The fixed that was reverted [92bd274](https://github.com/magnopus-opensource/connected-spaces-platform/pull/846/commits/92bd274249e08511aa040199f1d3cc8f4c0b8e2b) was at risk of undefined behaviour
> 
> Do you not think this calls the entire method into question?
> 

This is a result of `static_cast` where `dynamic_cast` should be used, due to limitations, I believe we have around WASM 

> I think that you're saying here that uses of this method get down the `if (exceptionType == Result)` path whilst calling the method such that the type of `ResultException->GetResult()` is not derived from `ResultsBase`, _despite_ that being the type `GetResult` stores, _and_ the cast to `ResultException` from `ExpectedExceptionBase` being valid and not unsafe. Can you make this make sense? If there's something the both of us are missing and you can't communicate it, you may need to setup a call.

If you cast the derived object `AssetResult` to `AssetCollectionResult` where `AssetResult` is x1 bytes and `AssetCollectionResult` is x2 bytes. baring in mind these are now heap allocation as share_ptr. Then return a copy of the `AssetCollectionResult` of size x2. The discrepancy can cause out-of-bounds read or an over-read, leading to Undefined Behavior

We do have guarantee that the `static_cast` from `ExpectedExceptionBase` to `ResultException`

MAG-ElliotMorris: I see, concrete examples helped.

This method is a bit of a hack then. It's like `ForceTheExceptionTypeIntoTheResultTypeIWant`. I'm uncomfortable with this for obvious reasons, and there are ways to deal even without dynamic_cast. My first thought would be to look at `InvokeIfExceptionInChain`, as that catches a single `std::exception`, but I suspect you could be more dynamic with it to support multiple derived types.

 But I grant that it _does_ fulfil the ticket and it does provide a better user experience than the `invalid` that was here before.

I'll approve this under this reasoning. But I think this is debt and will need to be removed in the medium term.
Leaving this conversation open as @MAG-AdamThorn also was involved, will leave to him to close.

MAG-ChristopherAtkinson: Thank you @MAG-ElliotMorris, but you are correct, there is potentially more we could do here, but it's value vs cost. I think in general we need to a large consideration around our error and exception strategy and this is not the place for that. ,

MAG-AdamThorn: Thank you for the detailed response Christopher, that helped clarify things.

This is a marked improvement so approving.

I am putting together something for tracking this kind of technical debt at the moment, so I will add this to it before sharing with the team.

PR #847, opened 2025-11-06
[OF-1804] build: Continuous Deployment preparatory commit

MAG-AdamThorn:
CSP is migrating to nightly releases of CSP. This is a preparatory commit which includes the following:

- A new `CHANGELOG.md` file. We are introducing a `CHANGELOG.md` file to our repo to track the changes made for each release.
- An update to our `README.md` file. The readme has been updated to include a link to this changelist as well as to the releases section of GitHub. The plan is to then include this readme with our npm packages so that npm will display it by default on the package page.
- A new `CopyReadmeToPackage.py` script will be introduced to the TeamCity `Publish CSP` build config. This script copies the `.github/README.md` file to the package directory. The plan is to replace our existing `GenerateReadMeWithLink` script which runs as part of the Publish with this one. Once replaced the old script will be removed. This script has been tested locally and will be validated in a temporary `Test Publish CSP` TeamCity config as part of this initiative.

### DIFF
--- a/teamcity/CopyReadmeToPackage.py
+++ b/teamcity/CopyReadmeToPackage.py
@@ -40,7 +40,8 @@ def main():
         print(f"Error: Source README.md not found at '{source_readme_path}'. Please ensure it exists.", file=sys.stderr)
         sys.exit(1)
 
-    # Clean the output directory if it exists
+    # The destination_directory is used in subsequent build steps to create the package that will be uploaded to npm
+    # We therefore clean the directory if it exists
     if os.path.exists(destination_directory):
         shutil.rmtree(destination_directory)
     


teamcity/CopyReadmeToPackage.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
```
MAG-ChristopherAtkinson, line 2: [Fix] Missing sys Import

teamcity/CopyReadmeToPackage.py
```diff
@@ -0,0 +1,52 @@
...
+    destination_readme_path = os.path.join(destination_directory, "README.md")
+    
+    # Check if the source README.md exists
+    if not os.path.exists(source_readme_path):
+        print(f"Error: Source README.md not found at '{source_readme_path}'. Please ensure it exists.", file=sys.stderr)
+        sys.exit(1)
+
+    # Clean the output directory if it exists
+    if os.path.exists(destination_directory):
+        shutil.rmtree(destination_directory)
```
MAG-ChristopherAtkinson, line 46: [Nit] This seems fine at a glance as this should only contain the readme. However, this call is redundant, if the goal is to ensure that the readme is updated would recommend targeting the file directly

teamcity/CopyReadmeToPackage.py
```diff
@@ -0,0 +1,52 @@
...
+
+import git
+
+from Config import config
+
+def main():
+    # Ensure our current working directory is the git repository root
+    repo = git.Repo(os.path.abspath(os.path.dirname(__file__)), search_parent_directories=True)
+    root_dir = repo.working_dir
+    os.chdir(root_dir)
```
MAG-ChristopherAtkinson: [Consider] the root directory is changed; however, the script is using absolute paths, meaning that this call does not affect the resulting logic

teamcity/CopyReadmeToPackage.py
MAG-AdamThorn: Well spotted!

Fix: [ef7be26](https://github.com/magnopus-opensource/connected-spaces-platform/pull/847/commits/ef7be2670f255546831c37d2803c284ad883e99e)

teamcity/CopyReadmeToPackage.py
MAG-AdamThorn: Subsequence build steps use this same directory to create the package for upload to npm. So it will contain more than just the readme file. It is best to clean up the whole thing.

teamcity/CopyReadmeToPackage.py
MAG-AdamThorn: Good shout, the use of absolute path here is redundant. I have also wrapped it in a try/except block in case the git root cannot be found for any reason.

Fix: [ef7be26](https://github.com/magnopus-opensource/connected-spaces-platform/pull/847/commits/ef7be2670f255546831c37d2803c284ad883e99e)

teamcity/CopyReadmeToPackage.py
MAG-ChristopherAtkinson: Thank you for the clarification, because of that, it makes sense to clean the whole directory. Can this be captured in the comment block above to prevent assumption and ambiguity

teamcity/CopyReadmeToPackage.py
MAG-ChristopherAtkinson: Great, happy to resolve this thread

teamcity/CopyReadmeToPackage.py
MAG-AdamThorn: Expanded on the comment here: [537406c](https://github.com/magnopus-opensource/connected-spaces-platform/pull/847/commits/537406ca0bc9ce1574609dbb6d098ab320d1a031)

PR #848, opened 2025-11-07
[NT-0] chore: Finish NO_EXPORT heirarchy

MAG-ElliotMorris:
This PR adds NO_EXPORT blocks to symbols that derive from, or are dependent on base constructs that also have NO_EXPORT.

There arn't too many, mostly it's just been forgotten. The reason I want to do this is because the SWIG wrapper actually performs a CPP build, but it also wants to not have to deal with non-exported functions. Stripping/ignoring just the non-exported ones makes the build fail because suddenly there's nothing to `override`.

This isn't a problem with the other generators because they don't perform CPP builds. It _would_ be a problem if we ever wanted to make Unreal consume our symbols properly.

Manually discovered via automated stripping of NO_EXPORT symbols, and then attempting to build a dependent cpp lib (SWIG interop layer) that #includes all of our headers.

### DIFF
--- a/Library/include/CSP/Multiplayer/NetworkEventBus.h
+++ b/Library/include/CSP/Multiplayer/NetworkEventBus.h
@@ -229,7 +229,8 @@ private:
     CSP_END_IGNORE
 
     // Map internal event values to the deserializers needed to unpack them
-    std::unique_ptr<csp::common::NetworkEventData> DeserialiseForEventType(NetworkEvent EventType, const std::vector<signalr::value>& EventValues);
+    CSP_NO_EXPORT std::unique_ptr<csp::common::NetworkEventData> DeserialiseForEventType(
+        NetworkEvent EventType, const std::vector<signalr::value>& EventValues);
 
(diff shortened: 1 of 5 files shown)


Library/include/CSP/Multiplayer/NetworkEventBus.h
```diff
@@ -229,7 +229,8 @@ class CSP_API NetworkEventBus
     CSP_END_IGNORE
 
     // Map internal event values to the deserializers needed to unpack them
-    std::unique_ptr<csp::common::NetworkEventData> DeserialiseForEventType(NetworkEvent EventType, const std::vector<signalr::value>& EventValues);
+    CSP_NO_EXPORT std::unique_ptr<csp::common::NetworkEventData> DeserialiseForEventType(
```
MAG-ElliotMorris, line 232: [Comment] `NetworkEvent` is inside a CSP_IGNORE block, hence why this is now NO_EXPORT.

PR #849, opened 2025-11-07
[OF-1804] build: Add GH Action for changelog update

MAG-AdamThorn:
Adding a new GitHub Action which listens for a `release-published` event from TeamCity. This event includes the following payload data:
- release version
- release date

This payload data is used to update the changelog file.

### DIFF
--- a/.github/workflows/update-changelog-on-tc-publish.yml
+++ b/.github/workflows/update-changelog-on-tc-publish.yml
@@ -14,8 +14,8 @@ jobs:
 
       - name: Configure Git
         run: |
-          git config user.name "automaton"
-          git config user.email "automaton@magnopus.com"
+          git config user.name "${{ secrets.GH_AUTOMATON_USERNAME }}"
+          git config user.email "${{ secrets.GH_AUTOMATON_EMAIL }}"
           git pull
 
       - name: Get Release Version and Date from Payload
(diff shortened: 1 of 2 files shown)


.github/workflows/update-changelog-on-tc-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+  update-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Configure Git
+        run: |
+          git config user.name "automaton"
+          git config user.email "automaton@magnopus.com"
```
MAG-ElliotMorris: [Fix] I know this is just an email and not a password, but you perhaps shouldn't have put this into the repo. Still opens up an attack vector. I'd advocate for putting this as a repository secret instead.

.github/workflows/update-changelog-on-tc-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+          echo "RELEASE_DATE=$RELEASE_DATE" >> $GITHUB_OUTPUT
+
+      - name: Update CHANGELOG.md
+        run: |
+          RELEASE_VERSION="${{ steps.get_release_info.outputs.RELEASE_VERSION }}"
+          RELEASE_DATE="${{ steps.get_release_info.outputs.RELEASE_DATE }}"
+          CHANGELOG_FILE=".github/CHANGELOG.md"
+
+          # Escape special characters for sed (like dots in version numbers)
+          ESCAPED_VERSION=$(echo "$RELEASE_VERSION" | sed 's/\./\\./g')
```
MAG-ElliotMorris: [Nit] "for sed" is a bit confusing. It looks like `sed` is the thing doing the escaping.

.github/workflows/update-changelog-on-tc-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+
+            # This `sed` command searches for the line containing "## [Unreleased]".
+            # When it finds it, it performs a 'c\' (change) operation.
+            # It replaces the *entire line* with the following multiline string:
+            # "## [Unreleased]\n\n## [$RELEASE_VERSION] - $RELEASE_DATE"
+            sed -i "/## \[Unreleased\]/c\## \[Unreleased\]\n\n## \[$RELEASE_VERSION\] - $RELEASE_DATE" "$CHANGELOG_FILE"
+
+          else
+            echo "Changelog already contains entry for $RELEASE_VERSION or '## [Unreleased]' not found. No update needed."
+          fi
```
MAG-ElliotMorris: [Consider] This grep and sed logic is a bit much to all be encoded in a github action, I can't parse it. You might consider using a real programming language to write a clearer flow to modify the changelog, and just invoke that from here.

.github/workflows/update-changelog-on-tc-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo "Updating CHANGELOG.md for release $RELEASE_VERSION."
+
+            # This `sed` command searches for the line containing "## [Unreleased]".
+            # When it finds it, it performs a 'c\' (change) operation.
+            # It replaces the *entire line* with the following multiline string:
+            # "## [Unreleased]\n\n## [$RELEASE_VERSION] - $RELEASE_DATE"
+            sed -i "/## \[Unreleased\]/c\## \[Unreleased\]\n\n## \[$RELEASE_VERSION\] - $RELEASE_DATE" "$CHANGELOG_FILE"
+
+          else
+            echo "Changelog already contains entry for $RELEASE_VERSION or '## [Unreleased]' not found. No update needed."
```
MAG-ElliotMorris: [Question] Shouldn't this abort the entire workflow? From my reading right we'll still end up doing a git add and commit? Or does that trigger the `No changes to commit." case?

.github/workflows/update-changelog-on-tc-publish.yml
```diff
@@ -0,0 +1,73 @@
...
+          fi
+
+      - name: Commit and Push Changes
+        run: |
+          git add "${CHANGELOG_FILE}"
+          git commit -m "[NT-0] docs: Update CHANGELOG.md for release ${{ steps.get_release_info.outputs.RELEASE_VERSION }}" || echo "No changes to commit."
```
MAG-ElliotMorris: [Consider] The `NT-0` and `docs:` tags here are just noise that don't really apply. The first is for tracking work in Jira, and the second is so we know how to populate the changelog. You could just not put them in and use `--no-verify`

.github/workflows/update-changelog-on-tc-publish.yml
MAG-AdamThorn: Agreed. I have made these GH secrets.

Fix: [1e0fd8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/849/commits/1e0fd8ee49086bf6128c906a65239817d8774914)

.github/workflows/update-changelog-on-tc-publish.yml
MAG-AdamThorn: Yep. I have moved this parsing/update logic to a separate python script which the GH Action now invokes. This makes it a lot easier to reason about.

Update: [1e0fd8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/849/commits/1e0fd8ee49086bf6128c906a65239817d8774914)

.github/workflows/update-changelog-on-tc-publish.yml
MAG-AdamThorn: Agreed. I have moved this parsing/update logic to a separate python script which the GH Action now invokes. This makes it a lot easier to reason about and has the added bonus of allowing me to test locally in isolation.

Update: [1e0fd8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/849/commits/1e0fd8ee49086bf6128c906a65239817d8774914)

.github/workflows/update-changelog-on-tc-publish.yml
MAG-AdamThorn: Agreed. This has not been replaced by a separate script which the action invokes, see above.

.github/workflows/update-changelog-on-tc-publish.yml
MAG-AdamThorn: Agreed.

Update: [1e0fd8e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/849/commits/1e0fd8ee49086bf6128c906a65239817d8774914)

PR #852, opened 2025-11-12
[OF-1804] ci: Add ssh deploy key

MAG-AdamThorn:
Add the SSH key and permissions required for the GitHub Action to bypass branch protection and execute the final git push.


MAG-ElliotMorris (approved):
Wave through again, although if this keeps happening we'll need to find a way to actually check if this works prior to merge.

PR #853, opened 2025-11-13
[OF-1807] feat: support text asset type to support ai chatbot

MAG-ChristopherAtkinson:
Introduce a new TEXT type to EAssetType, in support of an AI chatbot.

### DIFF
--- a/Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
+++ b/Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-#include "../AssetSystemTestHelpers.h"
 #include "../SpaceSystemTestHelpers.h"
 #include "../UserSystemTestHelpers.h"
 #include "Awaitable.h"
@@ -29,7 +28,6 @@
 
 #include "gtest/gtest.h"
 #include <chrono>
-#include <filesystem>
 #include <thread>
 
 using namespace csp::multiplayer;
@@ -49,10 +47,6 @@ CSP_PUBLIC_TEST(CSPEngine, AIChatbotTests, AIChatbotSpaceComponentTest)
     auto& SystemsManager = csp::systems::SystemsManager::Get();
     auto* UserSystem = SystemsManager.GetUserSystem();
     auto* SpaceSystem = SystemsManager.GetSpaceSystem();
-    auto* AssetSystem = SystemsManager.GetAssetSystem();
-
-    const char* TestAssetCollectionName = "CSP-UNITTEST-ASSETCOLLECTION-MAG";
-    const char* TestAssetName = "CSP-UNITTEST-ASSET-MAG";
 
     // Log in
     csp::common::String UserId;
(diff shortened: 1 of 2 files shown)


Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
```diff
@@ -73,6 +79,43 @@ CSP_PUBLIC_TEST(CSPEngine, AIChatbotTests, AIChatbotSpaceComponentTest)
...
+    fread(UploadFileData, UploadFileSize, 1, UploadFile);
+    fclose(UploadFile);
+
+    csp::systems::BufferAssetDataSource BufferSource;
+    BufferSource.Buffer = UploadFileData;
+    BufferSource.BufferLength = UploadFileSize;
+
+    BufferSource.SetMimeType("text/plain");
+
+    printf("Uploading asset data...\n");
```
MAG-AdamThorn: [Question] Is there a reason this isn't a csp log?

Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
```diff
@@ -73,6 +79,43 @@ CSP_PUBLIC_TEST(CSPEngine, AIChatbotTests, AIChatbotSpaceComponentTest)
...
+
+    // Upload data
+    UploadAssetData(AssetSystem, AssetCollection, Asset, BufferSource, Asset.Uri);
```
MAG-AdamThorn: [Consider] I question what the value is in doing this. In this test we don't download the file again to check it is valid and nor should we, as we check file upload/download elsewhere.

I would consider removing file upload and just verify that the asset collection Id can be set and retrieved correctly.

MAG-ChristopherAtkinson: this was only introduced to ensure we have a case where we try to upload a .txt file. The general tests for assets and collection cover the general test cases, but not explicit asset types. Not that this should affect behaviour, happy to remove as this was based on existing patterns that seem superfluous

Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
MAG-ChristopherAtkinson: Not at all, I overlooked this when creating the test based on existing cases

Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
MAG-ChristopherAtkinson: reverted the changes here: [0b6f410](https://github.com/magnopus-opensource/connected-spaces-platform/pull/853/commits/0b6f4105fcadc99f2092393a0118d888cf5a453b)

.github/CHANGELOG.md
MAG-ChristopherAtkinson, line 1: artifacts as a result or poor timing with a rebase, while some ongoing branch protection work was being tested. I will clean up the branch later and clean the history

Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
MAG-AdamThorn: I think the better option in this case would be to add test/s to the AssetSystem that validate that an Asset of this new type can be uploaded/downloaded.

I have approved what is here though as other types are tested.

MAG-mv (approved):

Tests/src/PublicAPITests/ComponentTests/AIChatbotComponentTests.cpp
MAG-ChristopherAtkinson: The implementation for upload remains the same, meaning adding tests for each asset type seems redundant. Also, we would need to include tests for each extension in that asset type, e.g. for Image and Thumbnail .png .jpg .jpeg etc

MAG-ChristopherAtkinson: If you are happy, I would like to resolve this conversation on the premise that we have sufficient test coverage for the asset system in general

PR #854, opened 2025-11-13
[NT-0] fix: List::Insert erasing 0th element

MAG-ElliotMorris:
Seems like it's been this way since the beginning of time.

The placement new (sort of redundant even now...) was erasing the 0th element no matter what. Internally, we only ever use `Insert` at the 0th element, so it's not noticed.

Externally, this could very well be the cause of random seeming crashes. Depends on how much use this type has.

There's two other bugs in this function, todo with memcpy being undefined behaviour on non-trivial types, and the placement new assignment just being a memory leak for any types that need to run destructors.

Found this during the SWIG stuff, there were no unit tests for this type in CSP proper.

### DIFF
--- a/Tests/src/InternalTests/CommonTypeTests/List.cpp
+++ b/Tests/src/InternalTests/CommonTypeTests/List.cpp
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


Tests/src/InternalTests/CommonTypeTests/List.cpp
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright 2023 Magnopus LLC
```
MAG-ElliotMorris: [nit] 2025

MAG-ElliotMorris: [6a49380](https://github.com/magnopus-opensource/connected-spaces-platform/pull/854/commits/6a4938082340af43b588e29faee0bc6a92a9c382)

PR #857, opened 2025-11-13
[NT-0] fix: List::Insert erasing 0th element

MAG-ElliotMorris:
Seems like it's been this way since the beginning of time.

The placement new (sort of redundant even now...) was erasing the 0th element (zero constructing over it) no matter what. Internally, we only ever use `Insert` at the 0th element, so it's not noticed.

Externally, this could very well be the cause of random seeming crashes. Depends on how much use this type has.

There's two other bugs in this function, todo with memcpy being undefined behaviour on non-trivial types, and the placement new assignment just being a memory leak for any types that need to run destructors.

Found this during the SWIG stuff, there were no unit tests for this type in CSP proper.

### DIFF
--- a/Library/include/CSP/Common/List.h
+++ b/Library/include/CSP/Common/List.h
@@ -265,7 +265,7 @@ public:
         std::memmove(ObjectArray + (Index + 1), ObjectArray + Index, sizeof(T) * After);
         ++CurrentSize;
 
-        T* ObjectPtr = &ObjectArray[0];
+        T* ObjectPtr = &ObjectArray[Index];
         new (ObjectPtr) T;
         ObjectArray[Index] = Item;
     }
--- a/Tests/src/InternalTests/CommonTypeTests/List.cpp
+++ b/Tests/src/InternalTests/CommonTypeTests/List.cpp
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2025 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


Library/include/CSP/Common/List.h
```diff
@@ -265,7 +265,7 @@ template <typename T> class CSP_API List
         std::memmove(ObjectArray + (Index + 1), ObjectArray + Index, sizeof(T) * After);
         ++CurrentSize;
 
-        T* ObjectPtr = &ObjectArray[0];
```
MAG-AdamThorn, line 268: Wow!

Tests/src/InternalTests/CommonTypeTests/List.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+/* Use a struct because trivial types might not express some behaviors */
+struct TestStruct
+{
+    int x = 0;
+    std::string str = "Default";
+};
+}
+
+CSP_INTERNAL_TEST(CSPEngine, CommonListTests, ListInsertAtStartTest)
```
MAG-AdamThorn, line 35: [Question] Is the thinking here that we expand on these tests later now that we have at least some test coverage? I don't want to block the actual work you are supposed to be doing, but I think we should capture that additional work as a ticket or fully test the class.

MAG-ElliotMorris: No, not really. These are just acceptance tests to prove that this change works. `List.h` is not our future, so I don't think it's particularly valuable building out a unit test suite for it. 

We'd be forced to fix the polymorphism issues if we did build out a full suite 😨. Although there's an argument that even in the interim, we should address issues in this type. It's a lead engineer call on whether you think the interim improvements are valuable enough.

MAG-AdamThorn: Yes ok that makes sense. I am happy with coverage of the changes made. Thanks for clarifying.

MAG-AdamThorn (approved):

MAG-mv (approved):

Tests/src/InternalTests/CommonTypeTests/List.cpp
MAG-ElliotMorris: From the tone of your above comment, presuming you're happy for me to resolve this conversation. ❤️

PR #859, opened 2025-11-17
[NT-0] feat: Add non-exported iterators to Map

MAG-ElliotMorris:
Using these for the SWIG stuff. The spelling is different to our standards because they need to adapt into STL compatible names, similar to what we do in List. Expect more of this stuff for the other interop types over the next few weeks. As this is strictly forwarding the underlying container, I didn't feel the need for additional tests, but am willing to be told that I'm wrong there.

Iterators are a cool pattern man, you expose these standard things and you get practically every algorithm you need for free, without needing to adapt the underlying type.

Feels like there's a language trying to come out where containers are just a standard set of iterators and a utility header like `<algorithm>`

### DIFF
--- a/Library/include/CSP/Common/Map.h
+++ b/Library/include/CSP/Common/Map.h
@@ -34,8 +34,6 @@ template <typename TKey, typename TValue> class CSP_API Map
 {
 public:
     using MapType = std::map<TKey, TValue>;
-    using iterator = typename MapType::iterator;
-    using const_iterator = typename MapType::const_iterator;
 
     /// @brief Constructs a map with 0 elements.
     Map() { Container = new MapType(); }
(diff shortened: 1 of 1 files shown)


MAG-ElliotMorris:
Got caught by the not running generate_solution trap. Seems the current wrapper generator cant handle public typedefs it dosen't expect (even if surrounded by CSP_START/END_IGNORE blocks). Fortuitously I can just do [this](https://github.com/magnopus-opensource/connected-spaces-platform/pull/859/commits/e03575ecc5471e6e13b87ee37d5e3e9b0733f010).

MAG-ElliotMorris:
Putting this back to draft til I can figure out the CI. Working locally :(

PR #860, opened 2025-11-18
[OB-4761] fix: unreal crashes on oko asset drag and drop without internet connection

MAG-ChristopherAtkinson:
This is quite a complicated problem; the PR here resolves the symptom but not the underlying cause. I will capture the overarching issues externally to this, but here is some additional information to highlight the main problem at a high level:

 - CSP, when experiencing an unexpected network interruption, will only receive a notification when attempting to operate. 
 - POCO will then send `ConnectionResetException` "Connection reset by peer" exception
   - POCO does support a `ConnectionAbortedException`, but it seems that the expected exception in an unexpected network interruption is `ConnectionResetException`
   - In any case, we trigger the same behaviour on all failure cases in `CSPWebSocketClientPOCO::HandleReceiveError`
 - The exceptions thrown do not always interrupt the current operation, meaning that we execute methods such as 'SendObjectMessage' with invalid data while in a disconnected state (e.g. anytime we require external data as part of the request, 'GenerateObjectIds')
 - Resulting in undefined behaviour from POCO when attempting to re-construct `websocket_factory` (Leaning towards a segmentation fault)

In the create entity multiplayer flow, we see this exact issue, where we invoke 'SendObjectMessage' event with an invalid ID due to a previous failed 'GenerateObjectIds' event. 

> [!NOTE]
> This is where the fix for the symptom is implemented, preventing execution of subsequent calls. I did a quick audit over similar functions to see if we had other potential edge cases, to no avail.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### 🍰 🙌 New Features
+
+- [NT-0] feat: Add non-exported iterators to Map by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/859
+  Add non-exported iterators in support of the SWIG generated interop API work.
+  
 ### 🐛 🔨 Bug Fixes
 
 - [OB-4761] fix: unreal crashes on oko asset drag and drop without internet connection by MAG-ChristopherAtkinson in https://github.com/magnopus-opensource/connected-spaces-platform/pull/860
@@ -20,7 +25,7 @@ All notable changes to this project will be documented in this file.
 
 ### 🐛 🔨 Bug Fixes
 
-- [NT-0] fix: List::Insert erasing 0th element by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/850
+- [NT-0] fix: List::Insert erasing 0th element by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/857
   Inserting elements into a List was always adding them as the 0th index rather than at the position specified.
 - [OB-4723] fix: exceptions in continuations return incorrect result by MAG-ChristopherAtkinson in https://github.com/magnopus-opensource/connected-spaces-platform/pull/846
   In the continuation flow, we would return an invalid result and therefore lose the associated result data. This change ensures the result data from the exception is preserved.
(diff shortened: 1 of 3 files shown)


MAG-ElliotMorris:
[Question] Is it doable to have a test that fails before this change, and passes now?

I understand the internet connection bit might make this hard, not sure how developed our mocks are in this area.

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -281,3 +285,42 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, TestErrorLogged
...
+            });
+
+    using MockEntityCreatedCallback = testing::MockFunction<void(SpaceEntity*)>;
+    MockEntityCreatedCallback MockCallback;
+
+    // Expect the callback gets nullptr (not the greatest error return...)
+    EXPECT_CALL(MockCallback, Call(nullptr));
+
+    // Expect that we log the error message once and only once
+    const csp::common::String ErrorMsg = "Failed to generate object ID.";
```
MAG-ChristopherAtkinson, line 319: In the event that the function failed to return when an exception is thrown, this would trigger multiple entries failing the test

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -70,7 +70,8 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, TestSuccessInRe
                 EXPECT_FALSE(Result.get_exception());
                 EXPECT_EQ(Result.get(), uint64_t(55));
             })
-        .then(async::inline_scheduler(), [](async::task<void> CheckForErrorsTask)
+        .then(async::inline_scheduler(),
```
MAG-ChristopherAtkinson, line 73: Apologies for the noise. It seems the file was formatted during my changes

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -281,3 +285,42 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, TestErrorLogged
...
+
+                Callback(Value, ExceptionPtr);
+
+                return async::make_task(std::make_tuple(Value, ExceptionPtr));
+            });
+
+    using MockEntityCreatedCallback = testing::MockFunction<void(SpaceEntity*)>;
+    MockEntityCreatedCallback MockCallback;
+
+    // Expect the callback gets nullptr (not the greatest error return...)
```
MAG-ElliotMorris, line 315: [Comment] You can say that again.

PR #861, opened 2025-11-19
[OF-1804] feat: Add changelog update enforcement GH Action

MAG-AdamThorn:
As part of our work on the Continuous Deployment initiative we have added a `.github/CHANGELOG.md` file. This file should be updated as part of every PR, of if an update to the changelog is not required, the `skip-changelog` PR label should be applied. This PR adds a new GitHub Action that enforces that.

In addition this PR also contains an update to the changelog as well as an update to some of our documentation which related to our release process. 

I have confirmed that the GitHub Action is working correctly while the PR was in draft by doing the following:

1. Commit # 1 contained just the GH `Changelog Check` Action and documentation change. The Action therefore failed as no changelog update or label was present. I added the label, which reran the action and it passed. Removing the label once more caused the Action to rerun and fail.
2. Commit # 2 included the update to the changelog. The Action automatically reran and passed. Adding and then removing the label caused the Action to rerun each time and it continued to pass with the changelog update present.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,7 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### 🍰 🙌 New Features
 
+- [NT-0] feat: Add non-exported iterators to Map by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/859
+  Add non-exported iterators in support of the SWIG generated interop API work.
 
 ## [6.11.0] - 2025-11-18_11-14-15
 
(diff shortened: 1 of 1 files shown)


Library/docs/source/manual/documentation/branching_strategy.md
```diff
@@ -39,10 +40,19 @@ Here is a high-level overview of how a change moves from development through to
...
 ### Release Deployment
-New CSP packages are published via the `Publish CSP` deploy step on Team City. This will create a new tag for the release on the `main` branch, create a new GitHub release and push build artifacts, and publish new `CSP for Unity` and `CSP for Web` NPM packages to npmjs. 
+
+CSP is published nightly via the automated `Publish CSP` deploy step on Team City. This deploy step will:
+
+-   Create a new tag for the release on the `main` branch.
+-   Create a new GitHub release and push build artifacts.
+-   Publish new `CSP for Unity` and `CSP for Web` NPM packages to npmjs.
+
+Please note that the while CSP releases follow the SemVer versioning format, they do not follow the SemVer versioning rules. Our release versions are structured as follows:
```
MAG-ElliotMorris, line 53: [nit] Major.Minor.Patch existed before the SemVer standardization, SemVer dosen't own that format so you don't specifically need to call this out. Although I can see the argument for doing it as I guess a lot of people would just assume nowadays.

MAG-AdamThorn: I see your point though I don't feel that the sentence necessarily reads like that. I will leave it as is.

PR #862, opened 2025-11-20
[NT-0] test: Update QuotaSystem test

MAG-AdamThorn:
Now that CHS have added the `GoogleGenAI` Quota tier feature, the `GetTierFeaturesQuota` test has been updated to remove temporary logic added to support the transition.

Please note: The new tier feature was not added at the end as expected and so its position has been updated in the enum to match.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -8,11 +8,16 @@ All notable changes to this project will be documented in this file.
 
 - [NT-0] feat: Add non-exported iterators to Map by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/859
   Add non-exported iterators in support of the SWIG generated interop API work.
-  
+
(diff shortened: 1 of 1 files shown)


Library/include/CSP/Systems/Quota/Quota.h
```diff
@@ -54,9 +54,9 @@ enum class TierFeatures
     TotalUploadSizeInKilobytes,
     Agora,
     OpenAI,
+    GoogleGenAI,
```
MAG-ElliotMorris, line 57: [Question] This has no backwards compatibility implications, given that `Shopify` and `TicketedSpace` are now one value lower?

MAG-AdamThorn: None that I am aware of or could find. There is nothing that expects a specific ordering. We don't expose an ordered list representing the enum values to clients and internally we use utilities to convert to/from enum/string for all operations.

PR #868, opened 2025-11-26
[OF-1758] feat: replace assertion in token expiration with meaningful log to notify users

MAG-ChristopherAtkinson:
RefreshIfExpired invokes a fatal log message in place of the existing assert as asserting on a refresh token failure is too aggressive because it provides no useful information to the user and offers no way to recover.
 - Include a test for the failure path of the refresh token

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -9,6 +9,11 @@ All notable changes to this project will be documented in this file.
 - [OB-4579] fix!: Issues around renaming hotspot groups by MAG-ChristopherAtkinson
   Rename event has been deprecated and removed at the services level. This is a breaking change due to the removal of the csp::common::ESequenceUpdateType::Rename.
 
+🍰 🙌 New Features
+
+  [OF-1758] feat: replace assertion in token expiration with meaningful log to notify users by MAG-ChristopherAtkinson
+  RefreshIfExpired invokes a fatal log message in place of the existing assert as asserting on a refresh token failure is too aggressive.
+
 ## [6.14.0] - 2025-12-02_17-43-07
 
 🔥 ❗Breaking Changes


Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -504,6 +504,30 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, RefreshTest)
...
+    // Log in
+    csp::common::String UserId;
+    LogInAsNewTestUser(UserSystem, UserId, true, true, TokenOptions);
+
+    RAIIMockLogger MockLogger {};
+    csp::systems::SystemsManager::Get().GetLogSystem()->SetSystemLevel(csp::common::LogLevel::Fatal);
+    csp::common::String Msg = "User authentication token refresh failed!";
+    EXPECT_CALL(MockLogger.MockLogCallback, Call(csp::common::LogLevel::Fatal, Msg)).Times(1);
+
+    std::this_thread::sleep_for(120s);
```
MAG-ChristopherAtkinson, line 527: [Comment] I have had timing inconsistencies; it seems there is a potential for the refresh token to be accepted after the elapsed time. I found that 120 is consistent, but 110 has intermittent failures.

MAG-ChristopherAtkinson:
Going to close the PR for the time being, as it seems that CHS flow may not support refresh tokens of such a short duration, meaning this is not possible to test reliably currently. Further investigations are required will provide more information on the associated ticket.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn, line 527: [fix] I can't remember exactly what was discussed and agreed with CHS, but they are planning to look into this when they have more time right? Do they have a ticket for that? We should add a comment here explaining why we are sleeping for 120 seconds and ideally link the ticket.

MAG-ChristopherAtkinson: Yeah CHS has tracked the investigation work. From our perspective, I will disable the test once the main review has taken place as well as update the changelog (waiting to update as this will prevent it from being accidentally merged)

I can add a comment block when disabling the test:

> This test is currently disabled pending investigation from MCS related to intermittent failures at short duration for refresh token expiry 
> Refer to CHS-5441 Investigate why refresh tokens with 5 second expiry only consistently fail after 120 seconds for more information

MAG-ChristopherAtkinson: Test has now been disabled based on the previous discussion in weekly sync: [7e8bcbf](https://github.com/magnopus-opensource/connected-spaces-platform/pull/868/commits/7e8bcbf58a04aa639a2523459db7d541789da25c)

PR #871, opened 2025-11-27
[NT-0] test: Prevent hanging on failure in uvu tests

MAG-ElliotMorris:
Do 2 things to achieve this :

- Ensure graceful shutdown of the puppeteer browser in the case it does not shut down correctly via an exceptional path.
- Shutdown the typescript loaded CSP at the end of the test, regardless of success of failure.

This dosen't fix any test issues, but makes sense to do regardless.

Can't really test this, as it requires failing tests. Works locally, can only assume it'll behave the same on CI. I'd normally make a failing test and run a quick CI job but with TC the way it is it dosen't seem worth it.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -20,6 +20,12 @@ All notable changes to this project will be documented in this file.
 - [NT-0] feat!: expose refresh token expiry length by MAG-ChristopherAtkinson in https://github.com/magnopus-opensource/connected-spaces-platform/pull/863
   Expose RefreshTokenExpiryLength to the token options to support clients configuring custom durations.
 
+### 🙈 🙉 🙊 Test Changes
+
+- [NT-0] fix: Prevent uvu (wasm test runner) hanging on failed tests, by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/871 
+  Achieves this by appropriately shutting down CSPFoundation at the end of the test suite run,
+  as well as guarding against puppeteer errors via try/finally.
+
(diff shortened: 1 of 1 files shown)


.github/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ All notable changes to this project will be documented in this file.
 - [NT-0] feat!: expose refresh token expiry length by MAG-ChristopherAtkinson in https://github.com/magnopus-opensource/connected-spaces-platform/pull/863
   Expose RefreshTokenExpiryLength to the token options to support clients configuring custom durations.
 
+### 🙈 🙉 🙊 Test Changes
+
+- [NT-0] fix: Prevent uvu (wasm test runner) hanging on failed tests, by MAG-ElliotMorris in https://github.com/magnopus-opensource/connected-spaces-platform/pull/871 
+  Achieves this by appropriately shutting down CSPFoundation at the end of the test suite run,
+  as well as guarding against puppeteer errors via try/finally.
 ### 🔨 🔨 Chore
```
MAG-AdamThorn, line 29: [nit] Nice to have a newline between the summary and the chore category.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/871/commits/727f7ce169ab01ddd618ec39c6ebe6f1c634ef43

MAG-ElliotMorris: Had to rebase away a conflict so this commit is probably gone now. It's dealt with nonetheless.

PR #872, opened 2025-11-27
[OF-1806] refac!: Add a new URL property and deprecate others in the AvatarSpaceComponent

MAG-DanielMurdolo:
To simplify the sharing of avatar component meshes that represent the spatial avatar between users of CSP, we've deprecated two properties and added a third to replace them.

AvatarMeshIndex and CustomAvatarUrl have now been marked as deprecated and their API functions removed.

AvatarUrl has been added in their place, and an API function added. This takes a string value that is used to point to a mesh location.


.github/CHANGELOG.md
```diff
@@ -8,14 +8,16 @@ All notable changes to this project will be documented in this file.
 
 - [NT-0] feat!: expose refresh token expiry length by MAG-ChristopherAtkinson in https://github.com/magnopus-opensource/connected-spaces-platform/pull/863
   Expose RefreshTokenExpiryLength to the token options to support clients configuring custom durations.
+- [OF-1806] refac!: Add a new URL property and deprecate others in the AvatarSpaceComponent in https://github.com/magnopus-opensource/connected-spaces-platform/pull/872
+  Simplify the setting and getting of the mesh URL for the avatar component
```
MAG-ElliotMorris: [Consider] Listing the api method deltas that make this a breaking change.

Tests/src/PublicAPITests/MultiPlayerTests.cpp
```diff
@@ -1623,3 +1623,65 @@ CSP_PUBLIC_TEST(CSPEngine, MultiplayerTests, TestMultiplayerDisconnectionWhenNew
...
+    RealtimeEngine->SetEntityFetchCompleteCallback([](uint32_t) {});
+
+    auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id, RealtimeEngine.get());
+
+    EXPECT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+    RealtimeEngine->SetRemoteEntityCreatedCallback([](csp::multiplayer::SpaceEntity* /*Entity*/) {});
+
+    // Create parent entity
+    csp::common::String ObjectName = "Object 1";
```
MAG-mv: [Consider] Use CreateTestObject instead of setting up an entity yourself. It makes tests easier to read.

.github/CHANGELOG.md
MAG-DanielMurdolo: Updated

MAG-mv (approved):

Tests/src/PublicAPITests/MultiPlayerTests.cpp
MAG-DanielMurdolo: Updated

PR #875, opened 2025-11-29
[NT-0] fix: Replace List implementation with std::vector

MAG-ElliotMorris:
This type is riddled with issues, the placement new manual management doesn't account for the difference between trivially copyable and non-trivially copyable types very well, and this is playing havok on wrapper gen stuff where ownership gets complicated.

To be honest, I couldn't figure out exactly what in this was causing an issue with the C# garbage collector. I think it's the std::copy copying types that in themselves contain non-trivially relocatable types (like maps). This was manifesting as an error similar to that map<string, string> release error we saw a while back.

Call this experimental, as I'm not sure it actually addresses the issue (I have a 100% repro using the C# bindings), but we'll see. It's fully non breaking at least.

### DIFF
--- a/Library/include/CSP/Common/List.h
+++ b/Library/include/CSP/Common/List.h
@@ -20,9 +20,8 @@
 
 #include <algorithm>
 #include <cassert>
-#include <cstring>
 #include <initializer_list>
-#include <utility>
+#include <vector>
 
 namespace csp::common
 {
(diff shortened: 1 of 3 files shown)


MAG-ElliotMorris:
Agents are disable for maintenance, was wanting to use CI to get a fuller coverage of this experimental change, guess that's not happening.

PR #876, opened 2025-12-01
[NT-0] fix: Fix SendReceiveNetworkEvent wasm test failure

MAG-mv:
I noticed that we don't have a matching C++ test that uses SendNetworkEvent, only SentdNetworkEventToClient. After changing to use this function instead, the wasm test consistatly passes.


MAG-mv:
@MAG-ElliotMorris Thanks I will change to draft. Yeah, this contains the other fix I mentioned which seems to be working. Still testing

PR #877, opened 2025-12-04
[OPE-2982] fix: Unity Android debug builds crash on login

mag-douglascooperhind:
The change fixes an issue where Unity Android debug builds crashed on login due to it failing to get a handle to the shared library. This adds Android specific library search names.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-🐛 🔨 Bug Fixes
+### 🐛 🔨 Bug Fixes
 
 - [OW-2312] fix: Add select_account for authorise URL prompt parameter.
   Google made changes to the way the oauth URL works, so we need to set prompt to select_account rather than none to restore the functionality. This change updates that parameter in the GetThirdPartyProviderAuthoriseURL function.


.github/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### 🐛 🔨 Bug Fixes
 
+- [OPE-2982] fix: Unity Android debug builds crash on login. Caused by Android not being able to find the CSP library file.
```
MAG-ElliotMorris: [Fix] Additionally note the specific lib name change that happened on android. People will be scanning this changelog and may need to know.

MAG-ElliotMorris (changes requested):

MAG-ChristopherAtkinson (approved):

Library/src/CSPFoundation.cpp
```diff
@@ -28,12 +28,21 @@
 
 #include <cstdio>
 #include <fmt/format.h>
-
+#if defined(CSP_ANDROID)
+// Android specific LIB_NAME definition.
+#if defined(DEBUG)
+#define LIB_NAME "libConnectedSpacesPlatform_D.so"
+#else
+#define LIB_NAME "libConnectedSpacesPlatform.so"
```
MAG-ElliotMorris, line 36: [Question] It's odd to me that the `.so` is included here. But not the `.dll` below. I'm open to this just being a platform difference, but we should understand why.

.github/CHANGELOG.md
mag-douglascooperhind: Added this extra detail

mag-douglascooperhind:
> As discussed, please confirm when you have manually run the android build (both release and debug) on a device and exercised at least one function from inside the binary (some sort of test suite would be better)

I think because I rebased the branch to create the PR I now need up update Unity's CSP to get it building correctly before I can properly test this. So I'm gonna switch to that now.

Library/src/CSPFoundation.cpp
mag-douglascooperhind: I've been Googling to try and find something that explicitly says the requirements of dlopen() but I'm struggling to find anything other than examples that show people using file extensions.

MAG-ElliotMorris: I did some duckduckgo'ing myself. Seems that this is correct. `.so` is recommended for portability on unix platforms as `dlopen` just uses exactly what you give it. I suspect various OS' flavours may actually handle file-opening differently and some demand it and some don't, but if you have it, it should always work.

Windows appends .dll automatically in `LoadLibrary`, unless you specifically tell it not to by adding a `.` to the file name/path.

PR #879, opened 2025-12-05
[OB-4579] fix!: issues around renaming hotspot groups

MAG-ChristopherAtkinson:
CHS deprecated and removed the use of the `Rename` event type in sequences; the hotspot sequence behaviour still expected it. This PR refactors the code to the new expected behaviour, defined as follows: 

 - `PUT / sequences` has an optional `NewKey` param, which, if provided, will trigger the following:
   - `ESequenceUpdateType::Update` event containing `Key` original group id that has been renamed, and `NewKey` group id which has been provided
   - No longer triggers a secondary `ESequenceUpdateType::Update` which was previously needed as the sequence would need updating for the changes to apply, as this is handled in the original `PUT / sequences`
   - No longer triggers a `ESequenceUpdateType::Rename` as this has been removed
   
> [!IMPORTANT]
> The original ticket also noted an issue with the return of the delete endpoint: _Deleting a hotspot group sends an ESequenceUpdateType.Delete the event, but with the sequence name and not the hotspot group name._
>
> However, in testing, I found that the group name was being returned as expected.

### DIFF
--- a/Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
+++ b/Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
@@ -104,9 +104,6 @@ void HotspotSequenceSystem::CreateHotspotGroup(
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
-            HotspotGroupResult ReturnValue(Result.GetResultCode(), Result.GetHttpResultCode());
-            // convert SequenceResult To HotspotGroupResultCallback
-            Callback(ReturnValue);
             return;
         }
 
@@ -138,9 +135,6 @@ void HotspotSequenceSystem::RenameHotspotGroup(
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
-            HotspotGroupResult ReturnValue(Result.GetResultCode(), Result.GetHttpResultCode());
-            // convert SequenceResult To HotspotGroupResultCallback
-            Callback(ReturnValue);
             return;
         }
 
@@ -163,9 +157,6 @@ void HotspotSequenceSystem::UpdateHotspotGroup(
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
-            HotspotGroupResult ReturnValue(Result.GetResultCode(), Result.GetHttpResultCode());
-            // convert SequenceResult To HotspotGroupResultCallback
-            Callback(ReturnValue);
             return;
         }
         auto Data = Result.GetSequence();
@@ -190,9 +181,6 @@ void HotspotSequenceSystem::GetHotspotGroup(const csp::common::String& GroupName
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
-            HotspotGroupResult ReturnValue(Result.GetResultCode(), Result.GetHttpResultCode());
-            // convert SequenceResult To HotspotGroupResultCallback
-            Callback(ReturnValue);
             return;
         }
         auto Data = Result.GetSequence();
@@ -214,9 +202,6 @@ void HotspotSequenceSystem::GetHotspotGroups(HotspotGroupsResultCallback Callbac
     {
         if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
         {
-            HotspotGroupsResult ReturnValue(Result.GetResultCode(), Result.GetHttpResultCode());
-            // convert SequenceResult To HotspotGroupResultCallback
-            Callback(ReturnValue);
             return;
         }
         auto Data = Result.GetSequences();


Library/src/Multiplayer/NetworkEventSerialisation.cpp
```diff
@@ -400,21 +400,7 @@ csp::common::SequenceChangedNetworkEventData DeserializeSequenceHotspotChangedEv
...
-        if (ParsedEvent.EventValues[2].GetReplicatedValueType() == csp::common::ReplicatedValueType::String)
-        {
-            ParsedEvent.HotspotData->NewName = GetSequenceKeyIndex(ParsedEvent.NewKey, 2);
-        }
-        else
-        {
-            LogSystem.LogMsg(csp::common::LogLevel::Error,
-                "SequenceHotspotChangedEvent - The expected new name of the hotspot sequence was not found in the event payload.");
-        }
-    }
```
MAG-ChristopherAtkinson, line 417: [Comment] There is no need to explicitly handle the rename case any further; if the `NewKey` is not provided, this will resolve as an empty string, which matches the expectation that the field is optional.

Library/include/CSP/Common/NetworkEventData.h
```diff
@@ -110,11 +110,10 @@ class CSP_API AccessControlChangedNetworkEventData : public NetworkEventData
 
 enum class ESequenceUpdateType
 {
-    Create,
-    Update,
-    Rename,
```
MAG-AdamThorn, line 115: [question] Should this be marked deprecated rather than being removed?
Are we still expecting to receive an update type from CHS with a value of 2?

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
```diff
@@ -133,38 +133,24 @@ void HotspotSequenceSystem::RenameHotspotGroup(
     auto SpaceId = SpaceSystem->GetCurrentSpace().Id;
     auto Key = CreateKey(GroupName, SpaceId);
     auto NewKey = CreateKey(NewGroupName, SpaceId);
-    auto SQ = this->SequenceSystem;
 
-    auto CB = [Callback, SQ, NewKey, NewGroupName, SpaceId](const SequenceResult& Result)
+    auto CB = [Callback, NewGroupName](const SequenceResult& Result)
     {
-        if (Result.GetResultCode() != csp::systems::EResultCode::Success)
+        if (Result.GetResultCode() == csp::systems::EResultCode::InProgress)
```
MAG-AdamThorn, line 136: [question] Why are we calling the callback here if the request is in progress?

Library/include/CSP/Common/NetworkEventData.h
MAG-ChristopherAtkinson: No, CHS no longer supports the Rename event at all; it was previously deprecated but not marked, so this is closer to removing a deprecated state that has reached end of life.

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
MAG-ChristopherAtkinson: This is a common pattern in hotspots, and I didn't give it much thought. Thinking about it, that does seem strange, as this should be a 1:1 return

MAG-ChristopherAtkinson: Yeah, this pattern was infested within the hotspot sequence system, as such has been remove: [2844430](https://github.com/magnopus-opensource/connected-spaces-platform/pull/879/commits/284443027997411fb56e8c001d1498174878d8b5)

I am surprised this hasn't caused more problems until now with the multiple potential callback being triggered

MAG-AdamThorn (approved):

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
MAG-mv: I dont think this should cause issues, we handle in progress callbacks in all of our tests, and clients should, too.

Library/include/CSP/Common/NetworkEventData.h
MAG-AdamThorn: So do we still need to account for it in the switch statement?

MAG-ChristopherAtkinson: I believe we will get a compiler warning or error when not all possible enum values are handled in a switch statement. if enabled anyway, but this is handled still as the raw int value

Library/src/Systems/HotspotSequence/HotspotSequenceSystem.cpp
MAG-ChristopherAtkinson: @MAG-mv from the approval, guessing you agree with the change to no longer return in-progress callbacks

Library/include/CSP/Common/NetworkEventData.h
MAG-AdamThorn: Not sure I follow as the enum value no longer exists. We need to get this in though so putting that conversation aside for now.

MAG-ChristopherAtkinson: Yeah was meaning here: https://github.com/magnopus-opensource/connected-spaces-platform/blob/284443027997411fb56e8c001d1498174878d8b5/Library/src/Multiplayer/NetworkEventSerialisation.cpp#L50

We log that we received the state but do not prevent execution

PR #880, opened 2025-12-08
[OF-1542] Callbacks Firing Twice

MAG-mv:
This PR fixed an issue with DestroyEntity where failure cases would case the client provided callback to fire twice due to not returning. The ticket also mentions CreateAvatar, which has already been fixed, and CreateObjectInternal which has been removed.

When testing CreateAvatar, I found an issue where the client provided callback wouldnt fire at all due to exceptions not being forwarded. This issue is explained more in the commit description

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -19,8 +19,13 @@ All notable changes to this project will be documented in this file.
 
 🍰 🙌 New Features
 
-  [OF-1758] feat: replace assertion in token expiration with meaningful log to notify users by MAG-ChristopherAtkinson
+- [OF-1758] feat: replace assertion in token expiration with meaningful log to notify users by MAG-ChristopherAtkinson
   RefreshIfExpired invokes a fatal log message in place of the existing assert as asserting on a refresh token failure is too aggressive.
+  
+🙈 🙉 🙊 Test Changes
+
+- [NT-0] test: Await exiting spaces in certain realtime engine tests by MAG-ElliotMorris.
+  Correct usage of the realtime engine requires that the exit space call returns fully before deleting the memory, these tests did not do that.
 
 ## [6.14.0] - 2025-12-02_17-43-07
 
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -178,13 +195,12 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, TestErrorInSend
...
             {
-                auto [Result, Exception] = Results;
-                EXPECT_TRUE(Exception);
+                EXPECT_TRUE(Id.get_exception());
                 try
                 {
-                    std::rethrow_exception(Exception);
+                    std::rethrow_exception(Id.get_exception());
                 }
                 catch (std::runtime_error error)
```
MAG-ChristopherAtkinson, line 205: [Fix] seems the CI is expecting a different error here, causing the test to fail: `mock exception` vs `Multiplayer error. mock exception` would look at https://github.com/magnopus-opensource/connected-spaces-platform/blob/574ec9d79f07b5193e17a70264c150f18999b3ee/Library/include/CSP/Multiplayer/ContinuationUtils.h#L50

.github/CHANGELOG.md
```diff
@@ -87,6 +87,8 @@ All notable changes to this project will be documented in this file.
   As part of ongoing Unity interop API work, we updated the 'MimeTypeHelper' to return a const ref string for the mime type to ensure correct SWIG code generation.
 - [OF-1524] fix: Wasm callbacks are now automatically forwarded to the main thread by MAG-mv in https://github.com/magnopus-opensource/connected-spaces-platform/pull/806
   This fixes crashes with the log callback if UserSystem::Login fails, and also the MultiplayerConnection::ConnectionInterrupted callback on wasm builds.
+- [OF-1542] fix: Prevent OnlineRealtimeEngine::DestroyEntity callback from firing twice on failure by MAG-mv in https://github.com/magnopus-opensource/connected-spaces-platform/pull/880
```
MAG-AdamThorn: [fix] we have agreed to no longer include a link to the PR in the changelog. This decision may have been made prior to use raising this PR. Please can you expand your description here to include the necessary information?

MAG-ChristopherAtkinson (approved):

PR #884, opened 2025-12-12
[NT-0] fix: Fix rare race condition in legacy leader election

MAG-ElliotMorris:
When interacting with enable/disable leader election immediately after joining a space, there is a possibility for a race condition if the deletion happens after the check for null to branch the leader election, but before the actual access.

This occurs most readily in `AddSecondScriptTest`, where we explicitly disable leader election.

This is a rather easy one to reason about, imagine if a user called `DisableLeaderElection` whilst the callback thread was past the if conditions on lines 776, but before the leader election object is used later in the function.

This is a bit of a patch-job, the true fix is not doing leader election in this function, but rather doing it in the OnFetched function which is when we "officially" are in a space, and is more of a transaction. @MAG-mv tells me this is possible to do once the legacy leader election stuff is removed.

Testing locally:
- Before: Fails reliably being run in a loop, normally in the 3-6th iteration.
- After: Passed 100 runs in a loop.

**Whilst possible it may be, I don't think this is really the cause of any persistant teamcity crashes.**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-🐛 🔨 Bug Fixes
+### 🐛 🔨 Bug Fixes
 
 - [NT-0] fix: Address rare race condition in legacy leadership election by MAG-ElliotMorris
   It was possible to trigger a memory race using `EnableLeaderElection` or `DisableLeaderElection` after having returned from `EnterSpace` in the OnlineRealtimeEngine.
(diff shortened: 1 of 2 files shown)


Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -478,6 +478,7 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
     // --------------------------------------------------------------------------
 
     std::recursive_mutex* TickEntitiesLock;
+    std::mutex LeadershipElectionLock;
```
MAG-ElliotMorris, line 481: [Question] Am I missing something with the need for the other lock to be a pointer? Seems like danger for no reason.

PR #885, opened 2025-12-16
Establish type hashing pattern

MAG-ElliotMorris:
Add std::hash specializations for our vector types, allowing them be stored in containers. The real reason for this however is so I can satisfy the IEquatable interface for C#.

This starts the pattern for these, we're going to need to do this for a fair chunk of our types, any that are conceptually "value-types".

We'll need one of these files for each of our 3 modules.
The reason this is centralized is twofold:

- std::hash specializations have to live in the std:: namespace so you don't get any information hiding by putting them along with the class anyhow
- I need to ignore these from the current wrapper gen, but _not_ for the new one. The only way I can really do this at the moment is on a file-by-file basis.

Also un-ignored some equality operators because I need these too in SWIG.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -4,7 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+🍰 🙌 New Features
 
+- [NT-0] feat: Exposed equality and inequality operators for `Vector2`, `Vector3`, `Vector4` and `Map` by MAG-ElliotMorris.
+  Also added hash functions in the `std` namespace for the Vector types.
 
 ## [6.17.0] - 2025-12-15_14-00-12
 


Library/include/CSP/Common/Hash.h
```diff
@@ -0,0 +1,55 @@
+/*
```
MAG-mv, line 17: [fix] Missing copyright

MAG-mv (changes requested):

Library/include/CSP/Common/Hash.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/885/changes/21010ce05c23542f7b54b047633d62278cd683cf

Library/include/CSP/Common/Vector.h
```diff
@@ -69,10 +69,8 @@ class CSP_API Vector2
     /// @param float Scalar
     Vector2 operator*(float Scalar) const;
 
-    CSP_START_IGNORE
```
MAG-SamBirley, line 72: [Question] How does these operators manifest in the Web API? Have we checked they produce valid syntax? I figure they must have been ignored for a reason before, and am assuming C# has already been tested, but worth at least checking the Web API is still kosher before considering merging.

Library/include/CSP/Common/Hash.h
```diff
@@ -0,0 +1,73 @@
...
+#include "CSP/CSPCommon.h"
+#include <CSP/Common/Vector.h>
+
+#include <functional>
+
+CSP_START_IGNORE
+namespace std
+{
+
+template <> struct hash<csp::common::Vector2>
```
MAG-SamBirley, line 39: [Question] Does this header file need to be in the public API? I think the answer is probably 'yes' for implementation reasons, but as this implies a kind of a utility class for equivalence checks, it's worth a conversation.

MAG-SamBirley: I can see it'd make the hash tests more painful, probably requiring hash getters per value type. 🤢

Library/include/CSP/Common/Vector.h
MAG-mv: We actually discussed this previously. We already has some comparison operators visible, so I would be suprised if this is an issue. AGreed we should check, though

MAG-mv (approved):
Approving assuming we're happy with the wasm output

Library/include/CSP/Common/Vector.h
MAG-ElliotMorris: Aye, spoke to Matt on this. There's loads of already visible operators.

The great and mysterious JS wrapper gen has this diff : 
image

It's just doc, guess I should at least write something on that. Will add a new commit shortly documenting the new operators. Also makes me think this probably does demand a changelog.

Do C++ operators become naturally callable from JS via emscripten compilation, even without any JS frontend for them? I ... sort of doubt it.

C# diff is even more wtf. Guess that's where the docs _would_ be.

image

Library/include/CSP/Common/Hash.h
MAG-ElliotMorris: The declarations do, and since this is a template implementation with a pretty inlineable implementation, standard practice is to put it all in the header.

> implies a kind of a utility class 

I don't think it does in the way you're thinking. `hash` in the `std` namespace is a special type, which makes containers and such work in more enhanced ways via its very existence (via C++ compile time branching).

This is something we'll use in interop wrapping, but I doubt it has public api implications. It might slightly benefit Unreal as they'll be able to naturally use any types with hashes as keys in hash-containers such as `std::unordered_map` and `std::set`.

Library/include/CSP/Common/Vector.h
MAG-SamBirley: Ok, well I guess it's at least consistent in it's pattern, even if that pattern is WTF. Thank you for checking. Will leave this thread open since you committed to another change here, but feel free to close it when you're done.

Library/include/CSP/Common/Hash.h
MAG-SamBirley: Do you think the header warrants being in some special directory that implies it's for STL compatibility?

Library/include/CSP/Common/Vector.h
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/885/changes/ef6636ab98472bf2014df72a9b0795409133a01c

Library/include/CSP/Common/Hash.h
MAG-ElliotMorris: No not really. Ideally these specializations would live in the headers of the types, but I don't think a central index public include is that wild of a pattern either.

I grant having to know to `#include "CSP/Common/Hash.h"` might be a minor footgun, I can imagine folks implementing their own hashes without realizing they are provided, that's my main concern. Putting it in a subdirectory almost makes that worse.

I can see what you're getting at, there is a category distinction to be made in the public include directory. Things like `SharedConstants`, `SharedEnums`, `fmt_Formatters`, etc, _feel_ like they should be seperated from "True" Api types like `LoginState` and `Map`, etc. 

I can't trivially see what that separation would be right now, I think the `Hash` file fits in the current set.

MAG-ElliotMorris: Closing as this wasn't a blocking comment, and Sam is off so the conversation can't progress.

@MAG-SamBirley If you think this needs further thought, happy to reopen the discussion.

PR #886, opened 2026-01-05
[OB-4961] fix: Fix some spaces not finding default scope

MAG-mv:
Changed the scope identification to only check if _autoGenerated is contained in the name, as older spaces dont use the space id

### DIFF
--- a/Library/src/Systems/Spaces/SpaceSystem.cpp
+++ b/Library/src/Systems/Spaces/SpaceSystem.cpp
@@ -305,28 +305,35 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
             .then(async::inline_scheduler(),
                 [this, FinishedGetScopeEvent, SpaceResult, RealtimeEngine](const csp::systems::ScopesResult& ScopesResult)
                 {
-                    // 2. Ensure we have the default, auto-generated scope.
-                    // This is done by checking for the global scope in the space, and that we only have 1 of these global scopes.
+                    // 2. Ensure we have single default, auto-generated scope.
+                    // This is done by checking for the global scope in the space, and that we only have 1 of these scopes, which is global.
                     // Right now, clients can't create scopes, so there should be a 0 chance of the above conditions failing.
                     const auto& Scopes = ScopesResult.GetScopes();
-                    const auto GlobalScopeCount
-                        = std::count_if(Scopes.begin(), Scopes.end(), [](const Scope& Scope) { return Scope.PubSubType == PubSubModelType::Global; });
 
-                    if (GlobalScopeCount < 1)
+                    if (Scopes.Size() < 1)
                     {
                         this->LogSystem->LogMsg(
-                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Space doesn't have a required global scope.");
+                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Space doesn't have a scope.");
                         FinishedGetScopeEvent->set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
-                            "SpaceSystem::RegisterScopesInSpace: Space doesn't have a global scope", MakeInvalid<csp::systems::SpaceResult>())));
+                            "SpaceSystem::RegisterScopesInSpace: Space doesn't have a scope", MakeInvalid<csp::systems::SpaceResult>())));
                         return;
                     }
 
-                    if (GlobalScopeCount > 1)
+                    if (Scopes.Size() > 1)
                     {
                         this->LogSystem->LogMsg(
-                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Multiple global scopes found. This version of CSP only supports spaces that have only a single global scope.");
+                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Multiple scopes found. This version of CSP only supports spaces that have only a single global scope.");
                         FinishedGetScopeEvent->set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
-                            "SpaceSystem::RegisterScopesInSpace: Space has multiple global scopes", MakeInvalid<csp::systems::SpaceResult>())));
+                            "SpaceSystem::RegisterScopesInSpace: Space has multiple scopes", MakeInvalid<csp::systems::SpaceResult>())));
+                        return;
+                    }
+
+                    if (Scopes[0].PubSubType != PubSubModelType::Global)
+                    {
+                        this->LogSystem->LogMsg(csp::common::LogLevel::Error,
+                            "SpaceSystem::RegisterScopesInSpace: Space doesn't contain a global scope.");
+                        FinishedGetScopeEvent->set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
+                            "SpaceSystem::RegisterScopesInSpace: Space doesn't contain a global scope.", MakeInvalid<csp::systems::SpaceResult>())));
                         return;
                     }
 


Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,8 +306,12 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
                 [this, FinishedGetScopeEvent, SpaceResult, RealtimeEngine](const csp::systems::ScopesResult& ScopesResult)
                 {
                     // 2. Ensure we have the default, auto-generated scope.
+                    // The default-generated scope always contains an id + "_autoGenerated".
+                    // With newer spaces, the id always seems to be the space id. However, with older ones, it seems to be a random id.
+                    // So to account for this, we check if the name contains "_autoGenerated" to identify the default scope.
+                    // This is NOT ideal, but we haven't fully implemented scopes at this point, and we need a way of getting the default scope.
                     auto DefaultScopeIt = std::find_if(ScopesResult.GetScopes().begin(), ScopesResult.GetScopes().end(),
-                        [](const Scope& Scope) { return Scope.Name == Scope.ReferenceId + "_autoGenerated"; });
+                        [](const Scope& Scope) { return Scope.Name.Contains("_autoGenerated"); });
```
MAG-ElliotMorris: [Question] Is there a test for this behaviour? If not [Consider] adding one.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,8 +306,12 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
                 [this, FinishedGetScopeEvent, SpaceResult, RealtimeEngine](const csp::systems::ScopesResult& ScopesResult)
                 {
                     // 2. Ensure we have the default, auto-generated scope.
+                    // The default-generated scope always contains an id + "_autoGenerated".
+                    // With newer spaces, the id always seems to be the space id. However, with older ones, it seems to be a random id.
+                    // So to account for this, we check if the name contains "_autoGenerated" to identify the default scope.
+                    // This is NOT ideal, but we haven't fully implemented scopes at this point, and we need a way of getting the default scope.
                     auto DefaultScopeIt = std::find_if(ScopesResult.GetScopes().begin(), ScopesResult.GetScopes().end(),
```
MAG-ElliotMorris: [Comment] It'd be nice if the object returned from `GetScopes` knew if it was a default scope. The information would have to come from CHS, which is the more appropriate source of answer for the "Is this a default scope" question. Doubt we can really do anything to that effect though, given how the teams are :(

MAG-AdamThorn: Yeah that's a neat idea. Worth discussing with CHS team. May be able to extend that to the concept of scope categories if that makes sense.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: ScopeLeadershipTest relies on this behaviour, otherwise SetOnVacatedAsScopeLeaderCallback and SetOnElectedScopeLeaderCallback will not work

MAG-mv: I can write a test specifically for the name

MAG-ElliotMorris: Probably redundant considering the discussions happening offline, feel free to close this comment when you change the implementation.

MAG-mv: Updated test in my latest commit to test for global scopes, as well: https://github.com/magnopus-opensource/connected-spaces-platform/pull/886/commits/31b6c22d8f1ea9e620b60f62cc27c7289a8de34d

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,12 +306,35 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
...
+                            "SpaceSystem::RegisterScopesInSpace: Space has multiple global scopes", MakeInvalid<csp::systems::SpaceResult>())));
+                        return;
+                    }
+
+                    auto DefaultScopeIt
+                        = std::find_if(Scopes.begin(), Scopes.end(),
                         [](const Scope& Scope) { return Scope.PubSubType == PubSubModelType::Global; });
 
-                    if (DefaultScopeIt == ScopesResult.GetScopes().end())
+                    if (DefaultScopeIt == Scopes.end())
```
MAG-AdamThorn: [fix] we already check if 'GlobalScopeCount < 1', is there a reason we are doing the same here for the 'find_if'?

.github/CHANGELOG.md
```diff
@@ -98,7 +98,8 @@ All notable changes to this project will be documented in this file.
   As part of ongoing Unity interop API work, we updated the 'MimeTypeHelper' to return a const ref string for the mime type to ensure correct SWIG code generation.
 - [OF-1524] fix: Wasm callbacks are now automatically forwarded to the main thread by MAG-mv in https://github.com/magnopus-opensource/connected-spaces-platform/pull/806
   This fixes crashes with the log callback if UserSystem::Login fails, and also the MultiplayerConnection::ConnectionInterrupted callback on wasm builds.
-
+- [OB-4961] fix: Some spaces could not find the default scope. by MAG-mv in https://github.com/magnopus-opensource/connected-spaces-platform/pull/886
```
MAG-ElliotMorris, line 105: [Comment] Reminder you don't have to put the PR link in, no harm having it though.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,10 +306,35 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
                 [this, FinishedGetScopeEvent, SpaceResult, RealtimeEngine](const csp::systems::ScopesResult& ScopesResult)
                 {
                     // 2. Ensure we have the default, auto-generated scope.
-                    auto DefaultScopeIt = std::find_if(ScopesResult.GetScopes().begin(), ScopesResult.GetScopes().end(),
-                        [](const Scope& Scope) { return Scope.Name == Scope.ReferenceId + "_autoGenerated"; });
+                    // This is done by checking for the global scope in the space, and that we only have 1 of these global scopes.
+                    // Right now, clients can't create scopes, so there should be a 0 chance of the above conditions failing.
```
MAG-ElliotMorris: [Nit] Do you mean below conditions?

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,10 +306,35 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
...
+                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Space doesn't have a global scope");
+                        FinishedGetScopeEvent->set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
+                            "SpaceSystem::RegisterScopesInSpace: Space doesn't have a global scope", MakeInvalid<csp::systems::SpaceResult>())));
+                        return;
+                    }
+
+                    if (GlobalScopeCount > 1)
+                    {
+                        this->LogSystem->LogMsg(
+                            csp::common::LogLevel::Error, "SpaceSystem::RegisterScopesInSpace: Space has multiple global scopes");
```
MAG-ElliotMorris: [Consider] These errors could be a tad better. You could state that this version of CSP only supports spaces that have only a single global scope.

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -306,10 +306,35 @@ std::function<async::task<SpaceResult>(const SpaceResult& SpaceResult)> SpaceSys
...
+                        FinishedGetScopeEvent->set_exception(std::make_exception_ptr(csp::common::continuations::ResultException(
+                            "SpaceSystem::RegisterScopesInSpace: Space has multiple global scopes", MakeInvalid<csp::systems::SpaceResult>())));
+                        return;
+                    }
+
+                    auto DefaultScopeIt
+                        = std::find_if(Scopes.begin(), Scopes.end(),
+                        [](const Scope& Scope) { return Scope.PubSubType == PubSubModelType::Global; });
```
MAG-ElliotMorris, line 343: [Consider] Probably redundant, you can just get `Scopes[0]` at this point.

MAG-mv: Not quite. There could be other scopes that arent Global in this array

MAG-ElliotMorris: Yeah, that is the code as-written ... but didn't we want to ensure that for this version of CSP, there is only a single global scope, you mention as much in the comment.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: I'm refferring to the conditions in the line above (which do refer to the conditions below)

.github/CHANGELOG.md
MAG-mv: oopsie

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/886/commits/2acc1e3792e994c56d74028f086226ef3f4efa5a

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/886/commits/2acc1e3792e994c56d74028f086226ef3f4efa5a

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-mv: We may be talking about slightly different things. I am ensuring there is only a single global scope, but there could theoretically be object scopes which wouldnt cause issues with this implementation. It also slightly futre proofs things

MAG-ElliotMorris: > but there could theoretically be object scopes which wouldnt cause issues with this implementation.

Do you have evidence of this? If this is the case then we're fine, but it strikes me that CSP is untested on this axis, so I don't see how we can say this with any confidence.

MAG-mv: After a discussion, we decided to only make a space valid if there is only a single scope, that is Global. Fixed here:

https://github.com/magnopus-opensource/connected-spaces-platform/pull/886/commits/9c1ed7ac2482099adf2968bdce16862ea4af1ba3

PR #942, opened 2026-04-08
Of 1844 add auth token asset requests

MAG-AdamThorn:
This change adds content security to our S3 Assets, by ensuring asset requests are authorized.

An `Authorization` header has been added to our file Web Requests with the authenticated users bearer token.

Tests have been added using a Mocked WebClient to ensure requests include the Authorization header when the user is authenticated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. For compile
 
 - [OF-1625] feat: Add enabled property to `CollisionSpaceComponent`. By @mag-lt.
 
-- [OF-1844] feat: Add `Authorization` header to asset requests. By @mag-lt. By @MAG-AdamThorn
+- [OF-1844] feat: Add `Authorization` header to asset requests. By @MAG-AdamThorn
   This change adds content security to our S3 Assets, by ensuring asset requests are authorized. An `Authorization` header has been added to our file Web Requests with the authenticated users bearer token.
 
 ###💫 💥 Code Refactors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file. For compile
 
 - [OF-1625] feat: Add enabled property to `CollisionSpaceComponent`. By @mag-lt.
 
+- [OF-1844] feat: Add `Authorization` header to asset requests. By @mag-lt. By @MAG-AdamThorn
+  This change adds content security to our S3 Assets, by ensuring asset requests are authorized. An `Authorization` header has been added to our file Web Requests with the authenticated users bearer token.
+
 ###💫 💥 Code Refactors
 
 - [OF-1848] refac: Add internal constructor to `ComponentBase` for constructing from a `Schema` structure. By @mag-lt
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -361,7 +361,7 @@ public:
 private:
     AssetSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
     CSP_NO_EXPORT AssetSystem(csp::web::WebClient* WebClient, csp::multiplayer::NetworkEventBus& EventBus,
-        csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem);
+        const csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem);
     ~AssetSystem();
 
     CSP_ASYNC_RESULT void DeleteAssetCollectionById(const csp::common::String& AssetCollectionId, NullResultCallback Callback);
--- a/Library/include/CSP/Systems/Assets/AssetSystem.h
+++ b/Library/include/CSP/Systems/Assets/AssetSystem.h
@@ -360,7 +360,8 @@ public:
 
 private:
     AssetSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
-    CSP_NO_EXPORT AssetSystem(csp::web::WebClient* WebClient, csp::multiplayer::NetworkEventBus& EventBus, common::LogSystem& LogSystem);
+    CSP_NO_EXPORT AssetSystem(csp::web::WebClient* WebClient, csp::multiplayer::NetworkEventBus& EventBus,
+        csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem);
     ~AssetSystem();
 
     CSP_ASYNC_RESULT void DeleteAssetCollectionById(const csp::common::String& AssetCollectionId, NullResultCallback Callback);
--- a/Library/src/Systems/Assets/AssetSystem.cpp
+++ b/Library/src/Systems/Assets/AssetSystem.cpp
@@ -351,13 +351,13 @@ AssetSystem::AssetSystem()
 }
 
 AssetSystem::AssetSystem(
-    web::WebClient* WebClient, multiplayer::NetworkEventBus& EventBus, csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem)
+    web::WebClient* WebClient, multiplayer::NetworkEventBus& EventBus, const csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem)
     : SystemBase(WebClient, &EventBus, &LogSystem)
 {
     PrototypeAPI = new chs::PrototypeApi(WebClient);
     AssetDetailAPI = new chs::AssetDetailApi(WebClient);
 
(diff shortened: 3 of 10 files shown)


Library/src/Web/RemoteFileManager.cpp
```diff
@@ -18,6 +18,29 @@
...
+#include <fmt/format.h>
+
+namespace
+{
+csp::common::Optional<csp::common::String> ConstructAuthorizationHeader()
+{
+    csp::common::Optional<csp::common::String> BearerToken;
+
+    auto& SystemsManager = csp::systems::SystemsManager::Get();
+    auto& UserSystem = *SystemsManager.GetUserSystem();
```
mag-lt: [Fix] Seems a shame to introduce new singleton usage when we'd ideally be getting rid of them. It looks to me that `RemoteFileManager` new depends on an `IAuthContext`, so we can pass a reference to one on construction. This will need to be passed down via the `AssetSystem` and ultimately get off of the UserSystem in `SystemsManager::CreateSystems`.

The test setup should be simpler following this, as we don't really need to setup all of Foundation as far as I can see, we can just have a stubbed/mocked IAuthContext.

Tests/src/InternalTests/RemoteFileManagerTests.cpp
```diff
@@ -0,0 +1,233 @@
...
+ * Tests that GetFile sends a GET request with the correct URI and Content-Type header.
+ * Confirm that it does not include an Authorization header when the user is not logged in.
+ */
+CSP_INTERNAL_TEST(CSPEngine, RemoteFileManagerTests, GetFileSendsCorrectRequestWhenNotLoggedIn)
+{
+    InitialiseFoundationWithUserAgentInfo(EndpointBaseURI());
+
+    csp::common::LogSystem* LogSystem = csp::systems::SystemsManager::Get().GetLogSystem();
+
+    WebClientMock* MockClient = new WebClientMock(80, ETransferProtocol::HTTP, LogSystem, true);
```
mag-lt: [Nit] No need to heap allocate at all (this could just be on the stack?), but if we're going to do so, prefer `make_unique`? No need to do handle deletion manually that way. (same across all the test cases)

mag-lt (changes requested):
Looks mostly fine, but pains me to see new singleton usage (hence a fix comment).

Tests/src/InternalTests/RemoteFileManagerTests.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    delete MockClient;
+
+    csp::CSPFoundation::Shutdown();
+}
+
+/*
+ * Test that GetResponseHeaders includes the Authorization header with a Bearer token when the user is logged in.
+ */
+CSP_INTERNAL_TEST(CSPEngine, RemoteFileManagerTests, GetResponseHeadersIncludesAuthorizationHeaderWhenLoggedIn)
```
MAG-ElliotMorris: [Comment] Good test coverage, you've got the full 2x2 matrix. 

We could consider looking into data-driven tests to avoid duplication for stuff like this, which I tend to tilt at when I see there's a matrix of conditions. I've not thought about it too deeply, don't think it's really worth going to the effort changing it now that the tests are written.

MAG-ElliotMorris (approved):
LGTM. I do agree that it'd be preferable not to insert a reliance on the singleton if we can avoid it.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file. For compile
 
 - [OF-1625] feat: Add enabled property to `CollisionSpaceComponent`. By @mag-lt.
 
+- [OF-1844] feat: Add `Authorization` header to asset requests. By @mag-lt. By @MAG-AdamThorn
```
mag-lt: [Fix] I'm humbled, but really the credit is all yours.

Library/src/Web/RemoteFileManager.cpp
MAG-AdamThorn: Excellent idea.

I have removed use of the SystemsManager singleton as suggested an am passing IAuthContext to the RemoteFileManager. I have also mocked this for testing: [193a658](https://github.com/magnopus-opensource/connected-spaces-platform/pull/942/changes/193a65802cf3c432a21ce8da34af2076b8e9098e)

Tests/src/InternalTests/RemoteFileManagerTests.cpp
MAG-AdamThorn: Agreed, updated to be stack allocated as part of same change: [193a658](https://github.com/magnopus-opensource/connected-spaces-platform/pull/942/changes/193a65802cf3c432a21ce8da34af2076b8e9098e)

Tests/src/InternalTests/RemoteFileManagerTests.cpp
MAG-AdamThorn: Nice callout, I have updated tests to be data-driven here: [207bb51](https://github.com/magnopus-opensource/connected-spaces-platform/pull/942/changes/207bb51248d5d5f24500d75bd1a94947d1ee9e1e)

CHANGELOG.md
MAG-AdamThorn: Fixed here: [d077d68](https://github.com/magnopus-opensource/connected-spaces-platform/pull/942/changes/d077d68109a45344c78105ea26acc5f7f471e474)

Library/include/CSP/Systems/Assets/AssetSystem.h
```diff
@@ -360,7 +360,8 @@ class CSP_API AssetSystem : public SystemBase
 
 private:
     AssetSystem(); // This constructor is only provided to appease the wrapper generator and should not be used
-    CSP_NO_EXPORT AssetSystem(csp::web::WebClient* WebClient, csp::multiplayer::NetworkEventBus& EventBus, common::LogSystem& LogSystem);
+    CSP_NO_EXPORT AssetSystem(csp::web::WebClient* WebClient, csp::multiplayer::NetworkEventBus& EventBus,
+        csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem);
```
mag-lt: [Nit] Could take the auth context as const ref, as it is only read from in this context

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -349,13 +350,14 @@ AssetSystem::AssetSystem()
 {
 }
 
-AssetSystem::AssetSystem(web::WebClient* WebClient, multiplayer::NetworkEventBus& EventBus, common::LogSystem& LogSystem)
+AssetSystem::AssetSystem(
+    web::WebClient* WebClient, multiplayer::NetworkEventBus& EventBus, csp::common::IAuthContext& InAuthContext, common::LogSystem& LogSystem)
```
mag-lt: [Nit] Could take the auth context as const ref, as it is only read from in this context

Library/src/Web/RemoteFileManager.cpp
```diff
@@ -45,8 +43,9 @@ csp::common::Optional<csp::common::String> ConstructAuthorizationHeader()
 namespace csp::web
 {
 
-RemoteFileManager::RemoteFileManager(csp::web::WebClient* InWebClient)
+RemoteFileManager::RemoteFileManager(csp::web::WebClient* InWebClient, csp::common::IAuthContext* InAuthContext)
```
mag-lt: [Nit] Pass as a reference vs pointer (and const ref at that, as is only read from). WebClient could also be taken by reference of course. I assume this is for some sort of local consistency, but would be nice to have references all the way down.

Library/src/Web/RemoteFileManager.cpp
```diff
@@ -25,16 +26,13 @@
 
 namespace
 {
-csp::common::Optional<csp::common::String> ConstructAuthorizationHeader()
+csp::common::Optional<csp::common::String> ConstructAuthorizationHeader(csp::common::IAuthContext* InAuthContext)
```
mag-lt: [Nit] Take auth context as const ref vs pointer

mag-lt (approved):
Nice, thanks for getting rid of the singleton.

Some nits about using ideally using references vs pointers, and some sprinkling some const-ness.

Also, a point that I probably don't articulate clearly enough in my original review is that `RemoteFileManager` doesn't seem to rely on anything that actually requires Foundation etc, so we don't actually need to be using any of our bespoke test infrastructure stuff, these could just be straightforward GTest tests/fixtures. In any case, nothing blocking.

_Edit_: Ah, looks like `PublicTestBaseInternalWithParam` doesn't actually use any of our internal stuff?

PR #943, opened 2026-04-14
[NT-0] test: Fix flackey ValidExpiryLengthInTokenOptionsTest

MAG-AdamThorn:
The test `ValidExpiryLengthInTokenOptionsTest` was failing when run against live tests on our CI, but was passing when run locally against local or live CHS. I was able to reproduce the issue by putting a breakpoint in the `SetNewLoginTokenReceivedCallback` and waiting for 5+ seconds, which was greater than the specified token expiry length. This then caused log out to fail as the user token had already expired.

For the fix I introduced a future/promise and moved calculation of the delta to outside the callback. I also increased the token expiry length to account for latency. The token expiry is now set to 30 seconds. I do not expect latency to approach this however the duration set has no bearing on the purpose of the test, which is simply to check that the `AccessExpiryTime` is correct within a generous threshold.

In addition to the above I have removed two calls to `NotifyRefreshTokenHasChanged()` from the UserSystem as this is already being done within the `LoginStateResult::OnResponse()`, which was resulting in the `UserSystem NewLoginTokenReceivedCallback` being fired twice.

### DIFF
--- a/Library/src/Systems/Users/UserSystem.cpp
+++ b/Library/src/Systems/Users/UserSystem.cpp
@@ -272,8 +272,6 @@ void UserSystem::Login(const csp::common::String& Email, const csp::common::Stri
         {
             if (LoginStateRes.GetResultCode() == csp::systems::EResultCode::Success)
             {
-                NotifyRefreshTokenHasChanged();
-
                 csp::multiplayer::MultiplayerConnection::ErrorCodeCallbackHandler ConnectionCallback
                     = [Callback, LoginStateRes](csp::multiplayer::ErrorCode ErrCode)
                 {
@@ -613,8 +611,6 @@ void UserSystem::LoginToThirdPartyAuthenticationProvider(const csp::common::Stri
     {
         if (LoginStateRes.GetResultCode() == csp::systems::EResultCode::Success)
         {
-            NotifyRefreshTokenHasChanged();
-
             csp::multiplayer::MultiplayerConnection::ErrorCodeCallbackHandler ConnectionCallback
                 = [Callback, LoginStateRes](csp::multiplayer::ErrorCode ErrCode)
             {
--- a/Tests/src/PublicAPITests/UserSystemTests.cpp
+++ b/Tests/src/PublicAPITests/UserSystemTests.cpp
@@ -530,27 +530,33 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, ValidExpiryLengthInTokenOptionsTest)
     auto* UserSystem = SystemsManager.GetUserSystem();
 
     auto TokenOptions = csp::systems::TokenOptions();
-    TokenOptions.AccessTokenExpiryLength = "00:00:05";
+    TokenOptions.AccessTokenExpiryLength = "00:00:30";
 
-    // Ensure that the token expiry time matched the provided token options
+    std::promise<csp::systems::LoginTokenInfoResult> TokenPromise;
+    std::future<csp::systems::LoginTokenInfoResult> TokenFuture = TokenPromise.get_future();
+
     UserSystem->SetNewLoginTokenReceivedCallback(
-        [](const csp::systems::LoginTokenInfoResult& Result)
+        [&TokenPromise](const csp::systems::LoginTokenInfoResult& Result)
         {
-            EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
-
-            const auto TokenInfo = Result.GetLoginTokenInfo();
-
-            const auto AccessExpiryTime = csp::common::DateTime(TokenInfo.AccessExpiryTime);
-            const auto CurrentTime = csp::common::DateTime::UtcTimeNow();
-
-            // Calculate the delta from the available time points as we do not support duration in DateTime
-            const auto Delta = AccessExpiryTime.GetTimePoint() - CurrentTime.GetTimePoint();
-            EXPECT_LE(Delta, 10s);
+            TokenPromise.set_value(Result);
         });
 
     // Log in
     csp::common::String UserId;
     LogInAsNewTestUser(UserSystem, UserId, true, true, TokenOptions);
+
+    ASSERT_EQ(TokenFuture.wait_for(30s), std::future_status::ready) << "NewLoginTokenReceivedCallback was not invoked";
+
+    const auto Result = TokenFuture.get();
+    EXPECT_EQ(Result.GetResultCode(), csp::systems::EResultCode::Success);
+
+    const auto TokenInfo = Result.GetLoginTokenInfo();
+    const auto AccessExpiryTime = csp::common::DateTime(TokenInfo.AccessExpiryTime);
+    const auto CurrentTime = csp::common::DateTime::UtcTimeNow();
+
+    // Calculate the delta from the available time points as we do not support duration in DateTime
+    const auto Delta = AccessExpiryTime.GetTimePoint() - CurrentTime.GetTimePoint();
+    EXPECT_LE(Delta, 40s);
 
     // Log out
     LogOut(UserSystem);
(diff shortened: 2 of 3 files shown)


Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -272,8 +272,6 @@ void UserSystem::Login(const csp::common::String& Email, const csp::common::Stri
         {
             if (LoginStateRes.GetResultCode() == csp::systems::EResultCode::Success)
             {
-                NotifyRefreshTokenHasChanged();
```
MAG-ElliotMorris, line 275: [Consider] Arguably should have a changelog entry.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -530,28 +530,34 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, ValidExpiryLengthInTokenOptionsTest)
...
-            const auto Delta = AccessExpiryTime.GetTimePoint() - CurrentTime.GetTimePoint();
-            EXPECT_LE(Delta, 10s);
+            TokenPromise.set_value(Result);
         });
 
     // Log in
     csp::common::String UserId;
     LogInAsNewTestUser(UserSystem, UserId, true, true, TokenOptions);
 
+    ASSERT_EQ(TokenFuture.wait_for(30s), std::future_status::ready) << "NewLoginTokenReceivedCallback was not invoked";
```
MAG-ElliotMorris, line 548: [Consider] Making this timeout longer than the expiry length, or making the expiry length shorter.

My understanding of this is that we expect a callback once the token time (30s) has expired. By timing out 30s after, we're cutting it rather close, but honestly I might have an incorrect interpretation.

mag-lt (approved):
The test code changes seem fine. I'm not convinced that we should be writing tests like this, which as far as I can see validate MCS behaviour. I'd much rather just write tests that validate we are producing the requests we intend to (i.e. sending the intended expiry), and that we can react to expected responses, vs actually writing a test for MCS specific code. However, I know we don't have consensus on this within the team. My two cents in any case.

I can see that changing to use future/promise style for awaiting the result has revealed the double notification (as setting a promise more than once throws?). This is the part of the PR that actually requires review in my opinion, and so looked it through a bit. I'm not clear on whether it is better for an external system to be the single place notifying via a singleton (i.e. [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/058fc3565e8f1141f3935efd0fdff484e47f5c19/Library/src/Systems/Users/Authentication.cpp#L151) vs the individual callbacks) I understand the tendency to prefer a centralised place, but I'm not 100% sure this should be that place. In any case, I can reason that it works as intended, so approving.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -530,28 +530,34 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, ValidExpiryLengthInTokenOptionsTest)
     auto* UserSystem = SystemsManager.GetUserSystem();
 
     auto TokenOptions = csp::systems::TokenOptions();
-    TokenOptions.AccessTokenExpiryLength = "00:00:05";
+    TokenOptions.AccessTokenExpiryLength = "00:00:30";
```
MAG-ElliotMorris, line 533: [Question] A little confused about specifically where latency was causing an issue. Was it failing before because the `EXPECT_LE` was overtime? In that case, does it make sense to increase this at all, don't we just want to increase the window we give it in our comparison?

MAG-ElliotMorris (commented):
Agree with your reasoning that specific timing has no bearing on the purpose of the test, so don't mind this change.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: The callback will fire immediately. The specified 30 second duration will just be used by MCS to calculate the `AccessExpiryTime`. We then check the delta between that and the current time.

The 30s in `TokenFuture.wait_for(30s)` is arbitrary, I am just specifying a time to wait beyond which we can assume things have failed.

mag-lt: I now have doubts that I understood the test code properly when reviewing.

[Question] What is this test doing (I have already deduced that it is testing some MCS behaviour vs any logic in our code. Correct me if I'm wrong on that). My understanding is that it asks for a token that will expire with the given duration in the request? So we're checking the server did that? Relative to the time we send the request at? I'll note that we are taking `CurrentTime` when receiving the response. Could that be why?

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: The test was failing specifically because the call to logout was failing. The call to logout was failing because the token had already expired. The token had expired as it took longer than 5 seconds after logging in to then call logout. I could reproduce this issue by sticking a breakpoint in the callback.

MAG-ElliotMorris: Right, gotcha. Key insight being that the callback fires immediately, not after the stated time. We could arguably just set that time to be hours then and it'd be the same test?

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ElliotMorris: I think you're right @mag-lt that this is almost exclusively testing CHS behaviour. We're registering a callback, checking that CHS fires one in response to login, and that they've populated a time value relative to an expiry that we've passed in.

Arguably, this tests that we send the `AccessTokenExpiryLength` correctly in the login request. But as you say, we _could_ test that more directly rather than waiting on the response.

MAG-ElliotMorris (approved):

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Update: [aa43825](https://github.com/magnopus-opensource/connected-spaces-platform/pull/943/changes/aa4382595280222c90ce9673faace1a3a84d444c)

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: Yes what you have both stated above aligns with my understanding of the test. I also agree that this test is validating CHS logic as much as our own. I am not sure we even need to check the delta, just that the AccessTokenExpiryLength has been set to something.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: Yes agreed. I just chose to increase it to something that is unlikely to be impacted by network latency. As discussed above I question the value of this test currently or at least how it validates the result of MCS logic.

MAG-AdamThorn:
> The test code changes seem fine. I'm not convinced that we should be writing tests like this, which as far as I can see validate MCS behaviour. I'd much rather just write tests that validate we are producing the requests we intend to (i.e. sending the intended expiry), and that we can react to expected responses, vs actually writing a test for MCS specific code. However, I know we don't have consensus on this within the team. My two cents in any case.
> 
> I can see that changing to use future/promise style for awaiting the result has revealed the double notification (as setting a promise more than once throws?). This is the part of the PR that actually requires review in my opinion, and so looked it through a bit. I'm not clear on whether it is better for an external system to be the single place notifying via a singleton (i.e. [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/058fc3565e8f1141f3935efd0fdff484e47f5c19/Library/src/Systems/Users/Authentication.cpp#L151) vs the individual callbacks) I understand the tendency to prefer a centralised place, but I'm not 100% sure this should be that place. In any case, I can reason that it works as intended, so approving.

I agree with you. I would like to evaluate these UserSystem tests more thoroughly though time is against us in this instance as we need to get a release out, so I elected to leave the test logic as is.

PR #945, opened 2026-04-15
[OB-5310] feat: Add audio controls to video player space component

MAG-ME:
Added new properties to the video space component

- Created a new interface `IAudioControlComponent` with common audio properties for audio and video player components
- Used the interface in the audio component
- Added Volume and AudioType, replicating the values from the audio component
- Updated the script binding to include the new properties
- Updated the Video Player tests to exercise the new properties in C++ and via a script

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. For compiled binaries, deployment packages, and version-specific artifacts, please visit our [GitHub Releases](https://github.com/magnopus-opensource/connected-spaces-platform/releases).
 
+## [6.35.0]
+
+### 🍰 🙌 New Features
+
+- [OB-5310] feat: Added Volume and AudioType properties to `VideoPlayerSpaceComponent`. By @MAG-ME.
+
 ## [6.33.0]
 
 ### 🙈 🙉 🙊 Test Changes
--- a/Library/include/CSP/Multiplayer/Components/Interfaces/IAudioControlComponent.h
+++ b/Library/include/CSP/Multiplayer/Components/Interfaces/IAudioControlComponent.h
@@ -32,7 +32,7 @@ enum class AudioType
 };
 
 /// @brief Controls the audio properties of a component.
-CSP_INTERFACE class CSP_API IAudioSourceComponent
+CSP_INTERFACE class CSP_API IAudioControlComponent
 {
 public:
     /// @brief Gets the type of the audio of this audio component.
@@ -70,7 +70,7 @@ public:
     virtual void SetVolume(float Value) = 0;
 
 protected:
-    virtual ~IAudioSourceComponent() = default;
+    virtual ~IAudioControlComponent() = default;
 };
 
 } // namespace csp::multiplayer
--- a/Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
+++ b/Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
@@ -26,6 +26,7 @@
 namespace
 {
 constexpr const float DefaultAttenuationRadius = 10.f; // Distance in meters
+constexpr const float DefaultVolume = 1.f;
 }
 
 namespace csp::multiplayer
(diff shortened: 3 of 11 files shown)


Library/include/CSP/Multiplayer/Components/Interfaces/IAudioControlComponent.h
```diff
@@ -0,0 +1,76 @@
...
+    /// @param Value Type of the audio of this audio component.
+    virtual void SetAudioType(AudioType Value) = 0;
+
+    /// @brief Gets the attenuation for the audio when a spatial audio type.
+    ///        The radius is the minimum distance between the origin of this audio component and
+    ///        the position of the player, from within which the player can start hearing
+    ///        the spatial audio in range.
+    ///        The radius is expressed in meters.
+    /// @return The minimum radius in meters from the origin of the audio component to hear the spatial audio.
+    virtual float GetAttenuationRadius() const = 0;
```
MAG-ElliotMorris, line 52: [Comment] Decent docs in this file, 👍

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
```diff
@@ -318,6 +320,45 @@ void VideoPlayerSpaceComponent::SetIsVirtualVisible(bool Value)
...
+{
+    if (Value >= 0.f && Value <= 1.f)
+    {
+        SetProperty(static_cast<uint32_t>(VideoPlayerPropertyKeys::Volume), Value);
+    }
+    else
+    {
+        if (LogSystem != nullptr)
+        {
+            LogSystem->LogMsg(csp::common::LogLevel::Error, fmt::format("Invalid value for volume ({:.2f}). Must be from 0.0 to 1.0", Value).c_str());
```
MAG-ElliotMorris, line 357: [Consider] Weakening this to `Warning`. `Errors` are generally harder problems than this that indicate a programming error or configuration error. Arguably that's true here, although I imagine folks might just pipe user input into this.

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
```diff
@@ -318,6 +320,45 @@ void VideoPlayerSpaceComponent::SetIsVirtualVisible(bool Value)
...
+void VideoPlayerSpaceComponent::SetAudioType(AudioType Value)
+{
+    SetProperty(static_cast<uint32_t>(VideoPlayerPropertyKeys::AudioType), static_cast<int64_t>(Value));
+}
+
+float VideoPlayerSpaceComponent::GetVolume() const { return GetFloatProperty(static_cast<uint32_t>(VideoPlayerPropertyKeys::Volume)); }
+
+void VideoPlayerSpaceComponent::SetVolume(float Value)
+{
+    if (Value >= 0.f && Value <= 1.f)
```
MAG-ElliotMorris, line 349: [Consider] I wonder if it's better UX just to set it to 1.0 if you pass in anything above 1, and 0.0f if you pass in anything below 0, getting rid of the need to error so strongly. Then again, I can also imagine that being confusing in it's own way. 🤔 I don't really mind this approach either.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. For compiled binaries, deployment packages, and version-specific artifacts, please visit our [GitHub Releases](https://github.com/magnopus-opensource/connected-spaces-platform/releases).
 
+## [6.35.0]
+
+### 🍰 🙌 New Features
+
+- [OB-5310] feat: Added Volume and AudioType properties to `VideoPlayerSpaceComponent`. By @MAG-ME.
```
MAG-ElliotMorris, line 9: [comment] I'm in two minds as to whether the move of `AudioType` to the interface technically makes this a breaking change, as you'd need different includes.

... thinking on it, not really, you'd always have the transient include anyway. This is fine.

MAG-ElliotMorris (approved):
Seems reasonable to me! Couple of considers, but i'm in two minds on them myself so feel free to make whatever decision you want, no need to request re-review unless you decide to change anything significant.

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
MAG-ME: This was a direct copy from the audio component. If we wanted to make this change we should do it there too.

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
MAG-ME: This was a direct copy from the audio component. If we wanted to make this change we should do it there too.

MAG-ElliotMorris: Ah i see, completely fair, just resolve these two convos 👍

MAG-mv (approved):
Great PR. Approved!

MAG-ElliotMorris:
Rebasing so we can merge.

PR #946, opened 2026-04-17
[OF-1849] feat: Bind components created w/ Schemas into scripts automatically

mag-lt:
Components defined in terms of a schema description will have their properties bound into the scripting system automatically where possible.

This moves a large portion of CSP's own components over to this pattern, with only a handful still being registered manually due to some remaining edge cases.

Historically, we bind into JS via quickjspp, a community library (not by the original author of quickjs) that adds pybind style c++ interface over the underlying C API, building up enough template infrastructure to ultimately map C++ classes and their methods onto JS objects. This pattern doesn't fit with what we're doing here, and so we have to drop to the underlying C API. Use of the C API is kept to a minimum, as some of quickjspp's supporting infrastructure for converting std library types is still useful.


Library/src/Multiplayer/Component/Property.h
```diff
@@ -15,24 +15,23 @@
...
-namespace csp::multiplayer::component
+namespace csp::multiplayer
 {
 
 /// @brief Represents an individual data field, or "property", within a component schema,
 /// consisting of a stable ID/key, a type kind and default value (via ReplicatedValue),
 /// and other metadata.
-///
-/// @tparam KeyType: The type of the stable ID value used for this property's unique key.
-template <typename KeyType> struct Property final
```
MAG-mv, line 30: [Question] Its a shame we have to remove compile time validation but theres no way around it for now. Do you see us changing it back when the new wrapper gen work is done? Not sure if it would be possible still

Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -208,7 +209,7 @@ class CSP_API ComponentBase
     csp::common::Map<uint32_t, csp::common::ReplicatedValue> Properties;
     csp::common::Map<uint32_t, csp::common::ReplicatedValue> DirtyProperties;
 
-    ComponentScriptInterface* ScriptInterface;
+    std::unique_ptr<ComponentScriptInterface> ScriptInterface;
```
MAG-mv, line 213: [Comment] You may need to CSP_START_IGNORE this. Otherwise you may get wrapper gen issues

Edit: looks like everything passed

Library/include/CSP/Multiplayer/ComponentSchema.h
```diff
@@ -38,6 +39,12 @@ class CSP_API ComponentSchema
     /// serialized and used in messages sent over the multiplayer connection.
     TypeIdType TypeId;
 
+    /// @brief A human-readable name describing this component (in `PascalCase`).
+    /// Must be unique across components registered with the engine. This name will be used for
```
MAG-mv, line 43: [Question] Is PascalCase a hard requirement? If so, should we create a setter function to either validate the name, or trim whitespace?

Library/include/CSP/Multiplayer/ComponentProperty.h
```diff
@@ -36,6 +37,12 @@ class CSP_API ComponentProperty
     /// multiplayer connection.
     KeyType Key;
 
+    /// @brief A human-readable name describing this propery (in `camelCase`).
+    /// Must be unique within the component (two properties should not have the same name).
+    /// This name will be used for generating script bindings i.e. a property with this name will
+    /// be exposed on the component in scripts.
+    csp::common::String Name;
```
MAG-mv, line 44: [Question] Is camelCase a hard requirement? If so, should we create a setter function to either validate the name, or trim whitespace?

Library/src/Multiplayer/Script/ComponentScriptInterface.cpp
```diff
@@ -90,4 +87,120 @@ void ComponentScriptInterface::SendPropertyUpdate()
...
+void ComponentScriptInterface::SetProperty(uint16_t Key, Value DesiredValue)
+{
+    if (!Component)
+    {
+        return;
+    }
+
+    struct Visitor final
+    {
+        std::optional<csp::common::ReplicatedValue> operator()(bool Value) const
```
MAG-mv, line 158: [Comment] Very similar conversion pattern to mcs types. I wonder if we can make  types that require less custom serialization when we add the new wrapper generators

Library/src/Multiplayer/Script/ComponentScriptInterface.cpp
```diff
@@ -90,4 +87,120 @@ void ComponentScriptInterface::SendPropertyUpdate()
...
+        {
+            return Value;
+        }
+
+        std::optional<csp::common::ReplicatedValue> operator()(const std::string& Value) const
+        {
+            return csp::common::String(Value.c_str());
+        }
+
+        std::optional<csp::common::ReplicatedValue> operator()(const std::vector<float>& Value) const
```
MAG-mv, line 178: [Consider] These functions already exist in MCSComponentPacker. We could potentially move these out and create a more generic ReplicatedValue conversion util

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
```diff
@@ -50,6 +50,8 @@ enum class AIChatbotVisualState
 class CSP_API AIChatbotSpaceComponent : public ComponentBase, public IPositionComponent
 {
 public:
+    CSP_NO_EXPORT static const ComponentSchema& GetSchema();
```
MAG-AdamThorn, line 53: [question] Why is this CSP_NO_EXPORT?

MAG-mv (commented):
Great PR. Will approve once questions are answered

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
```diff
@@ -0,0 +1,90 @@
...
+        ToPair(GaussianSplatSpaceComponent::GetSchema()),
+        ToPair(TextSpaceComponent::GetSchema()),
+        ToPair(HotspotSpaceComponent::GetSchema()),
+        ToPair(ScreenSharingSpaceComponent::GetSchema()),
+        ToPair(AIChatbotSpaceComponent::GetSchema()), 
+    };
+
+    for (const auto& Schema : AdditionalComponents)
+    {
+        Registry[Schema.TypeId] = Schema;
```
MAG-AdamThorn, line 84: [question] There is nothing to stop a user from specifying an Id already used by one of our existing/legacy components right? If so is that validation going to be added as part of a future ticket?

Library/src/Multiplayer/Components/AnimatedModelSpaceComponent.cpp
```diff
@@ -26,69 +26,86 @@ namespace csp::multiplayer
...
             true,
         },
         {
             static_cast<ComponentProperty::KeyType>(AnimatedModelPropertyKeys::ThirdPartyComponentRef),
+            {}, // not exposed to scripting
             "",
         },
         {
             static_cast<ComponentProperty::KeyType>(AnimatedModelPropertyKeys::IsShadowCaster),
+            {}, // not exposed to scripting
```
MAG-AdamThorn: [fix] I initially thought that we should either add these missing bindings or create a ticket to ensure we do not forget, but now I think we should just do it now.

I think we should not expose those that are deprecated properties, but everything else should likely be done. Happy to discuss if you think there are any that should not be.

Library/include/CSP/Systems/Script/ScriptSystem.h
```diff
@@ -44,6 +46,8 @@ class CSP_API ScriptSystem : public csp::common::IJSScriptRunner
     void Initialise();
     /// @brief Shuts down and deletes the JavaScript runtime context.
     void Shutdown();
+    
+    CSP_NO_EXPORT static std::shared_ptr<ScriptSystem> MakeInitialised();
```
MAG-AdamThorn, line 63: [fix] as this is public we should add some documentation, perhaps calling out its use in testing.

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
MAG-AdamThorn, line 486: [question] I appreciate this isn't part of your change, but I noticed that SCRIPT_NAMESPACE = "OKO". Is that what we want I wonder... ?

MAG-AdamThorn (changes requested):

Library/src/Multiplayer/Components/AnimatedModelSpaceComponent.cpp
mag-lt: So everything I've marked across all files with `// not exposed to scripting`, just add the names now to expose them (unless obviously deprecated)? And same for components that have not been exposed at all (`Collision` and `Reflection`)? I can do that. I deliberately tried to capture the existing behaviour, but in doing so it does highlight that most of these were probably accidental omissions.

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
mag-lt: Not sure to be honest, but I don't see this as conceptually related to these changes. This seems to be about how the scripts can import utilities we expose i.e. `import * as CSP from "CSP"` or `import * as OKO from "OKO"`. From a cursory read it looks like we have functionality for changing this and supporting "aliases" as to not break existing scripts, but I don't want to get involved with this here.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: The same reason as anything else we mark as `CSP_NO_EXPORT` i.e. to conveniently be able to add methods we need ourselves but don't consider part of the exposed public API.

For these legacy hardcoded components, I want to keep the schema definition inside the CPP, but I need a way to get it to add to the central registry. This was the most pragmatic way I could come up with for doing that (I don't consider this part of the `ComponentBase` interface, it is specific to these hardcoded components.). There is no harm in others getting at this, but that isn't reason to expose it.

MAG-AdamThorn: My question was more why weren't we exposing it, but your response makes sense - no reason to currently.

Library/src/Multiplayer/Components/AnimatedModelSpaceComponent.cpp
MAG-AdamThorn: Yeah exactly that, though I think it would also be worth exposing Collision and Reflection as they were likely missed in error.

Library/src/Multiplayer/Component/Property.h
MAG-ElliotMorris: Not read the PR fully so don't know if this is a more specific concern, but removing compile time validation is generally something we're going to have to accept to enable runtime registration. It's a tradeoff. It can be layered back in atop a runtime system via generation as you've said, but tbh I feel people jump too quickly to thinking that the compile time validation matters in those cases, vs an on-initialize check. 🤷

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
MAG-ElliotMorris: To be an annoying prodder. I don't see a reason why we wouldn't expose this eventually. Could serve as good examples to people trying to write their own schema, or sources for people to edit schemas to create similar components at runtime. 🤷

mag-lt: The public API needs some thought, I'm going to keep this is as `CSP_NO_EXPORT` for now. Remember that we are talking about deleting these specific classes entirely at some point. All schemas will be available from the registry in any case.

Library/src/Multiplayer/Component/Property.h
mag-lt: My original intent with this compile time check was two fold:
1. Express in the source text that the Key/Id is ultimately serialised, and there are associated constraints 
2. Especially as I templated this, statically enforce those constraints

Why did I template this in the first place? Mainly because of the `ComponentType` enum, which I envisage we have to do away with at some point, so my C++ brain leaned on templates as a way to switch that out at a later date when we flatten this to a `uint64_t` (`ComponentType` represents a static/closed/fix set of components, and we are switching to a dynamic/open set).

Now that I've had to move this over to an API type, I've been somewhat more pragmatic about the situation. Component type Ids are ultimately `uint64_t`s and property keys are `uint16_t`s, so just get that out in the open and hardcode/flatten them to that now (and as they're no longer templated, there is no encouragement to change them, so we don't need to check).

I've mentioning in the comments that these are ultimately serialised, which is a very soft way of achieving the first point still, and the second point is now no longer relevant.

I still think there is value (especially as I introduced the static checking in the first place) in trying to express those constraints in a way that is checked at compile time, in case a future developer doesn't realise those constraints exist, but ultimately decided it wasn't really worth it. I mainly arrived at that conclusion when I was about to change one of the lines to `static_assert(mcs::IsPackableIdV<uint64_t, uint64_t>)`, which looked a bit tautological. I could still `static_assert(mcs::IsPackableIdV<ComponentProperty::KeyType, mcs::PropertyKeyType>)`, and add an equivalent typedef in the `mcs` namespace for the component type, so that it doesn't seem so tautological, and it would satisfy my original intent, but at the moment I'm still leaning towards keeping this "simple".

MAG-mv: All good. Happy to keep things the way they are

Library/include/CSP/Systems/Script/ScriptSystem.h
mag-lt: Added a clarifying comment in abef9465f3dbda1523f179b47f45608a28dcd597

Library/src/Multiplayer/Components/AnimatedModelSpaceComponent.cpp
mag-lt: I've done the low hanging fruit properties in 8b9f03fa0df8296d567260dd85bc1f471a998923, quite a few didn't actually make sense still which I've documented in the commit, and a few are still prevented by some edge cases which I'm still looking into but will likely end up punting for now (i.e. not addressing in this PR).

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
mag-lt: Good point, I could change the order so that the our existing components are applied after such that they will overwrite. I'll think about this again in the context of validation steps in the current ticket I'm working on.

Library/include/CSP/Multiplayer/ComponentSchema.h
mag-lt: It is mainly to satisfy our existing convention.

A layer of validation for this makes some sense, but I'm not entirely sure what part of the system should do this at the moment. At the moment this schema definition is just data, and as such is quite easily constructed using normal aggregate initialisation (which reads quite nicely in the examples in the tests). I could push some of this into a wrapper type specifically for the names, but the convolutes the bindings a bit. I could also do the validation whilst setting up the scripting specifically, or during registration in the registry, or perhaps as part of actual schema validation once we have a JSON format for this.

I need to think about this a bit more, and am leaning on punting it to more general validation in the context of the ticket I'm working on at the moment. For now I'm personally okay with this being rather simple, and relying on the programmer following the convention (this is no worse than it already was in the traditional by hands bindings).

Library/include/CSP/Multiplayer/ComponentProperty.h
mag-lt: By convention, but we've not prevented this historically which leads to at least one edge case [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/7147ecaf583315533526727efc8d53b0b1edaa6a/Library/src/Multiplayer/Script/EntityScriptBinding.cpp#L308). We can improve the validation here or potentially enforce the convention for scripting (by programatically lower/upper casing the first character as needed, but that isn't necessarily as straight-forward as it sounds if you consider multi-byte unicode / anything non-ASCII).

I made some other comments about this [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/946#discussion_r3173847725) on the other PR comment, I think I'll punt this for now if that is okay for you and think about this as part of other validation.

MAG-AdamThorn (approved):

Library/include/CSP/Multiplayer/ComponentSchema.h
MAG-mv: Yep, im happy with that

MAG-mv (approved):

Library/src/Multiplayer/Script/ComponentScriptInterface.cpp
mag-lt: I'd missed that, thanks. I'm inclined to leave this as-is for now, mainly due to `mcs::ItemComponentData` covering more cases than are currently supported here. I'll accept the mild duplication for now over spending time teasing apart the bits that make sense in a way that can be used in both.

PR #947, opened 2026-04-17
[OF-1846] Add Emscripten support for new build system

MAG-mv:
Added emscripten support for the new build system. The work needed for this was:
- Turn off POCO for emscripten builds
- Update tinyspline to use cinterface (this is explained in the commit message)
- Added emscripten profile with compile flags
- Updates to docs


MAG-mv:
Skipping changelog as this isn't a public feature

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
```diff
@@ -54,10 +69,51 @@ csp::common::Vector3 SplineSpaceComponent::GetLocationAlongSpline(float Normalis
             Internalpoints.push_back(static_cast<double>(ListPoints[i].Z));
         }
 
-        tinyspline::BSpline spline = tinyspline::BSpline::interpolateCubicNatural(Internalpoints, 3);
-        const std::vector<tinyspline::real> SplineResult = spline.eval(NormalisedDistance).result();
+        std::vector<tsReal> internalPoints;
```
MAG-ElliotMorris: [fix] I don't really care too much about changing the interface of this one piece of code, even if it is more verbose. However I'd like a changelog entry because this is a change that could theoretically regress something we don't understand.

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
```diff
@@ -54,10 +69,51 @@ csp::common::Vector3 SplineSpaceComponent::GetLocationAlongSpline(float Normalis
...
+        tsStatus Status {};
+        TsSpline Spline;
+
+        tsError Err = ts_bspline_interpolate_cubic_natural(internalPoints.data(), internalPoints.size() / Dimension, Dimension, &Spline.s, &Status);
+
+        if (Err != TS_SUCCESS)
+        {
+            if (LogSystem != nullptr)
+            {
+                LogSystem->LogMsg(csp::common::LogLevel::Error, (std::string{ "SplineSpaceComponent::GetLocationAlongSpline spline error: "} + Status.message).c_str());
```
MAG-ElliotMorris: [nit] `fmt` is probably a bit more performant that constructing strings and using the + operator, but 🤷

Library/docs/source/manual/building/build-system.md
```diff
@@ -49,6 +49,8 @@ conan install . -s build_type=<CONFIG> --build=missing --profile:host=profiles/h
 
 ### Notes
 
+* When switching between different profiles, delete the generated `ConanPresets.json` file in the repository root.
```
MAG-ElliotMorris, line 52: [Consider] Annoying. I wonder if it might be possible to configure conan to generate the `ConanPresets.json` things outside of the default location, like at `build/<profile>/` or something. You'd have to explicitly reference them in the hand-written presets file because it won't find them automatically anymore, but it could deal with this problem as they would be isolated.

CMakeLists.txt
```diff
@@ -212,6 +212,18 @@ target_compile_options(csp-lib
 	    # Don't error on ignored qualifiers, such as returning const objects. These should be simple to remove at some point.
 	    -Wno-ignored-qualifiers
     >
+
+    $<$<PLATFORM_ID:Emscripten>:-pthread;-fwasm-exceptions>
+)
+
+
+
+target_link_options(csp-lib PRIVATE
```
MAG-ElliotMorris, line 221: [Comment] Glad to see this more minimal.

ThirdParty/signalrclient/CMakeLists.txt
```diff
@@ -40,11 +40,25 @@ target_include_directories(csp-signalr
...
-target_compile_definitions(csp-signalr 
-    # Allows us to link statically
-    PUBLIC NO_SIGNALRCLIENT_EXPORTS
-    # We use message pack for signalr serialization
-    PUBLIC USE_MSGPACK
+target_link_options(csp-signalr PUBLIC
+  $<$<PLATFORM_ID:Emscripten>:
+    -pthread
+    -fwasm-exceptions
+    -lwebsocket.js
```
MAG-ElliotMorris, line 77: [Comment] I don't have a suggestion, but it seems odd to me that we have to replicate these options rather than having them be propagated via a target. But it's only one or two times so I'm not overly fussed.

IntegrationTests/Desktop/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.15)
```
MAG-ElliotMorris, line 1: [Question] Do we plan to keep this directory around long term? Feels like it might become redundant once we're using the build system in anger.

Not that i'm against having it during the transition, makes sense.

MAG-ElliotMorris (changes requested):
Didn't pull and try to build. Going to be trying this very soon anyhow so should be able to work out kinks then.

Library/docs/source/manual/building/build-system.md
MAG-mv: Yeah, this can be done. The problem with doing this is that we need to hardcode the build folder in our presets due to us inheriting the conan presets

ThirdParty/signalrclient/CMakeLists.txt
MAG-mv: I originally only had this in csp-lib, but was getting errors with certain objects not havign pthread enabled

IntegrationTests/Desktop/CMakeLists.txt
MAG-mv: I think having a basic integraiton tests is useful to make sure we don't reak our install step

Library/docs/source/manual/building/build-system.md
MAG-ElliotMorris: 🤔 Bugger, yeah. So it's a choice between insisting on this deletion, or insisting people use `build` if they want to use the presets? I'd favour the latter, but I can see it is a tradeoff decision, so yours to make.

IntegrationTests/Desktop/CMakeLists.txt
MAG-ElliotMorris: Maybe [consider] adding a Readme.md to this directory then to indicate the intent.

CMakeLists.txt
MAG-mv: Yep. Most of the flags are only needed in the final executable

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/947/commits/c3bd8f4b5c7b0d815e2cf1ed4c91f62aa7b257d9

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
```diff
@@ -45,7 +60,7 @@ csp::common::Vector3 SplineSpaceComponent::GetLocationAlongSpline(float Normalis
     csp::common::List<csp::common::Vector3> ListPoints = GetWaypoints();
     if (ListPoints.Size() != 0)
     {
-        std::vector<tinyspline::real> Internalpoints;
+        std::vector<tsReal> Internalpoints;
```
MAG-AdamThorn: [fix] I pinged you about this already but I don't see it resolved in a later commit yet so will capture here so not forgotten. You have initialised two vectors that are the same thing; internalPoints and InternalPoints.

MAG-AdamThorn (changes requested):

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
MAG-mv: Fixed up here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/947/commits/118e056e93d53d878ce9f3145b80f11928c58943

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/947/commits/118e056e93d53d878ce9f3145b80f11928c58943

Library/docs/source/manual/building/build-system.md
MAG-mv: I did test this and it still has the same issue if you don't delete it, because it accumulates as you call install with different profiles.

It does make it simpler to say "just delete the build folder when running different profiles". But seeing as it doesn't solve the problem, I think keeping it as-is is ther best solution for now.

IntegrationTests/Desktop/CMakeLists.txt
MAG-mv: Added install docs + readme:

https://github.com/magnopus-opensource/connected-spaces-platform/pull/947/commits/e874e2cd5455570e2272598477d40dd828a961b8

https://github.com/magnopus-opensource/connected-spaces-platform/pull/947/commits/0f7298f64e943cc89bd7b3fe63b0c65309bf2fc6

Library/src/Multiplayer/Components/SplineSpaceComponent.cpp
MAG-ElliotMorris: You could probably have just mentioned this tinyspline change directly as a `refac`, but 🤷 , it's enough.

Library/docs/source/manual/building/build-system.md
MAG-ElliotMorris: Alright. I'm trusting this dosen't imply a need to `conan install` when switching between profiles, and it is just a file deletion. I'll probably still want to revisit this to try and figure out something nicer.

MAG-ElliotMorris (approved):

MAG-AdamThorn (approved):

PR #948, opened 2026-04-21
[OB-4980] fix!: Address issues with third-party authorization flow

MAG-AdamThorn:
The third-party authorization flow is divided into three parts, with two of those handled via calls to CSP. This change addresses an issue whereby a call to the first CSP method would log you out or leave you in an invalid login state if you did not complete the authorization flow.

In addition, the backend services are now constructing the authorization URL that users require directly, rather than CSP constructing it from data returned. This will allow the services to account for changes required to the URL depending on client type and provider.

This change has been validated with the Unity team via a feature publish.

**This change introduces the following breaking changes:**
  1. Spelling of method `UserSystem::GetThirdPartyProviderAuthoriseURL()` has been Americanised to `UserSystem::GetThirdPartyProviderAuthorizeURL()`.
  2. Method `UserSystem::GetThirdPartyProviderAuthorizeURL()` now takes an additional optional ClientType argument of type `EThirdPartyPlatform`.
  3. The values of enum `EThirdPartyPlatform` are now capitalized rather than upper case and the value `Web` has been added.

### DIFF
--- a/Library/include/CSP/Common/SharedEnums.h
+++ b/Library/include/CSP/Common/SharedEnums.h
@@ -29,7 +29,6 @@ namespace csp::common
 /// @brief Enum which represents possible login states of a CSP client.
 enum class ELoginState : uint8_t
 {
-    LoginThirdPartyProviderDetailsRequested,
     LoginRequested,
     LoggedIn,
     LogoutRequested,
--- a/Library/include/CSP/Common/SharedEnums.h
+++ b/Library/include/CSP/Common/SharedEnums.h
@@ -230,41 +230,6 @@ enum class EThirdPartyPlatform
     Web
 };
 
-inline std::string ThirdPartyPlatformToString(csp::systems::EThirdPartyPlatform Platform)
-{
-    std::string PlatformString;
-    switch (Platform)
-    {
-    case csp::systems::EThirdPartyPlatform::None:
-    {
-        PlatformString = "None";
-        break;
-    }
-    case csp::systems::EThirdPartyPlatform::Unreal:
-    {
-        PlatformString = "Unreal";
-        break;
-    }
-    case csp::systems::EThirdPartyPlatform::Unity:
-    {
-        PlatformString = "Unity";
-        break;
-    }
-    case csp::systems::EThirdPartyPlatform::Web:
-    {
-        PlatformString = "Web";
-        break;
-    }
-    default:
-    {
-        PlatformString = std::string("Unknown platform. Value") + std::to_string(static_cast<unsigned int>(Platform));
-        break;
-    }
-    }
-
-    return PlatformString;
-}
-
 /// @brief Code to indicate the result of a request.
 /// Request results should be checked for a success by clients before using any other accessors.
 enum class EResultCode : uint8_t
--- a/Library/include/CSP/Common/SharedEnums.h
+++ b/Library/include/CSP/Common/SharedEnums.h
@@ -221,14 +221,49 @@ enum class StereoVideoType
 namespace csp::systems
 {
 
-/// @brief Indicates special handling for any thirdparty platform
-/// @note We may remove this soon, as it's deceptive implying these are the only platforms we support.
+/// @brief Indicates the thirdparty platform
 enum class EThirdPartyPlatform
 {
-    NONE,
-    UNREAL,
-    UNITY
+    None,
+    Unreal,
+    Unity,
+    Web
 };
+
+inline std::string ThirdPartyPlatformToString(csp::systems::EThirdPartyPlatform Platform)
+{
+    std::string PlatformString;
+    switch (Platform)
+    {
+    case csp::systems::EThirdPartyPlatform::None:
+    {
+        PlatformString = "None";
+        break;
+    }
+    case csp::systems::EThirdPartyPlatform::Unreal:
+    {
+        PlatformString = "Unreal";
+        break;
+    }
+    case csp::systems::EThirdPartyPlatform::Unity:
+    {
+        PlatformString = "Unity";
+        break;
+    }
+    case csp::systems::EThirdPartyPlatform::Web:
+    {
+        PlatformString = "Web";
+        break;
+    }
+    default:
+    {
+        PlatformString = std::string("Unknown platform. Value") + std::to_string(static_cast<unsigned int>(Platform));
+        break;
+    }
+    }
+
+    return PlatformString;
+}
 
 /// @brief Code to indicate the result of a request.
 /// Request results should be checked for a success by clients before using any other accessors.
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -163,8 +163,8 @@ public:
     /// @param ClientType csp::common::Optional<EThirdPartyPlatform> : An optional parameter that allows the client to specify their platform for the
     /// third party authentication flow. This is used for some providers to determine the format of the Authorize URL.
     /// @param Callback StringResultCallback : callback that contains the Authorize URL that the Client should be navigating to for step 2.
-    CSP_ASYNC_RESULT void GetThirdPartyProviderAuthorizeURL(
-        EThirdPartyAuthenticationProviders AuthProvider, const csp::common::String& RedirectURL, StringResultCallback Callback);
+    CSP_ASYNC_RESULT void GetThirdPartyProviderAuthorizeURL(EThirdPartyAuthenticationProviders AuthProvider, const csp::common::String& RedirectURL,
+        const csp::common::Optional<EThirdPartyPlatform>& ClientType, StringResultCallback Callback);
 
     /// @brief Part two of the 3rd party authentication flow
     /// Note: The steps are as follows:
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -145,37 +145,45 @@ public:
     CSP_ASYNC_RESULT void LoginAsGuestWithDeferredProfileCreation(
         const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
-    /// @ingroup Third Party Authentication
-    /// @brief As a Connected Spaces Platform user the 3rd party authentication flow consists of two steps, first calling
-    /// GetThirdPartyProviderAuthoriseURL followed by LoginToThirdPartyAuthenticationProvider You can see a Sequence Diagram with all the parties
-    /// involved including what a Client should be calling and when here https://miro.com/app/board/uXjVPflpu98=/.
-
     /// @brief API to retrieve the Connected Spaces Platform supported 3rd party authentication providers
     /// @return Array of Connected Spaces Platform supported 3rd party authentication providers
     [[nodiscard]] csp::common::Array<EThirdPartyAuthenticationProviders> GetSupportedThirdPartyAuthenticationProviders() const;
 
-    /// @brief First step of the 3rd party authentication flow
-    /// If you call this API but for some reason you'd like to call this again, this is supported, the params you pass second time will replace the
-    /// ones you've passed initially
-    /// @param AuthProvider EThirdPartyAuthenticationProviders : one of the supported Authentication Providers
-    /// @param RedirectURL csp::common::String : the RedirectURL you want to be used for this authentication flow
-    /// @param Callback StringResultCallback : callback that contains the Authorise URL that the Client should be navigating next before moving to the
-    /// second Connected Spaces Platform Authentication step
-    CSP_ASYNC_RESULT void GetThirdPartyProviderAuthoriseURL(
+    /// @brief Part one of the 3rd party authentication flow
+    /// Note: The steps are as follows:
+    /// 1. This step. Call this method to retrieve the Authorization URL which is required for step 2.
+    /// 2. The caller should navigate to the Authorize URL retrieved in step 1, where they will authenticate with the 3rd party provider and retrieve
+    /// an auth token and state Id.
+    /// 3. Call @ref UserSystem::LoginToThirdPartyAuthenticationProvider() with the auth token and state Id retrieved in step 2.
+    ///
+    /// Calling this method will not impact your current login state.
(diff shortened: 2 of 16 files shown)


Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -146,6 +146,28 @@ csp::systems::Profile GetFullProfileByUserId(csp::systems::UserSystem* UserSyste
     return GetProfileResult.GetProfile();
 }
 
+// Decode a percent-encoded URL
+std::string DecodeURL(const std::string& EncodedURL)
```
MAG-AdamThorn: This utility function is removed in a later commit: [299a23e](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/299a23e8dce9194f8de345b8888f4e18d612fe1c)

Library/include/CSP/Common/SharedEnums.h
```diff
@@ -222,15 +221,50 @@ enum class StereoVideoType
...
+        break;
+    }
+    case csp::systems::EThirdPartyPlatform::Web:
+    {
+        PlatformString = "Web";
+        break;
+    }
+    default:
+    {
+        PlatformString = std::string("Unknown platform. Value") + std::to_string(static_cast<unsigned int>(Platform));
```
mag-lt: [Question] This value could (in theory, if the case was actually possible) end up getting sent to the server, is this deliberate?

[Comment] When switching on an enum class, I'd generally advocate these days for not adding a `default` case. With the right compiler flags, modern compilers will warn on a non exhaustive switch, which can be turned into an error with warnings as errors (so if a new case is added, it forces it to get handled. putting a default case supresses this). I'd typically write this like:

```
inline std::string ThirdPartyPlatformToString(csp::systems::EThirdPartyPlatform Platform)
{
    switch (Platform)
    {
    case csp::systems::EThirdPartyPlatform::None:
    {
        return "None";
    }
    case csp::systems::EThirdPartyPlatform::Unreal:
    {
        return "Unreal";
    }
    case csp::systems::EThirdPartyPlatform::Unity:
    {
        return "Unity";
    }
    case csp::systems::EThirdPartyPlatform::Web:
    {
        return "Web";
    }
    }

    std::unreachable(); // C++23 feature, but could have project specific macro with something like the `possible implementation` shown
                        // [here](https://en.cppreference.com/cpp/utility/unreachable)
                        // This satisfies the compiler (if it isn't already by the exhaustive switch),
                        // that this code cannot be reached, and so says "I know what I'm doing, don't warn me" 
}
```

May not be quite the right pattern here though, which looks to me based on the calling code that it should instead return a std::optional

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -532,144 +532,173 @@ csp::common::Array<EThirdPartyAuthenticationProviders> UserSystem::GetSupportedT
...
-
             StringResult ErrorResult(ProviderDetailsRes.GetResultCode(), ProviderDetailsRes.GetHttpResultCode());
             ErrorResult.SetValue("error");
             Callback(ErrorResult);
         }
     };
 
+    std::optional<csp::common::String> ClientTypeString;
+
+    if (ClientType.HasValue() && *ClientType != csp::systems::EThirdPartyPlatform::None)
```
mag-lt: [Consider] Based on this being the only usage of `ThirdPartyPlatformToString`, it looks like we don't want a string representation of the `None` case. We could get rid of this check, and make `ThirdPartyPlatformToString` instead return an optional, only returning values for the cases/strings we actually want to communicate

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
```diff
@@ -70,6 +70,7 @@ class POCOWebClient : public WebClient
     void Put(HttpRequest& Request);
     void Delete(HttpRequest& Request);
     void Head(HttpRequest& Request);
+    void Patch(HttpRequest& Request);
```
mag-lt, line 75: [Question] (not specifically about this code, but not sure where else to flag), does this need to be done for `EmscriptenWebClient` too (I assume yes)? Why is this being handled in the context of this PR for POCOWebClient if nothing is actually forcing it? I'm curious what flagged this (a failing test? a compiler error?)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -380,6 +383,64 @@ void POCOWebClient::Head(HttpRequest& Request)
     LogHttpResponseIfLoglevelVeryVerbose(LogSystem, "HEAD", Request, PocoResponse);
 }
 
+void POCOWebClient::Patch(HttpRequest& Request)
```
mag-lt, line 374: [Nit] A lot of duplicate code between this and `Post`, and to some extent `Get`. I'm pretty sure this could all by DRY-ed up into a single function.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -579,91 +594,104 @@ void UserSystem::LoginToThirdPartyAuthenticationProvider(const csp::common::Stri
...
-        CurrentLoginState.State = csp::common::ELoginState::Error;
+        CSP_LOG_ERROR_MSG("The provided ThirdPartyStateId is not correct"); // intentionally not too explicit about the error for security reasons
 
         csp::systems::LoginStateResult ErrorResult;
         ErrorResult.SetResult(csp::systems::EResultCode::Failed, (uint16_t)csp::web::EResponseCodes::ResponseBadRequest);
         Callback(ErrorResult);
     }
 
-    LoginStateResultCallback LoginStateResCallback = [=](const LoginStateResult& LoginStateRes)
+    if (CurrentLoginState.State == csp::common::ELoginState::LoggedOut || CurrentLoginState.State == csp::common::ELoginState::Error)
```
mag-lt: [Nit] The invalid state (i.e. the `else`) here could've been checked explicitly and then early return, vs introducing further nesting and a bigger diff here.

[Consider] Adding a test for this invalid case. It isn't clear to me on first read why it has been introduced here.

mag-lt (approved):

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -145,37 +145,45 @@ class CSP_API UserSystem : public SystemBase
...
+    /// Note: The steps are as follows:
+    /// 1. This step. Call this method to retrieve the Authorization URL which is required for step 2.
+    /// 2. The caller should navigate to the Authorize URL retrieved in step 1, where they will authenticate with the 3rd party provider and retrieve
+    /// an auth token and state Id.
+    /// 3. Call @ref UserSystem::LoginToThirdPartyAuthenticationProvider() with the auth token and state Id retrieved in step 2.
+    ///
+    /// Calling this method will not impact your current login state.
+    /// @param AuthProvider EThirdPartyAuthenticationProviders : The Authentication Provider you want to be authenticated with (e.g. Google, Apple,
+    /// Discord).
+    /// @param RedirectURL csp::common::String : the RedirectURL you want to be used for this authentication flow.
```
mag-lt: [Nit] The types here are (and were before these changes) inconsistent wrt the actual method signature (where they are actually const ref. same for `ClientType` below). As we discovered recently, writing these types in comments is actually redundant and doxygen can lift them out from the actual code.

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
mag-lt: As we discussed offline, I'd missed that this has been done, as EmscriptenWebClient is already factored in a way that doesn't require much new code to support the new request type

Library/include/CSP/Common/SharedEnums.h
MAG-AdamThorn: Ah, an excellent point! I have migrated this function to the UserSystem and updated it to return a std::optional so that it can be used directly.

Update: [6caa948](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/6caa9487b0558a5ab2f8c3f04be16f1ef77afbdf)

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Addressed as part of the same commit above.

Update: [6caa948](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/6caa9487b0558a5ab2f8c3f04be16f1ef77afbdf)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: Agreed, have introduced new functions with common code to reduce code duplication.

Update: [26397e3](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/26397e305e4b271392d7505e34cfc37ee45b23c6)

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Updated to check the invalid state up front to remove the nested logic. Not really possible to test this method in any meaningful way currently as it requires a manual step.

Update: [10b1a91](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/10b1a9123cb1ebdabdc3b07d982c714088712600)

Library/include/CSP/Systems/Users/UserSystem.h
MAG-AdamThorn: Removed the types from these two methods but not currently elsewhere in this class.

Update: [66a6fea](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/66a6feadd3c9f524fce2ded58be258fa836b6019)

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -93,6 +93,32 @@ bool CheckExpiryLengthFormat(const csp::common::String& ExpiryLength)
     return false;
 }
 
+inline std::optional<csp::common::String> ThirdPartyPlatformToString(const csp::common::Optional<csp::systems::EThirdPartyPlatform>& ClientType)
```
mag-lt: [Nit] `inline` is likely totally redundant here (we're within an anonymous namespace, so we have internal linkage within this single translation unit, so no ODR issues need solving here, whereas they did before when this was in a header)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
mag-lt: [Comment] I hadn't realised there was such different processing (sending responses without a body, streaming the body, sending it all in one go). Can't help but wonder if they could all just be streamed / handled uniformly. The resulting method bodies are still quite repetitive, but not the most important thing to fix right now especially as this work isn't even needed for the bug fix here despite being a large portion of the changes in the PR at this point 😅

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -380,6 +383,64 @@ void POCOWebClient::Head(HttpRequest& Request)
...
+{
+    CSP_PROFILE_SCOPED_FORMAT("PATCH %s", Request.GetUri().GetAsStdString().c_str());
```
MAG-mv, line 376: [Question] Do we not need to add this functionality for emscripten?

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -579,91 +594,104 @@ void UserSystem::LoginToThirdPartyAuthenticationProvider(const csp::common::Stri
...
         csp::systems::LoginStateResult ErrorResult;
         ErrorResult.SetResult(csp::systems::EResultCode::Failed, (uint16_t)csp::web::EResponseCodes::ResponseBadRequest);
         Callback(ErrorResult);
     }
 
-    LoginStateResultCallback LoginStateResCallback = [=](const LoginStateResult& LoginStateRes)
+    if (CurrentLoginState.State == csp::common::ELoginState::LoggedOut || CurrentLoginState.State == csp::common::ELoginState::Error)
     {
-        if (LoginStateRes.GetResultCode() == csp::systems::EResultCode::Success)
+        CurrentLoginState.State = csp::common::ELoginState::LoginRequested;
```
MAG-mv: [Question] Didn't we decide a while ago not to modify the login state in this function?

mag-lt (approved):

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-AdamThorn: The `EmscriptenWebClient` has been updated as part of this PR. It doesn't have lots of duplicate logic like the POCOWebClient with dedicated functions of reach verb, so it just needed an additional switch statement. Do you think something more is needed?

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: That was for the step 1 function in the flow - `GetThirdPartyProviderAuthorizeURL`. Step one should not and now does not have any impact on your login state. You can call it while logged in if you want and it will have no impact. This second method is now logging you in, so it should check your login state and effect it, just as the other login methods do.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Ah yes, a copy paste error. Thanks

Fix: [b0fd325](https://github.com/magnopus-opensource/connected-spaces-platform/pull/948/changes/b0fd3255787c89d581b9b377c1a1e38ede3a15b7)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-mv: Sounds good. I was worried that additional work was needed due to Poco needing POCOWebClient::Patch

Library/src/Systems/Users/UserSystem.cpp
MAG-mv: I see. Looks good!

MAG-mv (approved):

PR #951, opened 2026-04-28
[NT-0] fix: Add missing Client param for usersLogin_socialPost call

MAG-AdamThorn:
We were missing the `Client` param for the `chs_user::LoginSocialRequest` object sent to `(AuthenticationAPI)->usersLogin_socialPost()`.

This has been added as well as a test that confirms that the `ELoginState` is valid after a failed login attempt.

### DIFF
--- a/Tests/src/PublicAPITests/UserSystemTests.cpp
+++ b/Tests/src/PublicAPITests/UserSystemTests.cpp
@@ -146,7 +146,15 @@ csp::systems::Profile GetFullProfileByUserId(csp::systems::UserSystem* UserSyste
     return GetProfileResult.GetProfile();
 }
 
-void ValidateThirdPartyAuthorizeURL(const csp::common::String& AuthoriseURL, const csp::common::String& RedirectURL)
+struct AuthorizeURLTokens
+{
+    std::string StateId;
+    std::string ClientId;
+    std::string Scope;
+    std::string RetrievedRedirectURL;
+};
+
+void ValidateThirdPartyAuthorizeURL(const csp::common::String& AuthoriseURL, const csp::common::String& RedirectURL, AuthorizeURLTokens& OutTokens)
 {
     EXPECT_FALSE(AuthoriseURL.IsEmpty());
     EXPECT_NE(AuthoriseURL, "error");
@@ -157,10 +165,10 @@ void ValidateThirdPartyAuthorizeURL(const csp::common::String& AuthoriseURL, con
     const char* REDIRECT_URL_PARAM = "redirect_uri=";
     const char* INVALID_URL_PARAM_VALUE = "N/A";
 
-    std::string StateId = INVALID_URL_PARAM_VALUE;
-    std::string ClientId = INVALID_URL_PARAM_VALUE;
-    std::string Scope = INVALID_URL_PARAM_VALUE;
-    std::string RetrievedRedirectURL = INVALID_URL_PARAM_VALUE;
+    OutTokens.StateId = INVALID_URL_PARAM_VALUE;
+    OutTokens.ClientId = INVALID_URL_PARAM_VALUE;
+    OutTokens.Scope = INVALID_URL_PARAM_VALUE;
+    OutTokens.RetrievedRedirectURL = INVALID_URL_PARAM_VALUE;
 
     const auto& Tokens = AuthoriseURL.Split('&');
     for (size_t idx = 0; idx < Tokens.Size(); ++idx)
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -968,7 +978,39 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, GetAuthorizeURLForGoogleTestWithClie
...
+
+    EXPECT_EQ(UserSystem->GetLoginState().State, csp::common::ELoginState::LoggedOut);
+
+    // Attempt to login which we expect to fail
+    auto [ResLogin]
+        = AWAIT_PRE(UserSystem, LoginToThirdPartyAuthenticationProvider, RequestPredicate, Token, URLTokens.StateId.c_str(), false, true, {});
+    EXPECT_EQ(ResLogin.GetResultCode(), csp::systems::EResultCode::Failed);
+
+    // If ELoginState == Error that means it passed the check comparing the state Id passed in to the one cached from the call to
+    // GetThirdPartyProviderAuthorizeURL() and attempted to log in, which will fail.
```
mag-lt: [Nit] It'd be better if we could instead observe that it actually tried to make the web request and that the request was populated with the expected client id, otherwise we can't actually really be sure without being very intimate with the implementation details that we are actually testing the thing we intend to

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -146,7 +146,15 @@ csp::systems::Profile GetFullProfileByUserId(csp::systems::UserSystem* UserSyste
...
-void ValidateThirdPartyAuthorizeURL(const csp::common::String& AuthoriseURL, const csp::common::String& RedirectURL)
+struct AuthorizeURLTokens
+{
+    csp::common::String StateId;
+    csp::common::String ClientId;
+    csp::common::String Scope;
+    csp::common::String RetrievedRedirectURL;
+};
+
+void ValidateThirdPartyAuthorizeURL(const csp::common::String& AuthoriseURL, const csp::common::String& RedirectURL, AuthorizeURLTokens& OutTokens)
```
mag-lt: [Nit] The out param appears to only be used in one of the five cases it is called, and then it only uses one of the fields (StateId).

There are mixed concerns here, one being parsing out the query parameters from the URL, and then validating them for this particular URL. If our `Uri` type could handle splitting out the parameters (which I'm not suggesting adding here), the tests would likely be written more naturally.

In any case, the steps could still be split out here, where a helper would return a `std::optional<AuthorizeURLTokens>`, this would just be a pure C++ function with no GTest stuff interleaved. Tests would then just setup expectations against that directly, and in the case where you happen to need the values later on, you could do so as they'd already be in a variable locally that was already checked with an assertion to be non optional.

Though, at some level that is of course just a different way to achieve the same thing 😅

mag-lt (approved):

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ElliotMorris: [Nit] Prefer returns to out params in almost all cases. It's more ergonomic when you don't need the result outside the function if nothing else.

... Ah, I see Lucas has said pretty much the same thing, with some added spice.

MAG-ElliotMorris (approved):

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: I agree that would be much better. I created that test to help me validate behaviour as I was trouble shooting but in retrospect it is not appropriate to merge it. I have removed the test and will consider a better approach separately from this work.

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-AdamThorn: Excellent idea, I have separated out the steps as suggested.

Update: [c3f7431](https://github.com/magnopus-opensource/connected-spaces-platform/pull/951/changes/c3f7431b0fb901806e7ea187cbbd1b73d0f568bc)

MAG-AdamThorn: Yes, great minds and all that.

Update: [c3f7431](https://github.com/magnopus-opensource/connected-spaces-platform/pull/951/changes/c3f7431b0fb901806e7ea187cbbd1b73d0f568bc)

PR #953, opened 2026-05-01
[NT-0] fix: Incorrect client argument used for Social Login

MAG-AdamThorn:
Following a discussion with the backend services team it was discovered that the wrong client value was being cached for later use in the  LoginToThirdPartyAuthenticationProvider() method.

### DIFF
--- a/Library/src/Systems/Users/UserSystem.cpp
+++ b/Library/src/Systems/Users/UserSystem.cpp
@@ -562,6 +562,14 @@ csp::common::Array<EThirdPartyAuthenticationProviders> UserSystem::GetSupportedT
     return Providers;
 }
 
+void UserSystem::ResetThirdPartyAuthState()
+{
+    ThirdPartyAuthStateId = "";
+    ThirdPartyClientType = "";
+    ThirdPartyAuthRedirectURL = "";
+    ThirdPartyRequestedAuthProvider = EThirdPartyAuthenticationProviders::Invalid;
+}
+
 void UserSystem::GetThirdPartyProviderAuthorizeURL(EThirdPartyAuthenticationProviders AuthProvider, const csp::common::String& RedirectURL,
     const csp::common::Optional<EThirdPartyPlatform>& ClientType, StringResultCallback Callback)
 {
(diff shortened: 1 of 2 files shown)


Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -612,14 +619,23 @@ void UserSystem::GetThirdPartyProviderAuthorizeURL(EThirdPartyAuthenticationProv
         }
     };
 
+    // Reset the state related to third party authentication. This ensures that if the client calls GetThirdPartyProviderAuthorizeURL and the calls
+    // fails they will not be able to call LoginToThirdPartyAuthenticationProvider() with invalid state.
+    ResetThirdPartyAuthState();
```
MAG-ElliotMorris, line 624: [Comment] The idea of doing state management in a getter method (or tbh, at all), is unpleasant. I see that's how this method has been designed however, it's doing state management all over the shop :/

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -612,14 +619,23 @@ void UserSystem::GetThirdPartyProviderAuthorizeURL(EThirdPartyAuthenticationProv
...
+
     std::optional<csp::common::String> Client = ThirdPartyPlatformToString(ClientType);
+    if (Client.has_value())
+    {
+        ThirdPartyClientType = Client.value();
+    }
```
MAG-ElliotMorris, line 630: [Question] At this point, we've just reset all the auth state, and we're setting `ThirdPartyClientType` back. Do we need to set the other values too? I guess I'm confused why this one specific value (one of 4), is being set here, but not the others.

MAG-AdamThorn: The other members are being set in the callback above from the state returned from the call to the backend services, however this value is set from the value passed in by the client.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Agreed. This flow needs a rethink if I am honest.

Library/src/Systems/Users/UserSystem.cpp
MAG-ElliotMorris: Right I see. Makes it even weirder, having 2 sources of state. [Consider] renaming the method if such a break is feasible, because it's doing so much more than just getting a URL. 🤷

MAG-ElliotMorris (approved):

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: I won't be able to do that as part of this change unfortunately. But I agree that makes sense. We will be back in this area soon when support for Netflix SSO get added, so I will address this and general quirks with this flow then.

MAG-mv (approved):

PR #954, opened 2026-05-01
[NT-0] build: Cmake conveniences for downstream integrators.

MAG-ElliotMorris:
Pulled these changes out of the unity backwards integration branch, as they're somewhat different from the bulk of that change. They were all sourced from desires and necessities found integrating the new build system into the unity swig bindings.

Mostly convenience updates based on using the cmake build system and discovering some quirks. Makes downstream integration less burdensome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,7 +182,11 @@ target_compile_definitions(csp-lib
 
 # Warnings as errors flag
 # Sets WX/Werror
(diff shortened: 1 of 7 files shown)


MAG-AdamThorn (approved):

MAG-mv (approved):
Looks good. Thanks for the build system changes

PR #955, opened 2026-05-05
[OW-2449] feat: Add height parameter to the fog space component

MAG-ME:
Added height parameter to the fog space component

- Added new parameter to the component
- Added the parameter to the script binding for the component
- Updated the fog component test to include the new parameter


MAG-mv (approved):

Library/include/CSP/Multiplayer/Components/FogSpaceComponent.h
```diff
@@ -132,6 +133,16 @@ class CSP_API FogSpaceComponent : public ComponentBase, public IThirdPartyCompon
     /// @param Value float : Global density factor.
     void SetDensity(float Value);
 
+    /// @brief Gets Fog Height.
+    /// @note Controls the height above which the fog density starts to diminish.
```
MAG-ElliotMorris, line 137: [nit] Odd that stuff like this is documented as `note`, this is more just the regular body of the documentation. However I see that's the pattern in this file. 🤷

Library/src/Multiplayer/Components/FogSpaceComponent.cpp
```diff
@@ -64,6 +64,11 @@ const auto Schema = ComponentSchema {
             "density",
             0.4f,
         },
+        {
+            static_cast<ComponentProperty::KeyType>(FogPropertyKeys::Height),
+            "height",
+            0.0f,
```
MAG-ElliotMorris, line 70: [Consider] Are you sure you want this to default to 0? Will that make it such that when you add a fog component it will appear to have no effect as it has zero height? This is more of an OKO question than a CSP one.

MAG-ElliotMorris (approved):
LGTM. Good job navigating the in-progress data component transformations.

PR #956, opened 2026-05-06
[OF-1831] Build on Linux

MAG-mv:
This PR allows the entire CSP solution to build on Linux. At this point, it has not been tested if it can run, which will be the next part of this work.

This PR ended up being quite small after moving the work to fix the changes-meaning errors in a later PR. This work consists of:

- Adding GCC compile options to supress known errors
- Adding platform specific Linux code
- Fixing includes due to gcc having a different implementation
- Fixing issues due to gcc being stricter
- Adding conan profiles for Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,10 +218,19 @@ target_compile_options(csp-lib
 	    -Wno-ignored-qualifiers
     >
 
+    $<$<CXX_COMPILER_ID:GNU>:
+      -Werror
+      -Wextra
+      # Don't error when naming members the same as class
+      -Wno-error=changes-meaning
+      -Wno-error=deprecated-declarations
+      -Wno-missing-field-initializers
+      -Wno-ignored-qualifiers
+      -fpermissive
+    >
+
     $<$<PLATFORM_ID:Emscripten>:-pthread;-fwasm-exceptions>
 )
-
-
 
 target_link_options(csp-lib PRIVATE
   $<$<PLATFORM_ID:Emscripten>:
@@ -372,6 +381,16 @@ if(BUILD_TESTING)
 
   target_compile_options(csp-tests PRIVATE
     $<$<CXX_COMPILER_ID:MSVC>:/permissive->
+
+    $<$<CXX_COMPILER_ID:GNU>:
+      -Werror
+      -Wextra
+      -Wno-error=changes-meaning
+      -Wno-error=deprecated-declarations
+      -Wno-missing-field-initializers
+      -Wno-ignored-qualifiers
+      -fpermissive
+    >
   )
 
   # Ensure csp and MultiplayerTestRunner are built first
--- a/Library/include/CSP/CSPCommon.h
+++ b/Library/include/CSP/CSPCommon.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Magnopus LLC
+ * Copyright 2026 Magnopus LLC
 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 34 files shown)


Library/src/CSPFoundation.cpp
```diff
@@ -54,24 +54,23 @@
...
 #include <sys/stat.h>
+#include <sys/types.h>
 // For PATH_MAX
-#include <sys/syslimits.h>
+#include <limits.h>
+
+#if defined(CSP_ANDROID)
+    #define LOAD_OWN_MODULE() dlopen(LIB_NAME, RTLD_LAZY)
+#else
+    #define LOAD_OWN_MODULE() dlopen(nullptr, RTLD_LAZY)
```
mag-lt, line 72: [question] any idea why we've historically done this on Apple platforms (i.e. pass `nullptr` vs the lib name / path)? This implies to me that there is a fundamentally different assumption about how the library is linked on different platforms. Using dlopen like this with `nullptr` to me implies we expect the consumer of the library to have statically linked it into the main executable, and so we expect the symbols to be found there, vs in an explicitly loaded runtime library. If this were an apple specific thing, I suspect it was for one specific use case that we had, and I'm not sure we'd want this behaviour on linux too.

However, I honestly doubt that anyone is consuming the library in this way, i.e. via explicit runtime linking. I think they'd be limited to looking up symbols from the generated C API (i.e. when using `dlsym`, there is no way they're going to know the names of C++ symbols to lookup).

Library/src/CSPFoundation.cpp
```diff
@@ -157,21 +156,37 @@ std::string DeviceIdPath()
...
 
-    return CSPDataRoot;
+#elif defined(CSP_LINUX)
+
+    const char* Home = getenv("HOME");
+    return std::string(Home ? Home : "") + "/.local/share/MagnopusCSP/";
+
+#else
+
+    return {};
```
mag-lt: [consider] adding an assert(false) or something here, so a developer is guided here when targeting a new platform (or something has gone wrong with setting their platform flags).

ThirdParty/quickjs/src/quickjs-libc.c
```diff
@@ -63,7 +63,15 @@ typedef sig_t sighandler_t;
 #include <crt_externs.h>
 #define environ (*_NSGetEnviron())
 #endif
-#endif /* __APPLE__ */
+
+#elif defined(__linux__)
+// sighandler_t is not part of standard C POSIX, so may not be defined
```
mag-lt: [comment] I'm a bit surprised by needing to do this, as I have to imagine quickjs has been compiled on linux by many. I wonder if an include is missing somewhere or something, or if we're missing a compiler flag?

mag-lt (approved):

ThirdParty/quickjs/src/quickjs-libc.c
MAG-mv: I spent some time looking into this and found that the upstream CMakeLists defines gnu extensions using _GNU_SOURCE. I have fixed up my commit to do his instead:

https://github.com/magnopus-opensource/connected-spaces-platform/pull/956/commits/cb125b72533fc7fdbb7b5e5d98671d157801f348

Library/src/CSPFoundation.cpp
MAG-mv: I have no idea why this was done. We do need toi review the way this is done at some point.

Library/src/CSPFoundation.cpp
MAG-mv: Fixed up to add compile error:

[7d30185c89a5f955d9bf0e1e775d5780e0334038](https://github.com/magnopus-opensource/connected-spaces-platform/pull/956/changes/7d30185c89a5f955d9bf0e1e775d5780e0334038)

Library/src/CSPFoundation.cpp
mag-lt: Fair.

It looks like the C# wrapper generator uses this via `GetFunctionAddress`, and there were problems previously on [Android](https://github.com/magnopus-opensource/connected-spaces-platform/pull/808).

I'd personally have expected loading the library by path as Android is doing, as I've never personally used dlopen any other way (though as the path isn't parameterised, this is already hardcoding some assumptions about where the library will be on disk with relation to the consuming executable). I'd suggest doing so here, but I am admittedly speculating about the intended usage still. So, perhaps until we know this is best left as you've written here.

Library/include/CSP/CSPCommon.h
```diff
@@ -18,65 +18,87 @@
...
-#if defined(TARGET_OS_MAC)
-#define CSP_MACOSX
-#elif defined(TARGET_OS_IPHONE)
-#define CSP_IOS
-#endif
+    #if TARGET_OS_IPHONE
+        #define CSP_IOS
+    #elif TARGET_OS_MAC
+        #define CSP_MACOSX
+    #endif
```
MAG-ElliotMorris, line 31: [Comment] I wonder if this translation to `CSP` style macros is strictly necessary in the long run.

Library/include/CSP/CSPCommon.h
```diff
@@ -18,65 +18,87 @@
...
+    #define PRAGMA_WARNING_IGNORE_MSVC(WarningCode)
+    #define PRAGMA_WARNING_IGNORE_CLANG(WarningName)
+    #define PRAGMA_WARNING_IGNORE_GCC(WarningName) CSP_DO_PRAGMA(GCC diagnostic ignored #WarningName)
+#else
+    #define PRAGMA_WARNING_PUSH()
+    #define PRAGMA_WARNING_POP()
+    #define PRAGMA_WARNING_IGNORE_MSVC(WarningCode)
+    #define PRAGMA_WARNING_IGNORE_CLANG(WarningName)
+    #define PRAGMA_WARNING_IGNORE_GCC(WarningName)
+#endif
```
MAG-ElliotMorris, line 85: [Consider] Similarly, I wonder if these are really necessary. Made this one a consider because I don't think there's many uses of these in our codebase and I think it might be possible to just remove them. One would imagine you are normally disabling a warning for a specific compiler, which you can do with the compiler specific syntax. The cases where you would need to do this cross-platform seem rare to me, and even an antipattern.

Library/include/CSP/CSPCommon.h
```diff
@@ -92,10 +114,11 @@
 #define CSP_NO_DISPOSE
 #define CSP_FLAGS
 
-// TODO: Remove the following includes. I don't think we should be default including these everywhere
+// TODO: Remove the following includes. 
+// I don't think we should be default including these everywhere
 #include <stdint.h>
 #include <stdlib.h>
```
MAG-ElliotMorris, line 120: [Comment] One day ... noble TODO author ... one day.

Library/src/CSPFoundation.cpp
```diff
@@ -54,24 +54,23 @@
...
+#include <sys/types.h>
 // For PATH_MAX
-#include <sys/syslimits.h>
+#include <limits.h>
+
+#if defined(CSP_ANDROID)
+    #define LOAD_OWN_MODULE() dlopen(LIB_NAME, RTLD_LAZY)
+#else
+    #define LOAD_OWN_MODULE() dlopen(nullptr, RTLD_LAZY)
+#endif
```
MAG-ElliotMorris, line 73: [Consider] Adding a bit of docs for this, just to indicate the intent of why this is done, because it's confusing. You can link to the `dlopen` spec. https://pubs.opengroup.org/onlinepubs/9699919799/functions/dlopen.html

Specifically, say that the normal behaviour is for a null pointer to return the symbol table for the currently loaded process, but android diverges from that, and passes the specific library name.

I do not know the implications this has on static and dynamic linkage. ... But tbh I don't even know how this macro machinery is used. 🤔

From the dlopen spec:
> If file is a null pointer, dlopen() shall return a global symbol table handle for the currently running process image. This symbol table handle shall provide access to the symbols from an ordered set of executable object files consisting of the original program image file, any executable object files loaded at program start-up as specified by that process file (for example, shared libraries), and the set of executable object files loaded using dlopen() operations with the RTLD_GLOBAL flag. As the latter set of executable object files can change during execution, the set of symbols made available by this symbol table handle can also change dynamically.

Library/src/CSPFoundation.cpp
```diff
@@ -157,21 +156,37 @@ std::string DeviceIdPath()
...
-    char CSPDataRoot[PATH_MAX];
-    sprintf(CSPDataRoot, "%s/Library/MagnopusCSP/", getenv("HOME"));
-#endif
+    const char* Home = getenv("HOME");
+    return std::string(Home ? Home : "") + "/Library/MagnopusCSP/";
 
-    return CSPDataRoot;
+#elif defined(CSP_LINUX)
+
+    const char* Home = getenv("HOME");
```
MAG-ElliotMorris, line 185: [Consider] This will usually (as in 99%+) work, but it's not technically correct. 

Correctness here is weird though, as the [specs ](https://specifications.freedesktop.org/basedir/latest/)are de-facto rather than enforced. All the major desktop environments follow them, although there are always holdouts that do things their own way (firefox jumps to mind).

> $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

> $XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used.

Honestly, i'm not overly fussed about this, the same criticisms apply to our MacOS handling and that's been "fine" for ages. Would be good to get right at some point though.

CMakeLists.txt
```diff
@@ -218,11 +218,20 @@ target_compile_options(csp-lib
...
+    $<$<CXX_COMPILER_ID:GNU>:
+      -Werror
+      -Wextra
+      # Don't error when naming members the same as class
+      -Wno-error=changes-meaning
+      -Wno-error=deprecated-declarations
+      -Wno-missing-field-initializers
+      -Wno-ignored-qualifiers
+      -fpermissive
+    >
```
MAG-ElliotMorris, line 230: [Comment] It's going to be all too easy to just leave these here forever, but I really don't want to. Maybe @MAG-AdamThorn can figure out a way to force the issue for us as a team.

... Dealing with one of these at a time might make good low-hanging-fruit stuff.

Tests/src/MultiplayerTestRunnerProcess.h
```diff
@@ -121,5 +121,5 @@ class MultiplayerTestRunnerProcess
     std::optional<std::string> Endpoint;
 
     // Created in StartProcess
-    std::unique_ptr<TinyProcessLib::Process> ProcessHandle = nullptr;
+    std::unique_ptr<TinyProcessLib::Process> ProcessHandle;
```
MAG-ElliotMorris, line 124: [Question] This is a surprising change, can you explain more in depth why it is required?

My assumption would be that this isn't actually about the setting of a unique_ptr to nullptr, but is rather about the generation of the destructor or the move/copy operations. You should be able to null-construct a pointer even to a forward-declared type.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -696,7 +696,8 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, UpdateDisplayNameTest)
 
     // Attempt Update - bad display name
     {
-        UniqueTestDisplayName = csp::common::String("??//-\"#~*") + GetUniqueString().c_str();
```
MAG-ElliotMorris, line 699: [comment] Lol! this is a funny one.

MAG-ElliotMorris (approved):
Nice one.

Tests/src/MultiplayerTestRunnerProcess.h
MAG-mv: This is essentially calling a function that requires memory allocation. Because the TinyProcessLib::Process has only been forward dec;ared at this point, it gives "use of undefined type" error within unique_ptr.

This can be replicated in msvc by creatig a  function in a header that calls ```ProcessHandle.reset(nullptr);```

Library/include/CSP/CSPCommon.h
MAG-mv: It seems pretty standrd, but you're right it may not be needed

Library/src/CSPFoundation.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/956/commits/9c684dbc9d1b0839bd2e70989d412bb7682bfec6

profiles/host/linux_clang
```diff
@@ -0,0 +1,15 @@
...
+compiler.version=18
+compiler.libcxx=libstdc++11
+compiler.cppstd=17
+
+[tool_requires]
+cmake/3.31.6
+ninja/1.11.1
+
+[conf]
+ools.build:compiler_executables={"c":"/usr/bin/clang","cpp":"/usr/bin/clang++"}
```
mag-lt: [fix] `ools`, presumably a typo? and so if builds are working at all, this is presumably not needed? is it common to explicitly specify paths to tools on the system in shared profiles?

ThirdParty/signalrclient/include/signalrclient/http_client.h
```diff
@@ -8,6 +8,7 @@
 #include <functional>
 #include <map>
 #include <chrono>
+#include <exception>
```
mag-lt, line 11: [comment] looks like this has also been [fixed upstream](https://github.com/aspnet/SignalR-Client-Cpp/commit/8b0f7aab0c472d4230e68b4bbad84ae26c250a54) in the same way

profiles/host/linux_clang
MAG-mv: Oops. I failed copying this from Remote Desktop. Updated here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/956/commits/81aed4ec49af8282ae49c36c7e73aa127eb6f501

Tests/src/MultiplayerTestRunnerProcess.h
MAG-ElliotMorris: I don't think that is true. From cppreference : 

> unique_ptr may be constructed for an [incomplete type](https://en.cppreference.com/cpp/language/type#Incomplete_type) T, such as to facilitate the use as a handle in the [pImpl idiom](https://en.cppreference.com/cpp/language/pimpl). If the default deleter is used, T must be complete at the point in code where the deleter is invoked, which happens in the destructor, move assignment operator, and reset member function of unique_ptr. (In contrast, std::shared_ptr cannot be constructed from a raw pointer to incomplete type, but can be destroyed where T is incomplete). Note that if T is a class template specialization, use of unique_ptr as an operand, e.g. !p requires T's parameters to be complete due to [ADL](https://en.cppreference.com/cpp/language/adl).

MAG-ElliotMorris: I wouldn't be surprised if this is about `noexcept` on the special constructors, linux compilers can be precious about that. 

For clarity, `unique_ptr` zero initializes anyway so this change is fine, but I want us to understand why we're doing things.

MAG-mv: "unique_ptr may be constructed for an [incomplete type](https://en.cppreference.com/cpp/language/type#Incomplete_type) T"  This is assigning a value, though which causes the deleter to need to know the size of the type.

Also, we shouldnt need = nullptr, due to the default constuctor doing this anyway

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-mv: I was very confused originally. I also had the same issue when adding ??/ in a comment!

Library/src/CSPFoundation.cpp
MAG-mv: Alright, I'm going to leave this one for now

Library/include/CSP/CSPCommon.h
MAG-mv: I think theres only a couple of instances we use this. I don't want to change as part of this PR, though

Tests/src/MultiplayerTestRunnerProcess.h
MAG-ElliotMorris: Value assignment is not different to construction, you're still talking about copy initialization. The default deleter does need to know the size of the type currently, but I think that is because the special constructors in this type are not marked noexcept.

PR #957, opened 2026-05-06
[NT-0] build: Add component registry files for cmake build

MAG-ElliotMorris:
Just forgotten, as the build system isn't in a test harness yet.


mag-lt (approved):
Thanks. Headers don't strictly need to be tracked via CMake, this just ensures they show up in IDEs. I'm daily driving the CMake, but mostly using VSCode, so have only been adding what is necessary to actually build.

_Edit_: The `CSP_MULTIPLAYER_PUBLIC_INCLUDES` bits are important here, as they make up part of `CSP_PUBLIC_SOURCES`, which is ultimately used in the `install` [step](https://github.com/magnopus-opensource/connected-spaces-platform/blob/289c271103b622e22de83c0f1413511e192200f9/CMakeLists.txt#L254) (via a file set [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/289c271103b622e22de83c0f1413511e192200f9/CMakeLists.txt#L100))

MAG-AdamThorn (approved):

PR #958, opened 2026-05-07
[NT-0] build: Emscripten profiles

MAG-ElliotMorris:
Enablement for the building of the WASM bindings project. 

This is about being able to build a static .a file, devoid of tests, with an emscripten host. You still use a windows (or whatever platform you're on) build profile. This is genuine cross-compilation, I hadn't really been thinking of it that way before.

You'll note this isn't really about producing WASM, this is about producing a library cross compiled to the WASM runtime, using the emscripten compilers. WASM + JS will be produced in the bindings repo by consuming these symbols. Seems a bit overcomplex, but separates the build configuration of that project (as well as the declarations of the bindings), which is something Magnopus needs to break the inter-team responsibility shrugging that's been going on.

The mental model to have (and the way you have to configure cmake supports this), is that `.a` is library, whilst `.wasm/.js` is equivalent to an executable. It is when you link into the executable that you decide things like initial/maximum memory, whether this is also a node build, are we going to include the web filesystem api, etc. Not things that CSP as a library needs to have an opinion on.

I have taken the opportunity to reframe the presets around a consistent axis.

Not in a CI harness yet, tested locally as usual. Getting antsy about this ... I want my CI harnesses, they make me feel better.

> [!NOTE]
>
> To be honest, i'm not sure the single-config build presets are pulling their weight. It's the difference between typing `cmake --build --preset debug-static` or `cmake --build build\Debug`. I might argue that if we are to do presets like this, we should also add install presets so then you can mindlessly type `--preset` into every step using the same name, which _is_ the most caveman thing to do, but it also adds a _bunch_ of duplication to the presets file for not much actual gain. Minor issue, can talk about it in weekly sync.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,8 +216,6 @@ target_compile_options(csp-lib
 	    -Wno-missing-field-initializers
 	    # Don't error on ignored qualifiers, such as returning const objects. These should be simple to remove at some point.
 	    -Wno-ignored-qualifiers
-        # Real big deal, REAL BUG in our custom list types, memcopying virtual objects :(
-        -Wno-nontrivial-memcall
     >
 
     $<$<CXX_COMPILER_ID:GNU>:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,8 @@ target_compile_options(csp-lib
 	    -Wno-missing-field-initializers
 	    # Don't error on ignored qualifiers, such as returning const objects. These should be simple to remove at some point.
 	    -Wno-ignored-qualifiers
+        # Real big deal, REAL BUG in our custom list types, memcopying virtual objects :(
+        -Wno-nontrivial-memcall
     >
 
     $<$<CXX_COMPILER_ID:GNU>:
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -3,51 +3,74 @@
   "include": ["./ConanPresets.json"],
   "configurePresets": [
     {
-      "name": "default-static",
-      "displayName": "multi config static",
+      "name": "multi-static",
+      "displayName": "multi-config static (with tests)",
       "inherits": "conan-default",
       "cacheVariables": {
-        "CSP_BUILD_SHARED": "OFF"
+        "CSP_BUILD_SHARED": "OFF",
+        "BUILD_TESTING": "ON"
       }
     },
     {
-      "name": "default-shared",
-      "displayName": "multi config shared",
+      "name": "multi-shared",
+      "displayName": "multi-config shared",
       "inherits": "conan-default",
       "cacheVariables": {
         "CSP_BUILD_SHARED": "ON",
         "BUILD_TESTING": "OFF"
       }
     },
-    {
-      "name": "release-static",
-      "displayName": "release single config static",
-      "inherits": "conan-release",
-      "cacheVariables": {
-        "CSP_BUILD_SHARED": "OFF"
-      }
-    },
-    {
-      "name": "release-shared",
-      "displayName": "release single config shared",
-      "inherits": "conan-release",
-      "cacheVariables": {
-        "CSP_BUILD_SHARED": "ON",
-        "BUILD_TESTING": "OFF"
-      }
-    },
+
     {
       "name": "debug-static",
-      "displayName": "debug single config static",
+      "displayName": "single-config debug static",
       "inherits": "conan-debug",
       "cacheVariables": {
-        "CSP_BUILD_SHARED": "OFF"
+        "CSP_BUILD_SHARED": "OFF",
+        "BUILD_TESTING": "OFF"
+      }
+    },
+    {
+      "name": "debug-static-tests",
+      "displayName": "single-config debug static (with tests)",
+      "inherits": "conan-debug",
+      "cacheVariables": {
+        "CSP_BUILD_SHARED": "OFF",
+        "BUILD_TESTING": "ON"
       }
     },
     {
       "name": "debug-shared",
-      "displayName": "debug single config shared",
+      "displayName": "single-config debug shared",
       "inherits": "conan-debug",
+      "cacheVariables": {
+        "CSP_BUILD_SHARED": "ON",
+        "BUILD_TESTING": "OFF"
+      }
+    },
+
+    {
+      "name": "release-static",
+      "displayName": "single-config release static",
+      "inherits": "conan-release",
+      "cacheVariables": {
+        "CSP_BUILD_SHARED": "OFF",
+        "BUILD_TESTING": "OFF"
+      }
+    },
+    {
+      "name": "release-static-tests",
+      "displayName": "single-config release static (with tests)",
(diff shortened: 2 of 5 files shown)


CMakePresets.json
```diff
@@ -3,97 +3,130 @@
   "include": ["./ConanPresets.json"],
   "configurePresets": [
     {
-      "name": "default-static",
-      "displayName": "multi config static",
+      "name": "multi-static",
+      "displayName": "multi-config static (with tests)",
```
MAG-ElliotMorris, line 7: [Comment] I'm conceptualizing these (with tests) presets as more like the _complete_ build. If we added more projects, I imagine they would get enabled in these presets, leaving the other ones as just the library.

CMakePresets.json
```diff
@@ -3,97 +3,130 @@
   "include": ["./ConanPresets.json"],
   "configurePresets": [
     {
-      "name": "default-static",
-      "displayName": "multi config static",
+      "name": "multi-static",
```
MAG-ElliotMorris, line 6: [Comment] Happy to revert this name change if folk think `default` is the more expectable thing. I just wanted to capture the axis. I know this goes back on what we were discussing @MAG-mv so feel free to shout at me.

CMakeLists.txt
```diff
@@ -216,6 +216,8 @@ target_compile_options(csp-lib
 	    -Wno-missing-field-initializers
 	    # Don't error on ignored qualifiers, such as returning const objects. These should be simple to remove at some point.
 	    -Wno-ignored-qualifiers
+        # Real big deal, REAL BUG in our custom list types, memcopying virtual objects :(
+        -Wno-nontrivial-memcall
```
mag-lt: [consider] putting this around the offending code with compiler pragmas (or our preprocessor wrappers for that), instead of project wide, not the sort of thing we want to be hiding (as your comment attests)

profiles/host/emscripten
```diff
@@ -1,18 +1,32 @@
...
 tools.cmake.cmaketoolchain:generator=Ninja
 tools.build:cxxflags=["-pthread","-fwasm-exceptions"]
 tools.build:cflags=["-pthread","-fwasm-exceptions"]
-tools.build:exelinkflags=["-pthread"]
\ No newline at end of file
+tools.build:exelinkflags=["-pthread"]
+tools.build:compiler_executables={'c':'{{ emsdk }}/upstream/emscripten/emcc{{ ext }}', 'cpp':'{{ emsdk }}/upstream/emscripten/em++{{ ext }}'}
+
+[buildenv]
+CC=emcc
```
mag-lt: [question] Are these redundant with the `tools.build:compiler_executables` entries above?

mag-lt (approved):
Looks reasonable to me. I'm not keen on the project wide ignoring of that warning, but I know you aren't either. I've not looked into the underlying cause, but if we're looking into that in the team already then hopefully it won't be needed for long in any case.

profiles/host/emscripten
MAG-ElliotMorris: That is entirely possible, I'll look into that tommorow. The [buildenv] stuff is taken from an example so I may be solving the same problem in 2 ways.

CMakeLists.txt
MAG-ElliotMorris: Fair, I think this is only in one location, although it's a big one, I'm grateful Matt is doing the linux work in parralel and is fixing it ... I'm honestly surprised anything works at all with errors like this in play in our interop types. Will do this tommoz.

... Stuff like this is why I want to purge our interop types, they're gravely under-considered. Let's transition to standard types first. Although I know you are concerned about ABI, and I wouldn't be against re-opening the conversation to transition to PIMPL-around-std types all in the one step, once the wrapper generator is replaced such that we are empowered to do that.

profiles/host/emscripten
mag-lt: It is possible you can do less here. Supposedly there is better first class support in Conan for Emscripten since version 2.18. I'm finding the docs a bit sparse, but they imply the latest versions can be used without having to fallback to a system one. I've not tried it out, but docs [here](https://docs.conan.io/2/examples/cross_build/emscripten.html) and [here](https://github.com/conan-io/conan-toolchains/blob/main/conan_config/profiles/emsdk/README.md) strongly suggest it should work. What version of Conan do you have locally?

MAG-AdamThorn (approved):

profiles/host/emscripten
MAG-ElliotMorris: 5.0.2. The latest recipe currently available is 3.1.73. There is a PR to try to update to emdsk 4, but it's been stuck in review hell for a while : https://github.com/conan-io/conan-center-index/pull/27213/changes. I think that doc was updated too soon. Trying to run the conan fetch step with this profile set to a higher version fails to resolve the recipe. Alas.

I think we'd be missing an opportunity not to upgrade to v5 at this point.

CMakeLists.txt
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/958/changes/c8a8eaa19fb240edb3c1a89090a843b6944f573f

Was 2 locations at the end of the day, the List one is the one that worries me.

profiles/host/emscripten
MAG-ElliotMorris: You were right, didn't need the buildenv. I was hoping I could remove the templating and _only_ have the buildenv, if the variables were set in environment variables from the emsdk activation, but no joy from my testing.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/958/changes/3788878344acfad4e640d4da4f7e011f3b939a86

MAG-ElliotMorris:
This PR informed me that the unity integrations I've setup don't work on PRs from forks. Have addressed that here : https://github.com/magnopus-opensource/connected-spaces-platform/pull/960

CMakeLists.txt
mag-lt: Thanks for that, this makes it much more clear where the problems are. Looks like we don't use the `Variant` type at all? The `List` one could probably use `std::move_backward`. Anyway, not something to address here.

profiles/host/emscripten
MAG-ElliotMorris: ... i might take yet another look at this this morning, just found this comment : https://github.com/conan-io/conan-center-index/pull/29620#issuecomment-3932076504

mag-lt: I linked to that same `conan-toolchains` repo in the second link I posted above. I'm still confused what this actually is to be honest. My best guess is you still have to manually pull some code from that remote repository (perhaps via some conan feature, not sure whether that repo can be specified in a config file somewhere), and then it probably has some code that is similar to what you've written in this PR. If you can get it working though, it does seem to be the closest thing to the official way of doing it.

MAG-ElliotMorris: Oh sorry, i didn't realize there was a distinction between that and conan center at the time. Will take another look. If it's just a matter of configuring another index, that might be doable.

PR #959, opened 2026-05-08
[OF-1831] build on linux

MAG-mv:
This PR allows the entire CSP solution to build on Linux. At this point, it has not been tested if it can run, which will be the next part of this work.

This PR ended up being quite small after moving the work to fix the changes-meaning errors in a later PR. This work consists of:

Adding GCC compile options to supress known errors
Adding platform specific Linux code
Fixing includes due to gcc having a different implementation
Fixing issues due to gcc being stricter
Adding conan profiles for Linux


MAG-ElliotMorris (approved):

mag-lt (approved):
For archival purposes, my review comments are in #956

PR #961, opened 2026-05-13
[OB-5368] fix: Resolve login crash caused by race conditions

MAG-AdamThorn:
The Unreal client team hitting an exception when attempting to do the following:

- A user enters a complex space with many Entities. Before these have finished loading the user calls their local LogOut method, which internally calls the CSP SpaceSystem::ExitSpace() followed by UserSystem::Logout() methods. Without waiting for their local Logout method to return they click Login again.

This resulted in a number of issues initially related to race conditions caused by multiple threads trying to access the LoginState object being passed to the Login, Logout and Refresh ResponseHandlers.

To address this I did the following: 

1. Introduced a sentinel `EntityFetchCancellationToken` and mutex lock `EntityFetchBodyGuard` to the `OnlineRealtimeEngine`.

2. Updated usage of `LoginState` within the `UserSystem`, so that whenever a `ResponseHandler` is being created for Login, Logout or RefreshToken refresh, we lock the object to avoid race conditions - we do this by wrapping it in a new `LoginStateLockable` object.

This resolved the issue in Unreal. However, after adding new tests in CSP to validate this behaviour I encountered a new use-after-free race condition that can occur if the `MultiplayerRealtimeEngine` is destroyed while either `OnElectedScopeLeader` or `OnVacatedScopeLeader` callbacks are in-flight. This issue was addressed with the addition of a new LeaderElection mutex and supporting logic.

### DIFF
--- a/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
+++ b/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
@@ -387,6 +387,10 @@ public:
     // We should remove this in OF-1785
     CSP_NO_EXPORT void SetServerSideElectionEnabled(bool Value);
 
+    // Mutex will be used in the MultiplayerConnection to guard against a use-after-free race condition that can occur in the ON_ELECTED_SCOPE_LEADER
+    // or ON_VACATED_AS_SCOPE_LEADER SignalR callbacks if the MultiplayerRealtimeEngine is freed by another thread while the callback is in-flight.
+    CSP_NO_EXPORT std::shared_ptr<std::mutex> GetLeaderElectionGuard() const { return LeaderElectionGuard; }
+
     /*
      * Called when MultiplayerConnection recieved signalR events.
      */
@@ -505,6 +509,10 @@ private:
     // Guards against a race condition whereby the OnlineRealtimeEngine dtor is called after the SignalR thread executing the
     // CreateRetrieveAllEntitiesCallback callback has checked the cancellation token, but BEFORE it has finished executing the callback.
     std::shared_ptr<std::mutex> EntityFetchBodyGuard;
+    // Mutex guards against use-after-free race condition that can occur in the ON_ELECTED_SCOPE_LEADER / ON_VACATED_AS_SCOPE_LEADER SignalR callbacks
+    // in the MultiplayerConnection. The SignalR thread may read MultiplayerRealtimeEngine as non-null only for the engine to be freed by another
+    // thread. The destructor waits to acquire this mutex (blocking until any in-flight callbacks finish) before proceeding with teardown.
+    std::shared_ptr<std::mutex> LeaderElectionGuard;
 
     std::deque<csp::multiplayer::SpaceEntity*>* PendingAdds;
     std::deque<csp::multiplayer::SpaceEntity*>* PendingRemoves;
--- a/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
+++ b/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
@@ -499,6 +499,13 @@ private:
     std::recursive_mutex* TickEntitiesLock;
     std::mutex LeadershipElectionLock;
 
+    // Represents the lifetime of the OnlineRealtimeEngine. Used to ensure the CreateRetrieveAllEntitiesCallback callback is not executed
+    // after engine is destroyed.
+    std::shared_ptr<std::atomic<bool>> EntityFetchCancellationToken;
+    // Guards against a race condition whereby the OnlineRealtimeEngine dtor is called after the SignalR thread executing the
+    // CreateRetrieveAllEntitiesCallback callback has checked the cancellation token, but BEFORE it has finished executing the callback.
+    std::shared_ptr<std::mutex> EntityFetchBodyGuard;
+
     std::deque<csp::multiplayer::SpaceEntity*>* PendingAdds;
     std::deque<csp::multiplayer::SpaceEntity*>* PendingRemoves;
     std::set<csp::multiplayer::SpaceEntity*>* PendingOutgoingUpdateUniqueSet;
--- a/Library/include/CSP/Systems/Users/Authentication.h
+++ b/Library/include/CSP/Systems/Users/Authentication.h
@@ -21,6 +21,11 @@
 #include "CSP/Common/String.h"
 #include "CSP/Systems/SystemsResult.h"
 
+CSP_START_IGNORE
+#include <memory>
+#include <mutex>
+CSP_END_IGNORE
+
 namespace csp::common
 {
 
@@ -42,6 +47,9 @@ namespace csp::systems
 
 class UserSystem;
 class LoginStateResult;
+CSP_START_IGNORE
+struct LoginStateLockable;
+CSP_END_IGNORE
 
 /// @brief Data for access and refresh tokens, and their expiry times.
 class CSP_API LoginTokenInfo
@@ -67,15 +75,18 @@ public:
     [[nodiscard]] const csp::common::LoginState& GetLoginState() const;
     CSP_NO_EXPORT LoginStateResult(csp::systems::EResultCode ResCode, uint16_t HttpResCode)
         : csp::systems::ResultBase(ResCode, HttpResCode)
-        , State(nullptr) {};
+        , State(nullptr) {}
 
 private:
     LoginStateResult();
-    LoginStateResult(csp::common::LoginState* InStatePtr);
+    CSP_START_IGNORE
+    LoginStateResult(LoginStateLockable* LoginStateLock);
+    CSP_END_IGNORE
 
     CSP_NO_EXPORT void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
 
-    csp::common::LoginState* State;
+    std::shared_ptr<csp::common::LoginState> State;
+    std::shared_ptr<std::mutex> StateMutex;
 };
 
 /// @ingroup User System
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -29,6 +29,10 @@
 #include "CSP/Systems/Users/Profile.h"
 #include "CSP/Systems/Users/ThirdPartyAuthentication.h"
 
+CSP_START_IGNORE
+#include <memory>
+CSP_END_IGNORE
+
 namespace csp::web
 {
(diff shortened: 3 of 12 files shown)


Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -499,6 +499,13 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
     std::recursive_mutex* TickEntitiesLock;
     std::mutex LeadershipElectionLock;
 
+    // Represents the lifetime of the OnlineRealtimeEngine. Used to ensure the CreateRetrieveAllEntitiesCallback callback is not executed
+    // after engine is destroyed.
+    std::shared_ptr<std::atomic<bool>> EntityFetchCancellationToken;
```
MAG-ElliotMorris, line 504: [nit] CancellationToken may not be the best naming, considering there are constructs in the ecosystem with [this name](https://learn.microsoft.com/en-us/cpp/parallel/concrt/reference/cancellation-token-class?view=msvc-170) and they tend to represent the ability to cancel a task, moreso than a lifetime sentinal. ... athough to be fair, I don't think we've been using the term `sentinal` right either. I wonder what the actual correct name for this construct is?

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -214,6 +218,14 @@ OnlineRealtimeEngine::OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerC
 
 OnlineRealtimeEngine::~OnlineRealtimeEngine()
 {
+    EntityFetchCancellationToken->store(true);
+    {
+        // This attempts to acquire the guard mutex. If the CreateRetrieveAllEntitiesCallback callback body is currently executing (the cancellation
+        // token was valid), it holds this mutex. If so the destructor blocks here until that callback finishes and then releases it. Once the
+        // destructor acquires the lock, it immediately releases it and continues.
+        std::scoped_lock lock(*EntityFetchBodyGuard);
+    }
```
MAG-ElliotMorris, line 227: [Comment] Must admit I'm a bit antsy about doing this sort of work in the destructor. There's nothing formally that prevents us blocking destruction, but it's a bit unexpectable imo. (especially if any of these async operations can fail and throw, which is UB).

My intuition would prefer doing all the guarding in the callbacks themselves, but there's nothing technically wrong with this and I do understand it prevents us having to deal with the online realtime engine being destroyed mid-callback, which would probably require larger refactors of the callbacks to achieve in the other way.

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -67,15 +75,18 @@ class CSP_API LoginStateResult : public ResultBase
...
     LoginStateResult();
-    LoginStateResult(csp::common::LoginState* InStatePtr);
+    CSP_START_IGNORE
+    LoginStateResult(LoginStateLockable* LoginStateLock);
+    CSP_END_IGNORE
 
     CSP_NO_EXPORT void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
 
-    csp::common::LoginState* State;
+    std::shared_ptr<csp::common::LoginState> State;
```
MAG-ElliotMorris, line 88: [comment] Surprised the wrapper generator tolerates this.

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -67,15 +75,18 @@ class CSP_API LoginStateResult : public ResultBase
...
+    std::shared_ptr<std::mutex> StateMutex;
```
MAG-ElliotMorris, line 89: [question] Why do these need to be shared ownership? Who else owns them?

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -323,7 +328,12 @@ class CSP_API UserSystem : public SystemBase
     csp::services::ApiBase* PingAPI;
     csp::services::ApiBase* StripeAPI;
 
-    csp::common::LoginState CurrentLoginState;
+    // Both LoginStateResult and LogoutResult OnResponse methods write directly to CurrentLoginState, while AuthContext::RefreshToken() reads from it.
+    // A mutex is required to ensure synchronisation between these operations and avoid race conditions.
+    // All writes to the LoginState happen under LoginStateMutex lock.
+    // shared_ptr ownership ensures the LoginState and mutex outlive any in-flight response handlers that hold copies.
+    std::shared_ptr<csp::common::LoginState> CurrentLoginState;
+    std::shared_ptr<std::mutex> LoginStateMutex;
```
MAG-ElliotMorris, line 336: [Consider] We duplicate this pattern a lot, and we also have this `LoginStateLockable` which is another duplication of this pattern.

Would it not be less error-prone to alter `LoginState` itself to inherently lock on all its operations, rather than introducing these wrappers?

Library/src/Multiplayer/MultiplayerConnection.cpp
```diff
@@ -176,6 +176,13 @@ MultiplayerConnection::~MultiplayerConnection()
...
 {
     MultiplayerRealtimeEngine = OnlineRealtimeEngine;
+
+    // Mutex guards against use-after-free race condition that can occur if the MultiplayerRealtimeEngine is destroyed while either
+    // OnElectedScopeLeader or OnVacatedScopeLeader callbacks are in-flight.
+    if (OnlineRealtimeEngine)
+        LeaderElectionGuardWeak = OnlineRealtimeEngine->GetLeaderElectionGuard();
+    else
+        LeaderElectionGuardWeak.reset();
 }
```
MAG-ElliotMorris, line 186: [Consider] Conceptually, is this not the same problem as before, the OnlineRealtimeEngine has been destroyed but we still have callbacks that might be in flight? Seems overly complicated to have specific guards for specific behaviours, rather than having a single `RealtimeEngineLifetime` token you can get a reference to.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -499,6 +499,13 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
...
+    // Guards against a race condition whereby the OnlineRealtimeEngine dtor is called after the SignalR thread executing the
+    // CreateRetrieveAllEntitiesCallback callback has checked the cancellation token, but BEFORE it has finished executing the callback.
+    std::shared_ptr<std::mutex> EntityFetchBodyGuard;
```
MAG-ElliotMorris, line 507: [Consider] Framing these around the engine lifetime, rather than just entity fetch. They represent a broader concept than just entity fetching. A subsequent implementor might think that these can only be used to guard entity fetch and end up appending unnecessary complexity.

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
MAG-ElliotMorris: ... You know actually the more I think about this, the more I don't mind it as a pattern. Although I don't understand why there's two locking primitives, you may have to explain it to me.

The idea that you may not be able to destruct a realtime engine when you want to still does fill me with some dread though ... 🤔

MAG-ElliotMorris (commented):
Good job hunting down the cause, and I believe this fixes the problem, however this is rather piecemeal and confusing to me.

There's a few things I think could be thought about:
- There seems to be a lot of shared_ptr use without subsequent weak_ptr use that I can see, which is the only real reason to do this sort of thing I'm aware of in this context.
- We're repeating this same pattern quite a lot rather than encapsulating it, most notably in the LoginState changes, it makes it tricky to follow what conceptual issue is being solved here. Smells a little bit like whack'a'mole.
- Which issues we're solving, are there 3 distinct ones? Online realtime engine lifetime in entity fetch, race conditions on data accesses to LoginState, and Online realtime engine lifetime on leader elections? Maybe individual PR's would have been better for changes at this level of intricacy, I'm finding it hard to map the patterns to the problems.

I'm probably missing at least some essential complexity, it smells like the double lock structures are something to do with TOCTOU windows, but am completely unsure about that. Happy to talk on this face to face if you like, I'm sure some of this could be explained to me.

mag-lt:
[drive by comment] I've not looked at this in detail, and don't intend to review this unless an extra pair of eyes are needed, but I did want to say that I can't help but feel we're missing a core abstraction here.

In previous projects, for situations where a callback is given to some subsystem that has a longer lifetime, and so needs cancellation, I've used something like this [`ThreadSafeFunctor`](https://github.com/Tracktion/choc/blame/main/choc/threading/choc_ThreadSafeFunctor.h) abstraction, which hides similar machinery inside the functor itself i.e. if the callback is cleared in a destructor, it is done in a thread safe way either null-ing it before being called, or blocking until it runs to completion.

This does involve storing the callbacks on the class before handing off, and explicitly clearing them in the destructor (this specific implementation I linked to doesn't automatically cancel in an RAII fashion). This might help here. It is a header only class, so you could try it out to see if it solves this more elegantly, depending on your appetite for doing so. This would still be quite piecemeal, we likely need to get a better handle on our threading and lifetimes in any case. I should emphasise that it is still up to the author of the callback to ensure any access to shared data is synchronised (that is not what the "thread safe"-ness is about here, it is about being able to cancel the callback safely when multiple threads are involved).

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -214,6 +218,14 @@ OnlineRealtimeEngine::OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerC
 
 OnlineRealtimeEngine::~OnlineRealtimeEngine()
 {
+    EntityFetchCancellationToken->store(true);
```
MAG-mv, line 221: [Question] Have we tested this on wasm? Won't this block the main thread and cause errors?

MAG-ElliotMorris: Ah feck, I forgot to think about that.

MAG-ElliotMorris: It may not even manifest in wasm because of the way the signalR threads work on that target (wasm tests appear to pass), but it's not good if it _could_ manifest and we're putting locks in without doing the thinking and making it clear what the branch point is.

MAG-mv: Spoke with Adam yesterday. It did manifest unfortunatly

PR #962, opened 2026-05-14
[OF-1850] Allow CSP to run on linux #1

MAG-mv:
This will be split into 2 PRs due to trouble fixing a script issue with ScopeLeadershipScriptEventTest.

We had 3 seperate issues that was preventing tests from running on Linux:
- Incorrect casing for test files
- Invalid List::Append implementation due to using std::memmove on non-triable types
- Assuming ordering of unordered_map iteration (We fixed some of these issues already when running on apple-clang

### DIFF
--- a/Library/include/CSP/Common/List.h
+++ b/Library/include/CSP/Common/List.h
@@ -53,7 +53,7 @@ inline size_t next_pow2(size_t val)
 /// This class is implemented using an array and, as such, removing items is not cheap as it requires
 /// us to move all items after it down one space.
 ///
-/// @tparam T : Object type to store in the list
+/// @tparam T : Object type to store in the list. T is required to be move-assignable.
 template <typename T> class CSP_API List
 {
 public:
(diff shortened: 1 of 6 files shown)


Library/include/CSP/Common/List.h
```diff
@@ -275,22 +275,16 @@ template <typename T> class CSP_API List
...
-#if defined(__clang__)
-#pragma clang diagnostic pop
-#endif
-        ++CurrentSize;
+        // Shift elements from the new insertion point 1 index to the right.
+        std::move_backward(
+            ObjectArray + Index,
+            ObjectArray + CurrentSize,
+            ObjectArray + CurrentSize + 1
+        );
```
MAG-ElliotMorris, line 288: [Consider] This looks like the right change (although i'm not thinking hard about it), but it's a very hot path and thus has strong regression potential. You should mention it explicitly in the changelog to give clients a clue if behaviour changes do pop up ... even positive ones.

Library/include/CSP/Common/List.h
```diff
@@ -53,7 +53,7 @@ inline size_t next_pow2(size_t val)
 /// This class is implemented using an array and, as such, removing items is not cheap as it requires
 /// us to move all items after it down one space.
 ///
-/// @tparam T : Object type to store in the list
+/// @tparam T : Object type to store in the list. T is required to be move-assignable.
```
mag-lt, line 56: [comment] in C++20 T could be constrained with concepts: `template <std::movable T> class List`. Though, I doubt the wrapper generator would be happy with that even if we were on a later version of the language.

[consider] still expressing this constraint statically by adding a `static_assert` about T in the class body:

```
#include <type_traits>

template <typename T> class CSP_API List
{
    static_assert(std::is_move_assignable_v<T>); // can optionally put a string explanation as to why as a second argument to static_assert
public:
/// ...
};
```

mag-lt (approved):
Looks good. PR checks failing due to lack of changelog entry (or explicitly skipping via the label). It is probably worth adding something, even if this isn't the full story for Linux yet.

MAG-AdamThorn (approved):

Library/include/CSP/Common/List.h
MAG-mv: Added to changelog: https://github.com/magnopus-opensource/connected-spaces-platform/pull/962/commits/8fd02d3023ef065cc854dad178bfdc3b56d033df

Library/include/CSP/Common/List.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/962/commits/f76c9c9d28e242352a6a3d72e26606d709ec9b45

mag-lt: Looks like the wrapper gen doesn't like it. Thanks for trying, but the comment will have to do!

MAG-mv: Trying again with CSP_START_IGNORE/CSP_END_IGNORE

MAG-mv: All working!

PR #963, opened 2026-05-15
[Experiment] Apply variable formatting rules.

MAG-ElliotMorris:
I don't really intend to merge this, it's probably mess up the build system and interop work. I just want to see what sort of impact applying clang-tidy with new formatting rules across the codebase might have. Didn't take too long, a few hours, but I also havn't looked at the output almost at all (look mum, I'm vibe coding 😆 )


MAG-ElliotMorris:
This build locally, but I would be surprised if any of the CI builds actually pass. Not going to invest time right now into looking into it.

MAG-ElliotMorris:
**Final results**: The simple builds using the new build system passes, and the unity bindings are unaffected. 

Other platforms using the legacy build have minor errors in various places, certain files that are included on that platform that have some variable shadowing that needs to be dealt with. Also there is some test code that is erroring for similar reasons, potentially some interference with gtest naming. All seems rather minor though. The unity bindings prove that this changeset is non-breaking for the API expressed in the unity-bindings tests (non exhaustive, but a decent chunk). Softly proves that the entire api surface change is non-breaking because it _did_ build, but i'd want to look at the actual output files because SWIG can occasionally generate some nonsense that you don't see until runtime ... although I think this is mostly me being paranoid.

PR #964, opened 2026-05-15
[OF-1850] feat: support fPIC for Linux shared builds

MAG-mv:
fPIC needs to be explicitly set for Linux when building as shared. I followed the Conan docs to create an option for this. I also passed this to cmake to ensure the cmake toolchain uses this variable.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -7,14 +7,12 @@ class CSP(ConanFile):
 
     exports_sources = "CMakeLists.txt", "Library/*", "Tests/*", "cmake/*"
 
-    options = {
-        "fPIC": [True, False],
-    }
-
     default_options = {
-        # Set fPIC to true by default which is pretty standard.
-        # This isn't always needed but we don't want to shift the responsibility of setting it to callers.
-        "fPIC": True,
+        # We currently don't support shared dependencies.
+        "*:shared": False,
+
+        # Our dependencies are always linked statically, so force PIC.
+        "*:fPIC": True,
     }
 
     def layout(self):
@@ -81,10 +79,6 @@ class CSP(ConanFile):
         deps.generate()
 
         tc = CMakeToolchain(self)
-
-        # Ensure our fPIC option is passed to cmake
-        if self.settings.os != "Windows":
-            tc.variables["CMAKE_POSITION_INDEPENDENT_CODE"] = bool(self.options.fPIC)
 
         # Generate conan presets file
         tc.user_presets_path = 'ConanPresets.json'
--- a/conanfile.py
+++ b/conanfile.py
@@ -7,6 +7,16 @@ class CSP(ConanFile):
 
     exports_sources = "CMakeLists.txt", "Library/*", "Tests/*", "cmake/*"
 
+    options = {
+        "fPIC": [True, False],
+    }
+
+    default_options = {
+        # Set fPIC to true by default which is pretty standard.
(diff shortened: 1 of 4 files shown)


conanfile.py
```diff
@@ -7,6 +7,16 @@ class CSP(ConanFile):
...
 
+    options = {
+        "fPIC": [True, False],
+    }
+
+    default_options = {
+        # Set fPIC to true by default which is pretty standard.
+        # This isn't always needed but we don't want to shift the responsibility of setting it to callers.
+        "fPIC": True,
+    }
```
MAG-ElliotMorris, line 16: [Question] You state in the PR message : 

> fPIC needs to be explicitly set for Linux when building as shared.

However, the [conan docs example](https://docs.conan.io/2/tutorial/creating_packages/configure_options_settings.html) explicitly removes it from shared libs.

My belief is that this is because fPIC _should_ be implicit for shared libs. If you're seeing problems, I think that probably comes back to the whole not using the standard CMake variable for shared libs, creating unfortunate issues like this. (Although i'm not entirely sure that we didn't address this).

The conan docs in the above page state : 

> configure(): use this method to configure which options or settings of the recipe are available. For example, in this case, we delete the fPIC option, because it should only be True if we are building the library as shared (in fact, some build systems will add this flag automatically when building a shared library).

But then there's other docs, such as [here](https://docs.conan.io/2/reference/conanfile/methods/configure.html)

> Likewise, for a package containing a library, the fPIC option really only applies when the library is compiled as a static library, but otherwise, the fPIC option doesn’t make sense, so it should be removed

My reading of this is that this is about not being able to know if the package we're making is going to be compiled into a static or shared library by the consumer, so you need to express an FPIC option _only in the static case_, as that is the only case where FPIC could even conceivably be false. In shared mode, it must always be true (and will always be true, enforced by the build system, if we were using the correct flags).

Is my reading of this situation correct? I'm rather confused.

MAG-AdamThorn (approved):

conanfile.py
MAG-mv: It is very confusing. We're always linking our dependencies statically, so it should always be true. This isn't about whether we are linking csp via shared or static

>My belief is that this is because fPIC should be implicit for shared libs. If you're seeing problems, I think that probably comes back to the whole not using the standard CMake variable for shared libs

It's about us always linking our deps statically, not the cmake variables

MAG-ElliotMorris: Right, so this is an option we're passing into the build of our dependencies, not into CSP itself? That makes sense, it is the conanfile afterall.

MAG-ElliotMorris: I wonder if providing the option is necessary at all then? Must we not always have position independent code with our build strategy?

MAG-ElliotMorris: Approving because i'm less worried now I understand this is confined to dependency fetching, although worth considering the above point, and maybe making it clearer in the file that we always link our deps statically, and what the PIC implication is. "which is pretty standard." is a pretty poor explanation. Resolve conversation at your leisure.

MAG-ElliotMorris (approved):

conanfile.py
MAG-mv: I changed this a bit based on your comments and some additional research:

https://github.com/magnopus-opensource/connected-spaces-platform/pull/964/commits/86074c4c1339377a64cd65e18376872f21bb8744

PR #965, opened 2026-05-18
[OF-1850] feat: Add release with debug info presets

MAG-mv:
Adding additional single-config confgure presets + single and multi config build presets

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -75,6 +75,34 @@
         "CSP_BUILD_SHARED": "ON",
         "BUILD_TESTING": "OFF"
       }
+    },
+
+    {
+      "name": "relwithdebinfo-static",
+      "displayName": "single-config release with debug info static",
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -75,6 +75,34 @@
         "CSP_BUILD_SHARED": "ON",
         "BUILD_TESTING": "OFF"
       }
+    },
+
```
mag-lt, line 79: [Nit] Whitespace. Is this accidental, or intended to visually mark a new conceptual group or something (I see this has been done elsewhere in the file)?

mag-lt (approved):

CMakePresets.json
MAG-mv: its intended to seperate the configure preset types

MAG-AdamThorn (approved):

PR #967, opened 2026-05-18
[OF-1866] feat: Add public API for interacting with Components defined w/ Schema

mag-lt:
This adds the following public methods:
  - `IRealtimeEngine::GetComponentSchemaRegistry()`:
    - Returns a "registry" instance, a read-only type describing the Components known by the engine. This can be used to interrogate the shape of a Component at runtime. Currently this is populated by CSP with a corresponding Schema for each of the hardcoded Component types in the library. It is still not possible to register a new schema externally, but this will come in a future release.
  - `SpaceEntity::AddComponentByTypeId(uint64_t TypeId)`:
    - Allows adding a component that corresponds to a Schema with the given TypeId.
  - `ComponentBase::GetTypeId()`:
    - This returns the unique integer ID representing the component type. For traditional hardcoded components, this corresponds to casting the `ComponentType` to its underlying integer value. This supersedes `GetComponentType` in that it allows for representing an open vs closed set of components. Details of allocating these IDs in a unique fashion is still to be defined.
  - `ComponentBase::GetSchemaProperty(uint16_t Key)`:
    - Get the value of a property corresponding to the one declared with the given Key in the Schema. The Key must be in the Schema, or null is returned. This supersedes the hardcoded wrapper methods typically written, trading off static typing for the flexibility of defining components externally.
  - `ComponentBase::SetSchemaProperty(uint16_t Key, const csp::common::ReplicatedValue& Value)`:
    - Set the value of a property corresponding to the one declared with the given Key in the Schema. The Key must be in the Schema, and the value must match the type declared there or it is ignored. This supersedes the hardcoded wrapper methods typically written, trading off static typing for the flexibility of defining components externally.

The existing methods and concrete component classes are preserved for now, and not yet considered formally deprecated.

### DIFF
--- a/Library/include/CSP/Multiplayer/ComponentBase.h
+++ b/Library/include/CSP/Multiplayer/ComponentBase.h
@@ -183,6 +183,14 @@ public:
     // Used for handling behavior when a client first deletes the component.
     CSP_NO_EXPORT virtual void OnLocalDelete();
 
+    /// @brief Get a property value by schema key.
+    /// @return A pointer to the value, or nullptr if the key is not in this component's schema.
+    const csp::common::ReplicatedValue* GetSchemaProperty(uint16_t Key) const;
+
+    /// @brief Set a property value by schema key.
+    /// Silently ignored if the key is absent from this component's schema or the value type does not match the schema definition.
+    void SetSchemaProperty(uint16_t Key, const csp::common::ReplicatedValue& Value);
+
 protected:
     ComponentBase();
     ComponentBase(uint64_t TypeId, csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
@@ -228,6 +236,8 @@ protected:
     csp::common::LogSystem* LogSystem = nullptr;
 
     csp::common::Map<csp::common::String, EntityActionHandler> ActionMap;
+
+    std::unique_ptr<ComponentSchema> CachedSchema;
 
 private:
     void InitialiseProperties();
--- a/Library/include/CSP/Multiplayer/ComponentBase.h
+++ b/Library/include/CSP/Multiplayer/ComponentBase.h
@@ -129,6 +129,17 @@ public:
     /// @return The type of the component as an enum.
     ComponentType GetComponentType() const;
 
+    /// @brief Get the TypeId of the component.
+    /// @return A unique integer ID representing the component type.
+    ///
+    /// @note This method is conceptually similar to GetComponentType(), but handles an 
+    ///       open set of IDs. For built-in, statically defined components, this ID is 
+    ///       guaranteed to be the exact integer cast of its ComponentType enum value. 
+    ///       For dynamically registered components, it returns the ID specified by its schema.
+    ///
+    /// @see GetComponentType()
+    uint64_t GetTypeId() const;
+
     /// @brief Get a map of the replicated values defined for this component.
     ///
     /// The index of the map represents a unique index for the property,
@@ -174,6 +185,7 @@ public:
 
 protected:
     ComponentBase();
+    ComponentBase(uint64_t TypeId, csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
 
     const csp::common::ReplicatedValue& GetProperty(uint32_t Key) const;
     bool GetBooleanProperty(uint32_t Key) const;
@@ -206,7 +218,7 @@ protected:
 
     SpaceEntity* Parent;
     uint16_t Id;
-    ComponentType Type;
+    uint64_t Type;
     csp::common::Map<uint32_t, csp::common::ReplicatedValue> Properties;
     csp::common::Map<uint32_t, csp::common::ReplicatedValue> DirtyProperties;
 
--- a/Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
+++ b/Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
@@ -16,23 +16,25 @@
 
 #pragma once
 
-#include "CSP/Multiplayer/IComponentSchemaRegistry.h"
-
-#include <unordered_map>
+#include "CSP/CSPCommon.h"
+#include "CSP/Common/Array.h"
+#include "CSP/Multiplayer/ComponentSchema.h"
 
 namespace csp::multiplayer
 {
 
-class ComponentSchemaRegistryImpl final : public IComponentSchemaRegistry
+/// @brief Read-only interface for querying registered component schemas.
+class CSP_API IComponentSchemaRegistry
 {
 public:
-    ComponentSchemaRegistryImpl(const csp::common::Array<ComponentSchema>& AdditionalComponents);
+    virtual ~IComponentSchemaRegistry() = default;
 
-    csp::common::Array<ComponentSchema> GetAll() const override;
-    const ComponentSchema* Find(uint64_t TypeId) const override;
+    /// @brief Returns all registered schemas.
+    virtual csp::common::Array<ComponentSchema> GetAll() const = 0;
 
-private:
-    std::unordered_map<ComponentSchema::TypeIdType, ComponentSchema> SchemaMap;
+    /// @brief Finds the schema for the given TypeId.
+    /// @return A pointer to the schema if found, otherwise nullptr.
+    virtual const ComponentSchema* Find(uint64_t TypeId) const = 0;
 };
 
 } // namespace csp::multiplayer
--- a/Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
+++ b/Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
@@ -35,6 +35,13 @@ public:
     /// @brief Finds the schema for the given TypeId.
     /// @return A pointer to the schema if found, otherwise nullptr.
     virtual const ComponentSchema* Find(uint64_t TypeId) const = 0;
+
+protected:
+    CSP_NO_EXPORT IComponentSchemaRegistry() = default;
+    IComponentSchemaRegistry(const IComponentSchemaRegistry&) = delete;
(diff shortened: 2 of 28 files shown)


Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
```diff
@@ -0,0 +1,40 @@
...
+#include "CSP/Multiplayer/ComponentSchema.h"
+
+namespace csp::multiplayer
+{
+
+/// @brief Read-only interface for querying registered component schemas.
+class CSP_API IComponentSchemaRegistry
+{
+public:
+    virtual ~IComponentSchemaRegistry() = default;
```
MAG-mv, line 30: [Consider] Rule of 5 for interfaces for safety against slicing

MAG-mv: you may have to wrap in CSP_START/END_IGNORE

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -501,9 +504,30 @@ uint16_t SpaceEntity::GenerateComponentId()
     }
 }
 
-ComponentBase* SpaceEntity::InstantiateComponent(uint16_t InstantiateId, ComponentType InstantiateType)
+ComponentBase* SpaceEntity::InstantiateComponent(uint16_t InstantiateId, uint64_t TypeId)
 {
+    if (!IsLegacyComponentTypeId(TypeId))
+    {
+        const auto* Registry = EntitySystem != nullptr ? EntitySystem->GetComponentSchemaRegistry() : nullptr;
```
MAG-mv, line 512: [Comment] I'd love to remove the EntitySystem from the SpaceEntity in the future, and do these kind of operations on the RealtimeEngine itself

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
```diff
@@ -45,46 +47,59 @@
...
 
-ComponentSchemaRegistry MergeWithLegacyComponents(const csp::common::Array<ComponentSchema>& AdditionalComponents)
+ComponentSchemaRegistryImpl::ComponentSchemaRegistryImpl(const csp::common::Array<ComponentSchema>& AdditionalComponents)
 {
-    const auto ToPair = [](const ComponentSchema& Schema) {
-        return std::make_pair(Schema.TypeId, Schema);
-    };
+    for (const auto& Schema : AdditionalComponents)
+    {
+        SchemaMap[Schema.TypeId] = Schema;
```
MAG-AdamThorn: [question] are we concerned about there being conflict between the Id of a default component schema and one of these additional components? Or indeed a conflict between the Ids of two additional components.

MAG-AdamThorn: Feels like there would be value in alerting the user to such a conflict.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -373,7 +376,7 @@ ComponentBase* SpaceEntity::AddComponent(ComponentType AddType)
     }
 
     auto ComponentId = GenerateComponentId();
-    auto* Component = InstantiateComponent(ComponentId, AddType);
+    auto* Component = InstantiateComponent(ComponentId, TypeId);
 
     // If Component is null, component has not been instantiated, so is skipped. (Can this ever be null... seems a bit of a footgun not to just
```
MAG-AdamThorn: [fix] this comment should be updated as the component can now definitely be null.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -501,9 +504,30 @@ uint16_t SpaceEntity::GenerateComponentId()
...
+        if (Schema == nullptr)
+        {
+            if (LogSystem != nullptr)
+            {
+                LogSystem->LogMsg(csp::common::LogLevel::Warning, fmt::format("Unknown Component TypeId: {}", TypeId).c_str());
+            }
+            return nullptr;
+        }
+
+        auto* Component = new ComponentBase(*Schema, LogSystem, this);
```
MAG-AdamThorn, line 525: [question] should we not guard against the provided `InstantiateId` not already being in the registry?

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
mag-lt: The order of registration is flipped here vs before so that built-in components take precedence, which is something you flagged in the last PR when the registry was introduced.

ID clashes can theoretically happen. If we wanted to produce a failure here, that would ultimately need to result in constructing the Engine to fail (i.e. throw), which I'd rather not do.

 I'll add a log message, but ultimately the behaviour will be that the last write wins.

Library/include/CSP/Multiplayer/ComponentBase.h
```diff
@@ -229,6 +237,8 @@ class CSP_API ComponentBase
 
     csp::common::Map<csp::common::String, EntityActionHandler> ActionMap;
 
+    std::unique_ptr<ComponentSchema> CachedSchema;
```
MAG-mv, line 240: [Consider]  This isn't "wrong", but the word "cached" feels slightly off here considering the schema is the immutable definition of the component, even though I understand they are defined somewhere else.

MAG-mv (approved):
Looks really good. Just a couple of considers

Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
mag-lt: This is a pure abstract class / interface type (only pure virtual methods). The virtual destructor here is dealing with standard polymorphism concerns i.e. ensuring a derived class is properly deleted when all they have is a pointer to the base, but anything else I've typically seen considered a concern for the concrete class (i.e. rule of 5 stuff), and so can address on `ComponentSchemaRegistryImpl`.

Unrelated but here for completeness: from an ABI point of view, the virtual destructor is technically not portable / stable across compilers, but as we aren't attempting to have a stable ABI at the moment, that isn't a concern here (and obviously this isn't the only class with a virtual destructor on an API type).

I've tested this out locally (before opening the PR) and the wrapper gen appears to be fine without CSP_START/END_IGNORE, I can interact with the registery in both generated C# and JavaScript targets.

Library/include/CSP/Multiplayer/ComponentBase.h
mag-lt: This is one of these things were a convention of naming member variables would help (i.e. `m_` or otherwise). The first name for this was `Schema`, but that clashes in derived classes with the current pattern for declaring the static schema within those cpp files.

Library/src/Multiplayer/SpaceEntity.cpp
mag-lt: I was confused by the naming of `EntitySystem`, as really this _is_ the `Engine`, I assume it is a legacy remnant before `IRealtimeEngine` was introduced. The registry could be passed via constructor injection here vs the whole engine, but whilst the class uses the engine for other things, we get the registry from there. The nullptr check is also a bit pathological, really we should be holding `EntitySystem` by reference, but I'm accepting what is here for now.

Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
MAG-mv: Explicitly blocking copys and moves into an abstract class is still useful to prevent slicing. I was only the impression its recommended. Not a mssive issue either way.

Library/src/Multiplayer/SpaceEntity.cpp
MAG-mv: Yeah, RealtimeEngine used to be called SpaceEntitySystem

MAG-AdamThorn (changes requested):

Library/src/Multiplayer/SpaceEntity.cpp
mag-lt: As in, what happens if the registry doesn't contain schema for that ID? That is handled immediately above [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/53342a7f3676f011ec2146fc449d49f7ad7d4012/Library/src/Multiplayer/SpaceEntity.cpp#L521)? Is there another case I'm missing?

mag-lt: `InstantiateId` is a different ID entirely, which is relative to the entity.

Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
mag-lt: Addressed in f738f609a05c093fe932b9a098ef5eaf8eac0411

Library/src/Multiplayer/SpaceEntity.cpp
mag-lt: Hopefully addressed with 7fe9fccea5ec0d35a0f116a20416dba86f53f0eb

mag-lt:
Heads up @MAG-AdamThorn and @MAG-mv, in addition to the commits addressing the PR feedback, there are two new unreviewed commits: ac6833bc123edcad8575291bda8858b32d28816a, 02994411a99f5a170783e8c4964812bcd3d308ec. I moved over some of my recent tests (introduced in my previous script bindings PR) to use these new API methods, and in doing so it flagged a bug in ReplicatedValue.

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
mag-lt: Log added in cf71f980b0d2bd49bb26c9ef336b2c018201a391

MAG-AdamThorn (approved):

PR #968, opened 2026-05-20
OB-2975: Invalid Mutex Reference Causing Unity Crashes

MAG-SangeetShah:
### Ticket

https://magnopus.atlassian.net/browse/OB-2975

NOTE: I am not sure how complete this fix is and so have not added this ticket to the sprint. If we feel this does not need to be revisited, I will move the ticket accordingly.

### Summary Of Changes

- Set MultiplayerConnection & webClient pointers to null when the connection is closed
- Add Shutdown tests to verify
- Update Changelog

### Testing

Locally build and tested the change in Unity Editor. Crashes no longer occur on space exit

### DIFF
--- a/Library/src/Systems/SystemsManager.cpp
+++ b/Library/src/Systems/SystemsManager.cpp
@@ -245,10 +245,22 @@ void SystemsManager::DestroySystems()
     delete SpaceSystem;
     delete UserSystem;
     delete VoipSystem;
-    delete MultiplayerConnection; // Also deletes NetworkEventBus
     delete ScriptSystem;
     delete MultiplayerSystem;
+
+#ifdef CSP_WASM
+    delete MultiplayerConnection;
     delete WebClient;
+#else
+    // INTENTIONALLY LEAK MultiplayerConnection and WebClient.
+    // SignalR's detached scheduler thread and WebClient's ThreadPool workers may still
+    // reference their members (mutexes, etc.) after destruction. Leaking avoids
+    // "mutex lock failed: Invalid argument" crashes on shutdown.
+    // The OS reclaims all memory when the process exits.
+    MultiplayerConnection = nullptr;
+    WebClient = nullptr;
+#endif
+
     delete LogSystem;
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ All notable changes to this project will be documented in this file. For compile
 - [OW-2438] feat: Properties to support cinematic camera depth of field. By @MAG-ThomasGreenhalgh.
   Adds focus distance and depth of field enabled properties to the cinematic camera space component.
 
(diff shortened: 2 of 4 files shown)


Library/src/Systems/SystemsManager.cpp
```diff
@@ -245,10 +245,22 @@ void SystemsManager::DestroySystems()
...
     delete WebClient;
+#else
+    // INTENTIONALLY LEAK MultiplayerConnection and WebClient.
+    // SignalR's detached scheduler thread and WebClient's ThreadPool workers may still
+    // reference their members (mutexes, etc.) after destruction. Leaking avoids
+    // "mutex lock failed: Invalid argument" crashes on shutdown.
+    // The OS reclaims all memory when the process exits.
+    MultiplayerConnection = nullptr;
+    WebClient = nullptr;
+#endif
```
MAG-ElliotMorris, line 262: [Comment] I'm afraid you can't do this. Unreal shuts down CSP without exiting the process, on each space exit. This will cause Unreal to leak staggering amounts of data. 

This also causes the MultiplayerConnection and WebClient destructors not to run, which I imagine will cause regressions in space exit flows, but more vitally in websocket handshakes. It's possible this just completely breaks Unreal second space entry.

I am sorry, I know this behaviour is frustrating and we've been encouraging self-serve more and more, but this is a very significant change that we can't just make without understanding the consequences. In that vein, I'd also want to be pointed to specifically which members are referenced and where, rather than the overarching gesture at "mutexes, etc".

It's possible, although I don't like the idea, of doing something like this as an _extremely temporary_ measure, but we'll need to perform more testing on unreal, as well as get their permission to leak like this. 

We are pivoting to look at these issues now that clients seem to care about them (I'm confused why they've all started all of a sudden, these aren't regressions as far as I can tell, but 🤷)

MAG-SangeetShah: Fair enough. I was expecting such an answer as I don't have the full context as to how deeply this affects the other clients but thought I would give it a go.

I did test shutdown points in Unity quite extensively (repeated shutdowns, shutdowns with different requests in flight etc.) and it does fix the crashes on our side so hopefully this points the CSP team in the right direction when you do get to the ticket.

It is not worth adding this fix in if you feel this is an extreme measure. I would rather we manage it through priorities.

Thanks for indulging me regardless. I will close this PR but leave the branch ready for the ticket to be picked up.

MAG-ElliotMorris: If you care to do the work to understand the unreal side impacts of intentional leaking and non-destruction like this, and really need to fix these crashes very immediately whilst we work on the correct changes, it's still a possibility. Although I would favour solving this correctly. We're going to speak about if and how we can accelerate dealing with this in our backlog meeting tommorow.

PR #969, opened 2026-05-21
[OF-1071] feat: Add support for third party auth with token

MAG-AdamThorn:
Add support for third party authentication using a token provided by the third party provider.

This is an alternate path for SSO to the one already provided, in that it expects the application to request the authorization token from the third party provider directly before passing it to CSP.

A test has been added, but as noted in `UserSystemTests` this has to be run manually as it requires an SSO token from a third party provider to be stored in a text file in a specific location.

Results of the test run shown below:

image

### DIFF
--- a/Library/src/Systems/Users/UserSystem.cpp
+++ b/Library/src/Systems/Users/UserSystem.cpp
@@ -123,6 +123,43 @@ std::optional<csp::common::String> ThirdPartyPlatformToString(const csp::common:
     return std::nullopt;
 }
 
+// Contains shared logic for logging in with a 3rd party provider - used by both LoginToThirdPartyAuthenticationProvider and
+// LoginToThirdPartyAuthenticationProviderWithToken.
+void LoginSocial(csp::services::ApiBase* AuthenticationAPI, csp::common::LoginState* CurrentLoginState, csp::common::LogSystem* LogSystem,
+    const std::shared_ptr<chs_user::LoginSocialRequest>& Request, bool CreateMultiplayerConnection, csp::systems::LoginStateResultCallback Callback)
+{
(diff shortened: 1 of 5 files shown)


MAG-ElliotMorris (approved):
LGTM. Decent idea committing the manual test code as docs.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -123,6 +123,43 @@ std::optional<csp::common::String> ThirdPartyPlatformToString(const csp::common:
...
 }
 
+// Contains shared logic for logging in with a 3rd party provider - used by both LoginToThirdPartyAuthenticationProvider and
+// LoginToThirdPartyAuthenticationProviderWithToken.
+void LoginSocial(csp::services::ApiBase* AuthenticationAPI, csp::common::LoginState* CurrentLoginState, csp::common::LogSystem* LogSystem,
+    const std::shared_ptr<chs_user::LoginSocialRequest>& Request, bool CreateMultiplayerConnection, csp::systems::LoginStateResultCallback Callback)
+{
+    csp::systems::LoginStateResultCallback LoginStateResCallback = [=](const csp::systems::LoginStateResult& LoginStateRes)
+    {
+        if (LoginStateRes.GetResultCode() == csp::systems::EResultCode::Success)
```
MAG-mv, line 133: [Comment] This isn't an issue now and I'm not asking you to fix it. Juts sharing knowledge.

This will compile error when we move to c++20, as implicit capture of this via [=] is deprecated.
We will have to start doing [=, this].

Don't worry about this now, there are many instances in the codebase we are already doing this. It'll be quicker to fix these in one go.

MAG-mv (approved):
Looks good

PR #970, opened 2026-05-27
[OB-5440] fix: Lock in Send (Test)

MAG-ElliotMorris:
Lock in POCO Send, like we lock in Stop. Also account for potential deadlocks with stop, or the socket not having been created yet. I think the assert was trying to do that, but obviously that's not a syncronization primitive.

Just running this as a test to see what CI does.

### DIFF
--- a/Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
+++ b/Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
@@ -178,32 +178,47 @@ void CSPWebSocketClientPOCO::Send(const std::string& Message, CallbackHandler Ca
 {
     CSP_PROFILE_SCOPED();
 
-    assert(PocoWebSocket && "Web socket not created! Please call Start() before calling Send().");
+    auto Succeeded = false;
 
-    int Flags = Poco::Net::WebSocket::SendFlags::FRAME_BINARY; // Assume binary as we don't support JSON anymore
-    auto Remaining = Message.size();
-    auto Succeeded = true;
-
-    try
     {
-        while (Remaining > 0)
+        std::lock_guard<std::mutex> Lock(Mutex);
+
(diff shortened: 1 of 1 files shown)


Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
MAG-AdamThorn, line 137: [comment] I am not sure the below fix will work. The comment here states that the Mutex is intentionally being unlocked to avoid a ReceiveThread deadlock. I think Send needs it's own dedicated lock and then stop can wait for that as it does the receive thread before proceeding.

Library/src/Multiplayer/SignalR/POCOSignalRClient/POCOSignalRClient.cpp
```diff
@@ -178,32 +178,47 @@ void CSPWebSocketClientPOCO::Send(const std::string& Message, CallbackHandler Ca
...
+    auto Succeeded = false;
 
-    int Flags = Poco::Net::WebSocket::SendFlags::FRAME_BINARY; // Assume binary as we don't support JSON anymore
-    auto Remaining = Message.size();
-    auto Succeeded = true;
-
-    try
     {
-        while (Remaining > 0)
+        std::lock_guard<std::mutex> Lock(Mutex);
```
MAG-AdamThorn, line 184: Another issue is that if a thread calls Stop it will lock the mutex and then unlock again while waiting for the receive thread. It then doesn't check the lock again before tearing down the socket. If a send is sent while waiting for the receive thread, the socket could still be destroyed before the send has completed.

MAG-ElliotMorris:
Agree with Adams points above. Threw this up to see if CI panicked with an extra lock. This won't be the final solve.

MAG-ElliotMorris:
... and teamcity is refusing to run these tests. Whatever. Closing. Adams going to look at this properly.

PR #971, opened 2026-05-27
[OB-5464] chore: Remove legacy identifier type parsing.

MAG-ElliotMorris:
Made this change right in the web editor :P No doubt CI will tell me of tests that need updated.

Will merge this once Jake confirms the migration is done. No need to handle this backwards compatibility anymore as CHS has purged all the data of this format.

### DIFF
--- a/Library/src/Systems/Settings/SettingsSystem.cpp
+++ b/Library/src/Systems/Settings/SettingsSystem.cpp
@@ -886,22 +886,10 @@ void SettingsSystem::GetAvatarInfo(AvatarInfoResultCallback Callback)
             CSP_LOG_ERROR_MSG("Whilst avatar info was successfully returned with data, no avatar type was included.");
         }
 
-        // Avatar identifier type (used to identify or locate the actual avatar asset)
+        // Avatar identifier (used to identify or locate the actual avatar asset)
         if (Json.HasMember("identifier"))
         {
-            if (Json.HasMember("identifierType") && static_cast<VariantType>(Json["identifierType"].GetInt()) == VariantType::Integer)
-            {
-                // Integer type is no longer supported -- convert to string
-                InternalResult.SetAvatarIdentifier(std::to_string(Json["identifier"].GetInt()).c_str());
-            }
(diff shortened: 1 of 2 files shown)


MAG-ElliotMorris:
[This](https://magnopus.atlassian.net/browse/OW-2491) confirms that the migration has been performed.

MAG-ElliotMorris:
Have got a confirmation that the migration works, so making this ready to go.

Apologies for the weird merge commit in the history. Tried to do an in-browser changelog fix and made a mistake.

Appears that there were no tests for this backwards compat branch, so no tests for its removal either.

Library/src/Systems/Settings/SettingsSystem.cpp
```diff
@@ -889,22 +889,10 @@ void SettingsSystem::GetAvatarInfo(AvatarInfoResultCallback Callback)
             CSP_LOG_ERROR_MSG("Whilst avatar info was successfully returned with data, no avatar type was included.");
         }
 
-        // Avatar identifier type (used to identify or locate the actual avatar asset)
+        // Avatar identifier (used to identify or locate the actual avatar asset)
         if (Json.HasMember("identifier"))
```
mag-lt: [nit] I'd personally also check the type here (i.e. `&& Json["identifier"].IsString()`, as done for the `type` and `avatarVisible` fields either side of this), as RapidJSON will not check on access in release builds. However, as we control all the data here arguably we don't have to be so defensive?

mag-lt (approved):

Library/src/Systems/Settings/SettingsSystem.cpp
MAG-ElliotMorris: Sounds reasonable. We lose nothing and gain a bit of protection in certain modes. Will do it.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/971/changes/c4cbe7b9c725b4732143f3c04b205d8595de55bf

MAG-AdamThorn (approved):

PR #972, opened 2026-05-28
[OB-5368] fix!: Guard access to login state data via a mutex

MAG-AdamThorn:
The Login, Logout and Refresh Token response handlers access the `LoginState` object on different threads which was resulting in login state data race conditions. In addition the `LoginState` object was being passed to the response handlers by value, which resulted in the `LoginState` being destroyed when CSP was torn down while callbacks were still in flight.

To resolve this issue the `LoginState` data members have been moved to a new `LoginStateData` class, and access to it is controlled via a mutex guard. In addition a copy of the `LoginState` shared_ptr is now captured by the response handler callbacks to ensure it's lifetime persists for the duration of the callback.

**Breaking change**
The `LoginState` class now provides a new api for accessing the underlying data. While before you could get/set members such as `State` and `UserId` directly, now you you have to go through the dedicated getters. These are:

- GetUserId()
- GetDeviceId()
- GetAccessToken()
- GetRefreshToken()
- GetLoginStateValue()
- GetDefaultApplicationSettings()
- GetDefaultSettings()

### DIFF
--- a/Library/include/CSP/Common/LoginState.h
+++ b/Library/include/CSP/Common/LoginState.h
@@ -41,10 +41,10 @@ public:
     LoginState(const LoginState& OtherState);
     LoginState& operator=(const LoginState& OtherState);
 
-    /// @brief Get a copy of the current login state data.
+    /// @brief Get a thread-safe copy of the current login state data.
     /// This is a snapshot of the data at the time of the call, and will not update if the underlying state is mutated.
     /// @return A snapshot of the login state data.
-    CSP_NO_EXPORT LoginStateData GetSnapshot() const;
+    CSP_NO_EXPORT LoginStateData GetSnapshotThreadSafe() const;
 
     /// @brief Set the login state data. This will replace all existing data in the login state with the new data provided.
     CSP_NO_EXPORT void SetLoginStateData(csp::common::LoginStateData NewData);
--- a/Library/include/CSP/Common/LoginState.h
+++ b/Library/include/CSP/Common/LoginState.h
@@ -37,7 +37,7 @@ class CSP_API LoginState
 public:
     LoginState();
     ~LoginState();
-    
+
     LoginState(const LoginState& OtherState);
     LoginState& operator=(const LoginState& OtherState);
 
@@ -46,8 +46,9 @@ public:
     /// @return A snapshot of the login state data.
     CSP_NO_EXPORT LoginStateData GetSnapshotThreadSafe() const;
 
-    /// @brief Set the login state data. This will replace all existing data in the login state with the new data provided.
-    CSP_NO_EXPORT void SetLoginStateData(csp::common::LoginStateData NewData);
+    /// @brief Set the login state data while guarded by a mutex lock. This will replace all existing data in the login state with the new data
+    /// provided.
+    CSP_NO_EXPORT void SetLoginStateDataThreadSafe(csp::common::LoginStateData NewData);
 
     /// @brief Attempt to set the login state to LoginRequested if the current state is LoggedOut.
     /// This will prevent multiple concurrent login attempts through use of a mutex guard.
@@ -102,7 +103,7 @@ public:
 
 private:
     void CopyStateFrom(const LoginState& OtherState);
-    
+
     std::unique_ptr<LoginStateData> Data;
     CSP_START_IGNORE
     mutable std::mutex LoginStateMutex;
--- a/Library/include/CSP/Common/LoginState.h
+++ b/Library/include/CSP/Common/LoginState.h
@@ -17,49 +17,96 @@
 #pragma once
 
 #include "CSP/CSPCommon.h"
+#include "CSP/Common/List.h"
 #include "CSP/Common/Settings.h"
 #include "CSP/Common/SharedEnums.h"
 #include "CSP/Common/String.h"
 
+#include <memory>
+#include <mutex>
+
 namespace csp::common
 {
 
-class DateTime;
+class LoginStateData;
 
 /// @brief Data structure representing the user login state, including detection of access token expiry
+/// @inv All access to the underlying LoginState data happen under a mutex lock.
 class CSP_API LoginState
 {
 public:
     LoginState();
     ~LoginState();
-
+    
     LoginState(const LoginState& OtherState);
     LoginState& operator=(const LoginState& OtherState);
+
+    /// @brief Get a copy of the current login state data.
+    /// This is a snapshot of the data at the time of the call, and will not update if the underlying state is mutated.
+    /// @return A snapshot of the login state data.
+    CSP_NO_EXPORT LoginStateData GetSnapshot() const;
+
+    /// @brief Set the login state data. This will replace all existing data in the login state with the new data provided.
+    CSP_NO_EXPORT void SetLoginStateData(csp::common::LoginStateData NewData);
+
+    /// @brief Attempt to set the login state to LoginRequested if the current state is LoggedOut.
+    /// This will prevent multiple concurrent login attempts.
+    /// @return If the state was successfully changed to LoginRequested.
+    CSP_NO_EXPORT bool TrySetLoginRequested();
+
+    /// @brief Attempt to set the login state to LogoutRequested if the current state is LoggedIn.
+    /// This will prevent multiple concurrent logout attempts.
+    /// @return If the state was successfully changed to LogoutRequested.
+    CSP_NO_EXPORT bool TrySetLogoutRequested();
+
+    /// @brief Reset the login state after a failed login/logout attempt.
+    CSP_NO_EXPORT void ResetResponseLoginState(csp::common::ELoginState LoginState);
+
+    /// @brief Update the access token expiry time.
+    CSP_NO_EXPORT void UpdateAccessTokenExpiry(csp::common::String AccessTokenExpiryLength);
+
+    /// @brief Update the refresh token expiry time.
+    CSP_NO_EXPORT void UpdateRefreshTokenExpiry(csp::common::String RefreshTokenExpiryLength);
 
     /// @brief Check if the access token for the login is expired.
     /// @return Is the token expired.
     [[nodiscard]] bool RefreshNeeded() const;
 
-    ELoginState State;
-    csp::common::String AccessToken;
-    csp::common::String AccessTokenExpiryLength;
-    csp::common::String RefreshToken;
-    csp::common::String RefreshTokenExpiryLength;
-    csp::common::String UserId;
-    csp::common::String DeviceId;
+    /// @brief Get the user ID of the currently logged-in user.
+    /// @return The user ID, or an empty string if not logged in.
+    csp::common::String GetUserId() const;
 
-    /// @Brief Default, tenant-wide settings returned from the service, often used to store universal data such as feature flags.
-    csp::common::List<csp::common::ApplicationSettings> DefaultApplicationSettings;
+    /// @brief Get the device ID of the currently logged-in user.
+    /// @return The device ID, or an empty string if not logged in.
+    csp::common::String GetDeviceId() const;
 
-    /// @Brief Default settings relevant to the specific user returned from the service. Also known as "UserSettings".
-    /// @see csp::systems::SettingsSystem
-    csp::common::List<csp::common::SettingsCollection> DefaultSettings;
+    /// @brief Get the access token the currently logged-in user.
+    /// @return The access token, or an empty string if not logged in.
+    csp::common::String GetAccessToken() const;
 
-    CSP_NO_EXPORT void SetAccessTokenRefreshTime(const csp::common::DateTime& NewDateTime);
+    /// @brief Get the refresh token the currently logged-in user.
+    /// @return The refresh token, or an empty string if not logged in.
+    csp::common::String GetRefreshToken() const;
+
+    /// @brief Get the current login state value.
+    /// @return The current login state value.
+    csp::common::ELoginState GetLoginStateValue() const;
+
+    /// @brief Get the default tenant-wide application settings returned at login.
+    /// @return The default application settings.
+    csp::common::List<csp::common::ApplicationSettings> GetDefaultApplicationSettings() const;
+
+    /// @brief Get the default user settings returned at login.
+    /// @return The default user settings.
+    csp::common::List<csp::common::SettingsCollection> GetDefaultSettings() const;
 
 private:
     void CopyStateFrom(const LoginState& OtherState);
-
-    csp::common::DateTime* AccessTokenRefreshTime;
+    
+    std::unique_ptr<LoginStateData> Data;
+    CSP_START_IGNORE
+    mutable std::mutex LoginStateMutex;
+    CSP_END_IGNORE
 };
+
 } // namespace csp::common
--- a/Library/include/CSP/Common/LoginState.h
+++ b/Library/include/CSP/Common/LoginState.h
@@ -50,12 +50,12 @@ public:
     CSP_NO_EXPORT void SetLoginStateData(csp::common::LoginStateData NewData);
 
     /// @brief Attempt to set the login state to LoginRequested if the current state is LoggedOut.
-    /// This will prevent multiple concurrent login attempts.
+    /// This will prevent multiple concurrent login attempts through use of a mutex guard.
     /// @return If the state was successfully changed to LoginRequested.
     CSP_NO_EXPORT bool TrySetLoginRequested();
 
     /// @brief Attempt to set the login state to LogoutRequested if the current state is LoggedIn.
-    /// This will prevent multiple concurrent logout attempts.
+    /// This will prevent multiple concurrent logout attempts through use of a mutex guard.
     /// @return If the state was successfully changed to LogoutRequested.
     CSP_NO_EXPORT bool TrySetLogoutRequested();
 
--- a/Library/include/CSP/Common/LoginState.h
+++ b/Library/include/CSP/Common/LoginState.h
@@ -60,8 +60,8 @@ public:
     /// @return If the state was successfully changed to LogoutRequested.
     CSP_NO_EXPORT bool TrySetLogoutRequested();
 
-    /// @brief Reset the login state after a failed login/logout attempt.
-    CSP_NO_EXPORT void ResetResponseLoginState(csp::common::ELoginState LoginState);
+    /// @brief Construct a new default login state object after a failed login/logout attempt with the specified ELoginState value.
+    CSP_NO_EXPORT void ReinitializeResponseLoginState(csp::common::ELoginState LoginState);
 
     /// @brief Update the access token expiry time.
     CSP_NO_EXPORT void UpdateAccessTokenExpiry(csp::common::String AccessTokenExpiryLength);
--- a/Library/src/Common/LoginStateData.h
+++ b/Library/src/Common/LoginStateData.h
@@ -27,6 +27,8 @@
 namespace csp::common
 {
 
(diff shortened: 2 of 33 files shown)


Library/include/CSP/Common/LoginState.h
```diff
@@ -17,49 +17,96 @@
...
+    /// @brief Get a copy of the current login state data.
+    /// This is a snapshot of the data at the time of the call, and will not update if the underlying state is mutated.
+    /// @return A snapshot of the login state data.
+    CSP_NO_EXPORT LoginStateData GetSnapshot() const;
+
+    /// @brief Set the login state data. This will replace all existing data in the login state with the new data provided.
+    CSP_NO_EXPORT void SetLoginStateData(csp::common::LoginStateData NewData);
+
+    /// @brief Attempt to set the login state to LoginRequested if the current state is LoggedOut.
+    /// This will prevent multiple concurrent login attempts.
```
MAG-ElliotMorris: [Consider] Maybe say how it prevents that. I presume it is because the caller must check if this `Try` was succesfull and then not attempt to login if so? Might be worth documenting that as the contract. Arguably this implies [[nodiscard]]

Library/src/Common/LoginStateData.h
```diff
@@ -0,0 +1,97 @@
...
+#include "CSP/Common/SharedEnums.h"
+#include "CSP/Common/String.h"
+#include "Common/DateTime.h"
+
+#include <memory>
+
+namespace csp::common
+{
+
+class LoginStateData
```
MAG-ElliotMorris, line 32: [Nit] Could do with a sentence describing the intent of this type. So people know it exists for thread safety reasons around `LoginState`

Library/src/Systems/Quota/QuotaSystem.cpp
```diff
@@ -51,7 +53,9 @@ QuotaSystem::~QuotaSystem()
 
 void QuotaSystem::GetTotalSpacesOwnedByUser(FeatureLimitCallback Callback)
 {
-    if (Context->GetLoginState().State != common::ELoginState::LoggedIn)
+    auto Data = Context->GetLoginState().GetSnapshot();
```
MAG-ElliotMorris: [nit] I know this seems silly, but given that this is threading, I wonder if a little `//Take a mutex-protected snapshot and work with that data to avoid races` comment above these lines might aid us in our thinking in the future.

... arguably you could accomplish the same with a method rename. 🤷

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -540,34 +540,35 @@ void SpaceSystem::EnterSpace(const String& SpaceId, csp::common::IRealtimeEngine
...
+                  LocalSpace.OwnerId = LocalUser;
+                  LocalSpace.UserIds = { LocalUser };
+                  LocalSpace.ModeratorIds = { LocalUser };
+
+                  SpaceResult LocalSpaceResult {};
+                  LocalSpaceResult.SetSpace(LocalSpace);
+                  LocalSpaceResult.SetResult(EResultCode::Success, static_cast<uint16_t>(csp::web::EResponseCodes::ResponseOK));
+                  return LocalSpaceResult;
+              })
+              .then(async::inline_scheduler(), FireEnterSpaceEvent(CurrentSpace));
```
MAG-ElliotMorris, line 571: [comment] Annoying how your editor has decided to mess with this a bit, and then github renders the entire closure as a change that you have to read both sides of ... only to finally realize it's just a newline at the bottom. 😅

Library/src/Systems/Users/Authentication.cpp
```diff
@@ -141,27 +145,25 @@ void LoginStateResult::OnResponse(const services::ApiResponseBase* ApiResponse)
                 return;
             }
 
-            State->SetAccessTokenRefreshTime(RefreshTime);
+            Data.SetAccessTokenRefreshTime(RefreshTime);
+
+            State->SetLoginStateData(Data);
```
MAG-ElliotMorris: [nit] Similar as before, maybe a comment just explicitly calling out that we do this data set like this from local data to prevent a race. Although maybe i'm being too paranoid with this one.

Library/src/Systems/Users/Authentication.cpp
```diff
@@ -141,27 +145,25 @@ void LoginStateResult::OnResponse(const services::ApiResponseBase* ApiResponse)
...
 
             // Signal login to anyone interested
             events::Event* LoginEvent = events::EventSystem::Get().AllocateEvent(events::USERSERVICE_LOGIN_EVENT_ID);
             LoginEvent->AddString("UserId", AuthResponse->GetUserId());
             events::EventSystem::Get().EnqueueEvent(LoginEvent);
-
-            SystemsManager::Get().GetUserSystem()->NotifyRefreshTokenHasChanged();
         }
+
+        SystemsManager::Get().GetUserSystem()->NotifyRefreshTokenHasChanged();
```
MAG-ElliotMorris: [Question] This appears to have moved out a scope level, is this intentional? What's the reasoning here?

Library/src/Systems/Users/Authentication.cpp
```diff
@@ -185,11 +187,7 @@ void LogoutResult::OnResponse(const services::ApiResponseBase* ApiResponse)
     {
         if (State)
         {
-            State->State = ELoginState::LoggedOut;
-            State->AccessToken = "InvalidAccessToken";
-            State->RefreshToken = "InvalidRefreshToken";
-            State->UserId = "InvalidUserId";
-            State->DeviceId = "InvalidDeviceId";
+            State->ResetResponseLoginState(ELoginState::LoggedOut);
```
MAG-ElliotMorris: [Consider] Not sure what i'd call it to be fair, but this name seems a bit ambiguous about whether it's overwriting the login state with the new one, or constructing a brand new default object (which is what it is actually doing).

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -61,8 +58,9 @@ void StartMultiplayerConnection(csp::multiplayer::MultiplayerConnection& Multipl
     if (CreateMultiplayerConnection)
     {
         LogSystem.LogMsg(csp::common::LogLevel::Log, "Starting Multiplayer Connection");
-        MultiplayerConnection.Connect(
-            ConnectionCallback, MultiplayerURI, LoginStateRes.GetLoginState().AccessToken, LoginStateRes.GetLoginState().DeviceId);
+
+        auto Data = LoginStateRes.GetLoginState().GetSnapshot();
```
MAG-ElliotMorris: [nit] const auto

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -125,10 +123,15 @@ std::optional<csp::common::String> ThirdPartyPlatformToString(const csp::common:
...
+void LoginSocial(csp::services::ApiBase* AuthenticationAPI, std::shared_ptr<csp::common::LoginState> CurrentLoginState, csp::common::LogSystem* LogSystem,
     const std::shared_ptr<chs_user::LoginSocialRequest>& Request, bool CreateMultiplayerConnection, csp::systems::LoginStateResultCallback Callback)
 {
-    csp::systems::LoginStateResultCallback LoginStateResCallback = [=](const csp::systems::LoginStateResult& LoginStateRes)
+    // Local copy made and captured by value in callback below. This ensures the shared_ptr stays alive until the callback is executed via the
+    // ResponseHandler.
+    auto LoginStateRef = CurrentLoginState;
+
+    csp::systems::LoginStateResultCallback LoginStateResCallback
+        = [LogSystem, Callback, CreateMultiplayerConnection, LoginStateRef](const csp::systems::LoginStateResult& LoginStateRes)
```
MAG-ElliotMorris: [nit] I think you can invoke this copy with an inline capture too. Just a bit terser.

```
csp::systems::LoginStateResultCallback LoginStateResCallback
    = [LogSystem, Callback, CreateMultiplayerConnection, LoginStateRef = CurrentLoginState]
      (const csp::systems::LoginStateResult& LoginStateRes){...};
```

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -155,13 +158,29 @@ void LoginSocial(csp::services::ApiBase* AuthenticationAPI, csp::common::LoginSt
     };
 
     csp::services::ResponseHandlerPtr ResponseHandler = AuthenticationAPI->CreateHandler<csp::systems::LoginStateResultCallback,
-        csp::systems::LoginStateResult, csp::common::LoginState, chs_user::AuthDto>(LoginStateResCallback, CurrentLoginState);
+        csp::systems::LoginStateResult, csp::common::LoginState, chs_user::AuthDto>(LoginStateResCallback, LoginStateRef.get());
```
MAG-ElliotMorris: [Comment] This is no less safe than it was before, but it still seems like a bit of a smell to me. These handlers always confuse me a bit, but they do seem like they are async objects. In the future they could be perhaps be extended to hold shared pointer references so they also keep these resources alive as long as they need them by definition.

Speculation though, I'm not sure what the ownership model is like around here.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -155,13 +158,29 @@ void LoginSocial(csp::services::ApiBase* AuthenticationAPI, csp::common::LoginSt
...
 
     static_cast<chs_user::AuthenticationApi*>(AuthenticationAPI)->usersLogin_socialPost({ Request }, ResponseHandler);
 }
 
+void SetTokenOptions(const csp::common::Optional<csp::systems::TokenOptions>& TokenOptions, std::shared_ptr<chs_user::TokenOptions> Options,
```
MAG-ElliotMorris, line 165: [Question] I'm assuming I'm just lost in the diff, because I can't spot this being removed anywhere else. I see it's always been called however, so it must have existed somewhere. Can you just double check that it's my eyes at fault and not an accidental method duplication 🙏 .

Library/include/CSP/Common/LoginState.h
```diff
@@ -17,49 +17,96 @@
...
+    /// @brief Get the refresh token the currently logged-in user.
+    /// @return The refresh token, or an empty string if not logged in.
+    csp::common::String GetRefreshToken() const;
 
-    /// @Brief Default settings relevant to the specific user returned from the service. Also known as "UserSettings".
-    /// @see csp::systems::SettingsSystem
-    csp::common::List<csp::common::SettingsCollection> DefaultSettings;
+    /// @brief Get the current login state value.
+    /// @return The current login state value.
+    csp::common::ELoginState GetLoginStateValue() const;
```
MAG-ElliotMorris, line 94: [comment] weird conceptual overlap in the naming. Not that I have any better ideas.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -286,7 +315,7 @@ void UserSystem::SetNetworkEventBus(csp::multiplayer::NetworkEventBus& EventBus)
     RegisterSystemCallback();
 }
 
-const csp::common::LoginState& UserSystem::GetLoginState() const { return CurrentLoginState; }
+const csp::common::LoginState& UserSystem::GetLoginState() const { return *CurrentLoginState; }
```
MAG-ElliotMorris, line 315: [consider] I wonder if it's worth changing the interface here to return a value, presumably this is copyable now? Would eliminate a potential lifetime issue. Not certain on this one, just throwing it out there.

MAG-ElliotMorris (approved):
Looking good from what I can tell, although the proof will be in the pudding with this, hard to know on threading issues as a reviewer.

Comments are pretty much entirely documentary stuff, nothing fundamental.

One general note: this was somewhat difficult to review due to the inclusion of quite a lot of whitespace changes along with the actual logical changes all in the same commit. It made it difficult to parse what was a real change and what was noise. Not the end of the world, but I would have been able to give a better review if it was somewhat cleaner.

Library/include/CSP/Common/LoginState.h
MAG-AdamThorn: It will prevent multiple concurrent attempts because the first successful attempt (which is under a `scoped_lock`) will set the `ELoginState` to `LoginRequested` and so subsequent calls will fail.

I will add a bit to the comment to mention the mutex guard.

Update: [0b88579](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/0b88579a10d2316b4ecec3f4b7634a00f7316783)

Library/src/Systems/Quota/QuotaSystem.cpp
MAG-AdamThorn: Makes sense. I elected to rename the method to `GetSnapshotThreadSafe()`.

Update: [7fbbc26](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/7fbbc267833ccbf72d84ac92823cf21fd0927023)

Library/src/Systems/Users/Authentication.cpp
MAG-AdamThorn: As above I elected to rename the method to `SetLoginStateDataThreadSafe()`.

Update: [00319e7](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/00319e7c92f41e7f5016284cc8036090cb905a6d)

Library/src/Systems/Users/Authentication.cpp
MAG-AdamThorn: Well spotted, that was a mistake.

Fix: [bed7d88](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/bed7d8807cab36139ba17b4b0d0c362175ef77df)

Library/src/Systems/Users/Authentication.cpp
MAG-AdamThorn: I have changed the name to `ReinitializeResponseLoginState()` and updated the comment slightly to hopefully make it clearer.

Update: [c5cb524](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/c5cb5241fecb41063fbd961fd9b14d6dea59174c)

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Fixed here: [1929ea7](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/1929ea76d916220f2cb1ebdc0f53d1200c40b4e6)

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: Yep I agree, that reads better.

Update: [1929ea7](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/1929ea76d916220f2cb1ebdc0f53d1200c40b4e6)

Library/include/CSP/Common/LoginState.h
MAG-AdamThorn: Yes I agree. Ideally we should change the name of the object or the enum, though I am not going to do that now.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: This is a new utility function that was added to remove the repetitive code from the various login methods. So it is not a duplicate.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: I agree and I think we could look at extending these handlers. I didn't want to do that as part of this change as it is already dragging on a bit.

Library/src/Common/LoginStateData.h
MAG-AdamThorn: Update: [0b88579](https://github.com/magnopus-opensource/connected-spaces-platform/pull/972/changes/0b88579a10d2316b4ecec3f4b7634a00f7316783)

MAG-AdamThorn:
> One general note: this was somewhat difficult to review due to the inclusion of quite a lot of whitespace changes along with the actual logical changes all in the same commit. It made it difficult to parse what was a real change and what was noise. Not the end of the world, but I would have been able to give a better review if it was somewhat cleaner.

Yes my bad. I ended up making a bit of a pigs ear of it when moving my commits to separate PR branches. It was all nicely laid out but I ended up having to collapse the whole thing. I should have made more of an effort to separate them out again after the fact.

Library/src/Systems/Users/UserSystem.cpp
MAG-AdamThorn: That wouldn't work as LoginState owns the LoginStateData via a unique_ptr and controls access to that data via the mutex it also owns. If this method returned a copy then it would not receive subsequent updates made to it.

Library/src/Systems/Settings/SettingsSystem.cpp
```diff
@@ -595,7 +593,7 @@ void SettingsSystem::AddAvatarPortrait(const FileAssetDataSource& ImageDataSourc
                 }
             };
 
-            const auto UniqueAssetName = AVATAR_PORTRAIT_ASSET_NAME + UserId;
+            const auto UniqueAssetName = AVATAR_PORTRAIT_ASSET_NAME + UserSystem->GetLoginState().GetUserId();
```
MAG-mv, line 596: [Comment] Now we understand the lifetime of the login state better, in the future, we should inject the login state into these systems instead of relying on the entire user system

Library/src/Web/RemoteFileManager.cpp
```diff
@@ -26,11 +28,13 @@ namespace
 {
 csp::common::Optional<csp::common::String> ConstructAuthorizationHeader(const csp::common::IAuthContext& InAuthContext)
 {
+    auto Data = InAuthContext.GetLoginState().GetSnapshot();
```
MAG-mv, line 31: [Nit] const

Tests/src/InternalTests/RemoteFileManagerTests.cpp
```diff
@@ -50,8 +51,12 @@ TEST_P(GetFile, GetFileSendsCorrectRequest)
 
     // Construct a LoginState object with the correct state.
     csp::common::LoginState LoginState;
-    LoginState.State = std::get<0>(GetParam());
-    LoginState.AccessToken = std::get<1>(GetParam());
+    {
+        csp::common::LoginStateData Data = LoginState.GetSnapshot();
```
MAG-mv, line 55: [nit] const

MAG-mv (approved):

Library/src/Systems/Users/UserSystem.cpp
MAG-ElliotMorris: I must be getting confused with `Request->SetTokenOptions`. I don't fully understand the distinction, but it's fine.

MAG-ElliotMorris (approved):

Library/src/Systems/Settings/SettingsSystem.cpp
MAG-AdamThorn: Something to consider for sure.

Library/src/Web/RemoteFileManager.cpp
MAG-AdamThorn: I have another PR inbound after this one, will make this change there.

Tests/src/InternalTests/RemoteFileManagerTests.cpp
MAG-AdamThorn: I have another PR inbound after this one, will make this change there.

PR #973, opened 2026-06-02
[OF-1850] feat: Run on linux unreal

MAG-mv:
This PR allows us to build CSP using Unreals custom toolchain, which gives us more safety against abi issues. This work contains the following:

- Create a Unreal Conan profile
- Fixes assumtions that user data directories always exist
- Downgrade to OpenSSL 1.1 to match Unreals version, and our old build system version
- Fix additional compile errors due to Unreals toolchain being stricter

### DIFF
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -26,6 +26,7 @@
 #include "Debug/Logging.h"
 #include "Events/EventSystem.h"
 
+#include <filesystem>
 #include <cstdio>
 #include <fmt/format.h>
 #if defined(CSP_ANDROID)
@@ -108,6 +109,23 @@ bool FileExists(std::string Path)
 
     return (stat(Path.c_str(), &Stat) == 0 && S_ISREG(Stat.st_mode));
 #endif
+}
+
+bool EnsureFolderExists(const std::string& Path)
+{
+    std::error_code Ec;
+
+    if (std::filesystem::exists(Path, Ec))
+    {
+        return std::filesystem::is_directory(Path, Ec);
+    }
+
+    if (!std::filesystem::create_directories(Path, Ec) && Ec)
+    {
+        return false;
+    }
+
+    return true;
 }
 
 // Named "CreateFolder" to avoid conflicts with Win32's CreateDirectory macro
@@ -214,9 +232,9 @@ std::string LoadDeviceId()
 #else
     const std::string CSPDataRoot = DeviceIdPath();
 
-    if (!FolderExists(CSPDataRoot))
+    if (CSPDataRoot.empty() || !EnsureFolderExists(CSPDataRoot))
     {
-        CreateFolder(CSPDataRoot);
+        return {};
     }
 
     auto DeviceIdFilePath = std::string(CSPDataRoot) + "device.id";
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -82,27 +82,6 @@
 namespace
 {
 
-bool FileExists(std::string Path)
-{
-#if defined(CSP_WINDOWS)
-    auto Attr = GetFileAttributesA(Path.c_str());
-
-    return (Attr != INVALID_FILE_ATTRIBUTES && ((Attr & FILE_ATTRIBUTE_NORMAL) || (Attr & FILE_ATTRIBUTE_ARCHIVE)));
-#else
-    struct stat Stat;
-
-    return (stat(Path.c_str(), &Stat) == 0 && S_ISREG(Stat.st_mode));
-#endif
-}
-
-bool EnsureFolderExists(const std::string& Path)
-{
-    std::error_code Ec;
-    std::filesystem::create_directories(Path, Ec);
-
-    return !Ec; 
-}
-
 #if defined(CSP_WASM)
 // clang-format off
 EM_JS(char*, get_device_id, (), {
@@ -128,6 +107,36 @@ EM_JS(void, set_device_id, (const char* cDeviceId), {
 #endif
 
 #if !defined(CSP_WASM)
+bool FileExists(std::string Path)
+{
+#if defined(CSP_WINDOWS)
+    auto Attr = GetFileAttributesA(Path.c_str());
+
+    return (Attr != INVALID_FILE_ATTRIBUTES && ((Attr & FILE_ATTRIBUTE_NORMAL) || (Attr & FILE_ATTRIBUTE_ARCHIVE)));
+#else
+    struct stat Stat;
+
+    return (stat(Path.c_str(), &Stat) == 0 && S_ISREG(Stat.st_mode));
+#endif
+}
+
+bool EnsureFolderExists(const std::string& Path)
+{
+#if defined(CSP_ANDROID)
+    if (mkdir(Path.c_str(), 0777) == 0) {
+        return true;
+    }
+
+    return errno == EEXIST;
+#else
+
+    std::error_code Ec;
+    std::filesystem::create_directories(Path, Ec);
+
+    return !Ec; 
+#endif
+}
+
 std::string DeviceIdPath()
 {
     // For all platforms, we want to guarantee the current user has read/write access
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -257,7 +257,9 @@ std::string LoadDeviceId()
         assert(File != nullptr);
 
         char Uuid[33];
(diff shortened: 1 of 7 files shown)


Library/src/CSPFoundation.cpp
```diff
@@ -214,9 +232,9 @@ std::string LoadDeviceId()
 #else
     const std::string CSPDataRoot = DeviceIdPath();
 
-    if (!FolderExists(CSPDataRoot))
+    if (CSPDataRoot.empty() || !EnsureFolderExists(CSPDataRoot))
```
mag-lt: [question] is there reason to believe `CSPDataRoot` would be empty? All implementations of `DeviceIdPath()` don't appear to have a code path where they return an empty string?

Library/src/CSPFoundation.cpp
```diff
@@ -257,7 +257,9 @@ std::string LoadDeviceId()
         assert(File != nullptr);
 
         char Uuid[33];
-        fread(Uuid, sizeof(char), 32, File);
+        [[maybe_unused]] const size_t ReadSize = fread(Uuid, 1, 32, File);
+        assert(ReadSize != 32);
```
mag-lt: [fix] This appears to be asserting the opposite condition to what is expected? i.e. we want to assert that `ReadSize == 32` ?

Library/src/CSPFoundation.cpp
mag-lt, line 210: [question] Should we be failing initialisation if this happens? i.e. [here](https://github.com/magnopus-opensource/connected-spaces-platform/blob/011b7a95b331f61de7164944c4cc65d2cfbd888c/Library/src/CSPFoundation.cpp#L490). As this is an internal function, we could also communicate the error via an optional vs an empty string

Library/src/Multiplayer/SpaceTransform.cpp
```diff
@@ -78,11 +78,11 @@ csp::multiplayer::SpaceTransform csp::multiplayer::SpaceTransform::operator*(con
     const auto TRS2 = T2 * R2 * S2;
     const auto ComposedTRS = TRS1 * TRS2;
 
-    glm::vec3 TransformScale;
-    glm::quat TransformRotation;
-    glm::vec3 Translation;
-    glm::vec3 Skew;
-    glm::vec4 Perspective;
+    glm::vec3 TransformScale{ 1.0f, 1.0f, 1.0f };
```
mag-lt: [question] These are all out params, right? They're written to but not read from in `glm::decompose`? I'd just default / zero init if the compiler is complaining i.e. `glm::vec3 TransformScale{};`, as written these numbers seem important, but they're not.

Library/src/CSPFoundation.cpp
```diff
@@ -110,6 +111,23 @@ bool FileExists(std::string Path)
 #endif
 }
 
+bool EnsureFolderExists(const std::string& Path)
```
mag-lt, line 123: [nit] typically with file operations, you just want to go for it vs checking, as technically things could change after checking. so this function could probably be reduced to:

```cpp
bool EnsureFolderExists(const std::string& Path)
{
    std::error_code Ec;
    
    std::filesystem::create_directories(Path, Ec);

    return !Ec; 
}
```

`create_directories` can handle it already existing, and it isn't an error if it does.

Library/src/CSPFoundation.cpp
```diff
@@ -214,9 +232,9 @@ std::string LoadDeviceId()
 #else
     const std::string CSPDataRoot = DeviceIdPath();
 
-    if (!FolderExists(CSPDataRoot))
```
mag-lt, line 217: [fix] Unless I'm mistaken, `FolderExists` and `CreateFolder` are no longer used following rewriting this? If so, we should delete them.

conanfile.py
```diff
@@ -35,6 +35,9 @@ def requirements(self):
 
         # We use the Emscripten WebSockets API for Emscripten
         if self.settings.os != "Emscripten":
+            # Override Pocos default OpenSSL to use OpenSSL 1.1
+            # This is currently needed for support with Unreals OpenSSL lib.
+            self.requires("openssl/1.1.1t", override=True)
```
mag-lt, line 40: [comment] I've typically solved similar issues to this by linking libraries statically and ensuring their symbols aren't exported/visible (which requires specific incantations for each compiler, and is often more awkward than it should be). Though, with openssl, perhaps the right thing to do is try to ensure we're using the actual same one, i.e. somehow communicating here to pick up Unreal's bundled version, though I'm not quite sure how to express that. Perhaps there is a flag to optionally pick up the system one, which may resolve correctly given Unreal's sysroot?

In any case, I'm okay with this.

mag-lt (changes requested):
Having a separate Conan file makes sense to me, nice one for figuring this stuff out. Some minor details to address in some of the code changes here

Library/src/CSPFoundation.cpp
MAG-mv: You're right, fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/efc5d0360826f4ccfae5147b4456ea8c69a3921d

Library/src/CSPFoundation.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/efc5d0360826f4ccfae5147b4456ea8c69a3921d

Library/src/CSPFoundation.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/efc5d0360826f4ccfae5147b4456ea8c69a3921d

Library/src/CSPFoundation.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/efc5d0360826f4ccfae5147b4456ea8c69a3921d

Library/src/CSPFoundation.cpp
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/efc5d0360826f4ccfae5147b4456ea8c69a3921d

conanfile.py
MAG-mv: Yeah, agreed. This would be a nice thing to do. I'm sure theres a way to select the version through the profile

Library/src/Multiplayer/SpaceTransform.cpp
MAG-mv: Fixed up here: 3a2fa50aa6409f7931c3eb68cf60563a6e2f89af

Library/src/CSPFoundation.cpp
```diff
@@ -146,6 +107,36 @@ EM_JS(void, set_device_id, (const char* cDeviceId), {
...
+    struct stat Stat;
+
+    return (stat(Path.c_str(), &Stat) == 0 && S_ISREG(Stat.st_mode));
+#endif
+}
+
+bool EnsureFolderExists(const std::string& Path)
+{
+#if defined(CSP_ANDROID)
+    if (mkdir(Path.c_str(), 0755) == 0) {
```
mag-lt: [nit] style / formatting of the curly brackets. run clang-format?

[question] this is different than before, it was previously `mkdir(Path.c_str(), 0777)`, any reason for changing?

mag-lt (approved):

Library/src/CSPFoundation.cpp
MAG-mv: From what I understand 0777 creates global access, o any app could modify the directory, while 0755 creates permissions for the specific app.

MAG-mv: Thinking about it more, I should probably keep it as 0777 considering we have been suing this until now

MAG-mv: fixed up here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/973/commits/e4f2f9166fa636eae0cbb53278827ad6a44e460c

MAG-AdamThorn (approved):

Library/src/CSPFoundation.cpp
mag-lt: Yeah, I'd vote for keeping the permissions as they were for now. Formatting nit still applies, but will resolve in any case.

MAG-mv: I fixed up the commit and changed back to 0777 from 0755

PR #974, opened 2026-06-02
[OF-1867] feat: Support injecting custom component schemas via JSON at engine construction time

mag-lt:
This introduces a JSON schema format for describing components, exemplified in the tests. Both engine types now accept this JSON at construction time, registering the described schemas in the engine-wide registry. Constructors accepting `Array<ComponentSchema>` directly are also exposed, though it is worth discussing whether JSON should be the sole interface, including for registry introspection, as programmatic construction of `ComponentSchema` is likely to be more awkward than raw JSON particularly in JavaScript.

A JSON Schema definition for the format would be a useful follow-up to enable editor tooling support.           
  
So, following this clients can inject and use entirely new components. However, modifying or extending existing built-in components is not yet supported, as we conservatively instantiate hardcoded component classes for legacy
TypeIds to ensure their existing behaviour is preserved. We'll need to address this in follow up work, as that is surely the more common use case.

### DIFF
--- a/Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
+++ b/Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
@@ -70,8 +70,36 @@ public:
     /// @param RemoteScriptRunner csp::common::IJSScriptRunner& : Object capable of running a script.
     OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner);
 
-    CSP_NO_EXPORT OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
+    /// @brief OfflineRealtimeEngine constructor.
+    /// Creates an empty realtime engine with additional component schemas.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param AdditionalComponents Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
         const csp::common::Array<ComponentSchema>& AdditionalComponents);
+
+    /// @brief OfflineRealtimeEngine constructor.
+    /// Creates an empty realtime engine, registering additional component schemas from JSON.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param JsonSchemas Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    /// The list is a wrapper generator workaround for passing large strings. In practice a single element is
+    /// expected, containing a JSON array of schema objects. Multiple elements are supported for combining schemas
+    /// from independent sources. Entries that fail to parse are skipped with a warning.
+    OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
+        const csp::common::List<csp::common::String>& JsonSchemas);
+
+    /// @brief OfflineRealtimeEngine constructor.
+    /// Creates a realtime engine pre-populated from a scene description, with additional component schemas from JSON.
+    /// @param SceneDescription The scene description containing entities to populate in the engine.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param JsonSchemas Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    /// The list is a wrapper generator workaround for passing large strings. In practice a single element is
+    /// expected, containing a JSON array of schema objects. Multiple elements are supported for combining schemas
+    /// from independent sources. Entries that fail to parse are skipped with a warning.
+    OfflineRealtimeEngine(const CSPSceneDescription& SceneDescription, csp::common::LogSystem& LogSystem,
+        csp::common::IJSScriptRunner& RemoteScriptRunner, const csp::common::List<csp::common::String>& JsonSchemas);
 
     /// @brief OfflineRealtimeEngine destructor
     /// Removes entity script bindings and deregisters tick event listeners
--- a/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
+++ b/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
@@ -110,9 +110,31 @@ public:
     OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerConnection, csp::common::LogSystem& LogSystem,
         csp::multiplayer::NetworkEventBus& NetworkEventBus, csp::common::IJSScriptRunner& RemoteScriptRunner);
 
-    CSP_NO_EXPORT OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerConnection, csp::common::LogSystem& LogSystem,
+    /// @brief OnlineRealtimeEngine constructor.
+    /// Creates a realtime engine with additional component schemas.
+    /// @param InMultiplayerConnection The multiplayer connection to construct the engine with.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param NetworkEventBus Reference to the network event bus, used for leadership election messaging.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param AdditionalComponents Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerConnection, csp::common::LogSystem& LogSystem,
         csp::multiplayer::NetworkEventBus& NetworkEventBus, csp::common::IJSScriptRunner& RemoteScriptRunner,
         const csp::common::Array<ComponentSchema>& AdditionalComponents);
+
+    /// @brief OnlineRealtimeEngine constructor.
+    /// Creates a realtime engine with additional component schemas from JSON.
+    /// @param InMultiplayerConnection The multiplayer connection to construct the engine with.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param NetworkEventBus Reference to the network event bus, used for leadership election messaging.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param JsonSchemas Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    /// The list is a wrapper generator workaround for passing large strings. In practice a single element is
+    /// expected, containing a JSON array of schema objects. Multiple elements are supported for combining schemas
+    /// from independent sources. Entries that fail to parse are skipped
+    /// with a warning.
+    OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerConnection, csp::common::LogSystem& LogSystem,
+        csp::multiplayer::NetworkEventBus& NetworkEventBus, csp::common::IJSScriptRunner& RemoteScriptRunner,
+        const csp::common::List<csp::common::String>& JsonSchemas);
 
     /// @brief OnlineRealtimeEngine destructor
     CSP_NO_EXPORT ~OnlineRealtimeEngine();
--- a/Library/src/Multiplayer/ComponentSchema.cpp
+++ b/Library/src/Multiplayer/ComponentSchema.cpp
@@ -176,7 +176,7 @@ namespace
         };
     }
 
-    std::optional<csp::common::ReplicatedValue> TryParseDefaultValue(const std::string& Type, const rapidjson::Value& Value)
+    std::optional<csp::common::ReplicatedValue> TryParse(const std::string& Type, const rapidjson::Value& Value)
     {
         if (Type == "string")
         {
@@ -246,7 +246,7 @@ namespace
         const auto Key = static_cast<ComponentProperty::KeyType>(Value["key"].GetUint());
         const auto* Name = Value["name"].GetString();
         const auto* Type = Value["type"].GetString();
-        const auto DefaultValue = TryParseDefaultValue(Type, Value["defaultValue"]);
+        const auto DefaultValue = TryParse(Type, Value["defaultValue"]);
 
         if (!DefaultValue)
         {
--- a/Library/src/Multiplayer/ComponentSchema.cpp
+++ b/Library/src/Multiplayer/ComponentSchema.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "CSP/Multiplayer/ComponentSchema.h"
+#include "CSP/Common/Systems/Log/LogSystem.h"
 #include "Common/Convert.h"
 #include "Json/JsonSerializer.h"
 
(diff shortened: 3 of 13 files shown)


Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+
+        return csp::common::Vector4 {
+            Value[0].GetFloat(),
+            Value[1].GetFloat(),
+            Value[2].GetFloat(),
+            Value[3].GetFloat(),
+        };
+    }
+
+    std::optional<csp::common::ReplicatedValue> TryParseDefaultValue(const std::string& Type, const rapidjson::Value& Value)
```
MAG-ElliotMorris: [nit] Is `defaultValue` the best framing for this method do you think? I presume that's parsing the default value in the schema, but the abstraction level of this method seems more like a more general purpose `TryParse`.

... Then again, I don't necessarily mind the conveyance the method name does as to the original intent of the method, it just becomes wrong if we ever reuse this method for something else.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+#include <optional>
+#include <string>
+#include <vector>
 
 namespace csp::multiplayer
 {
 
+namespace
+{
+    void SerializeStringProperty(csp::json::JsonSerializer& Serializer, const csp::common::String& Value)
```
MAG-mv, line 35: [Comment] I'm interested why these can't be used with the json serializer.JsonDeserializer dos provide a DeserializeSafe function

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+    {
+        if (Type == "string")
+        {
+            return AsReplicatedValue(TryParseString(Value));
+        }
+
+        if (Type == "float")
+        {
+            return AsReplicatedValue(TryParseFloat(Value));
+        }
```
MAG-ElliotMorris, line 190: [Comment] I'm glad you've not decided to use explicit template instantiations or visitors to deduplicate this if/else. I get into some goodnatured arguments about this sometimes, but I think this is simpler and easier to maintain, despite the technically redundant runtime branching.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+
+        if (!Value.HasMember("key") || !Value["key"].IsUint())
+        {
+            return std::nullopt;
+        }
+
+        if (!Value.HasMember("name") || !Value["name"].IsString())
+        {
+            return std::nullopt;
+        }
```
MAG-ElliotMorris, line 250: [Consider] These parsing methods are defining format invariants, and I am grateful to see the preconditions checked. I wonder if there's a way to improve the authoring experience by emitting errors or logs here specifying exactly what member was missing, to help people who make typos or don't understand specifically how to build their json, etc.

MAG-ElliotMorris: std::expected, or a richer result type, would help here. Optional is a bit of a limited error handling object.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+
+        if (Type == "vec4")
+        {
+            return AsReplicatedValue(TryParseVec4(Value));
+        }
+
+        return std::nullopt;
+    }
+
+    std::optional<ComponentProperty> TryParseProperty(const rapidjson::Value& Value)
```
MAG-ElliotMorris: [Consider] Some inline doc on these richer methods explaining what JSON structure they are parsing, probably with json examples. Heck, maybe only examples. It's deducible from the code, but it'll be easier to grok if you can see the json.

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
```diff
@@ -70,9 +70,37 @@ class CSP_API OfflineRealtimeEngine : public csp::common::IRealtimeEngine
...
     OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner);
 
-    CSP_NO_EXPORT OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
+    /// @brief OfflineRealtimeEngine constructor.
+    /// Creates an empty realtime engine with additional component schemas.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param AdditionalComponents Component schemas to register alongside the built-in schemas in the engine-wide registry.
+    OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
         const csp::common::Array<ComponentSchema>& AdditionalComponents);
```
MAG-ElliotMorris, line 79: [Comment] I'm also in two minds about whether the C++ interface is worth it here. The software engineer in me says yes, it's more flexible and introspectable, but the ecosystem steward in me says no, it'll create more paths of intricate, low-observability, potentially problematic data. 🤔

MAG-ElliotMorris (approved):
Engineering wise I have little problem with this, solid implementation as usual from you. I think the fact that this PR is rather small and simple belies good architectural thinking leading up to this.

Ecosystem wise, I think you're right that this needs further iteration. Nonetheless, I have no problem with this interface going in as-is, as it is complete, albeit it might need polishing and sanding in the UX/workflow department.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: I'm glad. I had to exercise an unbelievable amount of self-control in order to not go down that path 😅

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: Good eye, I'll rename to `TryParse`

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
MAG-ElliotMorris: Ach this is a hard one. I could also make the case that the C++ interface should be the _only_ interface and we should leave parsing entirely to clients ... although that would make adoption friction to this quite high for clients as they are currently positioned, so I guess i'd lean away from _that_ at least ... maybe.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: > std::expected, or a richer result type, would help here. Optional is a bit of a limited error handling object.

Agreed, I'd prefer not to interleave the logging here, but without having a convenient way to pass back what failed to the caller so they can log it, it may be the pragmatic thing to do. I could of course use exceptions, but that makes me feel just as bad...

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -110,10 +110,32 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
...
         csp::multiplayer::NetworkEventBus& NetworkEventBus, csp::common::IJSScriptRunner& RemoteScriptRunner);
 
-    CSP_NO_EXPORT OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerConnection, csp::common::LogSystem& LogSystem,
+    /// @brief OnlineRealtimeEngine constructor.
+    /// Creates a realtime engine with additional component schemas.
+    /// @param InMultiplayerConnection The multiplayer connection to construct the engine with.
+    /// @param LogSystem Logger for status and debug output.
+    /// @param NetworkEventBus Reference to the network event bus, used for leadership election messaging.
+    /// @param RemoteScriptRunner Object capable of running a script.
+    /// @param AdditionalComponents Component schemas to register alongside the built-in schemas in the engine-wide registry.
```
MAG-mv, line 119: [Question] Am I understanding you commit comment correctly? Are you suggesting only having the constuctor for a json string? If so, I think I agree.

Library/src/Multiplayer/ComponentSchema.cpp
MAG-mv: Oh, is it because it only check if the key exists, and doesn't check the type?

Library/src/Multiplayer/ComponentSchema.cpp
MAG-ElliotMorris: Maybe a bit out there, but it might be rather trivial to take a backport of `std::expected` and use [that](https://github.com/TartanLlama/expected). It's even on [conancentre](https://github.com/conan-io/conan-center-index/tree/master/recipes/tl-expected)

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -15,10 +15,371 @@
...
+
+    std::optional<csp::common::Array<ComponentProperty>> TryParseProperties(rapidjson::Value::ConstArray JsonProperties)
+    {
+        auto Properties = std::vector<ComponentProperty> {};
+
+        for (const auto& Element : JsonProperties)
+        {
+            const auto Property = TryParseProperty(Element);
+
+            if (!Property)
```
MAG-AdamThorn, line 304: [Question] Is this correct? It means that if one of the properties fails to be parsed, for example is missing a default value, then all of the properties will fail - we will return a nullopt.

mag-lt: That is the design choice I've made here. It is quite conservative, taking the stance that if the schema given at construction time is not fully understood by the library, then that says something about the integrity of the schema. We could potentially loosen it, but I'm on the side of being strict here currently unless we can make the case not to be. In practice this either means a typo or type mismatch in one of the properties, or that somehow they got a schema from the future that supported more types and passed it at runtime to the engine constructor (unlikely).

Things are slightly different when receiving data about a component instance over the wire, where the behaviour should be to just ignore properties from a component that don't match the schema in the registry (assuming they came from a newer client that added a property), but I'm conceptualising that as a different thing.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: > Maybe a bit out there, but it might be rather trivial to take a backport of std::expected and use [that](https://github.com/TartanLlama/expected). It's even on [conancentre](https://github.com/conan-io/conan-center-index/tree/master/recipes/tl-expected)

Something to consider more broadly for sure, but I wouldn't want this to be the PR to introduce this. Even more out there would be updating to C++23 instead (also not in the context of this PR).

I've reluctantly just interleaved the logging for now in 2e93f0ac3281e47dc16ba739cd8b7f97fea51ca0, as I can't really convince myself not doing so is better (I prefer the code before, but this is obviously more useful).

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
mag-lt: I've wrestled with the same thing, and don't really have a good answer. I could be conservative and keep the `ComponentSchema` ones as `CSP_NO_EXPORT` for now, how does that sit with you?

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: My opinion on this is that is sort of jointly what the tests are for, to give examples. These are internal helper functions, which I personally wouldn't tend to document. I'm inclined to leave these as-is.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: Done in 7f68532c011e63c8bbc49bc57a81e51e

Library/src/Multiplayer/ComponentSchema.cpp
MAG-AdamThorn: Yes ok that makes sense.

Library/src/Multiplayer/ComponentSchema.cpp
MAG-mv: [Comment] I think an enum or result type would be useful to return from here, then we can just log in one place

MAG-mv (approved):

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: > Oh, is it because it only check if the key exists, and doesn't check the type?

Yeah, that is the crux of it. I considered trying to work that through into `JsonDeserializer` but opted not to.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: I agree generally, especially if this were shared functionality in a public header, but interleaving the logging here seemed the most pragmatic thing to do here to me without better general utility types.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
mag-lt: Yeah exactly, similar discussion [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/974#discussion_r3350358868).

I was also thinking more generally, especially when extending the component schema to have ranged values, enum/discrete options, whether we can flatten that all into an interface type that is ergonomic to use without good primitives (like a std::variant) to represent the different cases, and whether it would just be easier to have JSON across the board at the interface level (i.e. `ComponentSchema` is also currently the type used on the `IComponentSchemaRegistry`).

Library/src/Multiplayer/ComponentSchema.cpp
MAG-ElliotMorris: I'd lean towards over-documenting, especially with the lack of specific error-reporting in these richer parse methods, but that's merely a philosophical difference so you know, 🤷. Not going to split hairs.

MAG-ElliotMorris: Oh, I see you've interleaved logging. That reduces my fear on this.

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
MAG-ElliotMorris: Hmm ... 

This is one of the harder calls I've had to make. But if you were to ask me, I say leave them both as public methods. Any sort of `bounded` trait, or other such authoring utility, will necessarily need to be represented in the model, so it should scale with that naturally, and there's no reason to make introspection and modification of schema by the program itself more difficult than it has to be. 

Something that would relieve my fear on this is introducing some sort of identifying invariant, where schema must have some identifier that labels it somehow, like how we set the platform/sku/etc when we connect to CHS, so we can trace bad data in cases that we must. We'd need to upload at least this piece of data, (if not the whole schema) to space metadata in order to leverage this, which I think we should.

mag-lt: > Something that would relieve my fear on this is introducing some sort of identifying invariant, where schema must have some identifier that labels it somehow, like how we set the platform/sku/etc when we connect to CHS, so we can trace bad data in cases that we must. We'd need to upload at least this piece of data, (if not the whole schema) to space metadata in order to leverage this, which I think we should.

I can't currently see how this relates to the discussion of how to author the schemas i.e. via JSON, programatically, or both. I think this might be something to discuss in our call tomorrow,

MAG-ElliotMorris: It's because my my fear of a programmatic interface is unknown data (ie, unknown where it comes from, it's on the servers, it's breaking things, etc). You can see how an identifier would help tracability, because you can ask people "Who's schema is this", and they probably know because they were required to identify it before we'd accept it.

mag-lt: I see what you mean about wanting some audit trail, my main point above was that whether the schema is authored in JSON or ComponentSchema doesn't make much of a difference, an author can write whatever they like, and both programatically, the JSON is no more blessed.

I won't address this here, but this is obviously part of a wider topic regarding storing these on space metadata etc.

PR #976, opened 2026-06-05
[OB-5368] fix: Address forward declaration issue

MAG-AdamThorn:
`LoginStateData` was being forward declared in `LoginState.h`, but this type was being returned by value via a getter as well as being taken by value in a setter. This was working because these methods were marked `NO_EXPORT` and because the internal call sites included the `LoginStateData` header.

I have addressed this issue by having the getter return a unique ptr and the setter take a const ref.

In addition I have made a couple of updates for const correctness.


MAG-ElliotMorris (approved):
Reasonable and good.

MAG-mv (approved):
Best solution with what we have imo. We discussed previously, but it would eb great to have private implementations and only expose interfaces to solve this problem

PR #977, opened 2026-06-05
[OF-1850] fix: Fix script event test

MAG-mv:
This was failing due to invalid exceptions being thrown from quicjs on Linux. Our theory is that this is happening de to us running scripts on different threads they were created on. We will test removing this again once we have handled our callbacks

### DIFF
--- a/ThirdParty/quickjs/src/quickjs.c
+++ b/ThirdParty/quickjs/src/quickjs.c
@@ -75,7 +75,7 @@
 #define CONFIG_ATOMICS
 #endif
 
-#if !defined(EMSCRIPTEN) && !defined(_MSC_VER)
+#if !defined(EMSCRIPTEN) && !defined(_MSC_VER) && !defined(CSP_DISABLE_STACK_CHECK_QJS)
 /* enable stack limitation */
 #define CONFIG_STACK_CHECK
 #endif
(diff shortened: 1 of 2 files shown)


ThirdParty/quickjs/src/quickjs.c
```diff
@@ -75,7 +75,7 @@
 #define CONFIG_ATOMICS
 #endif
 
-#if !defined(EMSCRIPTEN) && !defined(_MSC_VER)
+#if !defined(EMSCRIPTEN) && !defined(_MSC_VER) && !defined(CSP_DISABLE_STACK_CHECK_QJS)
```
mag-lt, line 78: [nit] My read of this is that the new manual override flag only applies to platforms where the feature is enabled by default, which I think is fine and probably as we intended, but might be worth adding a note in the comment in the CMakeLists.txt (as the flag seems like a global override, but is only conditionally used)

mag-lt (approved):

MAG-AdamThorn (approved):

ThirdParty/quickjs/src/quickjs.c
MAG-mv: Updated here: [3b83591](https://github.com/magnopus-opensource/connected-spaces-platform/pull/977/commits/3b83591ed36b80b58dcd66bf91377b432c2f8a79)

PR #979, opened 2026-06-08
[OB-5368] fix: Guard against race conditions caused by realtime engine destruction

MAG-AdamThorn:
The `OnlineRealtimeEngine::CreateRetrieveAllEntitiesCallback` method returns a callback which is used to process all Entities fetched when joining a Space. While we support paged retrieval of Entities we do not currently use that feature. This means that when entering a Space with many Entities, the callback can take a long time to execute.

If during this time the `OnlineRealtimeEngine` is destroyed, CSP will hit an exception when trying to access it within the callback body. There are a number of potential solutions to this issue, but the situation is complicated by the fact that we:
    1. Do not own the OnlineRealtimeEngine or manage its lifetime.
    2. We cannot block the main thread.

We have work planned to introduce a task queue to shepherd callbacks to the main thread which will make this issue a lot easier to address. In the meantime this change introduces a sentinel which is checked at key points in the body of the callback. This does not address the potential for a race condition but it does limit the window within which it may occur.

In addition to this change, while investigating race conditions caused by destruction of the realtime engine, I encountered an issue when running the new test I had introduced (`EnterSpaceTeardownDuringEntityFetchTest`) relating to LeadershipElection.

The issue occurred in the `BindOnElectedScopeLeaderCallback()` and `BindOnVacatedScopeLeaderCallback()` methods of the `MultiplayerConnection`. When entering a Space the user passes in the `OnlineRealtimeEngine`, which internally is passed to the `MultiplayerConnection` via `SetOnlineRealtimeEngine()`. The two bind methods would check that the `MultiplayerRealtimeEngine` member was not nullptr before calling a method on them. However, when the `OnlineRealtimeEngine` is destroyed, the ptr that `MultiplayerConnection` holds is not set to nullptr.

The fix was was to call the `SetOnlineRealtimeEngine()` method again in the `OnlineRealtimeEngine` dtor passing nullptr, just as we do in our existing `ExitSpace` flow.

### DIFF
--- a/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
+++ b/Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
@@ -520,6 +520,9 @@ private:
 
     std::recursive_mutex* TickEntitiesLock;
     std::mutex LeadershipElectionLock;
+    // Guards against a race condition whereby the OnlineRealtimeEngine dtor is called after the SignalR thread executing the
+    // CreateRetrieveAllEntitiesCallback callback has checked the cancellation token, but BEFORE it has finished executing the callback.
+    std::shared_ptr<std::mutex> EngineLifetimeGuard;
 
     std::deque<csp::multiplayer::SpaceEntity*>* PendingAdds;
     std::deque<csp::multiplayer::SpaceEntity*>* PendingRemoves;
--- a/Library/src/Multiplayer/OnlineRealtimeEngine.cpp
+++ b/Library/src/Multiplayer/OnlineRealtimeEngine.cpp
@@ -170,6 +170,7 @@ OnlineRealtimeEngine::OnlineRealtimeEngine()
     , EventHandler(nullptr)
     , ElectionManager(nullptr)
     , TickEntitiesLock(new std::recursive_mutex)
+    , EngineLifetimeGuard(std::make_shared<std::mutex>())
     , PendingAdds(nullptr)
     , PendingRemoves(nullptr)
     , PendingOutgoingUpdateUniqueSet(nullptr)
@@ -204,6 +205,7 @@ OnlineRealtimeEngine::OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerC
     , EventHandler(new SpaceEntityEventHandler(this))
     , ElectionManager(nullptr)
     , TickEntitiesLock(new std::recursive_mutex)
+    , EngineLifetimeGuard(std::make_shared<std::mutex>())
     , PendingAdds(new(std::deque<csp::multiplayer::SpaceEntity*>))
     , PendingRemoves(new(std::deque<csp::multiplayer::SpaceEntity*>))
     , PendingOutgoingUpdateUniqueSet(new(std::set<csp::multiplayer::SpaceEntity*>))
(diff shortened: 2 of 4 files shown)


Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -520,6 +520,9 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
 
     std::recursive_mutex* TickEntitiesLock;
     std::mutex LeadershipElectionLock;
+    // Guards against a race condition whereby the OnlineRealtimeEngine dtor is called after the SignalR thread executing the
+    // CreateRetrieveAllEntitiesCallback callback has checked the cancellation token, but BEFORE it has finished executing the callback.
+    std::shared_ptr<std::mutex> EngineLifetimeGuard;
```
mag-lt, line 525: [question] why is this a shared pointer to a mutex? As far as I can see the held data (the mutex) is never actually used, and we're just leaning on the shared/weak pointer machinery to determine if the object is still alive?

You can just use a shared pointer to void for this type of token.

As you've pointed out, as we could be during destruction when this condition is checked and we're not passing a shared pointer to this/the whole class, this isn't going to prevent the issue entirely (the whole object isn't kept alive by obtaining a shared pointer from the weak pointer). You could make this the ~~first~~ last member to mitigate this a bit more, but still this isn't a robust solution.

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -224,6 +224,11 @@ OnlineRealtimeEngine::OnlineRealtimeEngine(MultiplayerConnection& InMultiplayerC
 
 OnlineRealtimeEngine::~OnlineRealtimeEngine()
 {
+    // We currently have a number of potential race conditions caused by callbacks trying to access the OnlineRealtimeEngine after it has been
```
mag-lt, line 227: [question] This seems more about an issue with lifetimes than threading to me, do you agree this is the case or am I missing something? I sort of feel like this comment more accurately describes the situation with the shared pointer, as the solution there isn't fully robust (and so I had attributed the comment in the PR description about reducing the window to be to do with that vs this).

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
MAG-ElliotMorris: > (the whole object isn't kept alive by obtaining a shared pointer from the weak pointer). You could make this the first member to mitigate this a bit more, but still this isn't a robust solution.

Not a bad thought. Might shave a percent off the potential crash surface.
I'm pretty happy with this temporary bodge though, this will probably stop unreal crashing in the immediate future. So long as we remember it _is_ a temporary bodge that doesn't provide any formal guarentee.

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -734,6 +747,11 @@ std::function<void(const signalr::value&, std::exception_ptr)> OnlineRealtimeEng
 
         for (const auto& EntityMessage : Items)
         {
+            if (!EngineLifetimeGuardWeak.lock())
+            {
+                return;
+            }
```
MAG-ElliotMorris, line 753: [Consider] Taking some timings for space entry of a large space and a small space. This is an atomic operation and it's possible it could be a noticeable pessimization when done in a loop.

I don't expect it will be, but might be worth just checking were not 10x'ing space entry time accidentally. Would be nice to have a small suite of tests just for this sort of stuff.

MAG-ElliotMorris (approved):
I know it's rough, and doesn't guarantee anything, but I quite like this approach just as a matter of unblocking and moving forward.

I might want to comment above every single weak-ptr lock and return that this is a bodge and shouldn't stay in long-term. Just to be _extra_ sure we don't fall into a false sense of security. But 🤷.

Lucas's point on the shared-ptr to mutex being redundant should probably be addressed prior to merge.

mag-lt (approved):
We know this is duct tape, if you've experimentally verified that this improves the situation then landing this is better than not. I must admit I'm a bit skeptical, but I can see how checking at various points that the object has died will avoid the issue, but we're clearly in a tangled mess that needs much more dedicated focus to untangle (and is outside of the context of review here).

If I'm correct in my comment about the mutex [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/979#discussion_r3374041824), which seems to be the case, please address before merging.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
mag-lt: turning this into a [fix], the mutex is misleading. I've already approved at this point, but conditionally (forgive the slightly awkward way of signalling that).

MAG-AdamThorn: Ah this is my fault. I had it as a `shared_ptr` to a mutex as in prior iterations I was doing something with that mutex. I no-longer am though so will change that to a void. Thanks.

mag-lt: > You could make this the first member to mitigate this a bit more

This isn't a good suggestion, I take it back. Making it the last member is probably the best bet (short of using `shared_from_this` on the class, or bagging up all the data into an impl struct and holding that by shared ptr) i.e. it signals we have started destruction. Anything else means any of the other members could've destructed already, making it not a very reliable proxy for the whole class.

PR #980, opened 2026-06-08
[OF-1878] feat: Add support for federated login

MAG-AdamThorn:
With the new federated login flow clients will be authenticating directly with MCS. They will not be authenticating through CSP as they currently do. This creates a problem, as with our current login flows we create the LoginState object via `LoginState::OnResponse`. This data is then used to; refresh the token, for secure asset download, when starting the Multiplayer Connection among other things.

To support federated login CSP needs to add a new `SetLoginDetails()` method that clients can call to pass the base64 encoded AuthDto json object to CSP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file. For compile
   practice a single element containing the full JSON array is expected. Equivalent constructors accepting
   `Array<ComponentSchema>` directly are also provided.
 
+- [OF-1878] feat: Add support for federated login. @MAG-AdamThorn.
+  A new federated login flow is being added that will allow client applications to authenticate directly with MCS. They will not be authenticating through CSP as they currently do. This creates a problem, as with our current login flows we create the LoginState object via `LoginState::OnResponse`. This data is then used to; refresh the token, for secure asset download, when starting the Multiplayer Connection among other things. To support federated login CSP has added a new `UserSystem::SetLoginDetails()` method that clients can call to pass the base64 encoded AuthDto json object to CSP.
+
 ### 🔥 ❗Breaking Changes
 
 - [OB-5368] fix!: Guard access to LoginState data with a mutex. By @MAG-AdamThorn.
--- a/Library/include/CSP/Systems/Users/Authentication.h
+++ b/Library/include/CSP/Systems/Users/Authentication.h
@@ -70,9 +70,11 @@ public:
         : csp::systems::ResultBase(ResCode, HttpResCode)
         , State(nullptr) {}
 
+    LoginStateResult(csp::common::LoginState* LoginState);
+
 private:
     LoginStateResult();
-    LoginStateResult(csp::common::LoginState* LoginState);
+
 
     CSP_NO_EXPORT void OnResponse(const csp::services::ApiResponseBase* ApiResponse) override;
 
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -206,8 +206,10 @@ public:
 
     /// @brief Set Login details after using federated login to authenticate with MCS.
     /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
-    /// @return bool : true if the login details were successfully set, false if there was an error parsing the json or setting the login details.
-    bool SetLoginDetails(const csp::common::String& LoginDetailsJson);
+    /// @param CreateMultiplayerConnection : Whether to create a multiplayer connection. If false, this session will not establish a SignalR.
+    /// @param Callback : callback that contains the result of the 3rd party authentication operation.
+    CSP_ASYNC_RESULT void SetLoginDetails(
+        const csp::common::String& LoginDetailsJson, bool CreateMultiplayerConnection, LoginStateResultCallback Callback);
 
     /// @brief Logout from Magnopus Cloud Services.
     /// @param Callback NullResultCallback : callback to call when a response is received
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -204,6 +204,11 @@ public:
         const csp::common::String& ThirdPartyToken, const csp::common::Optional<EThirdPartyPlatform>& ClientType, bool CreateMultiplayerConnection,
         const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
+    /// @brief Set Login details after using federated login to authenticate with MCS.
+    /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
+    /// @return bool : true if the login details were successfully set, false if there was an error parsing the json or setting the login details.
+    bool SetLoginDetails(const csp::common::String& LoginDetailsJson);
+
     /// @brief Logout from Magnopus Cloud Services.
     /// @param Callback NullResultCallback : callback to call when a response is received
     CSP_ASYNC_RESULT void Logout(NullResultCallback Callback);
--- a/Library/include/CSP/Systems/Users/UserSystem.h
+++ b/Library/include/CSP/Systems/Users/UserSystem.h
@@ -206,8 +206,10 @@ public:
 
     /// @brief Set Login details after using federated login to authenticate with MCS.
     /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
-    /// @param CreateMultiplayerConnection : Whether to create a multiplayer connection. If false, this session will not establish a SignalR.
-    /// @param Callback : callback that contains the result of the 3rd party authentication operation.
+    /// @param CreateMultiplayerConnection : Whether to create a multiplayer connection. If false, this session will not establish a SignalR
+    /// connection to backend services, and thus be unable to receive messages or events. This session will also be unable to enter online spaces via
+    /// a csp::multiplayer::OnlineRealtimeEngine. If true, this session will receive events, and may enter both online and offline spaces.
+    /// @param Callback : callback when asynchronous task finishes
     CSP_ASYNC_RESULT void SetLoginDetails(
         const csp::common::String& LoginDetailsJson, bool CreateMultiplayerConnection, LoginStateResultCallback Callback);
 
--- a/Library/src/Common/LoginStateData.cpp
+++ b/Library/src/Common/LoginStateData.cpp
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2026 Magnopus LLC
+
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "LoginStateData.h"
+
+#include "CSP/Common/Settings.h"
+#include "Common/Convert.h"
+#include "Common/DateTime.h"
+#include "Debug/Logging.h"
+#include "Services/ApiBase/ApiBase.h"
+#include "Services/UserService/Dto.h"
+
+namespace
+{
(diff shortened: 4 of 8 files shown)


MAG-SangeetShah (commented):
[Question] Once the login details are set, what login method do clients call? Currently, all the Login methods available would require clients to decode the base64 string (to extract tokens & user IDs for example) in order to log in.

[Consider] A Login method that takes the the base64 string as a parameter and uses it to log the user in directly (the clients also don't need to set the LoginStateData via a separate method)

MAG-AdamThorn:
> [Question] Once the login details are set, what login method do clients call? Currently, all the Login methods available would require clients to decode the base64 string (to extract tokens & user IDs for example) in order to log in.
> 
> [Consider] A Login method that takes the the base64 string as a parameter and uses it to log the user in directly (the clients also don't need to set the LoginStateData via a separate method)

This seems to have been a common misunderstanding. You will be authenticating with MCS via the federated login page. Once you have done that you will be authenticated with MCS and will not need to call a CSP login method. However, CSP obviously won't know anything about it, so you will call the new csp endpoint passing the base64 encoded json string that gets returned with the redirect url.

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -204,6 +204,13 @@ class CSP_API UserSystem : public SystemBase
         const csp::common::String& ThirdPartyToken, const csp::common::Optional<EThirdPartyPlatform>& ClientType, bool CreateMultiplayerConnection,
         const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
+    /// @brief Set Login details after using federated login to authenticate with MCS.
+    /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
+    /// @param CreateMultiplayerConnection : Whether to create a multiplayer connection. If false, this session will not establish a SignalR.
+    /// @param Callback : callback that contains the result of the 3rd party authentication operation.
```
MAG-ElliotMorris: [Question] Is the callback doc misleading? Are we doing authentication, or merely setting up CSP state?

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -204,6 +204,13 @@ class CSP_API UserSystem : public SystemBase
         const csp::common::String& ThirdPartyToken, const csp::common::Optional<EThirdPartyPlatform>& ClientType, bool CreateMultiplayerConnection,
         const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
+    /// @brief Set Login details after using federated login to authenticate with MCS.
+    /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
+    /// @param CreateMultiplayerConnection : Whether to create a multiplayer connection. If false, this session will not establish a SignalR.
```
MAG-ElliotMorris: [fix] Finish the doc, "will not establish a SignalR" ... presumably connection?

Library/include/CSP/Systems/Users/UserSystem.h
```diff
@@ -204,6 +204,13 @@ class CSP_API UserSystem : public SystemBase
         const csp::common::String& ThirdPartyToken, const csp::common::Optional<EThirdPartyPlatform>& ClientType, bool CreateMultiplayerConnection,
         const csp::common::Optional<bool>& UserHasVerifiedAge, LoginStateResultCallback Callback);
 
+    /// @brief Set Login details after using federated login to authenticate with MCS.
+    /// @param LoginDetailsJson : A json string containing the login details returned from the federated login.
```
MAG-ElliotMorris, line 208: [Consider] Any way to be more specific here. In particular I'd like to know what encoding we expect, and a bit more detail on what federated login we're talking about. Even "The magnopus federated login service" would serve.

If we don't know the encoding, what about "The encoded json received from a successful login to the Magnopus federated login service."?

Library/src/Common/LoginStateData.cpp
```diff
@@ -0,0 +1,111 @@
...
+    {
+        CSP_LOG_FORMAT(LogLevel::Error, "RefreshToken Expired: %s %s", AuthResponse->GetRefreshToken().c_str(),
+            AuthResponse->GetRefreshTokenExpiresAt().c_str());
+
+        return std::nullopt;
+    }
+
+    Data.SetAccessTokenRefreshTime(RefreshTime);
+
+    return std::make_optional<LoginStateData>(Data);
```
MAG-ElliotMorris, line 108: [nit] std::move(Data)

Library/src/Systems/Users/Authentication.cpp
```diff
@@ -90,65 +64,20 @@ void LoginStateResult::OnResponse(const services::ApiResponseBase* ApiResponse)
...
-
-            if (CurrentTime >= Expiry)
-            {
-                CSP_LOG_FORMAT(LogLevel::Error, "AccessToken Expired: %s %s", AuthResponse->GetAccessToken().c_str(),
-                    AuthResponse->GetAccessTokenExpiresAt().c_str());
+            auto Data = *DataOpt;
 
-                return;
-            }
+            State->SetLoginStateDataThreadSafe(Data);
```
MAG-ElliotMorris, line 76: [nit] Redundant copy, although not that important. could either use `auto&` or `std::move`. Move probably won't _actually_ avoid the copy knowing our types ... but you know, in theory it's nicer.

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -789,6 +790,100 @@ void UserSystem::LoginToThirdPartyAuthenticationProviderWithToken(EThirdPartyAut
     LoginSocial(AuthenticationAPI, CurrentLoginState, LogSystem, Request, CreateMultiplayerConnection, Callback);
 }
 
+void UserSystem::SetLoginDetails(const csp::common::String& LoginDetailsJson, bool CreateMultiplayerConnection, LoginStateResultCallback Callback)
```
MAG-ElliotMorris, line 793: [consider] This is griping over names, but I might just call this `FederatedLogin()`, `SetLoginDetails` is so specific to the implementation it's pretty useless. Reads like a utility setter.

Then you could call the string `FederatedLoginJsonToken` or something as well, which is also helpful.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file. For compile
   practice a single element containing the full JSON array is expected. Equivalent constructors accepting
   `Array<ComponentSchema>` directly are also provided.
 
+- [OF-1878] feat: Add support for federated login. @MAG-AdamThorn.
```
MAG-ElliotMorris, line 18: [nit] Could add your name here too. @MAG-mv :P

Library/include/CSP/Systems/Users/Authentication.h
```diff
@@ -70,9 +70,11 @@ class CSP_API LoginStateResult : public ResultBase
         : csp::systems::ResultBase(ResCode, HttpResCode)
         , State(nullptr) {}
 
+    LoginStateResult(csp::common::LoginState* LoginState);
```
mag-lt, line 73: [nit] `CSP_NO_EXPORT`

MAG-ElliotMorris (changes requested):
Looks correct enough, but just eyeballing aint that great.

Obviously I do not at all like the emergency reactiveness of this, we've not tested this enough. Nonetheless, it's the right thing to do for the moment.

Library/src/Common/LoginStateData.h
```diff
@@ -96,4 +102,6 @@ class LoginStateData
     csp::common::DateTime AccessTokenRefreshTime;
 };
 
+std::optional<LoginStateData> AuthDtoToLoginStateData(const csp::services::generated::userservice::AuthDto* AuthResponse);
```
mag-lt, line 105: [nit] take by const ref vs pointer

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -789,6 +790,100 @@ void UserSystem::LoginToThirdPartyAuthenticationProviderWithToken(EThirdPartyAut
...
+{
+    chs_user::AuthDto AuthDetails;
+
+    AuthDetails.FromJson(LoginDetailsJson);
+
+    if (AuthDetails.GetAccessToken().IsEmpty())
```
mag-lt, line 799: [nit] `FromJson` isn't pulling its weight here. I know this is in the generated services layer, so probably not worth trying to fix that now, but we might [consider] rolling these checks into `AuthDtoToLoginStateData`?

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -789,6 +790,100 @@ void UserSystem::LoginToThirdPartyAuthenticationProviderWithToken(EThirdPartyAut
...
+        csp::systems::LoginStateResult BadResult;
+        BadResult.SetResult(csp::systems::EResultCode::Failed, (uint16_t)csp::web::EResponseCodes::ResponseBadRequest);
+        Callback(BadResult);
+
+        return;
+    }
+
+    auto Data = *DataOpt;
+
+    CurrentLoginState->SetLoginStateDataThreadSafe(Data);
```
mag-lt, line 856: [nit] from here down to `SystemsManager::Get().GetUserSystem()->NotifyRefreshTokenHasChanged()` on line 866 are duplicated from `LoginStateResult::OnResponse`. Feels like there could be a common place to handle all of this stuff, though I appreciate that may be a larger refactor outside the scope of these changes.

Arguably as all this code is now in the user system, and that `OnResponse` eventually calls into the user system, this could just be the central place

Library/src/Systems/Users/UserSystem.cpp
```diff
@@ -789,6 +790,100 @@ void UserSystem::LoginToThirdPartyAuthenticationProviderWithToken(EThirdPartyAut
...
+
+    web::HttpAuth::SetAccessToken(
+        AuthDetails.GetAccessToken(), AuthDetails.GetAccessTokenExpiresAt(), AuthDetails.GetRefreshToken(), AuthDetails.GetRefreshTokenExpiresAt());
+
+    // Signal login to anyone interested
+    events::Event* LoginEvent = events::EventSystem::Get().AllocateEvent(events::USERSERVICE_LOGIN_EVENT_ID);
+    LoginEvent->AddString("UserId", AuthDetails.GetUserId());
+    events::EventSystem::Get().EnqueueEvent(LoginEvent);
+
+    SystemsManager::Get().GetUserSystem()->NotifyRefreshTokenHasChanged();
```
mag-lt, line 866: [nit] using a singleton to get... ourselves? we are in a method of the user system right now... relates to the above comment about duplication. Functionally this works, but this should probably be [fix]ed

Library/include/CSP/Systems/Users/UserSystem.h
mag-lt: [nit] arguably not 3rd party as it'll use our central login?

mag-lt (approved):
A few minor comments, but won't change anything functionally so approving regardless. I can't help but feel there is a way to centralise a lot of this logic, but very likely we won't find it today.

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Yeah, its not great tbf

MAG-mv: probably should just make it the same as the majority of the other callback docs for now

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/980/commits/ecaf0862416fd21df8561058dea9520d31279642``

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/980/commits/ecaf0862416fd21df8561058dea9520d31279642

MAG-mv (approved):

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: https://github.com/magnopus-opensource/connected-spaces-platform/pull/980/commits/ecaf0862416fd21df8561058dea9520d31279642

Library/include/CSP/Systems/Users/UserSystem.h
MAG-mv: https://github.com/magnopus-opensource/connected-spaces-platform/pull/980/commits/ecaf0862416fd21df8561058dea9520d31279642

MAG-ElliotMorris (approved):
Do not like the rush we have done to get this in.

PR #982, opened 2026-06-09
[NT-0] fix!: Rename to FederatedLogin, Fix result + optional accesses

MAG-ElliotMorris:
We rushed the initial change through a bit. Add some tests and work out some edge case errors, including potential crashes in the failure case.

Rename from `SetLoginDetails` -> `FederatedLogin`, as well as expand the documentation a tad.


MAG-ElliotMorris:
Probably will close this and solve these problems all together. There's issues with refresh token login too.

MAG-mv (approved):

mag-lt (approved):

PR #985, opened 2026-06-10
[OF-1840] Add PR checks workflow using Github Actions

MAG-mv:
This PR sets up a PR check workflow using Github Actions by building all of our supported platforms except wasm which is included in the next PR.

It then runs our Linux RelWithDebInfo tests. A comment was added explaining the reason we used Linux over Windows - Essentially the GitHub hosted Windows runners does not give you that same Linux container runtime that Dockwer Desktop does.

- Also adds missing iOS and visionIS Conan host profiles
- Also fixes a couple of bugs

### DIFF
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -57,12 +57,9 @@ runs:
         restore-keys: |
           ${{ env.CONAN_CACHE_PREFIX }}-
 
-    - name: Set CMake compatibility policy
-      shell: bash
-      run: |
-        echo "CMAKE_POLICY_VERSION_MINIMUM=3.5" >> "$GITHUB_ENV"
-
     - name: Conan install
+      env:
+        CMAKE_POLICY_VERSION_MINIMUM: "3.5"
       shell: bash
       run: |
         set -euxo pipefail
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -1,0 +1,99 @@
+name: "Build and Install"
+description: "Fetch dependencies with Conan, configure CMake, build, and install"
+
+inputs:
+  host_profile:
+    description: "The name of the Conan host profile to use. See profiles/host"
+    required: true
+
+  build_type:
+    description: "CMake build type [Debug, Release, RelWithDebInfo]"
+    required: true
+
+  configure_preset:
+    description: "CMake configure preset to use. Run cmake --list-presets to list available presets."
+    required: true
+
+  build_preset:
+    description: "CMake build preset to use. Run cmake --build --list-presets to list available presets."
+    required: true
+
+  build_folder:
+    description: "CMake build directory to install from. This can change depending on the generator used."
+    required: true
+
+  output_folder:
+    description: "Conan output folder"
+    required: true
+    default: ""
+
+  install_prefix:
+    description: "The location the CSP library will be installed to"
+    required: false
+    default: "install"
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: Install Conan
+      uses: conan-io/setup-conan@v1
+
+    # Setup Conan cache.
+    # Use a custom cache key instead of setup-conan's default key so matrix jobs do not conflict.
+    #
+    # The cache prefix uses: host_profile  and build_preset, which
+    # separates caches by platform, build type, and linkage.
+    #
+    # We also append a hash of our Conan recipe, host profile, and CMake presets so changes
+    # to these files creates a new cache key.
+    - name: Cache Conan packages
+      uses: actions/cache@v4
+      env:
+        CONAN_CACHE_PREFIX: conan-${{ inputs.host_profile }}-${{ inputs.build_preset }}
+      with:
+        path: ~/.conan2
+        key: ${{ env.CONAN_CACHE_PREFIX }}-${{ hashFiles('conanfile.py', format('profiles/host/{0}', inputs.host_profile)) }}
+        restore-keys: |
+          ${{ env.CONAN_CACHE_PREFIX }}-
+
+    - name: Conan install
+      shell: bash
+      run: |
+        set -eux
+
+        export CMAKE_POLICY_VERSION_MINIMUM="3.5"
+
+        PROFILE_PATH="profiles/host/${{ inputs.host_profile }}"
+
+        if [ ! -f "$PROFILE_PATH" ]; then
+          echo "Host profile not found: $PROFILE_PATH"
+          exit 1
+        fi
+
+        conan install . \
+          -of "${{ inputs.output_folder }}" \
+          -s build_type="${{ inputs.build_type }}" \
+          --build=missing \
+          --profile:host="$PROFILE_PATH"
+
+    - name: Configure CMake
+      shell: bash
+      run: |
+        set -eux
+        cmake --preset "${{ inputs.configure_preset }}"
+
+    - name: Build
+      shell: bash
+      run: |
+        set -eux
+        cmake --build --preset "${{ inputs.build_preset }}"
+
+    - name: Install
+      shell: bash
+      run: |
+        set -eux
+        cmake --install \
+          "${{ github.workspace }}/${{ inputs.build_folder }}" \
+          --config "${{ inputs.build_type }}" \
+          --prefix "${{ github.workspace }}/${{ inputs.install_prefix }}"
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -65,7 +65,7 @@ runs:
     - name: Conan install
       shell: bash
       run: |
-        set -eux
+        set -euxo pipefail
 
         PROFILE_PATH="profiles/host/${{ inputs.host_profile }}"
 
@@ -83,19 +83,19 @@ runs:
     - name: Configure CMake
       shell: bash
       run: |
-        set -eux
+        set -euxo pipefail
         cmake --preset "${{ inputs.configure_preset }}"
 
     - name: Build
       shell: bash
       run: |
-        set -eux
+        set -euxo pipefail
         cmake --build --preset "${{ inputs.build_preset }}"
 
     - name: Install
       shell: bash
       run: |
-        set -eux
+        set -euxo pipefail
         cmake --install \
           "${{ github.workspace }}/${{ inputs.build_folder }}" \
           --config "${{ inputs.build_type }}" \
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -57,12 +57,15 @@ runs:
         restore-keys: |
           ${{ env.CONAN_CACHE_PREFIX }}-
 
+    - name: Set CMake compatibility policy
+      shell: bash
+      run: |
+        echo "CMAKE_POLICY_VERSION_MINIMUM=3.5" >> "$GITHUB_ENV"
+
     - name: Conan install
       shell: bash
       run: |
         set -eux
-
-        export CMAKE_POLICY_VERSION_MINIMUM="3.5"
 
         PROFILE_PATH="profiles/host/${{ inputs.host_profile }}"
 
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -126,3 +126,46 @@ jobs:
           output_folder: build/emscripten-${{ matrix.config.preset_prefix }}
           build_folder: build/emscripten-${{ matrix.config.preset_prefix }}/build/${{ matrix.config.build_type }}
           install_prefix: install/emscripten/${{ matrix.config.preset_prefix }}
+  
+  build-android:
+    name: Build android ${{ matrix.config.build_type }} ${{ matrix.linkage.name }}
+    runs-on: ubuntu-latest
+    
+    strategy:
+      fail-fast: false
+      
+      matrix:
+        config:
+          - build_type: Debug
+            preset_prefix: debug
+
+          - build_type: Release
+            preset_prefix: release
+
+          - build_type: RelWithDebInfo
+            preset_prefix: relwithdebinfo
+
+        linkage:
+          - name: static
+            shared: false
+
+          - name: shared
+            shared: true
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          # Ensure we fetch tags so our automatic versioning script picks up the version.
+          fetch-tags: true
+
+      - name: Build and install
+        uses: ./.github/actions/build-and-install
+        with:
+          host_profile: android_arm
+          build_type: ${{ matrix.config.build_type }}
+          configure_preset: ${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
+          build_preset: ${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
+          output_folder: build/android-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
+          build_folder: build/android-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}/build/${{ matrix.config.build_type }}
+          install_prefix: install/android/${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -3,6 +3,10 @@ name: PR Checks
 # This workflow performs the following actions:
 #  - Builds all desktop platforms
 #  - Uploads the install tree as a workflow artifact
+#  - Builds csp as emscripten static lib
+#  - Builds csp for android
+#  - Builds csp for iOS and visionOS
+#  - Runs csp locsl tests on Linux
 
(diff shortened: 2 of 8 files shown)


.github/actions/build-and-install/action.yml
```diff
@@ -0,0 +1,90 @@
...
+    required: false
+    default: ""
+
+  install_prefix:
+    description: "Install prefix"
+    required: false
+    default: "install"
+
+runs:
+  using: "composite"
```
MAG-ElliotMorris, line 36: [Comment] A tad disappointed to see this as a composite action rather than an encapsulated reusable workflow that lets you build CSP in any configuration from any commit. Seems like a bit of a waste not to have gone the extra mile.

.github/actions/build-and-install/action.yml
```diff
@@ -0,0 +1,93 @@
...
+
+  output_folder:
+    description: "Conan output folder"
+    required: false
+    default: ""
+
+  install_prefix:
+    description: "Install prefix"
+    required: false
+    default: "install"
```
MAG-ElliotMorris, line 33: [Consider] I don't feel "Install prefix" is not a good enough description for a variable called `install_prefix`, it demands the user know what a cmake install prefix is, rather than saying something like "The location the CSP library will be installed to".

.github/actions/build-and-install/action.yml
```diff
@@ -0,0 +1,93 @@
...
+    description: "CMake build type [Debug, Release, RelWithDebInfo]"
+    required: true
+
+  configure_preset:
+    description: "CMake configure preset to use. Run cmake --list-presets to list available presets."
+    required: true
+
+  build_preset:
+    description: "CMake build preset to use. Run cmake --build --list-presets to list available presets."
+    required: true
```
MAG-ElliotMorris, line 19: [Consider] It might be advisable to present a dependency inverted interface here, with a known platform list and build mode list, rather than directly asking for the configure and build presets.

Firstly because you then don't require users to leave the actions context to figure out what to pass (good for workflow_dispatch especially)

Secondly then if we change the presets, there is only one point of change in the actions, rather than having to go to every call site, including potential cross-repo ones. This is the same sort of encapsulation desire from some other comments.

I'm dubious on this myself, it's sort of another expression of desire for different encapsulation, so might not be a good idea if we're not doing that.

.github/actions/build-and-install/action.yml
```diff
@@ -0,0 +1,93 @@
...
+          conan install . \
+            -of "$OUTPUT_FOLDER" \
+            -s build_type="${{ inputs.build_type }}" \
+            --build=missing \
+            --profile:host="$PROFILE_PATH"
+        else
+          conan install . \
+            -s build_type="${{ inputs.build_type }}" \
+            --build=missing \
+            --profile:host="$PROFILE_PATH"
```
MAG-ElliotMorris: [Question] what's the point of this branch, or the parameter in general? Dubious about any branching in an inline bash script, this one is too large for my liking.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+          build_folder: build/${{ matrix.platform.name }}-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}/build/${{ matrix.config.build_type }}
+          install_prefix: install/${{ matrix.platform.name }}/${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
+
+      - name: Upload build
+        if: matrix.config.preset_prefix == 'relwithdebinfo' && matrix.linkage.name == 'static'
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.platform.name }}-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}-build
+          path: build/${{ matrix.platform.name }}-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}/build/${{ matrix.config.build_type }}
+          if-no-files-found: error
```
MAG-ElliotMorris, line 93: [Consider] A shame to be doing this here imo. It's encapsulation breaking, this should be a concern of the build automation, not the checks.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+    
+  build-emscripten:
+    name: Build emscripten ${{ matrix.config.build_type }}
+    runs-on: windows-latest
+    
+    env:
+      EMSDK_VERSION: 5.0.2
```
MAG-ElliotMorris, line 100: [Comment] Would have been pleasant to include deduplicating this decision in this work. Another encapsulation break imo, this is the sort of thing that should be defined in a centralised build job.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+  tests-local-mcs:
+    name: Run Tests (Local MCS) Linux
+    ## Why Local MCS cannot run on GitHub Windows runners
+
+    # MCS Docker stack uses Linux container images.
+    # GitHub windows-2022 runners do not provide the same Docker Desktop Linux-container environment.
+    # When Docker runs there, it attempts to pull Windows container images.
+    # Because the MCS images are Linux images, Docker fails with "no matching manifest for windows(10.0.20348)/amd64"
+    # For this reason, we currently run our local tests on Linux.
+    runs-on: ubuntu-latest
```
MAG-ElliotMorris: [Consider] Think we should pin specific runner versions?

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+        uses: docker/setup-docker-action@v5
+
+      - name: Set up Docker Compose
+        uses: docker/setup-compose-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v4
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
MAG-ElliotMorris, line 284: [Question] What did the credential end up being on this? Are we using the CHS controlled account, or did you close the loop on getting something we control?

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+
+      - name: Docker diagnostics
+        shell: bash
+        run: |
+          docker version
+          docker info
+          docker compose version
+
+      - name: Pull CHS
```
MAG-ElliotMorris: [nit] Slightly inappropriate name, given that this also starts CHS.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+
+          git lfs ls-files
+          git lfs pull
+          git lfs checkout
+
+          cd "${{ github.workspace }}/build/linux-relwithdebinfo-static/build/RelWithDebInfo"
+
+          chmod +x ./MultiplayerTestRunner/MultiplayerTestRunner
+          chmod +x ./csp-tests
+          ./csp-tests
```
MAG-ElliotMorris, line 326: [Consider] It would be better if the git pull and the test was separate tasks, so we could see what fails more easily. We should really try and avoid inline bash scripts getting long as they become impossible to understand where errors are coming from, get over 4 lines and I start to worry.

MAG-ElliotMorris (approved):
First off, if you haven't ensured that any given step fails in the actions UI when it fails, please do so. I'm thinking of the bash steps in particular. This constantly catches me out and becomes a pain later.

Two cross cutting points worth further consideration. It would be good to talk about these prior to merge imo, a bit afeared that if we don't do them now we never will.

- Surprised not to see any caching, especially for conan dependencies. I think this is rather important. Perhaps I missed it.
- As you can tell from some of my comments, I was hoping for a different sort of encapsulation of the build jobs that would grant greater flexibility across the whole organization. This implementation seems focused on solving our issues, which is all that was asked, but it's a shame not to go a step further imo.

Other than that, this seems solid. Well done for solving all the issues that cropped up.

Tests/src/PublicAPITests/ComponentTests/ComponentTests.cpp
```diff
@@ -366,6 +366,10 @@ CSP_PUBLIC_TEST(CSPEngine, ComponentTests, SchemaComponentRoundtrip)
         Engine.ProcessPendingEntityOperations();
 
         AWAIT(SpaceSystem, ExitSpace);
+
+        // Ensure component data has been written to database by chs before entering the space again
+        // This is due to an enforced 2 second chs database write delay
+        std::this_thread::sleep_for(7s);
```
mag-lt, line 388: [Comment] I'd missed this, apologies. Generally, I'd like to try to find a way to avoid sleeping in tests (7 seconds is particularly harsh too, this adds up quickly), which may not be possible without API changes. For these types of things, I can imagine somehow setting up to be notified when the corresponding change happens and waiting on a timeout (perhaps already possible with self messaging?), or perhaps polling in a loop until a particular state is observed (again, with a timeout), or something else entirely. In any case, not suggesting doing anything more here

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+          docker info
+          docker compose version
+
+      - name: Pull CHS
+        shell: bash
+        env:
+          COMPOSE_ANSI: never
+          COMPOSE_PROGRESS: plain
+        run: |
+          set -euxo pipefail
```
mag-lt: [Question] I'm interested to know why we set these settings explicitly, and in a few different ways amongst the configs / different scripts. Has each option been added deliberately in the cases that it is there, and deliberately omitted in others (i.e. to not expose secrets?), or is it somewhat "cargo culted"? 

I've seen the following across the diff:
* `set -euxo pipefail`, once, here
* `set -eux`, ~7 times in various places
* No explicit options set, ~4 places

Apparently github by default sets `bash --noprofile --norc -eo pipefail {0}` when `shell: bash` is specified (see [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_iddefaultsrunshell)), so that'll be running for all bash commands already.

So, if I understand correctly we are sometimes additionally adding `-u` and `-x`? The former treats unset env vars as errors, and the latter prints every command before executing it. Why do we only sometimes want to do this?

.github/actions/build-and-install/action.yml
```diff
@@ -0,0 +1,93 @@
...
+  using: "composite"
+
+  steps:
+    - name: Install Conan
+      uses: conan-io/setup-conan@v1
+
+    - name: Set CMake compatibility policy
+      shell: bash
+      run: |
+        echo "CMAKE_POLICY_VERSION_MINIMUM=3.5" >> "$GITHUB_ENV"
```
mag-lt: [nit] This could probably be in an [`env` block](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#env) in the `Configure CMake` step? There are a couple of other similar cases using `>> "$GITHUB_ENV"` where that may also be true. Either works of course.

.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,323 @@
...
+          - name: windows
+            runner: windows-2022
+            host_profile: windows_ninja
+            
+          - name: linux
+            runner: ubuntu-latest
+            host_profile: linux_gcc
+
+          - name: macos
+            runner: macos-latest
```
mag-lt: [question] if we don't fix the version, how can we be sure that the toolchain doesn't change out from underneath us (i.e. the compiler could change, minimum OS SDK could change etc)? Are you confident conan + cmake is taking care of everything we care about here? Even if we are being explicit about min os / sdk target (which I'm not sure we are), I'm pretty sure the version of Xcode could change out underneath us, which is probably something we want to be in control of.

CMakeLists.txt
```diff
@@ -414,6 +414,10 @@ if(BUILD_TESTING)
       -Wextra
       -Wno-error=changes-meaning
       -Wno-error=deprecated-declarations
+      # uninitialized and maybe-uninitialized are needed our test project.
```
mag-lt, line 417: [Comment] It is a shame to apply this so broadly. This is the sort of thing that could be solved in a more specific way with compiler specific pragmas within the source text, wrapping the offending include (so, `gtest/gtest.h`? which could be moved to a separate header with the pragmas surrounding it, and then that header included instead in all places gtest/gtest.h is mentioned). I can't help but think a later version of gtest has probably also addressed whatever the issues were.

Didn't you work around similar issues before by treating some library headers as system headers?

mag-lt (approved):
I'm not very well versed in the specifics of Github Actions, so this PR and particularly @MAG-ElliotMorris's feedback were educational to me.

With the knowledge I do have, I don't see any particular blockers here, but have left some comments that would be worth clarifying before merging.

.github/workflows/pr-checks.yml
mag-lt: [Consider] Similar to the above, splitting this up has some other benefits: most of this seems platform agnostic, other than the very unix-y bit at the end with those `chmod +x` permissions (which seem odd, i'd have assumed the compiler would've spat something out that was executable?)

.github/actions/build-and-install/action.yml
MAG-mv: Fixed up here: https://github.com/magnopus-opensource/connected-spaces-platform/commit/2e0fcd75953f851e9a2f6aacada7d653b7e8aae8

.github/actions/build-and-install/action.yml
MAG-mv: This was to allow us to have an optional output folder. I have made this required now and removed the branch. Fixed up here: https://github.com/magnopus-opensource/connected-spaces-platform/commit/2e0fcd75953f851e9a2f6aacada7d653b7e8aae8

.github/workflows/pr-checks.yml
MAG-mv: Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/changes/ea0d2ad827689befd1f759e5ebe6cf787ab9e586

.github/workflows/pr-checks.yml
MAG-mv: Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/commits/26e6b02a28fc6b7d0f9b1c838ad9fe2c52c3b156

.github/workflows/pr-checks.yml
MAG-mv: lfs calls here were a mistake and fixed up here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/changes/f033fa2c168621b4b62d2c007f784dff080ebc83

Tests/src/PublicAPITests/ComponentTests/ComponentTests.cpp
MAG-mv: Agreed, this is aweful. This is ho we handle disconnection and reconnection all our other tests. I'm not sure the best way to address this at the moment

.github/workflows/pr-checks.yml
MAG-mv: I reviewed the bash shell options here. https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/commits/1fbb3c6a6d9293b6dee660720463ff625bcd7967

I added my reasoning in the commit description

CMakeLists.txt
MAG-mv: Yeah I dont like this but couldnt find a better solution that could be addressed easily.

I think fixing the way we use gtest and updating to a newer version is the best fix

.github/workflows/pr-checks.yml
MAG-mv: Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/commits/ea0d2ad827689befd1f759e5ebe6cf787ab9e586

.github/actions/build-and-install/action.yml
MAG-mv: We're going to address this when we create a reusable workflow

.github/workflows/pr-checks.yml
MAG-mv: We're going to address this when we create a reusable workflow

.github/workflows/pr-checks.yml
MAG-mv: We're going to address this when we create a reusable workflow

.github/actions/build-and-install/action.yml
MAG-mv: Fixed here: [e0dc969](https://github.com/magnopus-opensource/connected-spaces-platform/pull/985/commits/e0dc9691dbce538c56848643d3c8d8f455fc02f0)

.github/workflows/pr-checks.yml
mag-lt: Nice. I think we might need to do more if want want to sure a specific version of Xcode for example, but I'm not sure how much flexibility we have over that. This is fine for now I think.

.github/workflows/pr-checks.yml
MAG-mv: Using the jenkins ones. I'll get an IT ticket up to make us a magnopus docker account

mag-lt (approved):

PR #986, opened 2026-06-10
[NT-0] test: Refresh Token Test

MAG-ElliotMorris:
Add an automated refresh token test. Patching a hole in our test coverage.

~~Whilst shutting down/restarting CSP as part of this test, I noted there was some bad behaviour with a completely redundant shared ptr in Systems Manager. I have deleted it to prevent flakiness.~~
Actually there isn't, I think I was misinterpreting a failure from an earlier version of the test function that had a use after free. Have rerun the tests both with and without the removal 20x and can find no crashes.

Nonetheless, going to merge this, as it's a redundant unnecesary data member. Will split it out if anyone feels super strongly but tbh going through the whole PR/CI cycle is _ugh_

### DIFF
--- a/Library/src/Systems/SystemsManager.cpp
+++ b/Library/src/Systems/SystemsManager.cpp
@@ -133,7 +133,6 @@ csp::common::IRealtimeEngine* SystemsManager::MakeRealtimeEngine(csp::common::Re
 SystemsManager::SystemsManager()
     : WebClient(nullptr)
     , MultiplayerConnection(nullptr)
-    , RealtimeEngine(nullptr)
     , UserSystem(nullptr)
     , SpaceSystem(nullptr)
     , AssetSystem(nullptr)
--- a/Tests/src/PublicAPITests/UserSystemTests.cpp
+++ b/Tests/src/PublicAPITests/UserSystemTests.cpp
@@ -336,6 +336,43 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LogInAsGuestDeferredProfileCreationT
     LogOut(UserSystem);
 }
 
+CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LoginWithRefreshToken)
+{
+    auto* UserSystem = csp::systems::SystemsManager::Get().GetUserSystem();
+
+    // Login to backend services in order to get the information to get a refresh token
+    csp::common::String UserId;
+    LogInAsNewTestUser(UserSystem, UserId, false); // No need to create a multiplayer connection here
+
+    const auto RefreshToken = csp::web::HttpAuth::GetRefreshToken();
+
+    // Shutdown and restart CSP. To simulate closing the app/tab and wanting to auto-login via refresh
+    // This is just taken from the fixture code we run at the start of every test.
+    csp::CSPFoundation::Shutdown();
+
+    InitialiseFoundationWithUserAgentInfo(EndpointBaseURI());
+
+    csp::common::LogSystem* LogSystem = csp::systems::SystemsManager::Get().GetLogSystem();
+    LogSystem->SetSystemLevel(csp::common::LogLevel::VeryVerbose);
+    LogSystem->SetLogCallback([](csp::common::LogLevel, csp::common::String Message) { fprintf(stderr, "%s\n", Message.c_str()); });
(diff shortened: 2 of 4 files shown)


MAG-ElliotMorris:
Ach! Accidentally hit the ready for review button. I'm juggling like 10 itty bitty things right now. Sorry!

MAG-ElliotMorris:
Okay, actually ready this time.

MAG-ElliotMorris:
Got another instance of the actions UI not being informed that the teamcity job has finished. It completed successfully, see here : https://magnopus.teamcity.com/buildConfiguration/Olympus_Foundation_X64_TestsLocalChsInstance/325813?showLog=325813_36448_36448&logView=flowAware

Library/src/Systems/SystemsManager.cpp
```diff
@@ -133,7 +133,6 @@ csp::common::IRealtimeEngine* SystemsManager::MakeRealtimeEngine(csp::common::Re
 SystemsManager::SystemsManager()
     : WebClient(nullptr)
     , MultiplayerConnection(nullptr)
-    , RealtimeEngine(nullptr)
```
mag-lt, line 136: [Comment] I'm a bit suspicious of the following in the commit message:

> Can cause occasional UB when shutting down, due to the destructor of the realtime engine still having some unfortunate dependencies on systems, which may have been destructed before this.

Why would a default constructed shared pointer that is totally unused cause UB?

[Question] Can you remember what sort of issues you were seeing before removing this? I can't help but feel this is masking another issue, and that somehow by changing the layout of the class we're now getting lucky and whatever the actual underlying issue is (the implied UB) remains.

mag-lt (approved):
The test introduced here seems fine. I'm suspicious of the claim that the previously unused / redundant shared pointer was causing UB, and suspect if there is UB that it still remains, but this PR is supposedly about improving test coverage so I don't see any reason to block the test going in

Library/src/Systems/SystemsManager.cpp
MAG-ElliotMorris: You're probably right ... this is just an `IRealtimeEngine`, not an `OnlineRealtimeEngine`, which was probably in my mind at the time (which could manifest issues due to its destructor).

I'll try to repro it, run the thing in a loop with the change back in place (and not in place) and see if the crash occurs.

MAG-ElliotMorris: Yep, you're right, cant reproduce in any circumstance.

I have rewritten the test slightly since I observed this, possible there was some inadvertent use after free I was misdiagnosing. Going to merge anyhow though. The comments implying UB will be squashed away.

MAG-ElliotMorris:
Oh yeah. The .NET binding integration no longer allows fork PRs to function due to the need for a secret. This will be my last PR from a fork then.

Tests/src/PublicAPITests/UserSystemTests.cpp
```diff
@@ -336,6 +336,43 @@ CSP_PUBLIC_TEST(CSPEngine, UserSystemTests, LogInAsGuestDeferredProfileCreationT
...
+
+    // Login to backend services in order to get the information to get a refresh token
+    csp::common::String UserId;
+    LogInAsNewTestUser(UserSystem, UserId, false); // No need to create a multiplayer connection here
+
+    const auto RefreshToken = csp::web::HttpAuth::GetRefreshToken();
+
+    // Shutdown and restart CSP. To simulate closing the app/tab and wanting to auto-login via refresh
+    // This is just taken from the fixture code we run at the start of every test.
+    csp::CSPFoundation::Shutdown();
```
MAG-mv, line 351: [Question]  Is not calling logout here how we are simulating "closing csp"?  I'm suprised this isn't causing issues due to csp being in a half initialized state and re-logging in

MAG-mv (changes requested):

Tests/src/PublicAPITests/UserSystemTests.cpp
MAG-ElliotMorris: That's exactly what the refresh token flow does though. It handles re-joining after not having logged out.

Why do you think we are in a half-initialized state? Does shutdown not clear all the systems data. Are you talking about static data?

MAG-mv (approved):

PR #990, opened 2026-06-16
[NT-0] feat: Only send cookies if hosts match

MAG-mv:
This addresses and issue with cookies being sent to incorrect endpoints. We were previously storing all cookies and sending them in all future calls. Now, we always check to make sure the hosts match before sending.

This also increases our async timeout, as some calls are taking longer than the current 40 seconds.

This also updates InvalidURLTest to expect a 404 instead of 403. This is likely due to recent MCS changes that we will discuss

### DIFF
--- a/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
+++ b/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
@@ -222,7 +222,7 @@ void POCOWebClient::Send(HttpRequest& Request)
     }
 }
 
-void POCOWebClient::AddCookie(Poco::Net::HTTPRequest & PocoRequest, const std::string& Domain)
+void POCOWebClient::AddCookie(Poco::Net::HTTPRequest& PocoRequest, const std::string& Domain)
 {
     {
         std::scoped_lock Lock(CookiesMutex);
--- a/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
+++ b/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
@@ -242,6 +242,9 @@ void POCOWebClient::AddCookie(Poco::Net::HTTPRequest& PocoRequest, const std::st
 
         for (const auto& Cookie : *Cookies)
         {
+            // Only send cookies whose domain scope matches the request host.
+            // This prevents cookies set by one endpoint from being sent to
+            // another endpoint that did not expect them.
             if (Domain == Cookie.Domain)
             {
                 CookieCollection.add(Cookie.Name, Cookie.Value);
--- a/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
+++ b/Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
@@ -99,7 +99,7 @@ bool POCOWebClient::PrepareAndSendRequest(
         PocoRequest.add(Header.first.c_str(), Header.second.c_str());
     }
 
-    AddCookie(PocoRequest);
+    AddCookie(PocoRequest, ClientSession->getHost());
 
     switch (SendBodyMode)
     {
@@ -144,7 +144,14 @@ std::istream& POCOWebClient::ReceiveResponse(Poco::Net::HTTPClientSession* Clien
 
     {
         std::scoped_lock Lock(CookiesMutex);
-        PocoResponse.getCookies(*Cookies);
+
+        std::vector<Poco::Net::HTTPCookie> ResponseCookies;
+        PocoResponse.getCookies(ResponseCookies);
+
+        for (const auto& ResponseCookie : ResponseCookies)
+        {
+            Cookies->push_back(CookieData { ResponseCookie.getName(), ResponseCookie.getValue(), ClientSession->getHost() });
+        }
     }
 
     return ResponseStream;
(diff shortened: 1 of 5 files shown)


Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -215,7 +222,7 @@ void POCOWebClient::Send(HttpRequest& Request)
     }
 }
 
-void POCOWebClient::AddCookie(Poco::Net::HTTPRequest& PocoRequest)
+void POCOWebClient::AddCookie(Poco::Net::HTTPRequest & PocoRequest, const std::string& Domain)
```
MAG-AdamThorn: [fix] You have add an unnecessary space before the '&'.

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
```diff
@@ -31,6 +31,15 @@ class LogSystem;
 namespace csp::web
 {
 
+struct CookieData
+{
+    std::string Name;
+    std::string Value;
+
```
MAG-AdamThorn: [fix] Remove unnecessary white space.

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
```diff
@@ -81,7 +90,7 @@ class POCOWebClient : public WebClient
 
     Poco::Net::Context::Ptr PocoContext;
 
-    std::vector<Poco::Net::HTTPCookie>* Cookies;
```
MAG-AdamThorn, line 84: [Question] Is there a reason we are introducing our own Cookie struct rather than using the Poco one? The Poco one has a domain property that can be set.

MAG-AdamThorn (changes requested):

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
MAG-mv: The cookies we receive contain an empty domain, so I have to retrieve it from the ClientSession

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
MAG-mv: Fixed in: [eccb18a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/990/commits/eccb18a5da8dfad0e8a43a05b22af6a2bb00bf43)

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-mv: Fixed in: [eccb18a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/990/commits/eccb18a5da8dfad0e8a43a05b22af6a2bb00bf43)

Library/src/Common/Web/POCOWebClient/POCOWebClient.h
MAG-AdamThorn: Yes but could we not set the Poco::Net::HTTPCookie domain property from the ClientSession just as we set the CookieData property?

MAG-mv: I'm cautious about doing this because Poco::Net::HTTPCookie represents the cookie we get from the server, and modifying it changes it's meaning. It's not wrong that the host is empty, it means  that it is a host only cookie, and sohuld only be send accross the same domain.

MAG-AdamThorn (approved):

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
```diff
@@ -233,9 +240,12 @@ void POCOWebClient::AddCookie(Poco::Net::HTTPRequest& PocoRequest)
     {
         std::scoped_lock Lock(CookiesMutex);
 
-        for (const Poco::Net::HTTPCookie& Cookie : *Cookies)
+        for (const auto& Cookie : *Cookies)
         {
-            CookieCollection.add(Cookie.getName(), Cookie.getValue());
+            if (Domain == Cookie.Domain)
```
MAG-SamBirley, line 248: [Question] Worth a comment here explaining why this is important? Thinking that it took us long enough to unpick the logic this time that in another three years this'll save some further head-scratching.

MAG-SamBirley (commented):
Logic looks sound. One question about giving us a stronger paper-trail for the future should be need to reverse-engineer past decisions again.

MAG-SamBirley (commented):
Logic looks sound. One question around giving us a stronger paper trail for the future.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-mv: Yeah i'll add a comment

MAG-SamBirley (approved):
Approving now though to avoid blocking this change going in.

Library/src/Common/Web/POCOWebClient/POCOWebClient.cpp
MAG-mv: Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/990/commits/6ed2bc35609af1581fbb130230985a9230c11914

PR #991, opened 2026-06-22
[OF-1840] feat: Add legacy typescript project

MAG-mv:
We have removed the emscripten flags and project in the new project, as this is being handled in the new binding repo. However, we still need to temporarily support making wasm builds without the new binding repo.

This adds a new cmake executable project with emscripten flags from the previous system.

I have tested this new build against our wasm tests

### DIFF
--- a/LegacyEmscriptenWrapper/CMakeLists.txt
+++ b/LegacyEmscriptenWrapper/CMakeLists.txt
@@ -16,6 +16,10 @@
 #
 #   No SOURCES given to target
 
+if(NOT EMSCRIPTEN)
+    message(FATAL_ERROR "This project only supports Emscripten builds.")
+endif()
+
 cmake_minimum_required(VERSION 3.31)
 
 project(LegacyEmscriptenWrapper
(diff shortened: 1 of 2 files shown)


MAG-ElliotMorris (commented):
[Question] Can you be more clear on the intent of this. I don't understand if "support the old system" means just support a build, or if this is intended to build all of CSP in legacy emscripten mode, or if it's about the typescript tests.

In particular, the inclusion of `Dummy.cpp` makes me think this is a temporary stub that is laying the groundwork for future work. That would be fine, but I can't tell if that's the intent.

MAG-mv:
@MAG-ElliotMorris I have updated the description slightly but basically:

This allows us to continue supporting emscripten javascript builds within CSP, until the new binding repo is complete.  This additional project builds csp as a javascript lib, instead of just an emscripten static lib. 

This will be removed when the binding repo is complete and being consumed.

LegacyEmscriptenWrapper/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+cmake_minimum_required(VERSION 3.31)
```
MAG-ElliotMorris, line 23: [fix] Document here, (or even better, in a README.md) what this is for. It's rather unclear right now.

In particular, mention that this is to produce the JS/TS bindings using the legacy wrapper generator until the new wrapper generator can be bought online. ... Even from your additional description I'm not certain if this is an accurate representation, as I don't see an invocation of the legacy wrapper generator.

Is this just to append the old hard-coded flags? Is this an in-between step to generate emscripten produced JS prior to the wrapper generator invocation? Still confused.

Also important, mention why `Dummy.cpp` exists, because that is surprising. I can deduce from my experience it's due to the need to have an entry point to satisfy the compiler (although I'm not entirely convinced that's required, but I don't really mind in a temp solution), but I don't think that's clear to the general audience.

MAG-ElliotMorris (changes requested):

LegacyEmscriptenWrapper/CMakeLists.txt
MAG-mv: Updated here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/991/commits/60e8cf8419aca34d311e58d61bc007d5f1ba7a5b

MAG-ElliotMorris (approved):

LegacyEmscriptenWrapper/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+# csp static library generation.
+
+# Until the binding project is complete, this target creates a JavaScript
+# "executable" that links against symbols from the csp-lib static library and
+# produces the WebAssembly/JavaScript output.
+
+# We still need to generate typescript bindings, as emscripten only produces .wasm and .js artifacts.
+# Before running the CMake build step, generate the wrapper files with:
+#
+#   python3 Tools/WrapperGenerator/WrapperGenerator.py --generate_typescript
```
mag-lt, line 13: [Comment] We could add a custom command that does this when needed. Though, as we see this whole script as temporary it might not be worth it.

LegacyEmscriptenWrapper/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+cmake_minimum_required(VERSION 3.31)
+
+project(LegacyEmscriptenWrapper
+    DESCRIPTION "LegacyEmscriptenWrapper" VERSION 1.0.0
+    LANGUAGES C CXX)
+
+add_executable(csp-emscripten ${CMAKE_CURRENT_SOURCE_DIR}/Dummy.cpp)
+
+target_link_libraries(csp-emscripten PRIVATE
+  "$<LINK_LIBRARY:WHOLE_ARCHIVE,csp-lib>"
```
mag-lt, line 32: [Question] Why is `LINK_LIBRARY:WHOLE_ARCHIVE` needed? I'd expect that the main thing we needed exported are the generated C symbols, which should hopefully all be annotated with appropriate export directives (and so linking it "normally" would be fine)

LegacyEmscriptenWrapper/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+)
+
+option(WASM_WITH_NODE "Build wasm target with Node.js environment support" OFF)
+
+if(EMSCRIPTEN)
```
mag-lt: [nit] presumably we also don't want to add the target if we're not targeting `EMSCRIPTEN`, we could perhaps halt the script earlier and avoid the nesting here

mag-lt (approved):

LegacyEmscriptenWrapper/CMakeLists.txt
MAG-mv: Yeah, I didn't want to add additional complexity to something that won't be around long

LegacyEmscriptenWrapper/CMakeLists.txt
mag-lt: The `add_subdirectory` in the parent CMakeLists.txt is protecting against this, you could probably remove the additional check here or better yet effectively "assert" at the top of the script (and error with a `message(FATAL_ERROR ...)`)

MAG-mv: Yeah, I like this more. Updated here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/991/commits/b8675f6131e4e49e5ca572b4789ab317b571451f

LegacyEmscriptenWrapper/CMakeLists.txt
MAG-mv: I believe the issue is because the executable target is an empty dummy file, so without this, the linker doesn't pull the object files out of the static lib which causes the emscripten binary to be essentially empty.

I tried to manually include the generated_wrapper.h in the Dummy.cpp, but this causes other issues with missing private includes.

PR #994, opened 2026-06-23
[NT-0] ci: Create new branch for unity integration in PRs

MAG-ElliotMorris:
Create a branch to run the unity integrations against, rather than always using `csp-development`, as that can have overlap problems. If developers make pushes to the ephemeral branch, it is up to them to merge them into `csp-development`.

This is a new PR, there was a previous one, but I needed a new ref to test against. The changeset is almost identical. _except_ for how we're now using an off the shelf action to create the branch, rather than trying to do it through API commands. I was trying to avoid checking out the unity repo in order to branch, but it became rather complicated. This action does checkout the unity repo, which in the scheme of things is actually fine, and it documents what we're doing in the action much better than the direct REST api which is less than good.

I actually did end up with a fully working API version ... but it got so complex that I wanted to put it in a `.sh` script, which meant checking out this repo to access the script ... which made me realize I was getting tied in knots over nothing.

> [!TIP]
>
> The first commit shows the old way, which we migrated away from. Don't spend too much time thinking about it.

## Testing

- You can see the first run [here](https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/28027670067/job/82959533356), which creates a new branch.
- The second run [here ](https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/28036381624/job/82990749684?pr=994) does not fail, but does log that the branch creation is skipped as the branch already exists.
- You can see the branch is created on the target repo [here](https://github.com/magnopus-opensource/connected-spaces-platform-unity/tree/csp-development-work/csp-development-branch). (it even managed to deal with the weird slashes i do in my branch names)
- You can see the action using the new branch [here](https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/28036381624/job/82990781046?pr=994), although you'll have to expand into the checkout step a bit.

### DIFF
--- a/.github/workflows/unity-interop-tests.yml
+++ b/.github/workflows/unity-interop-tests.yml
@@ -1,12 +1,38 @@
 name: .NET Interop Binding Integration
 
 # Run a fully integrated build of the unity bindings, based on the CSP build from the current branch.
-# If breaking changes are made, push directly to the "csp-development" branch in the connected-spaces-platform-unity repo, which represents the breaking changeset of the next release.
+# Tests run against a dedicated branch in the connected-spaces-platform-unity repo, forked from
+# "csp-development" and named after the branch that triggered the run, e.g. "csp-development-my-branch".
+# If breaking changes are made, push to that branch to unblock these tests, then manually merge it
+# back into "csp-development" once the CSP change lands.
+# This workflow requires token permissions to create a branch, so will not run from forks.
+
 on:
   pull_request:
   workflow_dispatch:
 
 jobs:
+
+  # Creates the per-branch binding branch from csp-development, or reuses it if it already
+  # exists (so commits contributors have pushed to it survive re-runs). Any other failure
+  # (e.g. bad credentials, missing base branch) fails the job so it isn't silently swallowed.
+  create-binding-branch:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.DOTNET_REPO_ACCESS_TOKEN }}
+      REPO: magnopus-opensource/connected-spaces-platform-unity
+      BRANCH: csp-development-${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/unity-interop-tests.yml
```diff
@@ -1,19 +1,43 @@
...
+  # Creates the per-branch binding branch from csp-development, or reuses it if it already
+  # exists (so commits contributors have pushed to it survive re-runs). Any other failure
+  # (e.g. bad credentials, missing base branch) fails the job so it isn't silently swallowed.
+  create-binding-branch:
+    runs-on: ubuntu-latest
+    env:
+      # The usual pattern to support both PR and workflow_dispatch refs
+      BRANCH_TO: csp-development-${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref_name }}
+    steps:
+      - uses: GuillaumeFalourd/create-other-repo-branch-action@v1.5
```
mag-lt, line 25: [Comment] From a cursory look at `GuillaumeFalourd/create-other-repo-branch-action@v1.5`, it checks out the repo locally on the machine to create a new named branch and then pushes up. I'm not very experienced with actions, but I'd guess there would be a way to do this at least via the `gh` cli tool (which if I understand correctly is how repo stuff is typically done in actions) that would avoid the roundtrip on the runner (to be clear I'm not actually concerned with performance, just a bit surprised there isn't a more first class way of doing this).

mag-lt (approved):

.github/workflows/unity-interop-tests.yml
MAG-ElliotMorris: There is, you can see that addressed in the PR description. I spent a fair amount of time making it work in the alternate way before deciding it wasn't worth the hassle.

mag-lt: Shame on me

MAG-mv (approved):

PR #995, opened 2026-06-26
[OF-1884] feat: Support injecting updated schemas for built-in components

mag-lt:
This allows clients to inject updated schemas for built-in components, adding properties beyond what is built-in, with compatibility enforced while the schema-driven and hardcoded component APIs coexist.


Tests/src/InternalTests/ComponentSchemaTests.cpp
```diff
@@ -1251,3 +1277,77 @@ CSP_INTERNAL_TEST(CSPEngine, ComponentSchemaTests, FromJsonRejectsVec4PropertyWi
...
+        WithExtraProperty(csp::multiplayer::CinematicCameraSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::FiducialMarkerSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::GaussianSplatSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::TextSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::HotspotSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::ScreenSharingSpaceComponent::GetSchema()),
+        WithExtraProperty(csp::multiplayer::AIChatbotSpaceComponent::GetSchema()),
+    };
+
+    auto Fixture = TestFixture(csp::common::Convert(AllUpdated));
```
MAG-ElliotMorris: [Nit] Probably from a prior review, but `TestFixture` is a tad confusing to read in isolation, requires a look into the codebase to figure out if this is a gtest thing or a csp test framework thing, and I was surprised to find it's something specific to data components. I might rename this to make the data-component association clear.

[Comment] A bit surprised to not see this be using TEST_F or something more in-framework, but 🤷

Tests/src/InternalTests/ComponentSchemaTests.cpp
```diff
@@ -1251,3 +1277,77 @@ CSP_INTERNAL_TEST(CSPEngine, ComponentSchemaTests, FromJsonRejectsVec4PropertyWi
...
+
+    auto* Entity = Fixture.MakeEntity("Test Entity");
+    ASSERT_NE(Entity, nullptr);
+
+    for (const auto& Schema : AllUpdated)
+    {
+        const auto LastPropertyIndex = Schema.Properties.Size() - 1;
+        const auto ExtraKey = Schema.Properties[LastPropertyIndex].Key;
```
MAG-ElliotMorris: [Nit] Unused variables?

Tests/src/InternalTests/ComponentSchemaTests.cpp
```diff
@@ -1251,3 +1277,77 @@ CSP_INTERNAL_TEST(CSPEngine, ComponentSchemaTests, FromJsonRejectsVec4PropertyWi
...
+
+        auto* Component = Entity->AddComponent(static_cast<csp::multiplayer::ComponentType>(Schema.TypeId));
+        ASSERT_NE(Component, nullptr) << Schema.Name.c_str();
+
+        const auto* Value = Component->GetSchemaProperty(ExtraKey);
```
MAG-ElliotMorris: [Comment] I'm still of the opinion that `GetSchemaProperty` is going to be a confusing name for users vs just `GetProperty` or `GetComponentProperty` 

As I'm sure the reason they're not that is due to existing method naming, something like `GetPropertyData` might be preferable. Wouldn't mind talking about this as a team.

Tests/src/InternalTests/ComponentSchemaTests.cpp
```diff
@@ -65,8 +65,11 @@ class TestFixture final
         : ScriptSystem(csp::systems::ScriptSystem::MakeInitialised())
         , Engine(LogSystem, *ScriptSystem, Schemas)
     {
+        InitialiseFoundationWithUserAgentInfo(EndpointBaseURI());
     }
 
+    ~TestFixture() { csp::CSPFoundation::Shutdown(); }
+
```
MAG-ElliotMorris: [Consider] It feels that this betrays the philosophy of the internal tests not doing network handshakes. I would move this to a public API test if posible.

Library/include/CSP/Multiplayer/ComponentSchema.h
```diff
@@ -71,4 +71,11 @@ class CSP_API ComponentSchema
 csp::common::Array<ComponentSchema> ComponentSchemasFromJson(
     const csp::common::List<csp::common::String>& JsonDocuments, csp::common::LogSystem& LogSystem);
 
+/// @brief Checks whether a schema update is compatible with an existing schema.
+/// The check is conservative. The updated schema must be a superset of the original, without altering any existing property definitions.
+/// @param Original The existing schema to check against.
+/// @param Updated The candidate updated schema.
+/// @return True if the updated schema is a compatible update of the original, false otherwise.
+bool IsCompatible(const ComponentSchema& Original, const ComponentSchema& Updated);
```
MAG-ElliotMorris: [Consider] Maybe a `@note` mentioning that we may loosen how strict this will be in the future?

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -474,4 +475,25 @@ bool ComponentSchema::operator==(const ComponentSchema& Other) const
...
+        UpdatedByKey.emplace(Property.Key, Property);
+    }
+
+    return std::all_of(Original.Properties.begin(), Original.Properties.end(),
+        [&UpdatedByKey](const auto& Property)
+        {
+            const auto It = UpdatedByKey.find(Property.Key);
+            return It != UpdatedByKey.end() && It->second == Property;
+        });
+}
```
MAG-ElliotMorris: [Consider] I wonder if it's worth threading user feedback through here and now? I know it's technically redundant as there's only two potential reasons (and really only one) that this could report false right now, but even that might be worth telling the user about, you could output the specific name of the incompatible property, which would be helpful in teaching them what they've done incorrectly.

[Comment] Tight implementation btw 👏

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
```diff
@@ -54,6 +54,14 @@ class CSP_API AIChatbotSpaceComponent : public ComponentBase, public IPositionCo
 
     AIChatbotSpaceComponent(csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
 
+    /// @brief Creates an AI chatbot space component using the provided schema if it is compatible with the built-in schema.
+    /// @param UpdatedSchema The schema to use. Must be compatible with the built-in schema.
+    /// @param LogSystem The log system.
+    /// @param Parent The space entity that owns this component.
+    /// @return A new AIChatbotSpaceComponent if the schema is compatible, nullptr otherwise.
+    CSP_NO_EXPORT static std::unique_ptr<AIChatbotSpaceComponent> TryMake(
+        const ComponentSchema& UpdatedSchema, csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
```
MAG-ElliotMorris: [Comment] Good pattern for this imo.

[Consider] The "Updated" in `UpdatedSchema` implies that you'd only call this if you've schema that's changed, but I don't think that should be the philosophy, as we're trying to make this the primary construction path. I understand that it's technically updating from the built-in schema understanding, but even so.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
```diff
@@ -54,6 +54,14 @@ class CSP_API AIChatbotSpaceComponent : public ComponentBase, public IPositionCo
 
     AIChatbotSpaceComponent(csp::common::LogSystem* LogSystem, SpaceEntity* Parent);
 
+    /// @brief Creates an AI chatbot space component using the provided schema if it is compatible with the built-in schema.
```
MAG-ElliotMorris, line 57: [Nit] Define what compatible means. You can just do it in terms of the check.

> ... is compatible (compatibility being defined as a passing result from `WhereveritIs::IsCompatible()`) with the built in schema.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -505,14 +505,26 @@ uint16_t SpaceEntity::GenerateComponentId()
     }
 }
 
-ComponentBase* SpaceEntity::InstantiateComponent(uint16_t InstantiateId, uint64_t TypeId)
+template <typename T> static ComponentBase* TryMakeOrDefault(const ComponentSchema* Schema, csp::common::LogSystem* LogSystem, SpaceEntity* Parent)
```
MAG-ElliotMorris, line 508: [Consider] For the schema argument, if nullability is a valid interface mechanism that dictates behaviour, consider `optional` instead to better express that. I don't _think_ the same argument applies for `Parent` because it's just a passthrough and doesn't express any decisions.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -505,14 +505,26 @@ uint16_t SpaceEntity::GenerateComponentId()
...
 {
-    if (!IsLegacyComponentTypeId(TypeId))
+    if (Schema != nullptr)
     {
-        const auto* Registry = EntitySystem != nullptr ? EntitySystem->GetComponentSchemaRegistry() : nullptr;
+        if (auto Component = T::TryMake(*Schema, LogSystem, Parent))
+        {
+            return Component.release();
```
MAG-ElliotMorris, line 514: [Comment] I like to see this unique_ptr into release sort of thing. It can feel futile, but I think every little helps in the long arc.

Library/src/Multiplayer/SpaceEntity.cpp
```diff
@@ -505,14 +505,26 @@ uint16_t SpaceEntity::GenerateComponentId()
...
+    }
 
-        const auto* Schema = Registry != nullptr ? Registry->Find(TypeId) : nullptr;
+    return new T(LogSystem, Parent);
+}
 
+ComponentBase* SpaceEntity::InstantiateComponent(uint16_t InstantiateId, uint64_t TypeId)
+{
+    const auto* Registry = EntitySystem != nullptr ? EntitySystem->GetComponentSchemaRegistry() : nullptr;
+    const auto* Schema = Registry != nullptr ? Registry->Find(TypeId) : nullptr;
```
MAG-ElliotMorris, line 524: [Comment] Neat, I like the way this is expressed and flows into the `TryMake` methods. Very functional.

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
```diff
@@ -98,6 +93,18 @@ ComponentSchemaRegistryImpl::ComponentSchemaRegistryImpl(
...
     AddSchema(ScreenSharingSpaceComponent::GetSchema());
     AddSchema(AIChatbotSpaceComponent::GetSchema());
+
+    for (const auto& Schema : AdditionalComponents)
+    {
+        if (const auto It = SchemaMap.find(Schema.TypeId); It != SchemaMap.end() && !IsCompatible(It->second, Schema))
+        {
+            LogSystem.LogMsg(csp::common::LogLevel::Warning,
+                fmt::format("Injected schema for TypeId {} is not compatible with the built-in schema and will be ignored.", Schema.TypeId).c_str());
+            continue;
```
MAG-ElliotMorris, line 103: [comment] This is where the comment I made earlier about threading some sort of reason from `IsCompatible` would manifest

MAG-ElliotMorris (commented):
Good commit structuring, this tells the story well, seems intentionally designed :P

MAG-ElliotMorris (approved):
Should have taken a stance with my last review. I've just feel like I've noticed a couple of times recently folk not responding to [Considers] at all ... but that's no excuse to be untrusting.

Library/include/CSP/Multiplayer/ComponentSchema.h
```diff
@@ -71,4 +71,11 @@ class CSP_API ComponentSchema
 csp::common::Array<ComponentSchema> ComponentSchemasFromJson(
     const csp::common::List<csp::common::String>& JsonDocuments, csp::common::LogSystem& LogSystem);
 
+/// @brief Checks whether a schema update is compatible with an existing schema.
+/// The check is conservative. The updated schema must be a superset of the original, without altering any existing property definitions.
+/// @param Original The existing schema to check against.
+/// @param Updated The candidate updated schema.
+/// @return True if the updated schema is a compatible update of the original, false otherwise.
```
MAG-mv, line 80: [Question] Is the IsCompatible a temporary check whil we still have hardcoded components? Or will it be used in the future?

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
```diff
@@ -54,6 +54,14 @@ class CSP_API AIChatbotSpaceComponent : public ComponentBase, public IPositionCo
...
+    /// @param UpdatedSchema The schema to use. Must be compatible with the built-in schema.
+    /// @param LogSystem The log system.
+    /// @param Parent The space entity that owns this component.
+    /// @return A new AIChatbotSpaceComponent if the schema is compatible, nullptr otherwise.
+    CSP_NO_EXPORT static std::unique_ptr<AIChatbotSpaceComponent> TryMake(
```
MAG-mv, line 64: [Question] Can't we just create a single, templated TryMake function? They all seem to be doing the same thing

MAG-mv (approved):

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: There would be a way to do it for sure, but it would have other compromises.

I only really see a few options:
1. Make the new constructor that takes a Schema public (on each hardcoded component)
    * It is currently private, and the `TryMake` factory function leverages it. Combining these two is the sole mechanism for ensuring the component is valid on construction
    * We could make the constructor public and have it throw, and then the `TryMake` factory functions would be non-throwing variants, but I'd rather only have one way of doing these things, and personally prefer not using exceptions
2. Have some templated friend function that this logic is pushed into
    * technically possible, but comes with its own bit of boilerplate
3. Use a preprocessor macro

As currently written, `IsCompatible` is the main point of de-deduplication, and then there is a bit of repeated boilerplate for sure. `TryMakeOrDefault` in SpaceEntity.cpp does is a templated helper (mainly because there would be way too much churn in the switch statement there otherwise), and at the time I did consider trying to push all of this code into there, but in the ended opted out as I didn't really like any of the options above, so opted to just stamp it out.

Library/include/CSP/Multiplayer/ComponentSchema.h
mag-lt: Currently the main use is do prevent injecting schemas relating to hardcoded components unless they meet the critera. If hardcoded components are eventually removed this specific usage goes away, but the concept likely resurfaces if space entry schema negotiation gets built.

So, likely it will stay in some form and evolve as necessary.

Library/src/Multiplayer/SpaceEntity.cpp
mag-lt: I actually did try and use `optional` whilst writing this but couldn't come up with something that I thought was worth the extra boilerplate / didn't just add more noise.

I'd actually prefer if this was an `optional` all the way down, i.e. `Registry->Find()` would return one. However, we don't have a good story for returning optionals at the interop API level (I was surprised to find `common::Optional` doesn't really work properly as a return value). So, for now the nullness of the pointer encodes the optionality.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: I did name these `Schema` initially but there was a shadowing warning as error. I'll look again and perhaps fix in the other direction, as I agree the `Updated` prefix isn't quite right.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: Will address

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: This reminds me of our discussion recently about `std::expected` etc. Another use case for it. I'll likely thread the logger through again here in an upcoming commit

Library/src/Multiplayer/SpaceEntity.cpp
MAG-ElliotMorris: Aye ... fair. Without a good enough optional sink across the public api, it does make it tricky. I would never use `common::Optional` if I could avoid it. Would use `std::optional` though.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: There isn't any networking going on here as far as I'm aware, least of all because the offline engine is used throughout these tests.

This is just about singleton Foundation initialisation stuff / hidden dependencies, and is needed because `ConversationSpaceComponent` reaches into `SystemsManager` to get the conversation system. The `AgentInfo` stuff is perhaps misleading there.

The `TestFixture` dtor handles the corresponding shutdown.

I could move this to be a public API test, just so that it also does this singleton setup/teardown, but to be honest I'm trying to write my tests in a much more isolated way that only pull in what is strictly necessary for what these tests are covering.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: As discussed earlier, I will rename this. I won't do it in this PR though, I'll do a follow up one (likely tomorrow at this point).

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: These are used, `ExtraKey` is used on the `const auto* Value = Component->GetSchemaProperty(ExtraKey);` line, I'll move the lines down closer to that vs being at the top.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: The local `TestFixture` class is something I've been doing recently, first introduced in the `ComponentSchemaScriptBindingTests.cpp` [tests](https://github.com/magnopus-opensource/connected-spaces-platform/blob/a093fc373d3b42f6ab0f0be368923074b7669b82/Tests/src/InternalTests/ComponentSchemaScriptBindingTests.cpp#L38). Admittedly there was much more boilerplate and state that was more justifiable to push into a helper class than there is here, but I still find the setup a bit more ergonomic. 

I would personally prefer to not be using any of our custom macros, and just use the vanilla `TEST` across the board. I'm not personally much of a fan of `TEST_F`, I prefer rolling a local helper type and using it from a plain test over the forced inheritance GTest imposes, and without the ability to use the constructor of that class etc. 

There is always a tradeoff with readability when using test helpers, and ideally the reader wouldn't have to jump around too much to understand the body of any given test, but here I'd argue it isn't that bad once you know what `TestFixture` is.

Library/include/CSP/Multiplayer/ComponentSchema.h
mag-lt: Address with abd313198b4314295f503fc53a747e8403c4f7d3

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: Should be improved with 1165798a0ded645c8492dd79425ee06ef4eed972

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: Addressed with 8dd599a32f8fa12539fb320d85159ec8bf298c11, though now not so tight.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: Addressed with another rename in 11353612e324a5af731b7998d7464e5ad6c27677

Library/src/Multiplayer/ComponentSchemaRegistry.cpp
mag-lt: Also in 8dd599a32f8fa12539fb320d85159ec8bf298c11

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
mag-lt: Hopefully covered by 2aefe55faa38aa61ac8ebd7eaf89053f7c3f511f

Library/src/Multiplayer/ComponentSchema.cpp
MAG-ElliotMorris: I still think we should take a backport of the thing _today_ rather than holding off for a language upgrade. Which I know _seems_ like a trivial thing to do ... but we still havn't done it 🤷

Tests/src/InternalTests/ComponentSchemaTests.cpp
MAG-ElliotMorris: My point is that `TestFixure` reads like it should be named `DataComponentsTestFixure`. I would assume `TestFixure` is a gtest type due to how generic it reads, and you can't expect folk to always be in an editor where they can break into types when reading code.

Tests/src/InternalTests/ComponentSchemaTests.cpp
MAG-ElliotMorris: My bad, just didn't spot it. Ty for the  move.

Tests/src/InternalTests/ComponentSchemaTests.cpp
MAG-ElliotMorris: Fair point, although I'm not fully convinced that this test would be resilient to being in a zero-network environment, simply because this will end up calling the constructor of `POCOWebClient`. Perhaps that is merely registration, but the complexity unsettles me, as internal tests should not have that sort of potential error complexity.

I agree with your philosophy on only pulling in what's stricly neccesary, the point I'm gesturing at is that by initializing all of the runtime, you've not accomplished that and I worry that makes calling this an internal test a bit of a deception.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: The way I see it, we're in a specific test file, and this is a fixture relevant to it, no need for prefixing (a similar argument to types inside namespaces). For what it is worth, I'm rarely in an editor where I can jump to definition.

As this wasn't introduced here, I'm going to leave it as is.

Library/src/Multiplayer/ComponentSchema.cpp
MAG-ElliotMorris: Ach yeah, shame to thread the logger through, it makes the error reporting worse than accumulating it at the top as well. I might even [consider] an out variable in this case.

Library/include/CSP/Multiplayer/Components/AIChatbotComponent.h
MAG-ElliotMorris: Yep, that'll do it.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: > although I'm not fully convinced that this test would be resilient to being in a zero-network environment

With my network disabled at the OS level (no internet), and no local CHS server running:

```
./build/build/Debug/csp-tests --gtest_filter="*ComponentSchemaTests*" --gtest_brief=1
[==========] 64 tests from 1 test suite ran. (17 ms total)
[  PASSED  ] 64 tests.
```

I'd like to improve our initialisation (I know we all do), so that we don't need singletons at all. In this specific case, we could also amend `ConversationSpaceComponent` so that it has to be passed `ConversationSystemInternal` on construction vs getting it off the `SystemsManager`. However, both are outside the scope of these changes.

I can move this specific test elsewhere to a public test (probably `ComponentTests`), and rewrite it slightly in doing so to not rely on the local test fixture, but I personally don't see much value.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: I think this is one of the cases where there aren't really any good options. Wanting to know what failed changes how the logic needs to be written, and will naturally be more "branchy", with interleaving concerns (or doing two passes). For now, I've picked my poison (threading the logger through).

Tests/src/InternalTests/ComponentSchemaTests.cpp
MAG-ElliotMorris: It's your call, I'm not a fan of what I perceive as a break in category, but 🤷

Library/src/Multiplayer/ComponentSchema.cpp
MAG-ElliotMorris: I'm adding std::expected to the weekly sync again. We need a result type.

Tests/src/InternalTests/ComponentSchemaTests.cpp
mag-lt: I moved the test that this change related to (`UpdatedLegacySchemaExposesExtraProperty`) to the public `ComponentTests` in 36b9760cbaa299fb0186e3864a065d022f04edb7.

Not setting up/tearing down Foundation in tests that don't need it is significant enough, see below (17ms vs 3ms).

```
./build/build/Debug/csp-tests --gtest_filter="*ComponentSchemaTests*" --gtest_brief=1
[==========] 63 tests from 1 test suite ran. (3 ms total)
[  PASSED  ] 63 tests.
```

(this isn't just the difference from not running the moved test, keeping the setup/teardown in and just running the other 63 tests does have a minor perf hit from the setup/teardown)

PR #996, opened 2026-06-29
[OF-1840] feat Android Support

MAG-mv:
This adds a couple of fixes + host profile to allow CSP to build for Android. This includes:

- Added an extra param to fix __android_log_print. This was previously hidden using -Wno-error=format-security
- Linking Android log lib so shared libs build
- Added Android host profiles

### DIFF
--- a/Library/src/Common/Systems/Log/LogSystem.cpp
+++ b/Library/src/Common/Systems/Log/LogSystem.cpp
@@ -84,7 +84,7 @@ void LogSystem::LogMsg(const csp::common::LogLevel Level, const csp::common::Str
 #endif
 
 #if defined(CSP_ANDROID)
-        __android_log_print(ANDROID_LOG_VERBOSE, "CSP", InMessage.c_str());
+        __android_log_print(ANDROID_LOG_VERBOSE, "CSP", "%s", InMessage.c_str());
 #endif
     }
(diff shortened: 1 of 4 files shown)


Library/src/Common/Systems/Log/LogSystem.cpp
```diff
@@ -84,7 +84,7 @@ void LogSystem::LogMsg(const csp::common::LogLevel Level, const csp::common::Str
 #endif
 
 #if defined(CSP_ANDROID)
-        __android_log_print(ANDROID_LOG_VERBOSE, "CSP", InMessage.c_str());
+        __android_log_print(ANDROID_LOG_VERBOSE, "CSP", "%s", InMessage.c_str());
```
MAG-ElliotMorris, line 87: [comment] Heh. More code that we never actually invoked?

MAG-ElliotMorris (approved):

Library/src/Common/Systems/Log/LogSystem.cpp
MAG-mv: I believe it is invoked, but we were hiding compiler errors using -Wno-error=format-security

MAG-AdamThorn (approved):

PR #997, opened 2026-06-29
[OF-1840] feat: Add legacy wrapper gen macro

MAG-mv:
This finally adds a macro around the generated macro header, meaning we won't need to call generate_solution in the new build system for development..

CSPFoundation contains an autogenerated header file from the wrapper generator. We don't need to use this in circumstances where the wrapper generator isn't used, so guard with a macro.

### DIFF
--- a/Library/premake5.lua
+++ b/Library/premake5.lua
@@ -118,7 +118,8 @@ if not Project then
             "POCO_NET_NO_IPv6",
 			"LIBASYNC_STATIC",
             "LIBASYNC_CUSTOM_DEFAULT_SCHEDULER",
-            "FMT_HEADER_ONLY"
+            "FMT_HEADER_ONLY",
+            "USE_LEGACY_WRAPPER_GEN"
         }
 
         filter "platforms:not wasm"
--- a/Library/src/CSPFoundation.cpp
+++ b/Library/src/CSPFoundation.cpp
@@ -15,7 +15,10 @@
  */
 #include "CSP/CSPFoundation.h"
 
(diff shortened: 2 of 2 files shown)


Library/premake5.lua
```diff
@@ -118,7 +118,8 @@ if not Project then
             "POCO_NET_NO_IPv6",
 			"LIBASYNC_STATIC",
             "LIBASYNC_CUSTOM_DEFAULT_SCHEDULER",
-            "FMT_HEADER_ONLY"
+            "FMT_HEADER_ONLY",
+            "USE_LEGACY_WRAPPER_GEN"
```
mag-lt, line 122: [Question] Do we not also need to set this flag in some of our CMake targets too? (`csp-lib`?). If we were to switch over to the CMake build system and release using those build artefacts, we'd need the generated c api symbols to be exported for the DLL the legacy C# wrapper loads, and also for the emscripten / wasm build.

Are you intending to set this in follow up work? Or are we planning to align this CMake switchover with the newer binding efforts (and thus never set the flag in CMake)?

MAG-mv: I'm passing this as an argument at configurte time in the CI when we build for emscripten/C#.

mag-lt: In a subsequent PR?

MAG-mv: Yeah

mag-lt (approved):

MAG-AdamThorn (approved):

PR #1001, opened 2026-06-30
[OF-1785] refac!: Remove legacy leadership election logic

MAG-AdamThorn:
Now that server-side script leader election has been operating as the default election model for ~6 months, it is time to remove the legacy client-side leader election logic.

The following changes have been made:

1. The `ClientProxy.h/.cpp` and `ClientElectionManager.h/.cpp` files have been removed along with any reference to them.
2. The `OnlineRealtimeEngine` has been updated to remove the `ServerSideElectionEnabled` member and its setter.
3. The `SpaceSystem::RegisterScopesInSpace()` method no longer checks whether `ManagedLeaderElection` is enabled for the DefaultScope (please see note below).
4. The `ClientProxy.h/.cpp` and `ClientElectionManager.h/.cpp` files have been removed from the cmake/multiplayer_files.cmake file.
5. The `OnlineRealtimeEngine` methods `EnableLeaderElection` & `DisableLeaderElection` have been set to `CSP_NO_EXPORT`.

**Breaking change:**
- The public methods `OnlineRealtimeEngine::EnableLeaderElection()` and `OnlineRealtimeEngine::DisableLeaderElection()` have been made `CSP_NO_EXPORT`.

**Please note:**
We no-longer check the `ManagedLeaderElection` member of the default scope. That is because the backend services team have stated that leader election will always be enabled for the default scope, and we currently only support this single scope. This will change in the future, but work to add more scopes is not currently scheduled.


Library/src/Multiplayer/OnlineRealtimeEngine.cpp
MAG-ElliotMorris: [Fix] I don't think we can call this removed if we still have two booleans that mean the same thing. The `return 0` branch below is a fragile path that simply should not exist, unless I'm missing something.

MAG-ElliotMorris: Ah, I see this got removed in a subsequent commit. Very good. Feel free to close this convo without comment.

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
```diff
@@ -335,19 +335,19 @@ class CSP_API OnlineRealtimeEngine : public csp::common::IRealtimeEngine
...
+    CSP_NO_EXPORT void DisableLeaderElection();
 
     /// @brief Check if the Leader Election feature is enabled.
     /// @return true if enabled, false otherwise.
     bool IsLeaderElectionEnabled() const;
 
     /// @brief Debug helper to get the id of the currently elected script leader.
     /// This should be updated when we fully support scopes. We will need to pass in the scopeId we want the leader for.
-    /// @return The id of the leader.
+    /// @return The id of the leader if leader election is enabled, 0 if it is not.
```
MAG-ElliotMorris: [nit] Is it worth having the zero branch at all? Can leader selection ever be disabled? The answer to that question should probably be here in the docs.

MAG-ElliotMorris: [Question] I don't suppose `0` can ever be a leader ID? Potential conflict.

MAG-ElliotMorris (approved):
LGTM. The only two comments are about one line of doc 😅

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
```diff
@@ -1161,7 +1129,7 @@ uint64_t OnlineRealtimeEngine::GetLeaderId() const
         }
         else
         {
-            return ElectionManager->GetLeader()->GetId();
```
MAG-mv, line 1164: [Consider] We could simplify this logic even further by putting the ServerSideElectionEnabled inside IsLeaderElectionEnabled. This way we can rremove additional checks.

I understand this would need to change when we implement aoi, but fow now, this would be simpler at least. It's wortht thinking about, although admittedly this may cause issues i havent considered.

MAG-mv (approved):

Library/include/CSP/Multiplayer/OnlineRealtimeEngine.h
MAG-AdamThorn: I have removed the zero branch and updated the documentation to give clarity around how leader election can be disabled and why it is safe to assume it is enabled.

Update: [50fdcbe](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1001/changes/50fdcbe510feda9cb827a0ca919c6e411410467c)

MAG-AdamThorn: It cannot, at least for the OnlineRealtimeEngine.

Library/src/Multiplayer/OnlineRealtimeEngine.cpp
MAG-AdamThorn: This logic has now been simplified as part of a change made in service of a comment above.

Update: [50fdcbe](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1001/changes/50fdcbe510feda9cb827a0ca919c6e411410467c)

PR #1004, opened 2026-07-06
[Of-1840] feat: Add legacy emscripten build and tests to ci

MAG-mv:
This PR adds support for our existing Wasm builds on CI. This PR adds:

- A new action to build the csp legacy wasm project
- A new workflow step to build and run tests
- A new script to generate the wasm package by copying wrapper files and binaries into the package folder

Ive decided to leave RelWithDebInfo out of this work, as it will require additional work to the wrapper gen, which us unnecessary for now. This can be added with the  new binding layer.

### DIFF
--- a/.github/actions/build-legacy-emscripten/action.yml
+++ b/.github/actions/build-legacy-emscripten/action.yml
@@ -1,0 +1,75 @@
+name: "Build and Install"
+description: "Fetch dependencies with Conan, configure CMake, build, and install"
+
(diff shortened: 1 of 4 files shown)


MAG-ElliotMorris (approved):
Probably would have had more criticisms if I didn't expect this to be taken out in a month or so, but legacy be legacy 🤷

MAG-AdamThorn (approved):

PR #1005, opened 2026-07-09
[OF-1681] Add WAF Bypass Header

MAG-mv:
This adds the abillity to set the WAF bypass fubnctionality using a function on SystemsManager. I chose to implement it here as SystemsManager owns the WebClient which adds the header.

I also added a mechanism to add this header in our tests locally, without exposing secrets in our repo. For our desktop tests, it reads the X_WAF_BYPASS environment variable.

For wasm tests, it was a bit harder, as we needed to forward the value to our internal test pages. I found using the bash shell on Windows, it would read directly from the environment variable. In Powershell, I had to manually export it in the value in the console.

Finally, i updated our request logging to hide the WAF value.


Library/src/Systems/SystemsManager.cpp
```diff
@@ -130,6 +130,19 @@ csp::common::IRealtimeEngine* SystemsManager::MakeRealtimeEngine(csp::common::Re
...
+    {
+        WebClient->SetWAFBypass(Value.HasValue() ? std::make_optional(std::string{ Value->c_str() }) : std::nullopt);
+        return true;
+    }
+    else
+    {
+        return false;
+    }
+}
+
```
MAG-ElliotMorris: [Comment] Shame about this being a passthrough, but tbh I don't really have a better approach in mind that isn't a rearchitecture of some sort, which you're right not to do for this ticket.

Library/src/Systems/SystemsManager.cpp
```diff
@@ -130,6 +130,19 @@ csp::common::IRealtimeEngine* SystemsManager::MakeRealtimeEngine(csp::common::Re
...
+bool SystemsManager::__SetWAFBypass(const csp::common::Optional<csp::common::String>& Value)
+{
+    if (WebClient != nullptr)
+    {
+        WebClient->SetWAFBypass(Value.HasValue() ? std::make_optional(std::string{ Value->c_str() }) : std::nullopt);
+        return true;
+    }
+    else
+    {
+        return false;
```
MAG-ElliotMorris: [Consider] Throwing is probably more appropriate here. This is an invalid use branch.

MultiplayerTestRunner/src/Utils.cpp
```diff
@@ -71,6 +71,12 @@ csp::systems::Profile CreateTestUser(bool AgeVerified /* = true */, csp::systems
     }
 }
 
+csp::common::Optional<csp::common::String> GetWAFBypassEnv()
+{
+    const auto WAFBypassEnv = std::getenv("X_WAF_BYPASS");
```
MAG-ElliotMorris: [fix] I feel this is too generic a name for an environment variable, could conflict with other software doing the same thing. `MCS_X_WAF_BYPASS` would be preferable.

WasmTesting/html_tests/CreateAvatar.html
```diff
@@ -16,7 +16,8 @@
 
         console.log ('spaceId: ', spaceId)
 
-        await initializeCSP(useDebugCSP);
+        const wafBypass = window.__CSP_TEST_CONFIG__?.wafBypass;
```
MAG-ElliotMorris, line 19: [Question] Do you know if this is actually neccesary for these tests? If the bypass is just about the rate limiting now, these tests don't exactly do a huge amount of user creation all at once. 🤷

Library/include/CSP/Common/fmt_Formatters.h
```diff
@@ -88,7 +88,19 @@ template <> struct formatter<csp::web::HttpRequest> : formatter<std::string_view
         csp::common::String Headers;
         for (const auto& Header : RequestHeaders)
         {
-            Headers.Append(fmt::format("\t{}: {}\n", Header.first, Header.second).c_str());
+            std::string HeaderValue;
+
+            if (Header.first == "X-WAF-Bypass")
```
MAG-ElliotMorris, line 93: [consider] Making this case insensitive, I can imagine something like this getting lower-cased across the wire in some situations.

Library/include/CSP/Common/fmt_Formatters.h
```diff
@@ -88,7 +88,19 @@ template <> struct formatter<csp::web::HttpRequest> : formatter<std::string_view
...
+            {
+                // Hide WAF Bypass secret from logs.
+                HeaderValue = "*****";
+            }
```
MAG-ElliotMorris, line 97: [Consider] Doing something like `const std::vector<std::string> ForbiddenList` here, to allow extension. I sort of think we may already be leaking other secrets in logs, which is going to matter more now that we're on actions. Cookies? BearerTokens? They're more ephemeral than WAF-Bypass, but still attack vectors.

... suppose that was always an issue, not like Teamcity was private. Still, i'm getting paranoid looking at this now.

Library/src/Common/PlatformUtils.cpp
```diff
@@ -0,0 +1,40 @@
+#include "PlatformUtils.h"
+
+namespace csp::common
+{
+
+std::optional<std::string> GetEnvironmentVariableValue(const std::string& Name)
+{
+#if defined(_MSC_VER)
+    // Prevent C2220 MSVC CRT warning
```
MAG-ElliotMorris: [fix] This comment says "fix warnings as errors warning", There should be  an underlying reason beyond just wanting the warning to go away, otherwise you should have just disabled it.

Say which warning in particular and why, with a link if possible. C2220 is just the warnings as errors code. You mean C4996 presumably, in regards to `std::getenv` being considered deprecated by microsoft.

Library/src/Common/PlatformUtils.cpp
```diff
@@ -0,0 +1,40 @@
...
+#else
+    const char* Value = std::getenv(Name.c_str());
+
+    if (Value == nullptr)
+    {
+        return std::nullopt;
+    }
+
+    return std::string { Value };
+#endif
```
MAG-ElliotMorris: [fix] Branching like this makes this performative imo, I think you should revert this change and just disable the warning. Microsoft is wrong to be so cautious here. Linux "suffers" from the exact same problem so you're not actually fixing anything by branching.

You can see Microsofts reasoning dosen't even really apply beyond C++11 https://en.cppreference.com/cpp/utility/program/getenv

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ All notable changes to this project will be documented in this file. For compile
   Developers making changes within CSP should use `GetPropertyDirect`/`SetPropertyDirect` in the cases they'd
   traditionally have used the original methods.
 
+- [OF-1681] feat: Add SystemsManager::__SetWAFBypass to allow csp devs to bypass firewall restrictions in tests. By @MAG-mv
+  This is an internal csp function and shouldn't be used by clients.
```
MAG-ElliotMorris: [Question] Why not? Clients may have the same test requirements that we do.

MAG-ElliotMorris (changes requested):

Library/src/Systems/SystemsManager.cpp
MAG-mv: I tried to get this working by reading environment variables internally, but this wasn't possible on wasmwithout us changing the way we expose csp as a module

WasmTesting/html_tests/CreateAvatar.html
MAG-mv: I didn't see errors in these tests wihout it. But setting this up by default will help prevent future devs from missing this when they're actually needed

Library/src/Common/PlatformUtils.cpp
MAG-mv: Reverted here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1005/commits/30afec3cdc8a490f7210b137fa5be61553c6b1d5

Library/src/Common/PlatformUtils.cpp
MAG-mv: Reverted here: https://github.com/magnopus-opensource/connected-spaces-platform/commit/30afec3cdc8a490f7210b137fa5be61553c6b1d5

Library/src/Common/PlatformUtils.cpp
MAG-mv: Added error supression here: https://github.com/magnopus-opensource/connected-spaces-platform/commit/f2cd390022168474f50b09792880343e9e8fe3d0

Library/src/Systems/SystemsManager.cpp
MAG-mv: Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1005/commits/ab005c9e36aec626070461657f58c52cefa35d69

MultiplayerTestRunner/src/Utils.cpp
MAG-mv: Updated here: [13361dc](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1005/commits/13361dc42f68b0b34d1a39d58405801d1caf98cd)

Library/include/CSP/Common/fmt_Formatters.h
MAG-mv: These are HTTP Requests and aren't received over the write - they are generated by us

Library/include/CSP/Common/fmt_Formatters.h
MAG-mv: We don't log cookies and we are also hidingthe password.

I'm cautious of doing this now as we may want to obsfucate different values in different ways. Until we have examples of more values that are logged here other than password and waf bypass i'm going to leave this

CHANGELOG.md
MAG-mv: Fixed up to mention its a magnopus org function instead: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1005/commits/90ff9614be96b804c274c165f22e2c9924aa4398

Library/include/CSP/Common/fmt_Formatters.h
MAG-ElliotMorris: Fair enough, glad you're sure we're not leaking already.

Library/include/CSP/Common/fmt_Formatters.h
MAG-ElliotMorris: Fair enough.

WasmTesting/html_tests/CreateAvatar.html
MAG-ElliotMorris: It also adds unnecessary complexity. I'd argue YAGNI in this case, but 🤷 your call, I do see the argument.

MultiplayerTestRunner/src/Utils.cpp
MAG-ElliotMorris: Remember to update commit links if you do a force push, they become invalid a short time after the force push. For future readers, is : https://github.com/magnopus-opensource/connected-spaces-platform/pull/1005/changes/32ad4342eed3c2989c24b655181dbaf1708da382

MAG-ElliotMorris (approved):

Library/src/Common/PlatformUtils.cpp
MAG-ElliotMorris: I left a comment on the change but it's not showing up in the UI for me, apologies if dupe.

[Nit] This feels overly broad to disable the entire CRT hardening suite. A more local disable specifically targetting std::getenv (either if there's a specific code, or scoped disables around its use in source) may have been more virtuous. Only the multiplayer test runner though (to my surprise, do we not have the same error config across all our projects?) so therefore only a [nit] since it dosen't matter that much.

MAG-AdamThorn (approved):

PR #1007, opened 2026-07-17
[OF-1886] feat!: Expose event specializations via network event bus

MAG-AdamThorn:
NetworkEventBus
Add new dedicated Listen and StopListen methods for our event specializations, as well as dedicated std::unordered_map containers and callbacks for each event type.

Event Specializations are:
- `AccessControlChanged`
- `AssetDetailBlobChanged`
- `AsyncCallCompleted`
- `Conversation`
- `SequenceChanged`

The existing `ListenNetworkEvent` and `StopListenNetworkEvent` methods have been renamed to `ListenCustomNetworkEvent` and `StopListenNetworkEvent` to better reflect their purpose as a means for clients to register their own custom events. In addition, rather than take a `NetworkEventRegistration` as an argument they now take `EventReceiverId` and `EventName` strings directly. This aligns usage with the new event specialization methods.

The `NetworkEventBus` is used internally by several of our systems. These were all using the old `ListenNetworkEvent` method and then casting the resultant `NetworkEventData` to the appropriate event data type. These systems have been updated to call the new event specialization methods directly, removing the need to perform a static cast.

The `SpaceSystem` has a `DuplicateSpaceAsync` method which fires the `AsyncCallCompleted` event on completion of the operation. The `SpaceSystem` had dedicated API for registering a callback with this event and for responding to its firing. Now that the event specializations have been added to the `NetworkEventBus`, including Listen and StopListen methods for `AsyncCallCompleted`, the `SpaceSystem` no longer requires this functionality and it has been removed.

The existing `NetworkEventBus` tests have been updated to reflect the change to the `ListenNetworkEvent` and `StopListenNetworkEvent` method signatures. New tests have also been added to verify that the new Event Specialization methods work correctly. I have elected to test just those related to `AsyncCallCompleted` as they all delegate to shared logic internally.

In addition to the above I have:
- updated the `NetworkEventManagerImpl::SendNetworkEvent()` method to add support for deserialising the `StringMap` replicated value type.
- updated one of the `SpaceSystem` tests which was using the legacy AsyncCallCompleted methods.

**Breaking changes:**
- The NetworkEventBus `ListenNetworkEvent(NeworkEventRegistration)` > `ListenCustomNetworkEvent(String EventReceiverId, String EventName)`.
- The NetworkEventBus `StopListenNeworkEvent(NeworkEventRegistration)` > `StopListenNeworkEvent(String EventReceiverId, String EventName)`.
- Removed the `SetAsyncCallCompletedCallback()` and `OnAsyncCallCompletedEvent()` methods from the `SpaceSystem`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,36 @@ All notable changes to this project will be documented in this file. For compile
 
 ## [6.46.0]
 
+### 🔌 API Changes
+
+#### NetworkEventBus
+- 🔄 **Updated:** `ListenNetworkEvent(NeworkEventRegistration Registration, NetworkEventCallback Callback)` → `ListenCustomNetworkEvent(String EventReceiverId, String EventName, NetworkEventCallback Callback)`
+- 🔄 **Updated:** `StopListenNetworkEvent(NeworkEventRegistration Registration)` → `StopListenCustomNetworkEvent(String EventReceiverId, String EventName)`
+
+- ➕ **Added:** `ListenAccessControlChangedEvent(String EventReceiverId, AccessControlChangedEventCallback Callback)`
+- ➕ **Added:** `ListenAssetDetailBlobChangedEvent(String EventReceiverId, AssetDetailBlobChangedEventCallback Callback)`
+- ➕ **Added:** `ListenAsyncCallCompletedEvent(String EventReceiverId, String OperationName, AsyncCallCompletedEventCallback Callback)`
+- ➕ **Added:** `ListenConversationEvent(String EventReceiverId, ConversationEventCallback Callback)`
+- ➕ **Added:** `ListenSequenceChangedEvent(String EventReceiverId, SequenceChangedEventCallback Callback)`
+- ➕ **Added:** `StopListenAccessControlChangedEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenAssetDetailBlobChangedEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenAsyncCallCompletedEvent(String EventReceiverId, String OperationName)`
+- ➕ **Added:** `StopListenConversationEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenSequenceChangedEvent(String EventReceiverId)`
+
+#### SpaceSystem
+- ➖ **Removed:** `SetAsyncCallCompletedCallback()`
+- ➖ **Removed:** `OnAsyncCallCompletedEvent()`
+
+### 🔥 ❗Breaking Changes
+
+- [OF-1886] feat!: Expose event specializations via network event bus. By @MAG-AdamThorn
+  Added new dedicated Listen and StopListen methods for our event specializations, as well as dedicated std::unordered_map containers and callbacks for each event type. The existing `ListenNetworkEvent` and `StopListenNetworkEvent` methods have been renamed to `ListenCustomNetworkEvent` and `StopListenNetworkEvent` to better reflect their purpose as a means for clients to register their own custom events. In addition, rather than take a NetworkEventRegistration as an argument they now take EventReceiverId and EventName strings directly. This aligns usage with the new event specialization methods.
+  The `SpaceSystem` has a `DuplicateSpaceAsync` method which fires the AsyncCallCompleted event on completion of the operation. The SpaceSystem had dedicated API for registering a callback with this event and for responding to its firing. Now that the event specializations have been added to the NetworkEventBus, including Listen and StopListen methods for AsyncCallCompleted, the SpaceSystem no longer requires this functionality and it has been removed.
+  The existing NetworkEventBus tests have been updated to reflect the change to the `ListenNetworkEvent` and `StopListenNetworkEvent` method signatures. Tests have also been added to verify that the new Event Specialization methods work correctly. I have elected to test just those related to `AsyncCallCompleted` as they all delegate to shared logic internally.
+  Please see the API Changes section above for further details.
+
+
 ### 💫 💥 Code Refactors
 
 - [NT-0] refac!: Rename `GetSchemaProperty`/`SetSchemaProperty` on `ComponentBase` to `GetProperty`/`SetProperty`. By @mag-lt.
--- a/Library/include/CSP/Multiplayer/NetworkEventBus.h
+++ b/Library/include/CSP/Multiplayer/NetworkEventBus.h
@@ -59,9 +59,19 @@ namespace csp::multiplayer
 class EventDeserialiser;
 enum class ErrorCode;
 
-// The callback used to register to listen to events.
+// The callback used to register to listen to custom network events.
 // NetworkEventData lifetime is tied to the callback, do not attempt to store it via reference.
 typedef std::function<void(const csp::common::NetworkEventData& NetworkEventData)> NetworkEventCallback;
+
+// Event listener specializations
+// Callbacks used to register to listen to event specializations.
+// The lifetime of network event data types are tied to the callback, do not attempt to store it via reference.
+typedef std::function<void(const csp::common::AccessControlChangedNetworkEventData& AccessControlChangedEventData)> AccessControlChangedEventCallback;
+typedef std::function<void(const csp::common::AssetDetailBlobChangedNetworkEventData& AssetDetailBlobChangedEventData)>
+    AssetDetailBlobChangedEventCallback;
+typedef std::function<void(const csp::common::AsyncCallCompletedEventData& AsyncCallCompletedEventData)> AsyncCallCompletedEventCallback;
+typedef std::function<void(const csp::common::ConversationNetworkEventData& ConversationNetworkEventData)> ConversationEventCallback;
+typedef std::function<void(const csp::common::SequenceChangedNetworkEventData& SequenceChangedEventData)> SequenceChangedEventCallback;
 
 /*
  * @brief Details about a network event registrations to serve as a key in the event map.
@@ -74,10 +84,10 @@ class CSP_API NetworkEventRegistration
 {
 public:
     /// @brief Construct a NetworkEventRegistration
-    /// @param EventReceiverId csp::common::String : The identifying name for the event receiver, used for management purposes, allowing clients to
+    /// @param EventReceiverId : The identifying name for the event receiver, used for management purposes, allowing clients to
     /// register multiple interests in single events. May be any arbitrary unique string. This is distinct from client ID. A single client/application
     /// may register multiple receivers if they choose.
-    /// @param EventName csp::common::String : The identifying name for the event. May be any arbitrary string.
+    /// @param EventName : The identifying name for the event. May be any arbitrary string.
     NetworkEventRegistration(const csp::common::String& EventReceiverId, const csp::common::String& EventName)
         : EventReceiverId(EventReceiverId)
         , EventName(EventName)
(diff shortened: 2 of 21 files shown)


Library/include/CSP/Multiplayer/NetworkEventBus.h
```diff
@@ -59,10 +59,20 @@ namespace csp::multiplayer
 class EventDeserialiser;
 enum class ErrorCode;
 
-// The callback used to register to listen to events.
+// The callback used to register to listen to custom network events.
 // NetworkEventData lifetime is tied to the callback, do not attempt to store it via reference.
 typedef std::function<void(const csp::common::NetworkEventData& NetworkEventData)> NetworkEventCallback;
```
MAG-ElliotMorris: [Consider] Renaming the callback and the data type itself as well.

Library/src/Multiplayer/NetworkEventBus.cpp
```diff
@@ -31,20 +31,87 @@
 #include <memory>
 #include <optional>
 
+namespace
+{
+
+template <typename EventContainer, typename EventCallback>
+void RegisterEventCallback(csp::common::LogSystem& LogSystem, const csp::common::String& EventReceiverId, csp::common::String EventName,
```
MAG-ElliotMorris, line 38: [Nit] Putting parameters that are just for logging in a function signature can be a tad confusing, a wee inline comment explaining that might be helpful, especially as it's surprising why some of these arguments have preconditional checks and some don't.

Library/src/Multiplayer/NetworkEventBus.cpp
```diff
@@ -53,74 +120,276 @@ NetworkEventBus::NetworkEventBus(MultiplayerConnection* InMultiplayerConnection,
...
 
-void NetworkEventBus::ListenNetworkEvent(NetworkEventRegistration Registration, NetworkEventCallback Callback)
+void NetworkEventBus::ListenCustomNetworkEvent(csp::common::String EventReceiverId, csp::common::String EventName, NetworkEventCallback Callback)
 {
+    if (EventName.IsEmpty())
+    {
+        LogSystem.LogMsg(
+            csp::common::LogLevel::Error, "NetworkEventBus::ListenCustomNetworkEvent: Expected non-empty EventName. Registration denied.");
+        return;
+    }
```
MAG-ElliotMorris: [Consider] I wonder if you could follow the pattern with some other arguments and do the validation in `RegisterEventCallback` instead. We'd lose a small amount of log clarity true, but it would reduce quite a bit of duplication getting in the way of mentally parsing these specializations. Same for the Callback argument below. You could use the "EventName" field to give hints as to what event the logs are referring to in either case.

In theory, this could all just be a oneline passthrough.

Library/src/Multiplayer/NetworkEventBus.cpp
```diff
@@ -53,74 +120,276 @@ NetworkEventBus::NetworkEventBus(MultiplayerConnection* InMultiplayerConnection,
...
+                        "ReceiverId: {}. "
+                        "Deregistration denied.",
+                EventReceiverId)
+                .c_str());
+    }
+}
+
+void NetworkEventBus::StopListenAllNetworkEvents(const csp::common::String& EventReceiverId)
+{
+    auto StopListenMatchingEventReceivers = [&](const auto& RegisteredEvents, auto StopListenAction) -> size_t
```
MAG-ElliotMorris: [nit] You probably want to use `&&` as a default when passing callables in a generic context (`StopListenAction`). Although it won't really matter in this case.

Library/src/Multiplayer/NetworkEventBus.cpp
```diff
@@ -53,74 +120,276 @@ NetworkEventBus::NetworkEventBus(MultiplayerConnection* InMultiplayerConnection,
...
+    auto* RegistrationsIt = Registrations.begin();
+
+    // Copy the NetworkEventRegistration and increment pointer
+    auto CopyRegistrations = [&RegistrationsIt](const auto& EventMap)
+    {
+        for (const auto& [Key, Value] : EventMap)
+        {
+            *RegistrationsIt++ = Key;
+        }
+    };
```
MAG-ElliotMorris: [Nit] This is a bit "clever" to my eye. A plain index based version might read more obviously.

``` cpp
    size_t Index = 0;

    auto CopyRegistrations = [&Index, &Registrations](const auto& EventMap)
    {
        for (const auto& [Key, Value] : EventMap)
        {
            Registrations[Index++] = Key;
        }
    };
```

A comment that expresses that we're flattening lots of buffers into one might also help.

Library/src/Multiplayer/NetworkEventBus.cpp
```diff
@@ -53,74 +120,276 @@ NetworkEventBus::NetworkEventBus(MultiplayerConnection* InMultiplayerConnection,
...
+    else
+    {
+        LogSystem.LogMsg(csp::common::LogLevel::Verbose,
+            fmt::format("NetworkEventBus::StopListenAllNetworkEvents: Removed {} network event registration/s with EventReceiverId: {}.",
+                RegistrationsToRemoveCount, EventReceiverId)
                 .c_str());
     }
 }
 
 csp::common::Array<NetworkEventRegistration> NetworkEventBus::AllRegistrations() const
```
MAG-ElliotMorris, line 328: [Consider] I wonder if this API is useful anymore, given that there is no longer a single interface that can work with `NetworkEventRegistration` across all event types. The intent of the original API was to have this singular event management object, but as that's no longer the case, perhaps we should remove this.

CHANGELOG.md
```diff
@@ -4,6 +4,36 @@ All notable changes to this project will be documented in this file. For compile
...
+- ➕ **Added:** `ListenSequenceChangedEvent(String EventReceiverId, SequenceChangedEventCallback Callback)`
+- ➕ **Added:** `StopListenAccessControlChangedEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenAssetDetailBlobChangedEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenAsyncCallCompletedEvent(String EventReceiverId, String OperationName)`
+- ➕ **Added:** `StopListenConversationEvent(String EventReceiverId)`
+- ➕ **Added:** `StopListenSequenceChangedEvent(String EventReceiverId)`
+
+#### SpaceSystem
+- ➖ **Removed:** `SetAsyncCallCompletedCallback()`
+- ➖ **Removed:** `OnAsyncCallCompletedEvent()`
```
MAG-ElliotMorris, line 26: [Comment] Neat

MAG-ElliotMorris (approved):
Approving structurally, although obviously there's still test issues to be worked out.

Main thing to my mind is whether the remove-all function still makes sense.

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: We use it for our tests. I had considered updating the method to take an optional NetworkEvent value but that would mean making that enum public. I think I will do that.

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: Yep that is better, will make that change.

Library/include/CSP/Multiplayer/NetworkEventBus.h
```diff
@@ -74,10 +84,10 @@ class CSP_API NetworkEventRegistration
 {
 public:
     /// @brief Construct a NetworkEventRegistration
-    /// @param EventReceiverId csp::common::String : The identifying name for the event receiver, used for management purposes, allowing clients to
+    /// @param EventReceiverId : The identifying name for the event receiver, used for management purposes, allowing clients to
```
MAG-mv, line 87: [Comment] Thanks for cleaning up comments

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-mv: Agreed. It took me a while to figure out whas was going on here

Library/src/Multiplayer/NetworkEventManagerImpl.cpp
```diff
@@ -27,6 +27,74 @@
 #include <limits>
 #include <signalrclient/signalr_value.h>
 
+namespace
+{
+
+// Serialises a single ReplicatedValue into a [TypeId, [Field]] signalr::value that is expected by the MCS ItemComponentData wire format.
+signalr::value SerialiseReplicatedValueToSignalRValue(const csp::common::ReplicatedValue& Value)
+{
```
MAG-mv, line 35: [Comment] We really need to do the ticket to remove this class

MAG-mv (approved):

Library/include/CSP/Multiplayer/NetworkEventBus.h
MAG-AdamThorn: I have renamed the callback to `CustomNetworkEventCallback`.

I did not want to rename the `NetworkEventData` as it is the base class for the event specializations.

Update: [c567925](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1007/changes/c567925a8396726f3fbf91b348bbcd42c3932193)

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: Agreed, fix: [ac1891a](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1007/changes/ac1891a22f63ce8a550833e9acdb0f08882178e4)

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: Addressed here: [1c72dda](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1007/changes/1c72dda82f1a25bc479e33ee37e8d0d470f01dc0)

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: Agreed, addressed in the same commit listed above: [1c72dda](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1007/changes/1c72dda82f1a25bc479e33ee37e8d0d470f01dc0)

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: I did update this to make the NetworkEvent enum public so that it could be passed to the AllRegistrations method as an optional - the idea being that if a value was specified then only registrations for that event type would be returned. However, the wrapper generator does not support optional enum args and so I have undone this change and will leave the method as is.

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-AdamThorn: Update here: [41b6441](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1007/changes/41b6441c9aa17d78ced63e578a4df5d3b0687b57)

Library/src/Multiplayer/NetworkEventBus.cpp
MAG-ElliotMorris: Fair enough. I'm still dubious on the value of the `NetworkEventRegistration` type now that it's been removed from the bulk of the registration interfaces, but 🤷

PR #1009, opened 2026-07-21
[OF-1886] test: Address test synchonization issue

MAG-AdamThorn:
I discovered a test synchronization issue which was resulting in one of the NetworkEventBus tests (EventDispatchReplicatedValueException) being flacky on CI.

Previously the test was waiting 5s to confirm that the NetworkEventFuture was not being set (which it never was) and then tearing down the test, potentially before the error log we were waiting for had fired. The updated test now uses a promise/future to wait for the error log before exiting the test.


MAG-ElliotMorris (approved):
Good catch, this was probably my mistake initially not considering the log itself being off thread.

MAG-mv (approved):

PR #1010, opened 2026-07-21
[NT-0] test: Fix an issue with a POI test

MAG-AdamThorn:
One of our POI tests (`GetPOIInsideCircularAreaExcludeOtherTest`) was failing due to an ordering issue. The clean up of the POIs created as part of the test took place after the test assertions. Because the POI coordinates were not randomised, if a run of the test failed, the POIs would not be cleaned up, and they would be returned as matching POIs in subsequent runs.

The fix was to randomise the POI lat/lon positions and to clean up the POIs prior to the assertions so that they get removed  regardless.

### DIFF
--- a/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
+++ b/Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
@@ -290,26 +290,30 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaE
 
     LogInAsNewTestUser(UserSystem, UserId);
 
+    // Randomise the search origin each run so that any POI orphaned as a result of a prior failed run
+    // does not fall inside this run's search area and inflate the result count.
+    csp::systems::GeoLocation SearchLocationOrigin;
+    SearchLocationOrigin.Latitude = RandomRangeDouble(-80.0, 80.0);
+    SearchLocationOrigin.Longitude = RandomRangeDouble(-170.0, 170.0);
+    double SearchRadius = 130000;
+
     csp::systems::GeoLocation InsidePOILocation;
-    InsidePOILocation.Latitude = 45.0;
-    InsidePOILocation.Longitude = 160.0;
+    InsidePOILocation.Latitude = SearchLocationOrigin.Latitude + 1.0;
+    InsidePOILocation.Longitude = SearchLocationOrigin.Longitude;
 
     csp::systems::PointOfInterest InsidePointOfInterest;
     CreatePointOfInterest(POISystem, nullptr, InsidePOILocation, nullptr, InsidePointOfInterest);
 
+    // Placed on the opposite side of the globe from the search origin to ensure it will always be outside the search radius.
     csp::systems::GeoLocation OutsidePOILocation;
-    OutsidePOILocation.Latitude = -45.0;
-    OutsidePOILocation.Longitude = -160.0;
+    OutsidePOILocation.Latitude = -SearchLocationOrigin.Latitude;
+    OutsidePOILocation.Longitude = SearchLocationOrigin.Longitude >= 0.0 ? SearchLocationOrigin.Longitude - 180.0
+                                                                         : SearchLocationOrigin.Longitude + 180.0;
 
     csp::systems::PointOfInterest OutsidePointOfInterest;
     CreatePointOfInterest(POISystem, nullptr, OutsidePOILocation, nullptr, OutsidePointOfInterest);
 
     // Search for the newly created POI inside a circular area
-    csp::systems::GeoLocation SearchLocationOrigin;
-    SearchLocationOrigin.Latitude = 44.0;
-    SearchLocationOrigin.Longitude = 160.0;
-    double SearchRadius = 130000;
-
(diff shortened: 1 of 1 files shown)


Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -318,12 +322,13 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaE
 
     const csp::common::Array<csp::systems::PointOfInterest>& POICollection = Result.GetPOIs();
 
-    EXPECT_EQ(POICollection.Size(), 1);
-    EXPECT_EQ(POICollection[0].Name, InsidePointOfInterest.Name);
-
+    // Delete the POIs prior to the assertions to ensure that they are cleaned up even if the assertions fail.
```
mag-lt, line 329: [Question] Is this not all we need to do? The randomisation makes a clash less likely to happen if they aren't cleaned up, but this implies they will be now?

[Comment] We have more issues like this elsewhere, most of the tests are not robust to failing half way through and impact subsequent tests if they do fail (due to not cleaning up). In cases like this, without thinking too much about it, we basically want to ensure something happens regardless of the way we exit (success, error, exception being thrown etc). This could be solved various different ways, either by leaning more on the test framework to put things in GTest's `TearDown()` hooks, or misc different ways of using RAII. i.e. wrap the test data up in a struct locally and hook in cleanup in the destructor, or have a general `ScopeExit` helper for running arbitrary lambdas on destruction. Alternatively it could be done the other way round, where at the start of each test (or in a GTest `SetUp` hook) we ensure we are in a totally clean state.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -290,26 +290,30 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaE
 
     LogInAsNewTestUser(UserSystem, UserId);
 
+    // Randomise the search origin each run so that any POI orphaned as a result of a prior failed run
+    // does not fall inside this run's search area and inflate the result count.
+    csp::systems::GeoLocation SearchLocationOrigin;
```
mag-lt, line 295: [nit] I know these were just moved up, but both `SearchLocationOrigin` (if constructed via the construtor), and `SearchRadius` could be const

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
```diff
@@ -290,26 +290,30 @@ CSP_PUBLIC_TEST(CSPEngine, PointOfInterestSystemTests, GetPOIInsideCircularAreaE
...
     csp::systems::GeoLocation InsidePOILocation;
-    InsidePOILocation.Latitude = 45.0;
-    InsidePOILocation.Longitude = 160.0;
+    InsidePOILocation.Latitude = SearchLocationOrigin.Latitude + 1.0;
+    InsidePOILocation.Longitude = SearchLocationOrigin.Longitude;
 
     csp::systems::PointOfInterest InsidePointOfInterest;
     CreatePointOfInterest(POISystem, nullptr, InsidePOILocation, nullptr, InsidePointOfInterest);
 
+    // Placed on the opposite side of the globe from the search origin to ensure it will always be outside the search radius.
```
mag-lt: [Comment] `SearchRadius` not having any obvious units, and me not knowing the size of the globe, this wasn't actually particularly obvious to me. I think that can mostly be put down to my ignorance, but it wouldn't hurt to be overly explicit in the comment.

mag-lt (approved):
If I understand correctly, ensuring we clean up unconditionally is actually the fix here? If so, I'd advocate for just doing that. The random points just make a clash less likely to happen in a case where we don't start in a state where there are no POIs (and that could also probably be avoided yet another way, by having a unique asset collection per test)

Having said all that, I don't see a good reason to block this going in as-is.

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-AdamThorn: Yes absolutely, our tests need a thorough review. This change is targeted at this particular issue to get PR's and the release through.

In answer to your other point I feel the randomisation is required. If the locations are hardcoded, as they were previously, and then the test fails or falls over at any point prior to clean up then we will be left with these dangling POIs.

MAG-mv (approved):

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-AdamThorn: Have updated names and added comments to help make this clearer.

Change: [b808950](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1010/changes/b8089501311cf3eb3931bf132317cccec237306b)

Tests/src/PublicAPITests/PointOfInterestSystemTests.cpp
MAG-AdamThorn: I have made SearchRadius const. I elected not to make SearchLocationOrigin const as using the ctor made it less clear which argument was latitude and which was longitude.

Change: [b808950](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1010/changes/b8089501311cf3eb3931bf132317cccec237306b)

PR #1012, opened 2026-08-07
[OF-1845] Redistributable CSP SDK

MAG-mv:
This PR introduces a way of creating a redistributable "sdk", that packages all third-party dependencies into one directory. This allows consumers to link CSP without needing Conan, or to source third-party libs themselves to link.

To do this, it adds the additional steps:

Conan Deployment - A custom conan deployer has been added which is a python script that iterates all third-party packages and copies libs, public includes, and licenses into a staging directory. This also generates a cmake file that contains all third party lib names. This can be optionally ran at the conan install step.

CMake SDK generation - This takes the generated cmake file and directory and deployer, validates the structure. and installs the dependencies into the final install folder.

Shared builds currently work with basic integration tests, but I may need to extend the SDK generation to shared builds as well, specifically for public dependencies. I will know if this is necessary as part of the client testing. I will do this in a seperate PR if needed.


.github/workflows/android-build.yml
```diff
@@ -79,6 +79,7 @@ jobs:
           output_folder: build/android-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
           build_folder: build/android-${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}/build/${{ matrix.config.build_type }}
           install_prefix: install/android/${{ matrix.config.preset_prefix }}-${{ matrix.linkage.name }}
+          generate_sdk: ${{ matrix.linkage.name == 'static' && inputs.upload_install }}
```
MAG-ElliotMorris: [fix] I hesitate to call this an SDK, that has implications [far beyond](https://www.geeksforgeeks.org/software-engineering/what-is-an-sdk-sdk-vs-library-vs-framework/#what-is-software-development-kit) what is being done here. You should call this what this is, which is packaging the libraries.

.github/actions/build-and-install/action.yml
```diff
@@ -32,6 +32,11 @@ inputs:
     required: false
     default: "install"
 
+  generate_sdk:
+    description: "Whether to generate a redistributable static CSP SDK"
```
MAG-ElliotMorris: [fix] Say what that means. "static CSP SDK" isn't descriptive, descriptions should say what is actually happening, not gesture at undefined abstractions.

.github/actions/build-and-install/action.yml
```diff
@@ -71,17 +76,39 @@ runs:
...
+        )
+
+        if [ "${{ inputs.generate_sdk }}" = "true" ]; then
+          CMAKE_ARGS+=(
+            -DCSP_GENERATE_SDK=ON
+            -DCSP_SDK_DEPENDENCIES_DIR="${{ github.workspace }}/${{ inputs.output_folder }}/sdk-dependencies"
+          )
+        fi
+
+        cmake "${CMAKE_ARGS[@]}"
```
MAG-ElliotMorris: [Consider] If this is the best way to be appending these config variables. Maybe adding another entry as a cmake preset and letting if flow through via the existing input would be simpler with less need to branch.

Similar case for the above branching for conan adding the deployer, although I'm less sure what to do about that.

.github/actions/build-and-install/action.yml
```diff
@@ -71,17 +76,39 @@ runs:
...
+          --build=missing
           --profile:host="$PROFILE_PATH"
+        )
+
+        if [ "${{ inputs.generate_sdk }}" = "true" ]; then
+          CONAN_ARGS+=(
+            --deployer=./conan/csp_deployer.py
+            --deployer-folder="${{ inputs.output_folder }}/sdk-dependencies"
+          )
+        fi
```
MAG-ElliotMorris: [Consider] Stuff like this is surprising, and does little to explain what is happening here at the call site. Please add descriptive comments when it is not obvious why something is in the file. `csp_deployer` executed when you are generating an sdk could mean anything.

PR #591, opened 2025-01-17
[OF-1494] Enable warnings as errors (Pure Refactor)

MAG-ElliotMorris:
Enable warnings as errors across all our projects, at default warning level.
Update the codebase to pass these new checks.

All warnings related to generated code have been suppressed, as has all external/thirdparty warnings. This surfaces a conceptual mistake in making some of our thirdparty libraries premake projects rather than using their packaging, as now we're unable to easily distinguish them as thirdparty, leading to manual maintenance as has been done here.

Deprecation warnings (on platforms that treat them as errors on w3 or equivalent) are suppressed as errors, but will still output. 

> [!IMPORTANT]  
> This PR excludes certain warning fixes that have a _very small_ chance of causing regressions. ([One](https://github.com/magnopus-opensource/connected-spaces-platform/pull/584/commits/1713652cd041daadfe0ff35b01956e6fe6787ba7), [Two](https://github.com/magnopus-opensource/connected-spaces-platform/pull/584/commits/8301d1eb2745b848a8210076c34e196547842faa)) They will come immediately in a subsequent PR, in order to better isolate integration risk. The warnings have been locally suppressed in this PR.

PR #1013, opened 2026-08-10
[OB-5754] fix!: Expose base component name vs deprecated name to scripting for Image, ExternalLink, FiducialMarker, and VideoPlayer

mag-lt:
This regressed in 289c271103b622e22de83c0f1413511e192200f9.

I had missed when making those changes that the video player component had specific custom handling for exposing the name (introduced as part of the changes landed in a40ea2b5063e956dee4ab6aaceaedaa32c60804e, specifically the `DEFINE_SCRIPT_PROPERTY_STRING_ADAPTNAME` macro [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/591/changes/387507c66b2ff19cd033a36bc967433a87a6721e#diff-79b320b89915f938f7d35a76573e6f0368cddbf03dac43aeeada2d2062048470R30-R32)), and so mistakenly exposed the deprecated property instead. The solution is to not expose the deprecated property to scripting via the schema, allowing the component to pick up the base class's behaviour for exposing the component name to scripting as `name`.

More details in the fix commit.

### DIFF
--- a/Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
+++ b/Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
@@ -38,7 +38,7 @@ const auto Schema = ComponentSchema {
     csp::common::Array<ComponentProperty> {
         {
             static_cast<ComponentProperty::KeyType>(VideoPlayerPropertyKeys::Name_DEPRECATED),
-            "name",
+            {}, // not exposed to scripting
             "",
         },
         {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file. For compile
 
 ## [6.47.0]
 
+### 🐛 🔨 Bug Fixes
+
+- [OB-5754] fix: Expose base component name vs deprecated name to scripting for VideoPlayer. @mag-lt.
+  This addresses a regression introduced in 6.38.0 made in the context of binding components created w/ Schemas into scripts automatically.
+  The `name` property in scripts accidentally switched back to the older deprecated name property on the video player vs the name corresponding
+  to `ComponentBase::GetComponentName()`/`ComponentBase::SetComponentName()`. This has now been rectified.
+
 ###  🔨 🔨 Chore
 
 - [OF-1835] chore: Remove handling for old AsyncCallCompleted event data structure. By@MAG-AdamThorn
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,13 @@ All notable changes to this project will be documented in this file. For compile
 
 ### 🐛 🔨 Bug Fixes
 
-- [OB-5754] fix: Expose base component name vs deprecated name to scripting for VideoPlayer. @mag-lt.
+- [OB-5754] fix!: Expose base component name vs deprecated name to scripting for Image, ExternalLink, FiducialMarker, and VideoPlayer components. @mag-lt.
   This addresses a regression introduced in 6.38.0 made in the context of binding components created w/ Schemas into scripts automatically.
   The `name` property in scripts accidentally switched back to the older deprecated name property on the video player vs the name corresponding
   to `ComponentBase::GetComponentName()`/`ComponentBase::SetComponentName()`. This has now been rectified.
+  Additionally, `Image`, `ExternalLink`, and `FiducialMarker` have also been updated to return the `ComponentName` vs the deprecated name in scripts,
+  as these were never updated to do so. This is technically a breaking change (without doing a data migration on the server), but should be low impact
+  in practice.
 
 ###  🔨 🔨 Chore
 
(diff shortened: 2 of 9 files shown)


MAG-ElliotMorris (approved):
I presume you did a quick scan to check this was just VideoPlayerSpaceComponent and not a broader pattern?

MAG-SamBirley (commented):
Are we sure we want to leave the other components which still expose deprecated names as-is? Feels a little like leaving latent bugs on the table, although i do understand the rationale of only fixing what's immediately needed.

mag-lt:
> Are we sure we want to leave the other components which still expose deprecated names as-is? Feels a little like leaving latent bugs on the table, although i do understand the rationale of only fixing what's immediately needed.

I can't see it being deliberate in the first place (to have not updated them), but they're technically not part of the regression (as the video component was already using the `ComponentName`, this fix is specifically about putting that specific thing back how it was). Anyone using `.name` in a script for an `ExternalLink`, `FiducialMarker`,
and `Image` will currently be getting the deprecated names, so changing this to the `ComponentName` _could_ change the behaviour of an existing script, but I think it is the sensible / consistent change. Happy to make those changes if there is consensus to do so whilst in the area, but for the purposes of fixing a specific regression I personally lean towards only changing what is directly needed.

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
MAG-AdamThorn, line 193: [Question] Does this not also need to change?

mag-lt: This isn't related to what I understand the regression to be, which is about script bindings. `VideoPlayerSpaceComponent::GetName()`/`VideoPlayerSpaceComponent::SetName(...)` are deprecated methods for the underlying deprecated property (as marked in the [header](https://github.com/magnopus-opensource/connected-spaces-platform/blob/0d484af5f0fa2a7d93c20087ce8a3797062cb37d/Library/include/CSP/Multiplayer/Components/VideoPlayerSpaceComponent.h#L120)).

Calling `.name` on a component from a script will by default call through to `ComponentBase::GetComponentName()`/`ComponentBase::SetComponentName(...)`. The video player getter/setters for name are just other vestigial stuff.

The test added as part of these changes verifies the intended behaviour.

MAG-SamBirley:
> > Are we sure we want to leave the other components which still expose deprecated names as-is? Feels a little like leaving latent bugs on the table, although i do understand the rationale of only fixing what's immediately needed.
> 
> I can't see it being deliberate in the first place (to have not updated them), but they're technically not part of the regression (as the video component was already using the `ComponentName`, this fix is specifically about putting that specific thing back how it was). Anyone using `.name` in a script for an `ExternalLink`, `FiducialMarker`, and `Image` will currently be getting the deprecated names, so changing this to the `ComponentName` _could_ change the behaviour of an existing script, but I think it is the sensible / consistent change. Happy to make those changes if there is consensus to do so whilst in the area, but for the purposes of fixing a specific regression I personally lean towards only changing what is directly needed.

Ok, thanks for the rationalization!

Library/src/Multiplayer/Components/VideoPlayerSpaceComponent.cpp
MAG-AdamThorn: Gotcha

MAG-ElliotMorris:
Aye, making any further changes than this would require either a compatibility shim in-code or a data-migration. Not worth it.

MAG-AdamThorn:
> > Are we sure we want to leave the other components which still expose deprecated names as-is? Feels a little like leaving latent bugs on the table, although i do understand the rationale of only fixing what's immediately needed.
> 
> I can't see it being deliberate in the first place (to have not updated them), but they're technically not part of the regression (as the video component was already using the `ComponentName`, this fix is specifically about putting that specific thing back how it was). Anyone using `.name` in a script for an `ExternalLink`, `FiducialMarker`, and `Image` will currently be getting the deprecated names, so changing this to the `ComponentName` _could_ change the behaviour of an existing script, but I think it is the sensible / consistent change. Happy to make those changes if there is consensus to do so whilst in the area, but for the purposes of fixing a specific regression I personally lean towards only changing what is directly needed.

If we elect not to address the other components now then we should at least start a conversation with Solutions/Client teams to discuss making the same change elsewhere. If we don't we will just forget about it until someone encounters the same issue with another component with a deprecated name property. @mag-lt are you happy to start that convo?

MAG-AdamThorn (approved):

mag-lt:
> If we elect not to address the other components now then we should at least start a conversation with Solutions/Client teams to discuss making the same change elsewhere. If we don't we will just forget about it until someone encounters the same issue with another component with a deprecated name property. @mag-lt are you happy to start that convo?

I've flagged to Solutions on Slack, in case they're aware of any scripts that are directly or indirectly relying on the name of any of the components (specifically `ExternalLink`, `FiducialMarker`, and `Image`, which weren't switched over to bind to `ComponentName` when calling `.name`). I doubt a data migration was done at the time for any of the other components, but I don't have visibility on that.

I'll make the changes in the context of this PR for the other components if no concerns are flagged. So, leaving this open for a bit. If nothing comes back, or there are concerns, I'll merge this as-is.

mag-lt:
> I'll make the changes in the context of this PR for the other components if no concerns are flagged.

I've made these changes now for `Image`, `ExternalLink`, and `FiducialMarker` (in 125e8c9697b4c1859e0031e38e9582d0313caa9f), so we're now returning `ComponentName` across the board into scripts.

@MAG-AdamThorn, @MAG-ElliotMorris

PR #1017, opened 2026-08-19
[NT-0] chore: Create copilot-instructions.md

MAG-ElliotMorris:
On the theme of experimenting with this, thought it might make sense to get a context file in.

This is more of a suggestion than anything else, if folk want to see how the thing works before we steer it, then that's fine too!

https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions#creating-custom-instructions


MAG-ElliotMorris:
Just closing this, we should see what it's like first.

PR #1019, opened 2026-08-19
[OF-1888] feat: Extend component schemas with value constraints, string map properties, scripting opt-out, and an API to export built-in schemas as JSON

mag-lt:
Builds out the schema description so more of what our built-in components do can be expressed in schema rather than hand-written C++ (and gets us closer to removing them entirely).

- `range` (`min`/`max`) on `int`/`float` (for expressing bounded value extents)
- `options` (named values) on `int`/`float`/`string`, which gives a general way to express enum cases.
- `scriptable` flag on components and properties, to opt out of automatic script binding.

API + implementation changes:
- `csp::GetComponentSchemasJson()` dumps the built-in schemas as pretty printed JSON.
- General improvements to error reporting when parsing schema JSON. Failures now name the offending field and why, rather than logging something indirectly related.
- Removed the IComponentSchemaRegistry abstraction from the public API + schema/property structs, these are now internal details

Built-in components are updated to use the above: enum properties declared with named options, every property named (now that we can properly opt out of automatic script binding), `volume` constrained via schema.

JSON is now the only public way to inject a schema (doing so allows us to better model the schema internally, without being constrained by interop types etc. we can look into exposing a way to programmatically interrogate the schema down the line if it is needed).


mag-lt:
Here is a JSON dump of the built-in schemas, for reference:

```json
[
    {
        "typeId": 3,
        "name": "StaticModel",
        "properties": [
            {
                "key": 1,
                "name": "externalResourceAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "externalResourceAssetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 8,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 9,
                "name": "isShadowCaster",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "materialOverrides",
                "type": "stringToStringMap",
                "defaultValue": {}
            },
            {
                "key": 11,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 12,
                "name": "showAsHoldoutInAR",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 13,
                "name": "showAsHoldoutInVirtual",
                "type": "bool",
                "defaultValue": false
            }
        ]
    },
    {
        "typeId": 4,
        "name": "AnimatedModel",
        "properties": [
            {
                "key": 1,
                "name": "externalResourceAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "externalResourceAssetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "isLoopPlayback",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 7,
                "name": "isPlaying",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 8,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "animationIndex",
                "type": "int",
                "defaultValue": -1
            },
            {
                "key": 11,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 12,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 13,
                "name": "isShadowCaster",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 14,
                "name": "materialOverrides",
                "type": "stringToStringMap",
                "defaultValue": {}
            },
            {
                "key": 15,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 16,
                "name": "showAsHoldoutInAR",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 17,
                "name": "showAsHoldoutInVirtual",
                "type": "bool",
                "defaultValue": false
            }
        ]
    },
    {
        "typeId": 6,
        "name": "VideoPlayer",
        "properties": [
            {
                "key": 0,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "videoAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "videoAssetURL",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 4,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 5,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 7,
                "name": "isStateShared",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 8,
                "name": "isAutoPlay",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 9,
                "name": "isLoopPlayback",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 10,
                "name": "isAutoResize",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 11,
                "name": "attenuationRadius",
                "type": "float",
                "defaultValue": 10.0
            },
            {
                "key": 12,
                "name": "playbackState",
                "type": "int",
                "defaultValue": 0
            },
            {
                "key": 13,
                "name": "currentPlayheadPosition",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 14,
                "name": "timeSincePlay",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 15,
                "name": "videoPlayerSourceType",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "URLSource",
                        "value": 0
                    },
                    {
                        "name": "AssetSource",
                        "value": 1
                    },
                    {
                        "name": "WowzaStreamSource",
                        "value": 2
                    }
                ]
            },
            {
                "key": 16,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 17,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 18,
                "name": "meshComponentId",
                "type": "int",
                "defaultValue": 0,
                "scriptable": false
            },
            {
                "key": 19,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 20,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 21,
                "name": "stereoVideoType",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "None",
                        "value": 0
                    },
                    {
                        "name": "SideBySide",
                        "value": 1
                    },
                    {
                        "name": "TopBottom",
                        "value": 2
                    }
                ]
            },
            {
                "key": 22,
                "name": "isStereoFlipped",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 23,
                "name": "volume",
                "type": "float",
                "defaultValue": 1.0,
                "range": {
                    "min": 0.0,
                    "max": 1.0
                }
            },
            {
                "key": 24,
                "name": "audioType",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "Global",
                        "value": 0
                    },
                    {
                        "name": "Spatial",
                        "value": 1
                    }
                ]
            }
        ]
    },
    {
        "typeId": 8,
        "name": "ExternalLink",
        "properties": [
            {
                "key": 0,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "linkUrl",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 3,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 4,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "displayText",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 6,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 8,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 9,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 9,
        "name": "Avatar",
        "properties": [
            {
                "key": 0,
                "name": "avatarId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 1,
                "name": "userId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "state",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Idle",
                        "value": 0
                    },
                    {
                        "name": "Walking",
                        "value": 1
                    },
                    {
                        "name": "Running",
                        "value": 2
                    },
                    {
                        "name": "Flying",
                        "value": 3
                    },
                    {
                        "name": "Jumping",
                        "value": 4
                    },
                    {
                        "name": "Falling",
                        "value": 5
                    }
                ]
            },
            {
                "key": 3,
                "name": "avatarMeshIndex_DEPRECATED",
                "type": "int",
                "defaultValue": -1,
                "scriptable": false
            },
            {
                "key": 4,
                "name": "agoraUserId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 5,
                "name": "customAvatarUrl_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 6,
                "name": "isHandIKEnabled",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 7,
                "name": "targetHandIKTargetLocation",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 8,
                "name": "handRotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 9,
                "name": "headRotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 10,
                "name": "walkRunBlendPercentage",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 11,
                "name": "torsoTwistAlpha",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 12,
                "name": "avatarPlayMode",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Default",
                        "value": 0
                    },
                    {
                        "name": "AR",
                        "value": 1
                    },
                    {
                        "name": "VR",
                        "value": 2
                    },
                    {
                        "name": "Creator",
                        "value": 3
                    }
                ]
            },
            {
                "key": 13,
                "name": "movementDirection",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 14,
                "name": "locomotionModel",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Grounded",
                        "value": 0
                    },
                    {
                        "name": "FreeCamera",
                        "value": 1
                    }
                ]
            },
            {
                "key": 15,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 16,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 17,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 18,
                "name": "avatarUrl",
                "type": "string",
                "defaultValue": ""
            }
        ]
    },
    {
        "typeId": 10,
        "name": "Light",
        "properties": [
            {
                "key": 1,
                "name": "lightType",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "Directional",
                        "value": 0
                    },
                    {
                        "name": "Point",
                        "value": 1
                    },
                    {
                        "name": "Spot",
                        "value": 2
                    }
                ]
            },
            {
                "key": 2,
                "name": "color",
                "type": "vec3",
                "defaultValue": [
                    255.0,
                    255.0,
                    255.0
                ]
            },
            {
                "key": 3,
                "name": "Intensity",
                "type": "float",
                "defaultValue": 5000.0
            },
            {
                "key": 4,
                "name": "range",
                "type": "float",
                "defaultValue": 1000.0
            },
            {
                "key": 5,
                "name": "innerConeAngle",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 6,
                "name": "outerConeAngle",
                "type": "float",
                "defaultValue": 0.7853981852531433
            },
            {
                "key": 7,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 8,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 9,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "cookieAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 11,
                "name": "cookieAssetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 12,
                "name": "lightCookieType",
                "type": "int",
                "defaultValue": 2,
                "options": [
                    {
                        "name": "ImageCookie",
                        "value": 0
                    },
                    {
                        "name": "VideoCookie",
                        "value": 1
                    },
                    {
                        "name": "NoCookie",
                        "value": 2
                    }
                ]
            },
            {
                "key": 13,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 14,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 15,
                "name": "lightShadowType",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "None",
                        "value": 0
                    },
                    {
                        "name": "Static",
                        "value": 1
                    },
                    {
                        "name": "Realtime",
                        "value": 2
                    }
                ],
                "scriptable": false
            },
            {
                "key": 16,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 11,
        "name": "Button",
        "properties": [
            {
                "key": 1,
                "name": "labelText",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "iconAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 4,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 5,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 7,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 8,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 9,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 12,
        "name": "Image",
        "properties": [
            {
                "key": 0,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "imageAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "billboardMode",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Off",
                        "value": 0
                    },
                    {
                        "name": "Billboard",
                        "value": 1
                    },
                    {
                        "name": "YawLockedBillboard",
                        "value": 2
                    }
                ]
            },
            {
                "key": 8,
                "name": "displayMode",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "SingleSided",
                        "value": 0
                    },
                    {
                        "name": "DoubleSided",
                        "value": 1
                    },
                    {
                        "name": "DoubleSidedReversed",
                        "value": 2
                    }
                ]
            },
            {
                "key": 9,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "isEmissive",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 11,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 13,
        "name": "Script",
        "properties": [
            {
                "key": 1,
                "name": "scriptSource",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "ownerId",
                "type": "int",
                "defaultValue": 0
            },
            {
                "key": 3,
                "name": "scriptScope",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "Local",
                        "value": 0
                    },
                    {
                        "name": "Owner",
                        "value": 1
                    }
                ]
            }
        ],
        "scriptable": false
    },
    {
        "typeId": 14,
        "name": "Custom",
        "properties": [
            {
                "key": 0,
                "name": "applicationOrigin",
                "type": "string",
                "defaultValue": ""
            }
        ]
    },
    {
        "typeId": 15,
        "name": "Conversation",
        "properties": [
            {
                "key": 0,
                "name": "conversationId",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 2,
                "name": "isActive",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "title",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 8,
                "name": "resolved",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 9,
                "name": "conversationCameraPosition",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 10,
                "name": "conversationCameraRotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            }
        ]
    },
    {
        "typeId": 16,
        "name": "Portal",
        "properties": [
            {
                "key": 0,
                "name": "spaceId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 1,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true,
                "scriptable": false
            },
            {
                "key": 2,
                "name": "isActive",
                "type": "bool",
                "defaultValue": true,
                "scriptable": false
            },
            {
                "key": 3,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true,
                "scriptable": false
            },
            {
                "key": 4,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 5,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 6,
                "name": "radius",
                "type": "float",
                "defaultValue": 1.5
            }
        ]
    },
    {
        "typeId": 17,
        "name": "Audio",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "playbackState",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Reset",
                        "value": 0
                    },
                    {
                        "name": "Pause",
                        "value": 1
                    },
                    {
                        "name": "Play",
                        "value": 2
                    }
                ]
            },
            {
                "key": 2,
                "name": "audioType",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Global",
                        "value": 0
                    },
                    {
                        "name": "Spatial",
                        "value": 1
                    }
                ]
            },
            {
                "key": 3,
                "name": "audioAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 4,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 5,
                "name": "attenuationRadius",
                "type": "float",
                "defaultValue": 10.0
            },
            {
                "key": 6,
                "name": "isLoopPlayback",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 7,
                "name": "timeSincePlay",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 8,
                "name": "volume",
                "type": "float",
                "defaultValue": 1.0,
                "range": {
                    "min": 0.0,
                    "max": 1.0
                }
            },
            {
                "key": 9,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 10,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            }
        ]
    },
    {
        "typeId": 18,
        "name": "Spline",
        "properties": [
            {
                "key": 0,
                "name": "waypoints",
                "type": "float",
                "defaultValue": 0.0,
                "scriptable": false
            }
        ]
    },
    {
        "typeId": 19,
        "name": "Collision",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 2,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 3,
                "name": "collisionShape",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Box",
                        "value": 0
                    },
                    {
                        "name": "Mesh",
                        "value": 1
                    },
                    {
                        "name": "Capsule",
                        "value": 2
                    },
                    {
                        "name": "Sphere",
                        "value": 3
                    }
                ]
            },
            {
                "key": 4,
                "name": "collisionMode",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Collision",
                        "value": 0
                    },
                    {
                        "name": "Trigger",
                        "value": 1
                    }
                ]
            },
            {
                "key": 5,
                "name": "collisionAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 6,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 7,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 8,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            }
        ],
        "scriptable": false
    },
    {
        "typeId": 20,
        "name": "Reflection",
        "properties": [
            {
                "key": 0,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "reflectionAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "reflectionShape",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "UnitSphere",
                        "value": 0
                    },
                    {
                        "name": "UnitBox",
                        "value": 1
                    }
                ]
            },
            {
                "key": 7,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            }
        ],
        "scriptable": false
    },
    {
        "typeId": 21,
        "name": "Fog",
        "properties": [
            {
                "key": 0,
                "name": "fogMode",
                "type": "int",
                "defaultValue": 1,
                "options": [
                    {
                        "name": "Linear",
                        "value": 0
                    },
                    {
                        "name": "Exponential",
                        "value": 1
                    },
                    {
                        "name": "Exponential2",
                        "value": 2
                    }
                ]
            },
            {
                "key": 1,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 2,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 3,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 4,
                "name": "startDistance",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 5,
                "name": "endDistance",
                "type": "float",
                "defaultValue": 0.0
            },
            {
                "key": 6,
                "name": "color",
                "type": "vec3",
                "defaultValue": [
                    0.800000011920929,
                    0.8999999761581421,
                    1.0
                ]
            },
            {
                "key": 7,
                "name": "density",
                "type": "float",
                "defaultValue": 0.4000000059604645
            },
            {
                "key": 8,
                "name": "heightFalloff",
                "type": "float",
                "defaultValue": 0.20000000298023224
            },
            {
                "key": 9,
                "name": "maxOpacity",
                "type": "float",
                "defaultValue": 1.0
            },
            {
                "key": 10,
                "name": "isVolumetric",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 11,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 12,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 13,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 14,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 22,
        "name": "ECommerce",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "productId",
                "type": "string",
                "defaultValue": ""
            }
        ]
    },
    {
        "typeId": 23,
        "name": "FiducialMarker",
        "properties": [
            {
                "key": 0,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 1,
                "name": "markerAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "assetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 8,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 24,
        "name": "GaussianSplat",
        "properties": [
            {
                "key": 1,
                "name": "externalResourceAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "externalResourceAssetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 4,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 6,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 8,
                "name": "isShadowCaster_DEPRECATED",
                "type": "bool",
                "defaultValue": true,
                "scriptable": false
            },
            {
                "key": 9,
                "name": "tint",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 10,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 25,
        "name": "Text",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 2,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 3,
                "name": "text",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 4,
                "name": "textColor",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 5,
                "name": "backgroundColor",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 6,
                "name": "isBackgroundVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "width",
                "type": "float",
                "defaultValue": 1.0
            },
            {
                "key": 8,
                "name": "height",
                "type": "float",
                "defaultValue": 1.0
            },
            {
                "key": 9,
                "name": "billboardMode",
                "type": "int",
                "defaultValue": 0,
                "options": [
                    {
                        "name": "Off",
                        "value": 0
                    },
                    {
                        "name": "Billboard",
                        "value": 1
                    },
                    {
                        "name": "YawLockedBillboard",
                        "value": 2
                    }
                ]
            },
            {
                "key": 10,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 11,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 12,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 26,
        "name": "Hotspot",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 2,
                "name": "name_DEPRECATED",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 3,
                "name": "isTeleportPoint",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 4,
                "name": "isSpawnPoint",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 5,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 6,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 7,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 27,
        "name": "CinematicCamera",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 2,
                "name": "isEnabled",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 3,
                "name": "focalLength",
                "type": "float",
                "defaultValue": 0.03500000014901161
            },
            {
                "key": 4,
                "name": "aspectRatio",
                "type": "float",
                "defaultValue": 1.777999997138977
            },
            {
                "key": 5,
                "name": "sensorSize",
                "type": "vec2",
                "defaultValue": [
                    0.035999998450279236,
                    0.024000000208616257
                ]
            },
            {
                "key": 6,
                "name": "nearClip",
                "type": "float",
                "defaultValue": 0.10000000149011612
            },
            {
                "key": 7,
                "name": "farClip",
                "type": "float",
                "defaultValue": 20000.0
            },
            {
                "key": 8,
                "name": "iso",
                "type": "float",
                "defaultValue": 400.0
            },
            {
                "key": 9,
                "name": "shutterSpeed",
                "type": "float",
                "defaultValue": 0.016699999570846558
            },
            {
                "key": 10,
                "name": "aperture",
                "type": "float",
                "defaultValue": 4.0
            },
            {
                "key": 11,
                "name": "isViewerCamera",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 12,
                "name": "thirdPartyComponentRef",
                "type": "string",
                "defaultValue": "",
                "scriptable": false
            },
            {
                "key": 13,
                "name": "focusDistance",
                "type": "float",
                "defaultValue": 5.0
            },
            {
                "key": 14,
                "name": "depthOfFieldEnabled",
                "type": "bool",
                "defaultValue": false
            }
        ]
    },
    {
        "typeId": 28,
        "name": "ScreenSharing",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "rotation",
                "type": "vec4",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0,
                    1.0
                ]
            },
            {
                "key": 2,
                "name": "scale",
                "type": "vec3",
                "defaultValue": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            {
                "key": 3,
                "name": "isVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 4,
                "name": "isARVisible",
                "type": "bool",
                "defaultValue": true
            },
            {
                "key": 5,
                "name": "isShadowCaster",
                "type": "bool",
                "defaultValue": false
            },
            {
                "key": 6,
                "name": "userId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 7,
                "name": "defaultImageCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 8,
                "name": "defaultImageAssetId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 9,
                "name": "attenuationRadius",
                "type": "float",
                "defaultValue": 10.0
            },
            {
                "key": 10,
                "name": "isVirtualVisible",
                "type": "bool",
                "defaultValue": true
            }
        ]
    },
    {
        "typeId": 29,
        "name": "AIChatbot",
        "properties": [
            {
                "key": 0,
                "name": "position",
                "type": "vec3",
                "defaultValue": [
                    0.0,
                    0.0,
                    0.0
                ]
            },
            {
                "key": 1,
                "name": "voice",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 2,
                "name": "guardrailAssetCollectionId",
                "type": "string",
                "defaultValue": ""
            },
            {
                "key": 3,
                "name": "visualState",
                "type": "int",
                "defaultValue": 0
            }
        ]
    }
]
```

Library/include/CSP/Multiplayer/IComponentSchemaRegistry.h
```diff
@@ -1,47 +0,0 @@
-/*
```
MAG-ElliotMorris, line 1: [Comment] Glad to see this go in this form to be honest. I understand the intent of the interface re: implementation hiding, but it never really feels right to me having an interface for something that isn't going to have multiple concretions.

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
```diff
@@ -75,7 +77,7 @@ class CSP_API OfflineRealtimeEngine : public csp::common::IRealtimeEngine
     /// @param LogSystem Logger for status and debug output.
     /// @param RemoteScriptRunner Object capable of running a script.
     /// @param AdditionalComponents Component schemas to register alongside the built-in schemas in the engine-wide registry.
-    OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
+    CSP_NO_EXPORT OfflineRealtimeEngine(csp::common::LogSystem& LogSystem, csp::common::IJSScriptRunner& RemoteScriptRunner,
         const csp::common::Array<ComponentSchema>& AdditionalComponents);
```
MAG-ElliotMorris: [Consider] Whether these constructors are even needed at all anymore.

Library/src/Multiplayer/ComponentSchemaRegistry.h
```diff
@@ -34,13 +35,23 @@ std::optional<ComponentType> ToComponentType(uint64_t TypeId);
 
 bool IsLegacyComponentTypeId(uint64_t TypeId);
 
-class ComponentSchemaRegistryImpl final : public IComponentSchemaRegistry
+/// @brief The engine-wide index of registered component schemas.
+class ComponentSchemaRegistryImpl final
```
MAG-ElliotMorris: [Consider] If this need to be `Impl` anymore. It would be pleasant if it could just be `ComponentSchemaRegistry`

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -26,9 +24,8 @@ namespace csp::multiplayer
 /// @brief Represents an individual data field, or "property", within a component schema,
 /// consisting of a stable ID/key, a type kind and default value (via ReplicatedValue),
 /// and other metadata.
-class CSP_API ComponentProperty
+struct ComponentProperty final
```
MAG-ElliotMorris: [Comment] Ah. I see you're adding `final` intentionally. 

[nit] That it breaks with style (I've never really felt it achieved much tbh, should be more like a marker that Liskovs substitution has been violated), but i'm not precious.

Library/src/Multiplayer/OfflineRealtimeEngine.cpp
```diff
@@ -97,7 +97,7 @@ OfflineRealtimeEngine::OfflineRealtimeEngine(const CSPSceneDescription& SceneDes
 {
     std::scoped_lock EntitiesLocker(EntitiesLock);
 
-    auto DeserializedEntities = SceneDescription.CreateEntities(*this, LogSystem, RemoteScriptRunner);
+    auto DeserializedEntities = SceneDescription.CreateEntities(*this, *ComponentRegistry, LogSystem, RemoteScriptRunner);
```
MAG-ElliotMorris, line 102: [Question] This will always be a dereference on a non-null, but empty component registry? I'm presuming that's how it gets initialized via the delegating constructor, since it's not visible injected here?

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -642,6 +642,46 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, CreateAvatarSen
         LocomotionModel::Grounded, MockCallback.AsStdFunction());
 }
 
+static auto MakeSingleEntityCreationInvoker(uint64_t ObjectId = 42)
```
MAG-ElliotMorris, line 649: [Nit] A single line comment explaining why you might want to use this would help, for the person who might discover they have the same problem in the future.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -183,195 +205,211 @@ namespace
         };
     }
 
-    std::optional<PropertyValue> TryParse(const std::string& Type, const rapidjson::Value& Value)
+    struct ParseError final
+    {
+        std::string Description;
+    };
+
+    template <typename T> class ParseResult final
```
MAG-ElliotMorris: [Nit] a little context comment, even just "This would be unnecessary if we had std::expected" will go a long way here.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -183,195 +205,211 @@ namespace
...
+    public:
+        explicit ParseResult(T InValue)
+            : Storage(std::move(InValue))
+        {
+        }
+
+        ParseResult(ParseError InError)
+            : Storage(std::move(InError))
+        {
+        }
```
MAG-ElliotMorris: [Consider] This `explicit` asymmetry looks intentional, you should document why so I don't have to engage my brain.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -183,195 +205,211 @@ namespace
...
+
+    private:
+        std::variant<T, ParseError> Storage;
+    };
+
+    template <typename T> ParseResult<T> TryParseRequired(const rapidjson::Value& Object, const char* Key)
+    {
+        if (!Object.HasMember(Key))
+        {
+            return ParseError { fmt::format("required field '{}' is missing", Key) };
```
MAG-ElliotMorris: [Comment] Love this, good errors.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -183,195 +205,211 @@ namespace
...
     {
         if (!Value.IsObject())
         {
-            if (LogSystem)
-            {
-                LogSystem->LogMsg(csp::common::LogLevel::Warning, "TryParseProperty: not a JSON object");
-            }
-
-            return std::nullopt;
+            return ParseError { "property must be a JSON object" };
```
MAG-ElliotMorris, line 516: [Question] I can understand there not being on an object that failed to parse, but do you think there is any way to identify the malformed json to give the author an idea where to start looking?

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -22,6 +22,7 @@
 #include <string>
```
MAG-ElliotMorris, line 23: (This is just the first line of the commit in question)

[Comment] Good job structuring this PR to introduce `PlainValue` and then immediately extend in the next commit with `BoundedValue` and `EnumeratedValue`.

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -40,13 +41,82 @@ template <typename T> bool PlainValue<T>::operator==(const PlainValue& Other) co
 
 template <typename T> bool PlainValue<T>::operator!=(const PlainValue& Other) const { return !(*this == Other); }
 
+/// @brief An inclusive range of values, constraining what a bounded property may hold.
+template <typename T> struct Range final
```
MAG-ElliotMorris: [Consider] I've said it before in this review, but seeing `final` all over this change keeps making me think that this API is somehow unique, or used differently to the rest of the codebase, because otherwise why isn't everything `final`?

Now, maybe everything _should_ be final, but that doesn't stop the creeping thought which might add confusion. I'd advocate consistency with what's already here in this case.

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -40,13 +41,82 @@ template <typename T> bool PlainValue<T>::operator==(const PlainValue& Other) co
...
+template <typename T> bool BoundedValue<T>::operator==(const BoundedValue& Other) const { return Default == Other.Default && Range == Other.Range; }
+
+template <typename T> bool BoundedValue<T>::operator!=(const BoundedValue& Other) const { return !(*this == Other); }
+
+/// @brief A named value within the set an enumerated property is constrained to.
+template <typename T> struct SchemaOption final
+{
+    /// @brief A human-readable name for this option, for clients presenting a choice of values.
+    /// Carries no meaning at runtime; the value is what is replicated.
+    csp::common::String Name;
```
MAG-ElliotMorris, line 81: [Consider] Marking comments with doxygen tags implies that this is public API, which could confuse.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -80,6 +81,21 @@ namespace
     // TODO: implement along with map parsing.
     void SerializeDefaultValue(csp::json::JsonSerializer&, const std::unordered_map<std::string, std::string>&) { }
 
+    template <typename T> void SerializeConstraint(csp::json::JsonSerializer&, const PlainValue<T>&)
+    {
+        // Unconstrained, so there is nothing to write.
+    }
```
MAG-ElliotMorris, line 95: [Comment] I don' t really mind as such, but it strikes me as strange to have `SerializeDefaultValue` also serialize the member, but `SerializeConstraint` to be a specific extension based on the Value type. Perhaps this is what you were getting at in one of your commit/PR comments.

I wouldn't go to any real effort to change this now, it's not offensive or anything.

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -40,13 +41,82 @@ template <typename T> bool PlainValue<T>::operator==(const PlainValue& Other) co
...
+template <typename T> bool Range<T>::operator!=(const Range& Other) const { return !(*this == Other); }
+
+/// @brief A property value constrained to an inclusive range.
+template <typename T> struct BoundedValue final
+{
+    /// @brief The default value of this property, used to initialise it on a newly created component.
+    T Default;
+
+    /// @brief The range this property's value is constrained to.
+    csp::multiplayer::Range<T> Range;
```
MAG-ElliotMorris: [Consider] Seeing all these new types. `PlainValue`, `EnumeratedValue`, `ConstrainedValue`, `Range`, etc. They're quite general sounding names, I wonder if they should get their own `schema` namespace, to prevent them being confused with anything runtime.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -254,53 +270,215 @@ namespace
...
+        return ParseResult<PropertyValue> { EnumeratedValue<T> {
+            std::move(*Default),
+            std::move(*Options),
+        } };
+    }
+
+    template <typename T> inline constexpr bool SupportsRangeV = std::is_same_v<T, float> || std::is_same_v<T, int64_t>;
+
+    template <typename T>
+    inline constexpr bool SupportsOptionsV = std::is_same_v<T, float> || std::is_same_v<T, int64_t> || std::is_same_v<T, std::string>;
```
MAG-ElliotMorris, line 431: [nit] Little examples of what the schema json actually looks like sprinkled throughout this file, next to the parsing code, would help de-mystify a lot of this stuff.

You can deduce the meaning of these symbols and methods yourself, it just takes a minute.

Library/src/Multiplayer/ComponentBase.cpp
MAG-ElliotMorris: [fix] We should not silently reject out-of-range values. Component base has the logsystem, check if it is null (I'm not sure under what conditions it would be), then log something about that. I would say at `warning` level.

... Ideally we'd surface an error like that via `std::expected` but aye, do what we can.

Library/src/Multiplayer/ComponentBase.cpp
```diff
@@ -119,7 +119,7 @@ void ComponentBase::SetProperty(uint16_t Key, const csp::common::ReplicatedValue
         return;
     }
 
-    if (const auto* Property = FindSchemaProperty(*CachedSchema, Key); Property && Value.GetReplicatedValueType() == TypeOf(Property->Value))
+    if (const auto* Property = FindSchemaProperty(*CachedSchema, Key); Property && IsValidValue(*Property, Value))
```
MAG-ElliotMorris: [Comment] This is a behaviour "change" we should be sure to communicate clearly to clients. (It's not really changed, it's just potentially wider now depending on how they do their schema).

To be clear I think it's great, bounded values should be bounded, for all time.

Library/src/Multiplayer/ComponentSchema.cpp
```diff
@@ -78,8 +79,11 @@ namespace
         Serializer.SerializeMember("defaultValue", std::vector<float> { Vector.X, Vector.Y, Vector.Z, Vector.W });
     }
 
-    // TODO: implement along with map parsing.
-    void SerializeDefaultValue(csp::json::JsonSerializer&, const std::unordered_map<std::string, std::string>&) { }
+    void SerializeDefaultValue(csp::json::JsonSerializer& Serializer, const std::unordered_map<std::string, std::string>& Value)
+    {
+        Serializer.SerializeMember("type", "stringToStringMap");
```
MAG-ElliotMorris, line 88: [comment] Been doing some OCaml recently, and there it would be `stringMap of string` ... so you know, could be worse.

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
```diff
@@ -40,6 +40,85 @@
...
+namespace qjs
+{
+
+template <typename T> struct js_traits<std::unordered_map<std::string, T>>
+{
+    static JSValue wrap(JSContext* Context, const std::unordered_map<std::string, T>& Map) noexcept
+    {
+        try
+        {
+            auto Object = qjs::Value { Context, JS_NewObject(Context) };
```
MAG-ElliotMorris, line 50: [Consider]

I'm not sure you should bother going to the effort to make this change I'm about to suggest, even if it would work... scripting is such broken windows the fact you've done anything here at all is a marvel.

In the bindings layer, we build maps via getting the constructor out of the [global `this`.](https://github.com/magnopus-opensource/connected-spaces-platform-web/blob/de1a7f09a0b6c7db1c547b6f1c46b6d68e946fb7/src/bindings/containers/Map.h#L77)

This is probably doable here, although the quickJS api is less ergonomic, probably something like

```
qjs::Value globalThis { Context, JS_GetGlobalObject(Context) };
qjs::Value mapCtor = globalThis ["Map"];
```
Then call the constructor somehow to get the object.

You've mentioned doing this requires some specialization, so I'm sure i'm missing some quickJS'ey stuff.

Library/src/Multiplayer/ComponentProperty.h
```diff
@@ -141,6 +142,10 @@ struct ComponentProperty final
     /// a `std::string`.
     PropertyValue Value;
 
+    /// @brief Whether this property is exposed to scripting. Absent when the schema does not say,
+    /// which means exposed.
+    std::optional<bool> IsScriptable;
```
MAG-ElliotMorris, line 147: [Consider] Is `IsScriptable` in the positive the right orientation for this, considering the default? Perhaps `HideFromScripting` is better, then `nullopt` can mean `false`, which is more natural.

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
 All notable changes to this project will be documented in this file. For compiled binaries, deployment packages, and version-specific artifacts, please visit our [GitHub Releases](https://github.com/magnopus-opensource/connected-spaces-platform/releases).
 
+## [6.48.0]
+
+### 🍰 🙌 New Features
+
+- [OF-1888] feat!: Extend component schemas with value constraints, string map properties, scripting opt-out, and an API to export built-in schemas as JSON. By @mag-lt.
+
+  Added to component schema:
+  - `range` (`min`/`max`) on `int` and `float` properties, and `options` (distinct set of named values) on `int`, `float` and `string` properties, which provides a way of expressing enum cases in a more general way.
```
MAG-ElliotMorris: [consider] Note how properties set outside these constraints will be rejected.

MAG-ElliotMorris (changes requested):
Excellent work, you should be proud of this. Count this among the last cohort of significant, comprehensible changesets that will ever grace the internet.

One [fix], around the check to ignore constraints when setting properties. tl;dr I think we should log. You probably thought about that, so we can discuss if there's reasons not to do it.

Library/src/Multiplayer/ComponentSchema.cpp
MAG-mv: Passing through some kind of  parsing context would be nice. This way we could pinpoint things like the component that failed and the property name

Library/src/Multiplayer/ComponentProperty.h
MAG-mv: Kind of agree. From my perspective, if something doesn't have a virtual destructor, you probably shouldn't override it.

I don't really care either way

CHANGELOG.md
mag-lt: Addressed in f022bee24b9b778af1af3e5d010957a31b49fdc7

Library/src/Multiplayer/ComponentBase.cpp
mag-lt: Addressed in 1296c1aca2c082fef6ded067c7ebe5e3c6c9c39f (in support of this, I've lifted out the `Result` type from the JSON parsing code in b8bcfc3c87558357f75d63ca2ee436f584c48e58)

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: It was intentional, but I've opted to remove it in b8bcfc3c87558357f75d63ca2ee436f584c48e58

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: Indirectly addressed by b8bcfc3c87558357f75d63ca2ee436f584c48e58

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
mag-lt: Addressed in 786c20f0bc5e14d13bd37f615c0b34c023bd6fce

Library/src/Multiplayer/ComponentSchemaRegistry.h
mag-lt: Addressed in 426468707f994bd4efdcc77a67eee5ec2fb17a59

Library/src/Multiplayer/ComponentSchemaHelpers.h
```diff
@@ -169,29 +173,106 @@ inline csp::common::ReplicatedValue ToReplicatedValue(const PropertyValue& Value
...
 {
     const auto ActualValue = ReplicatedTypeMap<T>::From(Value);
+
+    if (!ActualValue)
+    {
+        return TypeMismatch<T>(Value);
+    }
+
     const auto Matches = [&ActualValue](const auto& Option) { return Option.Value == *ActualValue; };
+    const auto IsValid = std::all_of(SchemaValue.Options.begin(), SchemaValue.Options.end(), Matches);
```
MAG-mv: [Fix] I dont think this is correct? I believe all_of only returns true if it matches every element, when you arew only checking for a single match

MAG-mv (changes requested):

Library/src/Multiplayer/ComponentSchemaHelpers.h
mag-lt: Thanks. I had this as `any_of` in 7937b9acd6737b6f5dee8da17d23f469df599c85, and was playing around with inverting it to `none_of` in the lead up to 1296c1aca2c082fef6ded067c7ebe5e3c6c9c39f, and seems I didn't have a clear enough head late friday when abandoning that. This highlights the test coverage isn't good enough, so will look in to that before changing this impl.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: There is a somewhat useful error composed here via the chained result types across the misc functions in play. So, with `csp::multiplayer::ComponentSchemasFromJson` as the entry point (which is what is used for parsing the raw strings passed on engine construction), for a component like this:

```json
{
    "typeId": 666,
    "name": "TestComponent",
    "properties": [
        {
            "key": 0,
            "name": "stringProperty",
            "type": "string",
            "defaultValue": "hello"
        },
        "not an object",
        {
            "key": 1,
            "name": "floatProperty",
            "type": "float",
            "defaultValue": 1.5
        }
    ]
}
```

The message printed in the log is: `ComponentSchemasFromJson: skipping entry: TestComponent: properties[1]: property must be a JSON object`

Context is built up across the returned results vs injecting some context object to build it up on.

Library/src/Multiplayer/OfflineRealtimeEngine.cpp
mag-lt: All constructors delegate to the `OfflineRealtimeEngine(csp::common::LogSystem&, csp::common::IJSScriptRunner&, const csp::common::Array<ComponentSchema>&)` overload which [initialises + populates the registry](https://github.com/magnopus-opensource/connected-spaces-platform/blob/f022bee24b9b778af1af3e5d010957a31b49fdc7/Library/src/Multiplayer/OfflineRealtimeEngine.cpp#L131). 

This particular constructor passes the raw JsonSchemas to a delegating ctor, which will parse them out as `ComponentSchema` structs, which then go through to the constructor mentioned above. By the time this line runs, the registry has been initialised and populated.

So, this will always be non-null (it was originally heap-allocated / stored as a pointer because of the runtime polymorphism when this was an interface, but continues to be now because it is forward declared).

Library/src/Multiplayer/ComponentBase.cpp
mag-lt: This is called out in the changelog f022bee24b9b778af1af3e5d010957a31b49fdc7. Unconstrained values are handled as before (their types are checked). Traditional hardcoded components don't go via this path, so this only impacts clients when they adopt the schema related approach.

Library/src/Multiplayer/ComponentProperty.h
mag-lt: > Now, maybe everything should be final, but that doesn't stop the creeping thought which might add confusion. I'd advocate consistency with what's already here in this case.

Fair, I'll go with consistency for now (will push a commit shortly)

This is ultimately personal preference, and whilst I do prefer it, I'm not sure I have a strong enough argument, other than I have an aversion to inheritance. You end up with stuff like [this](https://github.com/magnopus-opensource/connected-spaces-platform/blob/0a0c8c13639b44957365e114fa41bdff64d5f6bf/Library/include/CSP/Multiplayer/MultiplayerHubMethods.h#L56), which I'd argue is not at all expected, but in any case I can see that the general consensus is to use `final` sparingly.

Library/src/Multiplayer/ComponentProperty.h
mag-lt: I can see that, however there are quite a few private headers doing this. Will leave as-is for now.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: Yeah I couldn't really find a nice way to cut this, and haven't managed to settle/find anything I strongly prefer over this. Will leave as-is.

Library/src/Multiplayer/ComponentSchema.cpp
mag-lt: I generally think the tests serve as the example, but I can see that doesn't help local readability / reasoning, but they do at least have to stay up to date if the code changes. I'm inclined to leave this as-is for now.

Library/src/Multiplayer/Script/EntityScriptBinding.cpp
mag-lt: The specialisation stuff I mentioned is mainly just about the `js_traits` `wrap`/`unwrap` stuff, and you're right that this is possible, in a similar way to how you describe, but there'd be some more details for calling the `Map` methods, iterating over the `entries` etc, none of which is built-in. All do-able, but gotta work it all through. However, I've not found tried to work this through yet, and at this point I think this will remain as-is.

Library/src/Multiplayer/ComponentProperty.h
mag-lt: I personally think that `scriptable` is the right framing (or any other variation on the naming, I still think about it in the positive i.e. `exposeToScripting`), just because generally I think that flags in that direction are easier to understand. Here though, `true` is the default, and this (`std::optional`) avoids having to write that out (as in to JSON) in what appears to be the overwhelming majority of cases (which I appreciate is probably the reasoning for suggesting to flip the framing/naming). I could've expressed this differently, having this just be a boolean, and choosing to not serialise the field if it is set to true, but preferred to separate the two.

Library/src/Multiplayer/ComponentProperty.h
mag-lt: There is another thread about this [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1019#discussion_r3846156583), resolving this and keeping discussion to the other comment.

Library/include/CSP/Multiplayer/OfflineRealtimeEngine.h
mag-lt: The unfortunate truth (for me) is that this remains almost exclusively for testing at this point, which I generally don't like as a justification, but will reluctantly do so here anyway. I'm going to leave this in place for now, but would generally like to find a better way to separate our public interfaces from our implementation so that we can have richer ways of constructing things internally but not expose them via the public api

Library/src/Multiplayer/ComponentSchemaHelpers.h
mag-lt: Fixed in 802ba9c3f711301f1319e54173bdf915d12fffe9

Library/src/Multiplayer/ComponentProperty.h
mag-lt: Removed in cad765c7a2007f5956f9285169dff585e8384466

Library/src/Multiplayer/ComponentProperty.h
mag-lt: I've moved all the schema related types inside to `csp::multiplayer::schema` in b095199964df05cfdfc35332ca9e15e61e0ace0c

MAG-mv (approved):

MAG-ElliotMorris (approved):

MAG-ElliotMorris:
(Apologies if this is duped, thought i left this somewhere else but can't find it.)

[Consider] Some of the formatting, like converting replicated value types to strings, might have been better expressed as custom `fmt` formatter implementations, as we've done elsewhere for a few things. Not something you should look at now, but keep it in mind for next time.

PR #1020, opened 2026-08-20
[OF-1845] Add custom Conan deployer

MAG-mv:
This PR adds a custom Conan deployer to copy all third-party dependencies in a specified directory. This allows us to create a  static CSP package that can be linked without consumers needing Conan to resolve dependencies.

This also updates our CI to usethe new deployers for releases.

### DIFF
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -68,6 +68,8 @@ runs:
         INPUTS_HOST_PROFILE: ${{ inputs.host_profile }}
         INPUTS_OUTPUT_FOLDER: ${{ inputs.output_folder }}
         INPUTS_BUILD_TYPE: ${{ inputs.build_type }}
+        INPUTS_THIRD_PARTY_LIBS: ${{ inputs.package_third_party_libs }}
+        INPUTS_INSTALL_PREFIX: ${{ inputs.install_prefix }}
       shell: bash
       run: |
         set -euxo pipefail
@@ -87,10 +89,10 @@ runs:
           --profile:host="$PROFILE_PATH"
         )
 
-        if [ "${{ inputs.package_third_party_libs }}" = "true" ]; then
+        if [ "${INPUTS_THIRD_PARTY_LIBS}" = "true" ]; then
           CONAN_ARGS+=(
             --deployer=./conan/csp_deployer.py
-            --deployer-folder="${{ inputs.install_prefix }}/lib/third_party"
+            --deployer-folder="${INPUTS_INSTALL_PREFIX}/lib/third_party"
           )
         fi
 
--- a/.github/actions/build-and-install/action.yml
+++ b/.github/actions/build-and-install/action.yml
@@ -32,6 +32,11 @@ inputs:
     required: false
     default: "install"
 
+  package_third_party_libs:
+    description: "Whether to copy all of csp's third party dependencies into the package."
+    required: false
+    default: "false"
+
 runs:
   using: "composite"
 
@@ -71,12 +76,22 @@ runs:
           exit 1
         fi
 
-        conan install . \
-          -of "${{ inputs.output_folder }}" \
-          -s build_type="${{ inputs.build_type }}" \
-          --build=missing \
+        CONAN_ARGS=(
+          .
+          -of "${{ inputs.output_folder }}"
+          -s build_type="${{ inputs.build_type }}"
+          --build=missing
           --profile:host="$PROFILE_PATH"
+        )
 
+        if [ "${{ inputs.package_third_party_libs }}" = "true" ]; then
+          CONAN_ARGS+=(
+            --deployer=./conan/csp_deployer.py
+            --deployer-folder="${{ inputs.install_prefix }}/lib/third_party"
+          )
+        fi
+
+        conan install "${CONAN_ARGS[@]}"
     - name: Configure CMake
       shell: bash
       run: |
--- a/conan/csp_deployer.py
+++ b/conan/csp_deployer.py
@@ -1,0 +1,123 @@
+"""
+CSP SDK dependency deployer.
+
+This deploys the third-party dependencies required for csp to be linked
+without requiring Conan in the consumer project.
+
+This can be invoked using the following args at the conan install step:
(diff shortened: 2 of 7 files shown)


conan/csp_deployer.py
```diff
@@ -0,0 +1,123 @@
...
+            for source in library_directory.rglob("*"):
+                if not source.is_file():
+                    continue
+
+                # Ensure file is a .lib or .a
+                if source.suffix.lower() not in STATIC_LIBRARY_SUFFIXES:
+                    continue
+
+                # Copy into provided deployer-folder
+                # Use # _copy_file_unique to throw if we find libs with the same name
```
mag-lt: [nit] the function actually does more than check the name (it compares the contents). I'd drop this comment, or document the function itself more explicitly

conan/csp_deployer.py
```diff
@@ -0,0 +1,123 @@
...
+)
+
+# Ensures a file doesn't exist at the destination before copying
+def _copy_file_unique(source: Path, destination: Path) -> Path:
+    destination.mkdir(parents=True, exist_ok=True)
+
+    output = destination / source.name
+
+    if output.exists():
+        if output.read_bytes() != source.read_bytes():
```
mag-lt: [nit] this will read the whole file into memory. potentially just doing a stat would be enough here (name, size, timestamp) as a proxy for sameness, and i'm sure python has something built-in that can either do that or compare files in a chunked fashion under the hood avoiding reading in the whole thing, but it is probably negligible (only checked in the case of collision, and we've probably not got many? and the files are probably quite small)

.github/actions/build-and-install/action.yml
```diff
@@ -74,12 +79,22 @@ runs:
...
-          --build=missing \
+        CONAN_ARGS=(
+          .
+          -of "${INPUTS_OUTPUT_FOLDER}"
+          -s build_type="${INPUTS_BUILD_TYPE}"
+          --build=missing
           --profile:host="$PROFILE_PATH"
+        )
+
+        if [ "${{ inputs.package_third_party_libs }}" = "true" ]; then
```
mag-lt: [fix] following the gh actions static analysis [stuff](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1018) we did yesterday, this is getting [flagged](https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/32468417723/job/96729895876?pr=1020#step:3:137). Will need to be lifted out as an env var (has different expansion / quoting behaviour), which I believe `zizmor` can auto apply a fix for.

.github/actions/build-and-install/action.yml
```diff
@@ -74,12 +79,22 @@ runs:
...
+          CONAN_ARGS+=(
+            --deployer=./conan/csp_deployer.py
+            --deployer-folder="${{ inputs.install_prefix }}/lib/third_party"
```
mag-lt: Same as other comment [here](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/changes#r3830255958)

conan/csp_deployer.py
```diff
@@ -0,0 +1,123 @@
+"""
+CSP SDK dependency deployer.
+
+This deploys the third-party dependencies required for csp to be linked
+without requiring Conan in the consumer project.
+
+This can be invoked using the following args at the conan install step:
+--deployer=csp_sdk_deployer --deployer-folder=<output-directory>
```
mag-lt: [question] `csp_sdk_deployer` is this name somehow associated with this script? or should this be `/path/to/csp_deployer.py`?

conan/csp_deployer.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if not package_folder_value:
+            continue
+
+        package_folder = Path(package_folder_value)
+
+        # dependency.cpp_info.libdirs contains the library directories
+        # Search recursively for static libs
+        for library_directory_value in dependency.cpp_info.libdirs:
+            library_directory = Path(library_directory_value)
```
mag-lt, line 90: [question] are libdirs relative? if so, this probably restricts running this script to a particular directory?

mag-lt (approved):
A few nits, and flagging that CI is failing due to the gh static analysis flagging some things (which will need to be addressed to turn the checks green, so approving). Otherwise, nothing blocking.

conan/csp_deployer.py
MAG-mv: Fixed in: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/commits/319b6c50366a79dfba95baa3a0eeb7a19ce331fd

conan/csp_deployer.py
MAG-mv: They're absolute paths

conan/csp_deployer.py
MAG-mv: That should be the actual path to the script. Fixed here: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/commits/1a77c2b6cf4e3d4e0a15acd403615d2d45dea70f

conan/csp_deployer.py
MAG-mv: I updated to use a built in function. I have set shallow=false which still compares the file data, as I'm paranoid about relying just on metadata. The scripts runs very fats either way. https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/commits/d1fd3183fb1f6b81d55cdac2795d7de5d8d3acba

.github/actions/build-and-install/action.yml
MAG-mv: Fixed here https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/commits/10c08815c509fbb0a7390a4c4238219cd4bfd2c9

.github/actions/build-and-install/action.yml
MAG-mv: Fixed here https://github.com/magnopus-opensource/connected-spaces-platform/pull/1020/commits/10c08815c509fbb0a7390a4c4238219cd4bfd2c9

MAG-AdamThorn (approved):

PR #1021, opened 2026-08-24
[OB-5315] fix: Deleting a script on remote client continues to execute message subscriptions

MAG-ElliotMorris:
There is a divergence in how we update script bindings, depending on if the source of change came from a local or remote client. 

A local client calls [OnSourceChanged](https://github.com/magnopus-opensource/connected-spaces-platform/blob/80a42efece1a456ea5fbe35efa4a12062c978717/Library/src/Multiplayer/Script/EntityScript.cpp#L195), (via [SetScriptSource](https://github.com/magnopus-opensource/connected-spaces-platform/blob/80a42efece1a456ea5fbe35efa4a12062c978717/Library/src/Multiplayer/Components/ScriptSpaceComponent.cpp#L82)) whilst a remote update merely calls [Bind and Invoke](https://github.com/magnopus-opensource/connected-spaces-platform/blob/80a42efece1a456ea5fbe35efa4a12062c978717/Library/src/Multiplayer/Components/ScriptSpaceComponent.cpp#L107).

The cause of the bug, where remote script deletions do not stop running scripts, is likely obvious in that context. The clearing of the message buffers is never performed in response to remote updates. Change the implementation such that remote updates call the same binding path as local updates.

This seems like rather a significant fix ... surprised this hasn't come up before. I am a tad worried about Hyrums law style manifestations downstream of this, especially relating to what the resetting of the context might imply. Don't really think we can do much to guard against that right now, script usage tends to be confined within solutions and we don't have a lot of visibility there.

### Testing

One regression test is included in which the multiplayer test runner is used to delete the script source, replicating the reported bug. Prior to the change, this errored on an expectation that no log from the script (logging on a function registered to the tick message) would be received, but after the change, passes as no log is output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ All notable changes to this project will be documented in this file. For compile
   Additionally, `Image`, `ExternalLink`, and `FiducialMarker` have also been updated to return the `ComponentName` vs the deprecated name in scripts,
   as these were never updated to do so. This is technically a breaking change (without doing a data migration on the server), but should be low impact
   in practice.
+  
+- [OB-5315] fix: Route remote script updates through the same binding path as local script updates go through. (OnSourceChanged). By @MAG-ElliotMorris
+  Prior to this, we were not clearing the `MessageMap` or `PropertyMap` containers as we should, which meant any `subscribeToMessage`
+  events would keep executing if the script was modified on a remote client.
+  This also causes us to call `ResetContext` when the script is modified, which also seems like the right thing to be doing.
 
 ###  🔨 🔨 Chore
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,13 @@ All notable changes to this project will be documented in this file. For compile
   - The realtime engine constructors taking `Array<ComponentSchema>` are now `CSP_NO_EXPORT`. Use the `List<String>` JSON overloads instead.
   - No clients had adopted the above, so this shouldn't have any practical impact
 
+### 🐛 🔨 Bug Fixes
+  
+- [OB-5315] fix: Route remote script updates through the same binding path as local script updates go through. (OnSourceChanged). By @MAG-ElliotMorris
+  Prior to this, we were not clearing the `MessageMap` or `PropertyMap` containers as we should, which meant any `subscribeToMessage`
+  events would keep executing if the script was modified on a remote client.
+  This also causes us to call `ResetContext` when the script is modified, which also seems like the right thing to be doing.
+
 ## [6.47.0]
 
 ### 🐛 🔨 Bug Fixes
@@ -33,11 +40,6 @@ All notable changes to this project will be documented in this file. For compile
   Additionally, `Image`, `ExternalLink`, and `FiducialMarker` have also been updated to return the `ComponentName` vs the deprecated name in scripts,
   as these were never updated to do so. This is technically a breaking change (without doing a data migration on the server), but should be low impact
   in practice.
-  
-- [OB-5315] fix: Route remote script updates through the same binding path as local script updates go through. (OnSourceChanged). By @MAG-ElliotMorris
-  Prior to this, we were not clearing the `MessageMap` or `PropertyMap` containers as we should, which meant any `subscribeToMessage`
-  events would keep executing if the script was modified on a remote client.
-  This also causes us to call `ResetContext` when the script is modified, which also seems like the right thing to be doing.
 
 ###  🔨 🔨 Chore
 
--- a/Tests/src/PublicAPITests/ScriptSystemTests.cpp
+++ b/Tests/src/PublicAPITests/ScriptSystemTests.cpp
@@ -26,6 +26,7 @@
 #include "Debug/Logging.h"
 #include "Multiplayer/NetworkEventManagerImpl.h"
 #include "Multiplayer/Script/EntityScriptBinding.h"
+#include "MultiplayerTestRunnerProcess.h"
 #include "RAIIMockLogger.h"
 #include "TestHelpers.h"
 #include "quickjspp.hpp"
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ All notable changes to this project will be documented in this file. For compile
   Additionally, `Image`, `ExternalLink`, and `FiducialMarker` have also been updated to return the `ComponentName` vs the deprecated name in scripts,
   as these were never updated to do so. This is technically a breaking change (without doing a data migration on the server), but should be low impact
   in practice.
+  
+- [OB-5315] fix: Route remote script updates through the same binding path as local script updates go through. (OnSourceChanged). By @MAG-ElliotMorris
```
mag-lt: [fix] (though minor) this is in the section for 6.47.0, which has already been [released](https://github.com/magnopus-opensource/connected-spaces-platform/releases/tag/v6.47.0), needs a new `[6.48.0]` section

Tests/src/PublicAPITests/ScriptSystemTests.cpp
```diff
@@ -1352,6 +1353,112 @@ TEST_P(ModifyExistingScript, ModifyExistingScriptTest)
...
+
+    // Don't proceed with the test until we've fully entered the space, we want a clean slate for script ticking.
+    std::promise<void> EntityFetchCompletePromise;
+    std::future<void> EntityFetchCompleteFuture = EntityFetchCompletePromise.get_future();
+    RealtimeEngine->SetEntityFetchCompleteCallback([&EntityFetchCompletePromise](uint32_t) { EntityFetchCompletePromise.set_value(); });
+
+    auto [EnterResult] = AWAIT_PRE(SpaceSystem, EnterSpace, RequestPredicate, Space.Id, RealtimeEngine.get());
+    ASSERT_EQ(EnterResult.GetResultCode(), csp::systems::EResultCode::Success);
+
+    EntityFetchCompleteFuture.wait_for(3s);
```
mag-lt: [fix] This should be wrapped in an ASSERT_EQ checking for `std::future_status::ready`, or using the existing `WaitForFuture` test helper (and wrapping in ASSERT_TRUE), otherwise we proceed regardless of having met the precondition

Tests/src/PublicAPITests/ScriptSystemTests.cpp
```diff
@@ -1352,6 +1353,112 @@ TEST_P(ModifyExistingScript, ModifyExistingScriptTest)
...
+    // Launch the multiplayer test runner process, it will clear the script source, and we'll recieve the replication back here
+    MultiplayerTestRunnerProcess RemoteClient
+        = MultiplayerTestRunnerProcess(MultiplayerTestRunner::TestIdentifiers::TestIdentifier::CLEAR_SCRIPT_SOURCE)
+              .SetSpaceId(Space.Id.c_str())
+              .SetLoginEmail(RemoteUser.Email.c_str())
+              .SetPassword(GeneratedTestAccountPassword)
+              .SetEndpoint(EndpointBaseURI())
+              .SetTimeoutInSeconds(60);
+
+    std::future<void> RemoteClientReady = RemoteClient.ReadyForAssertionsFuture();
```
mag-lt, line 1435: [fix] this is unused (unless I'm missing something and it somehow causes something to run?). I presume it isn't needed in this case because the side effect the runner will cause is checked (the script source becoming empty), vs explicitly checking that the ready future has been set first before doing any test specific assertions.

For consistency with other tests, the future is the thing to check first, and then any specific assertions, but whatever you prefer (if keeping the current style then drop this variable).

Tests/src/PublicAPITests/ScriptSystemTests.cpp
```diff
@@ -1352,6 +1353,112 @@ TEST_P(ModifyExistingScript, ModifyExistingScriptTest)
...
+    // Log in local user
+    csp::common::String UserId;
+    LogIn(UserSystem, UserId, LocalUser.Email, GeneratedTestAccountPassword);
+
+    // Create space
+    const char* TestSpaceName = "CSP-UNITTEST-SPACE-MAG";
+    const char* TestSpaceDescription = "CSP-UNITTEST-SPACEDESC-MAG";
+
+    char UniqueSpaceName[256];
+    SPRINTF(UniqueSpaceName, "%s-%s", TestSpaceName, GetUniqueString().c_str());
```
mag-lt: [nit] I know this is just copied from other tests, but something like `const auto UniqueSpaceName = fmt::format("{}-{}", TestSpaceName, GetUniqueString())` is a bit more contemporary

mag-lt (changes requested):
The change itself looks good. I've left a fix comment for the change log, and a couple in the test case. All are rather minor in the grand scheme of things, but still need addressing

Tests/src/PublicAPITests/ScriptSystemTests.cpp
MAG-ElliotMorris: Oh, that's stilly of me, I thought wait_for threw if it didn't get a success after the time.

Tests/src/PublicAPITests/ScriptSystemTests.cpp
MAG-ElliotMorris: Ach yeah, i'm a moron.

Got a bit hung up on this idea that a replication message may not have reached us by the time that the test runner finished that I never bothered to even assert the test runner had finished. I suppose the subsequent check guards it anyhow, but that's no excuse for sloppiness.

 (There used to be a bit where I was doing `wait_for` in a loop like we do elsewhere, to simulate continual ticks like a game loop, but i decided I didn't need it)

Tests/src/PublicAPITests/ScriptSystemTests.cpp
MAG-ElliotMorris: You think `wait_for` is overly paranoid vs just `wait`? Just trying to stop infinite hangs but I suppose the test framework will do that anyway.

mag-lt: I think `wait_for` is appropriate. I don't believe GTest itself has built-in mechanism for timing out (it'd need to run all the tests in a separate process or something), though running via ctest may give us something.

Here though, I think `wait_for` as you had it is fine, just think we ought to surface the issue explicitly if the condition is not met.

MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1021/changes/70d4619fb960df00dbba2709933a3dbb07028c90

CHANGELOG.md
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1021/changes/93ee75d2768cb7b7f8e29824de0d78012299b8ba

Tests/src/PublicAPITests/ScriptSystemTests.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1021/changes/5c2803b46736c5ca9bf90b937b6121a03604e8d9

Tests/src/PublicAPITests/ScriptSystemTests.cpp
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1021/changes/dd99fa5f8751f9e93736122bef5d01721f5e9248

mag-lt (approved):

MAG-AdamThorn (approved):

PR #1022, opened 2026-08-24
[NT-0] ci: Fix quoting issue in CI preventing GTEST_FILTER from being expanded

MAG-ElliotMorris:
We haven't been running tests for the last few days, since that zizmor change that hoisted variables.

Our test invocations, uniquely, used single quoted strings,. Double quoted strings are necessary for variable expansion in bash. This is actually why it's more secure (I think), as previously it was the github actions pre-processor doing this substitution, which is less restrictive.

You can see the tests run now : https://github.com/magnopus-opensource/connected-spaces-platform/actions/runs/32732542048/job/97456344385?pr=1022


mag-lt (approved):
Good catch, thanks for fixing.

MAG-mv (approved):

PR #1024, opened 2026-08-26
[OF-1882] fix: Guard RealtimeEngine destructor calling NetworkEventBus

MAG-ElliotMorris:
In the case where an online realtime engine was destructed after Shutdown() had been called (not unusual when exiting play mode in an editor with a GC'd language), depending on if leader election was enabled or not, there was a possibility of getting a null dereference.

This adds a guard, breaking modularity, but no more than it already is as the script runner does the exact same thing here. This isn't a true fix, as lack of syncronisation could still cause a crash if action happens to occur between the set of the `isInitialised` bool and the call here, but this radically reduces the window. We still need to make this formally correct, which we can do by changing how we own our connection objects. This closes the window significantly for now though.

Test added reproduces the issue 100% prior to fix, now passes.

### DIFF
--- a/Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
+++ b/Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
@@ -816,10 +816,20 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, ConstructFromJs
 
     EXPECT_CALL(*SignalRMock, Invoke).WillRepeatedly(MakeSingleEntityCreationInvoker());
 
-    auto [Entity] = AWAIT(&Engine, CreateEntity, "SchemaEntity", csp::multiplayer::SpaceTransform {}, csp::common::Optional<uint64_t> {});
+    auto [Entity] = AWAIT(&Engine, CreateEntity, "SchemaEntity", csp::multiplayer::SpaceTransform { }, csp::common::Optional<uint64_t> { });
     ASSERT_NE(Entity, nullptr);
 
     EXPECT_NE(Entity->AddComponentByTypeId(AllPropertyTypesSchemaId), nullptr);
     EXPECT_NE(Entity->AddComponentByTypeId(EmptySchemaId), nullptr);
     EXPECT_EQ(Entity->AddComponentByTypeId(InvalidSchemaId), nullptr);
+}
+
+CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, RealtimeEngineCanBeDestructedAfterSystemsShutdown)
+{
+    auto& SystemsManager = csp::systems::SystemsManager::Get();
+    csp::multiplayer::OnlineRealtimeEngine* RealtimeEngine { SystemsManager.MakeOnlineRealtimeEngine() };
+    RealtimeEngine->EnableLeaderElection(); // Neccesary to make LeaderElectionManager not null internally, such that we trigger the appropriate
+                                            // destruction path in realtime engine
+    csp::CSPFoundation::Shutdown();
+    delete RealtimeEngine; // Should not crash
 }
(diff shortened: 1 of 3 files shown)


Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
```diff
@@ -816,10 +816,20 @@ CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, ConstructFromJs
...
 }
+
+CSP_PUBLIC_TEST_WITH_MOCKS(CSPEngine, OnlineRealtimeEngineTests, RealtimeEngineCanBeDestructedAfterSystemsShutdown)
+{
+    auto& SystemsManager = csp::systems::SystemsManager::Get();
+    csp::multiplayer::OnlineRealtimeEngine* RealtimeEngine { SystemsManager.MakeOnlineRealtimeEngine() };
+    RealtimeEngine->EnableLeaderElection(); // Neccesary to make LeaderElectionManager not null internally, such that we trigger the appropriate
+                                            // destruction path in realtime engine
+    csp::CSPFoundation::Shutdown();
+    delete RealtimeEngine; // Should not crash
```
mag-lt, line 834: [Comment] When explicitly written out like this (teardown dependencies that `RealtimeEngine` depends on, whilst it is still alive) this sort of spells out that it is the owner/callers fault i.e. user error / API misuse.

This would be even clearer if using the actual constructor we export (vs the non-exported convenience one on systems manager) that forces dependencies to be passed:

```cpp

auto Engine = csp::multiplayer::OnlineRealtimeEngine {
    *SystemsManager.GetMultiplayerConnection(),
    *SystemsManager.GetLogSystem(),
    *SystemsManager.GetEventBus(),
    *SystemsManager.GetScriptSystem(),
};
```

Clearly systems manager (and `NetworkEventBus` etc) needs to live longer. We can blame GC-ed languages, but in my view that just points to that trying to hook in to their non-deterministic, often threaded, destroy/cleanup methods is perhaps futile (the general wisdom with Java for example as far as I've picked up from previous projects I've worked on is to never use finalizers, and usually just suck it up and have some explicit destroy method).

Having clients construct the engine themselves is also perhaps a footgun, vs actually making the library / systems manager responsible for it, i.e. `SystemsManager.MakeOnlineRealtimeEngine()` as written above, as that may give more flexibility for the implementation to do some heroics to coordinate destruction / deal with the coupling between the two.

Anyway, I know I'm preaching to the choir, and we need to get a handle on this stuff more generally anyway.

mag-lt (approved):
A somewhat reluctant approval (to a somewhat reluctant fix). Clearly the pragmatic thing to do now, but as I said in the [comment](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1024/changes#r3870374400) I left, this is arguably user error (and in any case not surprising, considering we sort of open up the possibility for this by-design).

Tests/src/PublicAPITests/OnlineRealtimeEngineTests.cpp
MAG-ElliotMorris: I'm loathe to throw away the flexibility of client ownership of the engine, as I believe it will cause us to have to explicitly design composability patterns for the usage of multiple engines in parallel. 

I tend to lean the other way, thinking that we internally have a problem with _needing_ any long-term instantiated memory outside of the engine. The engine should simply own everything that is runtime instantiable. "Systems" are REST CRUD, and can be static (and majorly trimmed/reconceptualized). In some ways, this reduces down to an argument about naming when you get down to it, as this does imply some sort of non-owning "view" "Engine -> SpaceView", which is pretty much the same split as "Systems -> Engine" on this axis. (Although I feel does edge it out on others, although that is intuition and not well defined in my mind)

The thing with GC'd languages is extra complex when you consider that they are GC'd languages in embedded runtimes inside game engines. That's not to say there isn't going to be a "onPlayExit" hook, but I can imagine that being tricker. Teardown is not the same as application exit when "play mode" is involved, and that's not to mention editor embedded scripting.

I'm not against the idea, I just think it's a distinction without a difference. We either move stuff into the engine, or move the engine into the stuff, and I think the former leads to a better, more clearly defined library with clearer tranches of functionality and responsibility.

MAG-AdamThorn (approved):

PR #1025, opened 2026-08-27
[NT-0] docs: Data-driven components overview

mag-lt:
Some high-level docs on what data-driven components are, how to define them, and the core API for getting/setting properties etc.

### DIFF
--- a/Library/docs/source/manual/documentation/data_driven_components.md
+++ b/Library/docs/source/manual/documentation/data_driven_components.md
@@ -1,0 +1,227 @@
+# Data-Driven Components
+
+In contrast to adding a new component by writing code [in C++](adding_components), which necessitates modifying the CSP library itself, building and distributing a new release etc, a "data-driven" component is one described by a schema, specifically a JSON document, that can be defined externally from the library and injected at runtime, allowing clients to author their own components for use in their connected applications.
+
+The schema describes the component, its properties, their types, any constraints on values, names for exposing to the script system etc, and then CSP takes care of replicating the component state as it always has, and provides a dynamic runtime API for reading and writing properties, ensuring that values conform to those described in the schema when set programmatically. 
+
+Below is an example of such a component which we'll reference throughout:
+
+```json
+{
+  "typeId": 808,
+  "name": "DistortionAudioEffect",
+  "properties": [
+    {
+      "key": 0,
+      "name": "gain",
+      "type": "float",
+      "defaultValue": 0.25,
+      "range": {
+        "min": 0.0,
+        "max": 1.0
+      }
+    },
+    {
+      "key": 1,
+      "name": "level",
+      "type": "float",
+      "defaultValue": 0.5,
+      "range": {
+        "min": 0.0,
+        "max": 1.0
+      }
+    },
+    {
+      "key": 2,
+      "name": "mode",
+      "type": "int",
+      "defaultValue": 0,
+      "options": [
+        {
+          "name": "Vintage",
+          "value": 0
+        },
+        {
+          "name": "Modern",
+          "value": 1
+        }
+      ]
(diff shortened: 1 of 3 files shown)


Library/docs/source/manual/documentation/data_driven_components.md
```diff
@@ -0,0 +1,227 @@
...
+
+```cpp
+const auto DistortionAudioEffectTypeId = uint64_t { 808 };
+
+auto* Component = entity->AddComponentByTypeId(DistortionAudioEffectTypeId);
+```
+
+### Reading Component Properties
+
+Traditionally, there has been a concrete class for each component type that CSP provides, and the caller was expected to cast to that (knowing the mapping from `ComponentType` to class), and could then call distinct methods for reading and writing properties, in a type-safe way.
```
MAG-mv, line 141: [Consider] Theres a few instances where you've mentioned how we did things previously. We could seperate the justificaiton for moving to data-driven components over traditional ones in a seperate docs. This will keep the actualt data-driven component docs a bit shorter.

I'm not too fussed about this

MAG-mv (approved):

Library/docs/source/manual/documentation/data_driven_components.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+## Schema Reference
+
+A data-driven component conforms to the JSON structure detailed below
+### Component Schema object
+
+| Field | | |
+|---|---|---|
+| `typeId` | required | Integer (`uint64_t` on the wire, but as this is JSON we're limited to 53 bits). Must be unique across everything registered with an engine. |
```
MAG-AdamThorn: [Consider] I think the 'uniqueness' could do with being expanded upon. Consider linking to the `Component TypeId allocation` section below.

Library/docs/source/manual/documentation/data_driven_components.md
```diff
@@ -0,0 +1,227 @@
+# Data-Driven Components
+
+In contrast to adding a new component by writing code [in C++](adding_components), which necessitates modifying the CSP library itself, building and distributing a new release etc, a "data-driven" component is one described by a schema, specifically a JSON document, that can be defined externally from the library and injected at runtime, allowing clients to author their own components for use in their connected applications.
```
MAG-AdamThorn: [nit] These opening paragraphs would be easier to read it they were broken up into more sentences.

Library/docs/source/manual/documentation/data_driven_components.md
MAG-AdamThorn: That is a good point. Doing that would also mean we don't need to come back and remove this historical information once the old components are removed.

MAG-AdamThorn (approved):

Library/docs/source/manual/documentation/data_driven_components.md
mag-lt: Linked to the later section in 1ea0db67c5d18549d7f8c12813bbe8f39725bf69

Library/docs/source/manual/documentation/data_driven_components.md
mag-lt: Tightened these up a bit in 2dc7d6a3051a298dc844662bc591a8c190a2253c

Library/docs/source/manual/documentation/data_driven_components.md
mag-lt: Yeah, I struggled to find a good balance here, as I think the actual audience for this in the first instance does truly need a bit of a migration guide (which this doubles as). I'm going to leave this as is, as I can't think how to split it meaningfully at the moment, and so will have to accept that it will date itself over time and need updating.

PR #1028, opened 2026-09-04
[NT-0] docs: Add build system documentation

MAG-mv:
This is part 1 of the build system architecture documentation. The second part will cover the CI

### DIFF
--- a/Library/docs/source/manual/documentation/build_system.md
+++ b/Library/docs/source/manual/documentation/build_system.md
@@ -1,0 +1,203 @@
+# Build System Architecture
+
+The CSP build system is based around CMake, which handles configuration, compilation, linking, testing, and installation.
+
+Conan is recommended for dependency management, toolchain configuration, and packaging of third-party libraries. We provide Conan profiles and custom tooling to make these workflows easier, but Conan is not required by the CMake build system.
+
+## High-Level Architecture
+
+When using Conan for dependency management, the typical build system flow is:
+
+```
+Conan Profile
+      │
+      ▼
+ conan install
+      │
+      ├── Resolve third-party dependencies
+      │
+      └── Generate CMake toolchain
+              │
+              ▼
+         CMake configure
+              │
+              ▼
+          CMake build
+              │
+              ▼
+         CMake install
+```
+
+## Conan
+
+Our Conan recipe declares the external packages required by CSP, such as:
+
+```
+RapidJSON
+fmt
+Async++
+msgpack-cxx
+tinyspline
+GLM
+Poco
+```
+
+Platform-specific dependencies are also handled here. For example, Poco is not used for Emscripten builds.
+
+The recipe also configures individual dependencies to work with CSP correctly, such as:
+
+```
+Disabling unused Poco components.
+Disabling Boost support in msgpack-cxx.
+Disabling the tinyspline C++ interface.
+Forcing dependencies to use static linkage.
+Enabling position-independent code.
+```
+
+This means the majority of dependency specific knowledge belongs in Conan rather than being duplicated throughout the CMake project.
+
+## Conan-Generated CMake Integration
+
+The recipe uses two Conan generators:
+
+```
+CMakeDeps
+CMakeToolchain
+```
+
+CMakeDeps generates the CMake package information required so we can find dependencies like so:
+
+```find_package(Poco REQUIRED)```
+
+CMakeToolchain takes the selected Conan configuration and creates a CMake toolchain.
+
+The way to understand how these systems inetract is:
+
+> Conan determines what dependencies and build environment CMake sees. CMake determines how CSP is built.
+
+
+## Conan Profiles
+
+Profiles define the build environment for each supported platform and toolchain combination.
+
+Current profiles include targets such as:
+
+```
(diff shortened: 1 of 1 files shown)


Library/docs/source/manual/documentation/build_system.md
```diff
@@ -0,0 +1,203 @@
...
+CMakeToolchain
+```
+
+CMakeDeps generates the CMake package information required so we can find dependencies like so:
+
+```find_package(Poco REQUIRED)```
+
+CMakeToolchain takes the selected Conan configuration and creates a CMake toolchain.
+
+The way to understand how these systems inetract is:
```
mag-lt, line 74: [fix] typo: inetract -> interact

Library/docs/source/manual/documentation/build_system.md
```diff
@@ -0,0 +1,203 @@
...
+tinyspline
+GLM
+Poco
+```
+
+Platform-specific dependencies are also handled here. For example, Poco is not used for Emscripten builds.
+
+The recipe also configures individual dependencies to work with CSP correctly, such as:
+
+```
```
mag-lt, line 49: [nit] here and and various other places in the doc a code block is used instead of bullet points, which is an unconventional choice, and in most cases that I can see would read better as bullet points

Library/docs/source/manual/documentation/build_system.md
```diff
@@ -0,0 +1,203 @@
...
+Building tests.
+Installing headers and binaries.
+Generating CSP's CMake package configuration.
+```
+
+Vendored CSP dependencies such as SignalR and QuickJS are included using ```add_subdirectory()```.
+
+Other source dependencies, such as csp-services, are brought into the build using FetchContent.
+
+The main CSP target is:
```
mag-lt, line 136: [nit] spreading this over 14 lines seems a bit excessive. If there were more detail it might make sense and make it easier to parse the signal from the noise, but here I think this could likely just be a reasonably short sentence on a single line

Library/docs/source/manual/documentation/build_system.md
```diff
@@ -0,0 +1,203 @@
...
+
+```SHARED```
+
+depending on the value of:
+
+```CSP_BUILD_SHARED```
+
+## Platform Abstraction
+
+The build system uses compiler expressions rather than creating multiple build projects for each platform.
```
mag-lt, line 154: [Question] What is a "compiler expression" in this context?

We use a combination of configuration-time conditionals (i.e. `if(ANDROID)`) and generator expressions (i.e. `$<$<PLATFORM_ID:Windows>:CSP_WINDOWS>`) in order to do various different things conditionally depending on compiler and target platform.

This is all relatively conventional cmake stuff and I'm not quite sure what this section is intending to communicate to the audience, I don't think I would expect there to be "multiple build projects for each platform", is that contrasting it somehow with premake?

Library/docs/source/manual/documentation/build_system.md
```diff
@@ -0,0 +1,203 @@
...
+    ├── Async++
+    ├── tinyspline
+    └── ...
+```
+
+For this reason, CSP provides a custom Conan deployer that works by iterating Conan's resolved dependency graph and collects the libraries to copy into the final install package.
+
+## Further reading
+
+To learn how to use the build system, a document is available [here](../building/build-system.md).
```
mag-lt, line 203: [nit] having both `build_system.md` (the file added in this review) and a `build-system.md` file elsewhere in the tree stands out here. Makes me wonder what this new doc is really adding, as the other doc contains the actual details I'd be interested in as a developer on the project.

In any case, seems this document is perhaps more of a `build_system_architecture.md` or `build_system_background.md` or something?

mag-lt (changes requested):
I'm wondering if this doc, at least in its current form, is really justified over the existing [build-system.md](https://github.com/magnopus-opensource/connected-spaces-platform/blob/main/Library/docs/source/manual/building/build-system.md) document.

A lot of the information is reasonably generic and I'd personally just glean from parsing the CMakeLists.txt. The existing doc already details what I need to do to check out and build the project, which is the main thing I'd personally be looking for. If we're intending to capture more of the "why", then I think some we should boil this new document down to that (although, again, I think this could probably just be combined into the other existing doc). The `High-Level Architecture` bit and the preceding intro/overview are perhaps closest to this (i.e. "the why") currently.

For example there are statements like:

> Vendored CSP dependencies such as SignalR and QuickJS are included using add_subdirectory().

and

> Other source dependencies, such as csp-services, are brought into the build using FetchContent.

Without any explanation as to why, which sort of just makes this a prose version of the CMakeLists.txt

Library/docs/source/manual/documentation/build_system.md
MAG-ElliotMorris: I tend to agree with Lucas here. I don't see the thematic difference between these two documents, y'all should merge them I reckon.

MAG-ElliotMorris:
Matts going to reframe this, so un-requesting my review.

MAG-mv:
Putting back o draft so I can refram this around our intentions of using Conan/Cmake

PR #1030, opened 2026-09-08
[OF-1896] feat!: Add double support to `ReplicatedValue`

mag-lt:
This adds double support to `ReplicatedValue`, and works through the implications of that for component schema and script bindings, network event bus etc.

The most contentious part of this is in the network event serialisation (in `NetworkEventManagerImpl.cpp` and `NetworkEventSerialisation.cpp`), where a breaking change is introduced regarding how `float` values are handled. Previously `float`s were being sent over the wire as doubles (specifically a `NULLABLE_DOUBLE`, despite CSP always actually sending a value). Trying to fit within that existing nullable pattern after introducing an actual double precision type creates a clash which is reconciled here by the breaking change, marking `float`s as `FLOAT`s and `double`s as `DOUBLE`s. Old and new clients won't interop on float event args. A more conservative non breaking change is possible (leave `float`s as `NULLABLE_DOUBLE`, and just use `DOUBLE` for double), but as we want to reconcile this code with the other `MCSType`s at some point (where nullable types aren't used), I suspect a breaking change is inevitable.


Library/include/CSP/Common/ReplicatedValue.h
```diff
@@ -30,7 +30,7 @@ namespace csp::common
 CSP_START_IGNORE
```
MAG-mv, line 30: [Comment] I hope this commit was relatively easy to figure out. I redesigned the entire entity serialization process after going through hell of adding new types with the previous system.

Tests/src/InternalTests/ComponentSchemaScriptBindingTests.cpp
```diff
@@ -259,6 +260,11 @@ CSP_INTERNAL_TEST(CSPEngine, ComponentSchemaScriptBindingTests, SupportedTypesGe
                     "vec4Property",
                     PlainValue<csp::common::Vector4> { csp::common::Vector4(1.0f, 2.0f, 3.0f, 4.0f) },
                 },
+                {
+                    7,
+                    "doubleProperty",
+                    PlainValue<double> { 1.0 + std::numeric_limits<double>::epsilon() },
```
MAG-ElliotMorris, line 266: [Question] Why the epsilon addition here and elsewhere? Just curious, could learn something. Do you need epsilon to reliably initialize 1.0 or something?

MAG-ElliotMorris (approved):
Looks by the numbers from my end. Matt will be the real judge though.

I'm not certain anyone is even building features atop custom events. It's a feature I really _want_ to be used to build features, but I'm not sure anyone realizes you're even allowed to do it. Might have to give up on that dream. What I mean to say is, I don't think the break will be a problem.

Great that we get in front of the video player timestamp problem before it completely throws off some complex PRD/Demo assumptions folk will have been making.

Tests/src/InternalTests/ComponentSchemaScriptBindingTests.cpp
MAG-mv: I assume its to test that the double precision is being preserved

MAG-mv (approved):
Oops. I reviewed this yesterday but didn't hit submit. Looks good to me! I think changing the CHS types is fine, as they are events and not data that is stored. Otherwise we would have to do some kind of migration

Tests/src/InternalTests/ComponentSchemaScriptBindingTests.cpp
mag-lt: > I assume its to test that the double precision is being preserved

Yeah, this.

All I'm really looking for is something a double can hold that a float can't, but couldn't come up with a nice magic number off the top off my head, 1 + epsilon is such a number. Epsilon is the difference between 1 and the next representable number (defined by the IEEE spec, which is why it is standard across JS and C++ etc). It is 1 + epsilon because a single precision float can actually fully represent the raw epsilon value itself, the precision limit is only triggered when it is added to 1 (I don't pretend to understand all the details of floating point numbers, but they're not evenly spaced meaning there is more precision below 1 than above it).

MAG-ElliotMorris: Ah right, makes sense cheers.

Library/include/CSP/Common/ReplicatedValue.h
mag-lt: The work was mostly guided by the compiler when adding a new type to ReplicatedValue (as in a bunch of stuff doesn't compile until it is all addressed), in large part because of that rework you did.

PR #1031, opened 2026-09-11
[NT-0] chore: Remove unreferenced `ExternalLinkActions` enum

MAG-ElliotMorris:
Completely unreferenced. Spotted this during bindings.


mag-lt (approved):

mag-lt (approved):
I've never fully understood what these are, possibly documentation? There are a couple of others (`VideoPlayerActions`, `AnimatedModelActions`), which are also unreferenced here. I guess client code could be referencing them, but for what? The "actions" stuff is string based (i.e. `ComponentBase::RegisterActionHandler` etc). I see no issue with removing these, but I guess I could be missing something.

MAG-ElliotMorris:
If they're unreferenced, I don't think we should be keeping them around. There's no value in providing enums to clients. I didn't realize there were more, i'll take a gander. The fact that there are more though does give me a slight pause, will run a search in client repos.

My suspicion is that this is a consequence of up-front-design/not finishing things rather than doing stuff iteratively.

MAG-AdamThorn (approved):

MAG-ThomasGreenhalgh:
I think @mag-lt may be right in that they could originally have been related to the `registerActionHandler` mechanism, however that uses strings for action identifiers. 
In the web client we have only two actions currently: `'PlayAnimation'` and `'PauseAnimation'` for animated model components.

PR #1032, opened 2026-09-15
[OF-1841] fix: Remove public dep requirements for consumers

MAG-mv:
Consumers do need need to locate these libs anymore when consuming csp via a shared lib.
I have proven this by getting Unity to link against CSP and run without them.


MAG-ElliotMorris (approved):

MAG-ElliotMorris:
[Consider] Oh, probably does want a changelog entry. It's just a potential error if you're wrong, and folks won't know why if there's no entry.

MAG-AdamThorn (approved):

PR #1033, opened 2026-09-15
[NT-0] Increase warning levels for lib target + enable warnings as errors in tests 

mag-lt:
Following on from low hanging fruit session earlier.

This enables warnings as errors for the tests, and rolls out consistent warning levels across all platforms for both the lib and test target. In the process we remove one of the warning suppressions across the board (relating to const value return types), but more remain (so more work to do there in another session)

The suppressions differ between lib and test targets as appropriate, trying to keep these to a minimum.


Library/include/CSP/Common/Array.h
```diff
@@ -205,7 +205,7 @@ template <typename T> class CSP_API Array
 
     /// @brief Returns the number of elements in the array.
     /// @return const size_t
```
MAG-AdamThorn: [fix] Update comment and remove mention of 'const'.

Library/include/CSP/Common/List.h
```diff
@@ -333,7 +333,7 @@ template <typename T> class CSP_API List
 
     /// @brief Returns the number of elements in the array.
     /// @return const size_t
```
MAG-AdamThorn: [fix] Update comment and remove mention of 'const'.

MAG-AdamThorn (changes requested):

MAG-AdamThorn:
Two minor fixes but looks good otherwise, thanks for taking this on @mag-lt

Library/include/CSP/Common/Array.h
mag-lt: Address in fa819cd3bbcf8bbd16383812fbec1ffc9eeb16c4

Library/include/CSP/Common/List.h
mag-lt: Address in fa819cd3bbcf8bbd16383812fbec1ffc9eeb16c4

MAG-mv (approved):

MAG-AdamThorn (approved):

PR #1034, opened 2026-09-16
[NT-0] chore: Bindings motivated const correctness changes

MAG-ElliotMorris:
You have to contort yourself a tad in the web bindings when things arn't const correct. Make the easy fixes. We're sort of behind, embind assumes move-capable objects in its philosophical foundation, so does not countenance non const mutable references for basic paths.

That isn't to say we can't support them, we can, it's just it requires out-of-pattern stuff and I'd rather avoid doing that too much when we're not actually using them for their intended use.

This involves some copies, but they're not going to be significant, and closes potential memory safety issues.


CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ All notable changes to this project will be documented in this file. For compile
   sent using the wire type for doubles, leaving no way to tell the two apart. Each is now sent using the
   corresponding wire type.
   Clients need to be updated together. Entity and component replication are unaffected.
+  
+- [NT-0] chore: Various bindings-motivated const correctness changes by @ElliotMorris
```
mag-lt, line 27: [nit] These are technically breaking changes, at least from an ABI perspective (though I suspect from an API perspective most client code is probably compatible)

Library/src/Systems/Spaces/SpaceSystem.cpp
```diff
@@ -1082,7 +1082,7 @@ void SpaceSystem::RemoveUserFromSpace(const String& SpaceId, const String& UserI
     static_cast<chs::GroupApi*>(GroupAPI)->groupsGroupIdUsersUserIdDelete({ SpaceId, UserId }, ResponseHandler);
 }
 
-void SpaceSystem::AddSiteInfo(const String& SpaceId, Site& SiteInfo, SiteResultCallback Callback)
+void SpaceSystem::AddSiteInfo(const String& SpaceId, Site SiteInfo, SiteResultCallback Callback)
```
mag-lt: [comment] it is quite odd that these methods were modifying the `SiteInfo` before (by setting the given SpaceId 😵‍💫 ). I don't imagine clients were relying on that (i.e. for their passed in info to be modified), but this is technically a behaviour break if they were...

Arguably we only need to be passed the SiteInfo (as const ref), and the SpaceId is redundant.

Library/src/Systems/Assets/AssetSystem.cpp
```diff
@@ -1401,7 +1401,7 @@ CSP_ASYNC_RESULT_WITH_PROGRESS void AssetSystem::RegisterAssetToLODChain(
 }
 
 void AssetSystem::CreateMaterial(const csp::common::String& Name, const csp::systems::EShaderType ShaderType, const csp::common::String& SpaceId,
-    csp::common::Map<csp::common::String, csp::common::String>& Metadata, const csp::common::Array<csp::common::String>& AssetTags,
+    csp::common::Map<csp::common::String, csp::common::String> Metadata, const csp::common::Array<csp::common::String>& AssetTags,
```
mag-lt, line 1404: [comment] Somehow this change seems more obviously correct than what I noted about the SiteId stuff in my other [comment](https://github.com/magnopus-opensource/connected-spaces-platform/pull/1034/changes/839f8150ed871cb1626ff8b8583da27824f46dce#r4026815680) i.e. patching in the shader metadata is just additional metadata we want to add to the created material, the caller wouldn't expect to know about it and we should've been creating a copy internally all along.

mag-lt (approved):
I think it is great that the web bindings work is surfacing this stuff (you had a similar const-ness one recently). Small changes like this do make a real difference.

The mutable ref -> value changes are the more contentious parts here, but I think in all cases seem like the original API was wrong and these were never conceptually out params (and so I agree with the copies).

This is technically a breaking change from my point of view, might want to flag that.

MAG-AdamThorn (approved):

CHANGELOG.md
MAG-ElliotMorris: I don't believe we consider non-API breaks as breaking, it's a constant problem in C++ library development deciding on what breaking actually means. I think our attitude for the moment should be "Will the client teams need to think about it".

I have altered the SiteInfo stuff to be an API break, so i'll call that out at least.

Library/src/Systems/Spaces/SpaceSystem.cpp
MAG-ElliotMorris: I did change this. The tests are even stranger now, but it's not like they weren't strange before.

https://github.com/magnopus-opensource/connected-spaces-platform/pull/1034/commits/0946eaf179daa4768db0e5a08d6c884090eecca4

CHANGELOG.md
MAG-ElliotMorris: https://github.com/magnopus-opensource/connected-spaces-platform/pull/1034/commits/37d20805864ca7b1cfe16d18034ad98f067f73ad

mag-lt (approved):
Changing to pass just the `SiteInfo` is much cleaner, thanks for working that through. The tests are fine in my opinion, there is probably something that can be done to tidy up the `AddSiteInfo` test helper but I don't think it is worth it here.

PR #1035, opened 2026-09-16
[NT-0] chore: Print entity name in failure message.

MAG-ElliotMorris:
Not clean, but quick, external disconnected state like this is a bit gross. In play the build, we were having trouble doing a reproduction in a large space for this error, because we had no clue what the entity was. This will help if it happens again.


MAG-ElliotMorris:
Logic error, name will always be empty, I have an empty head.

